PR #10, opened 2015-12-07
don't reset fontsize and line height.

zilongshanren:
Fix https://github.com/fireball-x/fireball/issues/762


zilongshanren:
@pandamicro  please review

PR #103, opened 2015-12-21
refactor to use data after trimmed

dabingnn:


dabingnn:
问题：
https://github.com/fireball-x/fireball/issues/1059
https://github.com/fireball-x/fireball/issues/1041
都已经修复

PR #131, opened 2015-12-25
Support Scheduler in Component instead of Node

pandamicro:


pandamicro:
For https://github.com/fireball-x/fireball/issues/673

PR #159, opened 2015-12-29
Fix text wrap issue for system Label

pandamicro:


pandamicro:
For fix https://github.com/fireball-x/fireball/issues/1273

PR #189, opened 2016-01-05
V0.7 label treat noneas clamp

dabingnn:


dabingnn:
@nantas @pandamicro 修复Label的overflow为NONE报红字的bug

PR #199, opened 2016-01-07
Fix issues in JSB v3.3

pandamicro:


pandamicro:
For fix https://github.com/fireball-x/fireball/issues/1413

PR #203, opened 2016-01-07
Made Action independent from retain release

pandamicro:


pandamicro:
Fixed https://github.com/fireball-x/fireball/issues/1450

PR #223, opened 2016-01-11
fix scrollview brake with large content size.

zilongshanren:
fixed: https://github.com/fireball-x/fireball/issues/1487


zilongshanren:
@pandamicro 求review

PR #229, opened 2016-01-12
Fix crash issue in JSB v3.9 caused by event listener release

pandamicro:


pandamicro:
Fixed https://github.com/fireball-x/fireball/issues/1518

PR #234, opened 2016-01-12
Update actions documention

pandamicro:


pandamicro:
https://github.com/fireball-x/fireball/issues/1535

PR #254, opened 2016-01-15
fix label does not work on sougou browser

dabingnn:


dabingnn:
https://github.com/cocos-creator/fireball/issues/1557
https://github.com/cocos-creator/fireball/issues/1558

PR #255, opened 2016-01-15
should not drawImage if size zero

dabingnn:


dabingnn:
https://github.com/cocos-creator/fireball/issues/1582

dabingnn:
https://github.com/cocos-creator/fireball/issues/1557

jareguo:
Prior to Gecko 7.0 (Firefox 7.0 / Thunderbird 7.0 / SeaMonkey 2.4), Firefox threw an exception if any of the coordinate values was non-finite or zero. As per the specification, this no longer happens.

PR #259, opened 2016-01-17
Fix bmfont texture not handled in Canvas

pandamicro:


pandamicro:
@dabingnn

pandamicro:
Fixed https://github.com/cocos-creator/fireball/issues/1558

PR #264, opened 2016-01-18
fix label component init font size issue.

zilongshanren:
fix https://github.com/cocos-creator/fireball/issues/1612

fix https://github.com/cocos-creator/fireball/issues/1573


zilongshanren:
@jareguo

PR #271, opened 2016-01-20
fix sprite

dabingnn:
1. sizeMode is not inited
2. trimmode is not synced to scale9 sprite


dabingnn:
fix https://github.com/cocos-creator/fireball/issues/1630#issuecomment-172857881

PR #334, opened 2016-02-17
fix reset scrollbar null exception

zilongshanren:
fix https://github.com/cocos-creator/fireball/issues/1840#issuecomment-184991182


zilongshanren:
@jareguo 
Done.

PR #342, opened 2016-02-18
support negtive value for filled sprite

dabingnn:


dabingnn:
for issue https://github.com/cocos-creator/fireball/issues/1946

PR #346, opened 2016-02-18
Fix comment and JSB issues

pandamicro:


pandamicro:
Fixed https://github.com/cocos-creator/fireball/issues/1839

PR #355, opened 2016-02-20
Add font downloader

pandamicro:


pandamicro:
https://github.com/cocos-creator/fireball/issues/1499

PR #592, opened 2016-04-12
minor tweaks

jareguo:
- 避免 component.name 和 node.name 同名引发的学习障碍
- gulp: rename fast-build to build-dev (https://github.com/cocos-creator/fireball/pull/2693)

@cocos-creator/engine-admins

### DIFF
--- a/cocos2d/core/components/CCComponent.js
+++ b/cocos2d/core/components/CCComponent.js
@@ -230,14 +230,15 @@ var Component = cc.Class({
 
         name: {
             get: function () {
-                return this._name || this.node.name;
-                //var className = cc.js.getClassName(this);
-                //var index = className.lastIndexOf('.');
-                //if (index >= 0) {
-                //    // strip prefix
-                //    className = className.slice(index + 1);
-                //}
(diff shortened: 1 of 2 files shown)


cocos2d/core/components/CCComponent.js
```diff
@@ -230,14 +230,15 @@ var Component = cc.Class({
...
-                //return this.node.name + '<' + className + '>';
+                if (this._name) {
+                    return this._name;
+                }
+                var className = cc.js.getClassName(this);
+                var trimLeft = className.lastIndexOf('.');
+                if (trimLeft >= 0) {
+                    className = className.slice(trimLeft + 1);
+                }
+                return this.node.name + '<' + className + '>';
```
pandamicro, line 241: Are you sure about this ? Now value get is not the same with value set.

jareguo, line 241: if value was set, it will return this._name directly

jareguo:
https://github.com/cocos-creator/fireball/pull/2693 已合并，该 PR 先同步合并

PR #642, opened 2016-05-04
fix layout errors when child's width is negative

zilongshanren:
Re: cocos-creator/fireball#2576

- Fix layout error when children's size < 0

@cocos-creator/engine-admins

### DIFF
--- a/cocos2d/core/components/CCLayout.js
+++ b/cocos2d/core/components/CCLayout.js
@@ -408,7 +408,7 @@ var Layout = cc.Class({
         var row = 0;
         var containerResizeBoundary;
 
-        var maxHeightChildAnchor = cc.p(0, 0);
+        var maxHeightChildAnchorY = 0;
 
         var newChildWidth = this.cellSize.width;
         if (this.type !== Type.GRID && this.resizeMode === ResizeMode.CHILDREN) {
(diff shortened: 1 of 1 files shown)


cocos2d/core/components/CCLayout.js
```diff
@@ -408,7 +408,7 @@ var Layout = cc.Class({
         var row = 0;
         var containerResizeBoundary;
 
-        var maxHeightChildAnchor;
```
pandamicro, line 411: Didn't see anywhere to use `maxHeightChildAnchor.x` and it's not a good practice to create new Point object here, we don't need to, if only y is needed, please use `maxHeightChildAnchorY` otherwise, please use `maxHeightChildAnchorX` and `maxHeightChildAnchorY`

It's really much faster to use numbers than objects

cocos2d/core/components/CCLayout.js
```diff
@@ -534,7 +534,7 @@ var Layout = cc.Class({
         var secondMaxWidth = 0;
         var column = 0;
         var containerResizeBoundary;
-        var maxWidthChildAnchor;
```
pandamicro, line 537: Same remark as before

zilongshanren, line 537: ok, thanks for the tip.

cocos2d/core/components/CCLayout.js
zilongshanren, line 411: It's changed...

PR #665, opened 2016-05-12
fix memory leak in JSB

jareguo:
Re: cocos-creator/fireball#2950

@cocos-creator/engine-admins

### DIFF
--- a/cocos2d/core/CCNode.js
+++ b/cocos2d/core/CCNode.js
@@ -502,6 +502,14 @@ var Node = cc.Class({
         this._releaseAllActions();
 
         // Remove all listeners
+        if (CC_JSB && this._touchListener) {
+            this._touchListener.release();
+            this._touchListener = null;
+        }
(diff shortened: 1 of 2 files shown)


cocos2d/core/CCNode.js
```diff
@@ -502,6 +502,12 @@ var Node = cc.Class({
         this._releaseAllActions();
 
         // Remove all listeners
+        if (this._touchListener) {
+            this._touchListener.release();
```
pandamicro, line 506: 在设置一下 this._touchListener = null; 比较好

cocos2d/core/CCNode.js
```diff
@@ -1052,7 +1061,7 @@ var Node = cc.Class({
                     onTouchMoved: _touchMoveHandler,
                     onTouchEnded: _touchEndHandler
                 });
-                this._touchListener.retain();
+                    this._touchListener.retain();
```
pandamicro: 为啥多出来几个空格？

PR #689, opened 2016-05-18
fix scrollview layout as content issue

zilongshanren:
Re: cocos-creator/fireball#3037

-  When Layout's resizeMode is Container ，don't listen to 'size-changed' event
-  fix scrollview boundary calculate error

@cocos-creator/engine-admins

### DIFF
--- a/cocos2d/core/components/CCLayout.js
+++ b/cocos2d/core/components/CCLayout.js
@@ -357,7 +357,10 @@ var Layout = cc.Class({
             this.node.setContentSize(this._layoutSize);
         }
 
-        this.node.on('size-changed', this._resized, this);
+        if(this.resizeMode !== ResizeMode.CONTAINER) {
+            this.node.on('size-changed', this._resized, this);
+        }
+
         this.node.on('anchor-changed', this._doLayoutDirty, this);
         this.node.on('child-added', this._childrenAddOrDeleted, this);
         this.node.on('child-removed', this._childrenAddOrDeleted, this);
--- a/cocos2d/core/components/CCLayout.js
+++ b/cocos2d/core/components/CCLayout.js
@@ -357,9 +357,7 @@ var Layout = cc.Class({
             this.node.setContentSize(this._layoutSize);
         }
 
-        if(this.resizeMode !== ResizeMode.CONTAINER) {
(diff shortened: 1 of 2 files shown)


cocos2d/core/components/CCLayout.js
```diff
@@ -357,7 +357,10 @@ var Layout = cc.Class({
             this.node.setContentSize(this._layoutSize);
         }
 
-        this.node.on('size-changed', this._resized, this);
+        if(this.resizeMode !== ResizeMode.CONTAINER) {
```
zilongshanren: because in ResizeMode.Container, the layout will change node's size, so there is no need to cause layout to dirty...

cocos2d/core/components/CCScrollView.js
```diff
@@ -593,6 +593,11 @@ var ScrollView = cc.Class({
 
     _calculateBoundary: function() {
         if (this.content) {
+            //refresh content size
+            var layout = this.content.getComponent(cc.Layout);
+            if(layout) {
+                layout.lateUpdate();
```
jareguo, line 599: 干得漂亮

cocos2d/core/components/CCScrollView.js
```diff
@@ -1027,18 +1031,15 @@ var ScrollView = cc.Class({
...
-            if(cc.isValid(this.content)) {
-                this.content.on('size-changed', this._calculateBoundary, this);
-            }
             this.node.on('size-changed', this._calculateBoundary, this);
-            this._calculateBoundary();
         }
     },
 
+    start: function() {
+        this._calculateBoundary();
```
jareguo, line 1039: 那是否意味着只有 start 过后才能调用 scrollToXXX ？

zilongshanren, line 1039: 不会，调用任何 scrollToXX之前还会再调用一次 calculateBoundary

cocos2d/core/components/CCScrollView.js
zilongshanren, line 599: 感觉这个地方是不是还要判断一下 layout的 resize mode 是否是 Container 啊？

PR #738, opened 2016-05-26
spine should be visible after adding to existing node.

jareguo:
修复 1.1 的 bug，spine 动态添加到场景中的节点会渲染不出来

@cocos-creator/engine-admins


nantas:
cocos-creator/fireball#3171

PR #770, opened 2016-06-06
Fix scrollview scrollto api

zilongshanren:
Re: cocos-creator/fireball#3270

- fix scrollview `scrollToOffset` when content size is less than view size
- improve layout none mode anchor and size precision

@cocos-creator/engine-admins

### DIFF
--- a/cocos2d/core/components/CCScrollView.js
+++ b/cocos2d/core/components/CCScrollView.js
@@ -502,8 +502,19 @@ var ScrollView = cc.Class({
     scrollToOffset: function(offset, timeInSecond, attenuated) {
         var maxScrollOffset = this.getMaxScrollOffset();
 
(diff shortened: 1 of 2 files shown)


cocos2d/core/components/CCScrollView.js
```diff
@@ -715,19 +731,24 @@ var ScrollView = cc.Class({
 
         var scrollSize = this.node.getContentSize();
         var contentSize = this.content.getContentSize();
-        var bottomDeta = Math.abs(this._getContentBottomBoundary() - this._bottomBoundary);
-        var leftDeta = Math.abs(this._getContentLeftBoundary() - this._leftBoundary);
+        var bottomDeta = this._getContentBottomBoundary() - this._bottomBoundary;
+        bottomDeta = -bottomDeta;
```
jareguo, line 735: 这里取反的话，为什么不直接 `var bottomDeta = this._bottomBoundary - this._getContentBottomBoundary();`

zilongshanren, line 735: 是可以直接反过来减。。。

PR #774, opened 2016-06-07
fix error when deactivating detached RendererInSG

jareguo:
修复代码移除 Mask 节点时报错

@cocos-creator/engine-admins


nantas:
cocos-creator/fireball#3311

PR #781, opened 2016-06-09
Chrome on Android is supporting WebGL since v. 30

mbunger:
Enabling webGL on Android running newer version of Chrome. As stated here [webGL](http://googlechromereleases.blogspot.dk/2013/08/chrome-beta-for-android-update_22.html) Chrome on Android supports WebGL from version 30 and newer.

### DIFF
--- a/cocos2d/core/platform/CCSys.js
+++ b/cocos2d/core/platform/CCSys.js
@@ -308,7 +308,8 @@ sys.EDITOR_PAGE = 102;
 sys.EDITOR_CORE = 103;
 
 /**
- * BROWSER_TYPE_WECHAT
+ *
+ _WECHAT
  * @property {String} BROWSER_TYPE_WECHAT
  * @readOnly
  * @default "wechat"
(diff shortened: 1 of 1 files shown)


cocos2d/core/platform/CCSys.js
```diff
@@ -551,6 +551,8 @@ else {
             browserType = sys.BROWSER_TYPE_MAXTHON;
         else if (browserType === "opr")
             browserType = sys.BROWSER_TYPE_OPERA;
+        else if (browserType === "chrome")
+            browserType = sys.BROWSER_TYPE_CHROME;
```
pandamicro: There is no need to do this, the `typeReg2.exec(ua)` will assign directly `chrome` to `browserType`

cocos2d/core/platform/CCSys.js
```diff
@@ -708,6 +710,15 @@ else {
                         _supportWebGL = true;
                     }
                     break;
+                case sys.BROWSER_TYPE_CHROME:
+                    // Chrome on android supports WebGL from v. 30
+                    var browserVer = parseFloat(sys.browserVersion);
```
pandamicro: Can you take this out of switch ? Now two case are using the same variable, to avoid multiple declaration, it's better to take it out

pandamicro:
Thanks @mbunger

PR #811, opened 2016-06-17
fix bmfont and ttf shrink in editor

zilongshanren:
Re: cocos-creator/fireball#3071  https://github.com/cocos-creator/fireball/issues/3389

1. fix bmfont and ttf shrink font size doesn't restore to the original size when changing string.
2. fix bmfont shrink/clamp/resize_height issue when first added to the editor.
3. fix system font markup control shouldn't be unmarked when it's already use system font.

@cocos-creator/engine-admins


zilongshanren:
thanks

PR #813, opened 2016-06-20
add 'crop' mode to sprite

xuewuli:
Re: cocos-creator/fireball#

add 'crop' mode to sprite

@cocos-creator/engine-admins

### DIFF
--- a/cocos2d/core/sprites/CCScale9Sprite.js
+++ b/cocos2d/core/sprites/CCScale9Sprite.js
@@ -154,6 +154,115 @@ var simpleQuadGenerator = {
     }
 };
 
+var cropQuadGenerator = {
+    _rebuildQuads_base: function (spriteFrame, contentSize, colorOpacity, isTrimmedContentSize) {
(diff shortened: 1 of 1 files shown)


pandamicro:
What's crop mode ? can you be more specific ?

xuewuli:
sorry，it's should be called 'center crop'.
https://developer.android.com/reference/android/widget/ImageView.ScaleType.html

pandamicro:
@nantas what do you think about this feature ?

@xuewuli We have recently refactored the quad data structure in cc.Scale9Sprite, to have better performance and less memory usage, you will need to store the vertices and uvs data into sprite's `_vertices` and `_uvs` data, please refer to the latest version of Scale9Sprite implementation

nantas:
Recently I found this mode useful when studying a project made by old cocos2d-html5, in the old engine you can set a boundingBox to a raw texture to manually 'crop' it during runtime, and it's not possible in Cocos Creator.

So I guess we could use some help here.

PR #879, opened 2016-07-06
stop scroll when scrollview is disabled

zilongshanren:
Re: cocos-creator/fireball#3457

- 3457

@cocos-creator/engine-admins


cocos2d/core/components/CCScrollBar.js
```diff
@@ -325,6 +327,11 @@ var Scrollbar = cc.Class({
         }
     },
 
+    reset: function() {
```
jareguo, line 330: 这个 reset 有用吗？

zilongshanren, line 330: 有用呀。。。只是 reset state 没有 reset position

PR #912, opened 2016-07-14
only do layout when layout is active in scrollview

zilongshanren:
Re: cocos-creator/fireball#3626

- 

@cocos-creator/engine-admins

### DIFF
--- a/cocos2d/core/components/CCScrollView.js
+++ b/cocos2d/core/components/CCScrollView.js
@@ -756,7 +756,7 @@ var ScrollView = cc.Class({
         if (this.content) {
             //refresh content size
             var layout = this.content.getComponent(cc.Layout);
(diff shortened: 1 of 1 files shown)


cocos2d/core/components/CCScrollView.js
```diff
@@ -756,7 +756,7 @@ var ScrollView = cc.Class({
         if (this.content) {
             //refresh content size
             var layout = this.content.getComponent(cc.Layout);
-            if(layout) {
+            if(layout && layout.activeInHierarchy) {
```
jareguo: activeInHierarchy 应该是 node 才有的 API，组件的是 enabledInHierarchy

zilongshanren: ok

PR #1033, opened 2016-08-12
update ccclass editor attributes

nantas:


nantas:
cocos-creator/fireball#2160

PR #1038, opened 2016-08-15
fix auto-release-utils error JSB not cc.Audio

knoxHuang:
Re: cocos-creator/fireball#3993

-  JSB 中不需要检测 cc.Audio

@cocos-creator/engine-admins


jareguo:
LGTM

PR #1277, opened 2016-11-10
Solve the warning in TMXXMLParser.

natural-law:
@cocos-creator/engine-admins


feijing566:
@natural-law I think Line870 should solve too

jareguo:
😂

PR #1500, opened 2017-02-16
Improve the initialization logic of Label with BitmapFont.

natural-law:
Re: cocos-creator/fireball#5234

 * Improve the initialization logic of Label with BitmapFont.

@cocos-creator/engine-admins

### DIFF
--- a/cocos2d/core/components/CCLabel.js
+++ b/cocos2d/core/components/CCLabel.js
@@ -375,7 +375,12 @@ var Label = cc.Class({
                     var isAsset = value instanceof cc.Font;
 
                     if (this.font instanceof cc.BitmapFont) {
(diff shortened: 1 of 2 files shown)


cocos2d/core/components/CCLabel.js
```diff
@@ -469,7 +469,12 @@ var Label = cc.Class({
 
         var sgNode;
         if (this.font instanceof cc.BitmapFont) {
```
jareguo, line 476: 上面还有一个 `this.font instanceof cc.BitmapFont`，那边不用做相同判断吗？

natural-law: 哦。我看到了。在 font 属性的 set 处理中也使用了 spriteFrame 属性。我也改一下吧

natural-law: 已修改。

PR #1619, opened 2017-04-27
fix cc.Integer

jareguo:
 * 修复 `@property(cc.Integer)` 会引起报错的 bug

@cocos-creator/engine-admins


jareguo:
cc @nantas

PR #1752, opened 2017-06-23
set transform updated for native

2youyou2:
Re: cocos-creator/fireball#5922

 * set transform updated for native

需要先合并 https://github.com/cocos-creator/cocos2d-x-lite/pull/670

@cocos-creator/engine-admins


pandamicro (approved):
等 lite 仓库合并

PR #1941, opened 2017-08-29
should ignore animation playOnLoad if already play an animation

2youyou2:
Re: cocos-creator/fireball#5869

 * should ignore animation playOnLoad if already play an animation

@cocos-creator/engine-admins

### DIFF
--- a/cocos2d/core/components/CCAnimation.js
+++ b/cocos2d/core/components/CCAnimation.js
@@ -169,10 +169,12 @@ var Animation = cc.Class({
     },
 
     start: function () {
-        var isPlaying = this._animator && this._animator.isPlaying;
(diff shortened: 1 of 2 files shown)


cocos2d/core/components/CCAnimation.js
```diff
@@ -169,7 +169,8 @@ var Animation = cc.Class({
     },
 
     start: function () {
-        if (!CC_EDITOR && this.playOnLoad && this._defaultClip) {
+        var isPlaying = this._animator && this._animator.isPlaying;
+        if (!CC_EDITOR && !isPlaying && this.playOnLoad && this._defaultClip) {
```
jareguo: 把 var isPlaying = this._animator && this._animator.isPlaying; 放到原来的 if 里再判断一次吧，提前被用户 play 的几率远远小于默认 play 的几率。

PR #2068, opened 2017-10-25
fix potential memory leak when node detached from parent without cleanup

jareguo:


cocos2d/core/utils/scene-graph-helper.js
```diff
@@ -27,15 +27,23 @@
...
-                parent.removeChild(node);
+                parent.removeChild(sgNode);
             }
-            node.release();
-            if (this._sgNode._entity) {
-                this._sgNode._entity = null;
+            else {
+                // cleanup was skipped when its node was detaching
+                if (CC_JSB) {
+                    sgNode.cleanup();
```
jareguo, line 39: C++ 本来就是递归调用

PR #2077, opened 2017-10-28
working on texture asset

jareguo:
Re: cocos-creator/fireball#6082

@cocos-creator/engine-admins


jareguo:
差异都混在一起了…… 合回 master 再 review 吧

PR #2146, opened 2017-11-29
Fix running in editor issue

pandamicro:
Re: cocos-creator/fireball#6639


cocos2d/core/components/CCRenderComponent.js
```diff
@@ -45,8 +45,12 @@ var _defaultVertexFormat = new gfx.VertexFormat([
 var RenderComponent = cc.Class({
     name: 'RenderComponent',
     extends: Component,
+
+    editor: CC_EDITOR && {
+        executeInEditMode: true
+    },
     
-    ctor: function() {
+    ctor: function () {
```
jareguo, line 53: You can just use `ctor () { ... }`

PR #2148, opened 2017-11-29
 background color should be (0,0,0,0), or will cause white border

2youyou2:
Changelog:
 * background color should be (0,0,0,0), or will cause white border


pandamicro:
麻烦尽量用 rebase，重新提交一下

PR #2192, opened 2017-12-18
fix getDisplayedColor api

knoxHuang:
Re: cocos-creator/fireball#

Changelog:
 * 修复 getDisplayedColor api


jareguo:
fix https://github.com/cocos-creator/creator-api-docs/issues/14

PR #2214, opened 2017-12-26
Fix for cc.Slider

Cerber-Ursi:
Found and easily fixed a bug with touch and handle positions mismatch in Slider.

### DIFF
--- a/cocos2d/core/components/CCSlider.js
+++ b/cocos2d/core/components/CCSlider.js
@@ -206,7 +206,7 @@ var Slider = cc.Class({
         if (!this.handle) { return; }
         var handlelocalPos;
         if (this.direction === Direction.Horizontal) {
-            handlelocalPos = cc.p((this.progress - this.node.anchorX) * this._maxRange * 2, 0)
+            handlelocalPos = cc.p((this.progress - this.node.anchorX) * this._maxRange * 2, 0);
         }
         else {
             handlelocalPos = cc.p(0, (this.progress - this.node.anchorY) * this._maxRange * 2);
@@ -224,6 +224,7 @@ var Slider = cc.Class({
                 this._maxRange = this.node.height / 2 - this.handle.node.height * this.handle.node.anchorY;
             }
         }
+    }
 
 });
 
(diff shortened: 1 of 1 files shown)


cocos2d/core/components/CCSlider.js
```diff
@@ -207,14 +206,24 @@ var Slider = cc.Class({
...
-            handlelocalPos = cc.p(0, -this.node.height * this.node.anchorY + this.progress * this.node.height);
+            handlelocalPos = cc.p(0, (this.progress - this.node.anchorY) * this._maxRange * 2);
         }
         var worldSpacePos = this.node.convertToWorldSpaceAR(handlelocalPos);
         this.handle.node.position = this.handle.node.parent.convertToNodeSpaceAR(worldSpacePos);
-    }
+    },
+    
+    _setMaxRange: function() {
+        if (!this._maxRange) {
```
knoxHuang: _maxRange can only be set once ? 
will _maxRange be updated if I dynamically modify the size of node or handle?

Cerber-Ursi: No, it's set only once per slider for now, thanks for paying attention to this. Perhaps it'll be more appropriate just to remove this conditional and update _maxRange on every progress update (almost as it is in master).

knoxHuang: Can you update the code? Thank you very much. 👍

knoxHuang:
I don't think this is a bug. For example, the volume control in Windows system is like this. but it's possible to add an attribute to let users choose which way to express.

tim 20171227180942

jareguo:
Thanks for the PR, but we can not merge now. Because the PR breaks the compatibility for old projects. If we need to change the feature, please make sure the behavior and representation of user's old scene will not change.

For your new request, I think you can add a new property called something like "Clamp"?  Clamp is false by default. When it turns on, the range will change.

Cerber-Ursi:
I'm speaking about another moment. Look at these three screenshots - they were all taken while dragging the slider. You can see that the handle isn't moving the same way as the cursor does.
screen1
screen2
screen3
Although, I see your point. It seems that we must fix this issue another way - changing not the handle positioning, but the cursor position checks.

jareguo:
Agree, we should fix the cursor first.

jareguo:
@knoxHuang please fix the cursor position.

jareguo:
I will close this due to the fix in https://github.com/cocos-creator/engine/pull/2245. The PR fixed the cursor position.
Thanks for your help!

PR #2234, opened 2018-01-03
CCWebView.destroy() should have a return value

simon129:
CCWebView.destroy() inherent from CCObject, it's destroy() should return true or false.
Add return value to improve creator.d.ts


jareguo:
Thanks!

PR #2271, opened 2018-01-18
Resolve a FIXME in AramatureDisplay.js, CCFactory is a singlton for JSB either now.

dumganhar:


cocos2d/deprecated.js
```diff
@@ -682,4 +682,8 @@ if (CC_DEV) {
         },
     });
 
+    if (dragonBones !== undefined) {
```
jareguo: 抱歉说错，这边要用 typeof dragonBones !== 'undefined'，否则如果裁剪掉 dragonBones 的话这个变量会报错

jareguo (changes requested):

PR #2285, opened 2018-01-26
dragonBones EventObject Type issue.

evan-hwang:
hi, i'm cocos lover.
    
dragonBones EventObject Type is string. so there is type isuue. it is as in the following

    [code]
    this._armatureDisplay.addEventListener(dragonBones.EventObject.COMPLETE, this._animationEventHandler, this);

    [alert message]
    [ts] Argument of type 'string' is not assignable to parameter of type 'EventObject'.
    (property) dragonBones.EventObject.COMPLETE: string

    so, i think EventObject need dragonBones.EventObject Type.
    plz comment to my oppinion.

Re: cocos-creator/fireball#

Changelog:
 *

### DIFF
--- a/extensions/dragonbones/lib/dragonBones.d.ts
+++ b/extensions/dragonbones/lib/dragonBones.d.ts
@@ -1598,55 +1598,55 @@ declare namespace dragonBones {
          * 动画开始。
          * @version DragonBones 4.5
          */
-        static START: string;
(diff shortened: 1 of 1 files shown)


jareguo:
Thanks for the PR. But we can't merge it. This issue is already fixed in #2284, we changed the definition for addEventListener.

PR #2289, opened 2018-01-29
Reverts dragonBones CCFactory change to avoid assert.

dumganhar:


dumganhar:
Send to wrong branch, close this one.

PR #2341, opened 2018-02-16
Fix types of CCPolygonCollier and CCPhysicsChainCollider

nshmura:
Types of `PolygonCollider` and `PhysicsChainCollider` in creator.d.ts seemds to be wrong.
I thank `points` should be Vec2[], not [Vec2].

Changelog:
 * Change types of points in CCPolygonCollier and CCPhysicsChainCollider to Vec2[].


jareguo:
Thanks!

PR #2356, opened 2018-02-24
Event manager set correct node to dirty when change sibling index

pandamicro:
Re: cocos-creator/fireball#6977


cocos2d/core/CCNode.js
```diff
@@ -1040,7 +1040,7 @@ var Node = cc.Class({
             for (; i < len; i++) {
                 sibling = siblings[i]._sgNode;
                 sibling._arrivalOrder = i;
-                eventManager._setDirtyForNode(sibling);
+                eventManager._setDirtyForNode(sibling[i]);
```
jareguo: 不对吧？

PR #2365, opened 2018-02-26
fixed label assembler bug

fusijie:
fixed label assembler bug.


pandamicro:
https://github.com/cocos-creator/fireball/issues/6988

pandamicro (approved):

PR #2377, opened 2018-02-28
fixed canvas-renderer-walker exports bug.

fusijie:
fixed canvas-renderer-walker exports bug.


pandamicro:
https://github.com/cocos-creator/fireball/issues/6995

PR #2422, opened 2018-03-12
fixed richText

fusijie:
remove setPositionX setPositionY & setSkew.


pandamicro:
For https://github.com/cocos-creator/fireball/issues/7022

PR #161, opened 2015-12-30
only doLayout when count of children > 0

zilongshanren:

PR #2498, opened 2018-04-03
Merge runtime builtin scripts.

dumganhar:


cocos2d/core/platform/instantiate-jit.js
```diff
@@ -162,7 +162,14 @@ function equalsToDefault (def, value) {
             (def.constructor === Object && value.constructor === Object)
         ) {
             try {
-                return JSON.stringify(def) === JSON.stringify(value);
+                if (CC_JSB) {
+                    //FIXME: JSON.stringify will take lots of time for compare object at runtime on JSB.
+                    // We need to consider a new way for checking whether two objects are equal.
```
dumganhar: @jareguo needs to fix this.

pandamicro (approved):

cocos2d/core/platform/instantiate-jit.js
jareguo: Okey, then just use
```
return Array.isArray(def) && Array.isArray(value) && def.length === 0 && value.length === 0;
```
in line #161:

dumganhar: But this is only for checking whether the length of two arrays is equal.
How about two array length is equal but their contents are not?

jareguo: so length should be 0

PR #2531, opened 2018-04-13
use native object to improve performance

minggo:


pandamicro:
https://github.com/cocos-creator/fireball/issues/7131

PR #2571, opened 2018-04-20
window.__gl -> window.__ccgl, window.canvas -> window.__cccanvas. Touch logic is moved to EventTarget

dumganhar:


pandamicro:
Re: cocos-creator/fireball#7005

PR #2585, opened 2018-04-24
adjust codes to unify js_loadImage binding codes

minggo:


minggo:
This PR works with https://github.com/cocos-creator/cocos2d-x-lite/pull/1210

pandamicro (approved):

PR #2587, opened 2018-04-24
add record pipe to record game for android instant apps

wuzhiming:
add record pipe to record game for android instant apps


wuzhiming:
move to editor static/simulator

cocos2d/core/load-pipeline/asset-record-pipe.js
```diff
@@ -0,0 +1,109 @@
+/****************************************************************************
+ Copyright (c) 2016 Chukong Technologies Inc.
```
jareguo, line 2: 去掉 16 年这行吧

PR #2596, opened 2018-04-25
fix audioSource destroy bug

PPpro:
Re: cocos-creator/fireball#

audioSource 组件在没有play 之前就进行销毁操作，这时候 _element === null
直接调用 this.element.destroy()  会报错。这里加多一个判断

@jareguo


jtianling (commented):
the mean is not so clear.

should be 
```
if (this._element)  this._element.destory()
```

if we surely don't care about why a element is null, I advice add a safeDestory function, could use it in many places.

PPpro:
yeah, thanks for your advice. I'll consider it.

PR #2602, opened 2018-04-26
optimization ScrollView scale changed

knoxHuang:
Re: cocos-creator/fireball#

Changelog:
 * 优化 ScrollView scale changed


jareguo:
性能优化请放到 next 分支

PR #2628, opened 2018-05-03
fix typo in RenderTextureTest

unace:
It looks like a bug.


pandamicro:
Actually we stopped using visual tests, but thanks anyway

PR #2716, opened 2018-05-22
Support setUniform when dirty

pandamicro:
Re: cocos-creator/fireball#7131


pandamicro:
https://github.com/cocos-creator/fireball/issues/7297

PR #2758, opened 2018-05-29
fixed error when switch scene in editor

2youyou2:
Re: cocos-creator/fireball#7413

Changelog:
 * fixed error when switch scene in editor

需要 @VisualSJ  先修改场景没有调用 onRestore 的问题再合

### DIFF
--- a/cocos2d/core/CCNode.js
+++ b/cocos2d/core/CCNode.js
@@ -1184,6 +1184,7 @@ var Node = cc.Class({
         // Recycle math objects
         mathPools.mat4.put(this._matrix);
         mathPools.mat4.put(this._worldMatrix);
+        this._matrix = this._worldMatrix = null;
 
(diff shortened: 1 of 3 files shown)


cocos2d/core/CCNode.js
```diff
@@ -2933,6 +2934,19 @@ var Node = cc.Class({
     onRestore: CC_EDITOR && function () {
         this._onRestoreBase();
 
+        /*
+         * TODO: Refine this code after completing undo/redo 2.0.
+         * The node is destroy when entering prefab edit mode, 
+         * but it is still used when exiting prefab edit mode, 
+         * so the built-in properties of the node need to be restored.
+        */
```
jareguo: The node will be destroyed when deleting in the editor,
but it will be reserved and reused for undo.

pandamicro (approved):

PR #2761, opened 2018-05-29
Fix video player inline play issue in x5

pandamicro:
Re: cocos-creator/fireball#7410


pandamicro:
参考资料：https://x5.tencent.com/tbs/guide/video.html

PR #2783, opened 2018-05-31
fixed CCNode event

2youyou2:
Re: cocos-creator/fireball#7400

Changelog:
 * fixed CCNode event

### DIFF
--- a/cocos2d/core/event/event-target.js
+++ b/cocos2d/core/event/event-target.js
@@ -198,8 +198,13 @@ proto.once = function (type, callback, target) {
  * @method emit
  * @param {String} type - event type
  * @param {*} [detail] - whatever argument the message needs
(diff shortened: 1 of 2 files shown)


pandamicro (approved):

cocos2d/core/event/event-target.js
```diff
@@ -209,7 +214,7 @@ proto.emit = function (type, detail) {
         event.detail = detail;
         // Event.AT_TARGET
         event.eventPhase = 2;
-        event.target = event.currentTarget = this;
+        event.target = event.currentTarget = target;
```
pandamicro: 这里要兼容以前的用法，target || this

pandamicro (changes requested):

pandamicro (approved):

PR #2941, opened 2018-07-04
Fix render engine issue in strict mode

pandamicro:


pandamicro:
For https://github.com/cocos-creator/fireball/issues/7802

PR #2945, opened 2018-07-05
Fix editor cannot open scene while inner asset missing

pandamicro:
Re: cocos-creator/fireball#7738

### DIFF
--- a/cocos2d/core/load-pipeline/uuid-loader.js
+++ b/cocos2d/core/load-pipeline/uuid-loader.js
@@ -173,7 +173,9 @@ function loadDepends (pipeline, item, asset, depends, callback) {
         if (CC_EDITOR && missingAssetReporter) {
             missingAssetReporter.reportByOwner();
         }
-        callback(errors, asset);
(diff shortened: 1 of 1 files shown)


cocos2d/core/load-pipeline/uuid-loader.js
```diff
@@ -173,7 +173,9 @@ function loadDepends (pipeline, item, asset, depends, callback) {
         if (CC_EDITOR && missingAssetReporter) {
             missingAssetReporter.reportByOwner();
         }
-        callback(errors, asset);
+        // Emit dependency errors in runtime, but not in editor,
+        // because editor need to open the scene / prefab to let user fix missing asset issues
+        callback(CC_EDITOR ? null : errors, asset);
```
jareguo: 最好是
```
if (CC_EDITOR && missingAssetReporter) {
  missingAssetReporter.reportByOwner();
  callback(null, asset);
}
else {
  callback(errors, asset);
}
```

pandamicro: 哦 ok

PR #3058, opened 2018-08-08
Fix spine issues for v2.0

pandamicro:
Re: cocos-creator/fireball#8106


pandamicro:
https://github.com/cocos-creator/fireball/issues/8116

PR #3059, opened 2018-08-09
Modify the create.d.ts

ColinCollins:
1. fixed the problem that CCDebug doesn't contain cc.debug.DebugMode.
2. adjust the marco file, forbid the user to use the cc.KEY straightly.
3. include others change.


PPpro:
关联 issue:
https://github.com/cocos-creator/fireball/issues/8108

pandamicro (approved):

PR #3066, opened 2018-08-10
fix editbox font color

PPpro:
Re: cocos-creator/fireball#8135

- 修复 editBox 字体颜色设置 失效
- 修复 color fromHSV() 计算失误

@cocos-creator/admins


PPpro:
https://github.com/cocos-creator/engine/issues/3065

PR #3085, opened 2018-08-14
fix sprite update renderData onEnable

PPpro:
Re: https://github.com/cocos-creator/fireball/issues/8114

@cocos-creator/admins

### DIFF
--- a/cocos2d/core/components/CCSprite.js
+++ b/cocos2d/core/components/CCSprite.js
@@ -444,7 +444,7 @@ var Sprite = cc.Class({
         }
         
         this._updateAssembler();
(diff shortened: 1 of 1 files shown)


cocos2d/core/components/CCSprite.js
```diff
@@ -445,6 +445,8 @@ var Sprite = cc.Class({
         
         this._updateAssembler();
         this._renderData.uvDirty = true;
```
pandamicro: 这句删掉

pandamicro (changes requested):

pandamicro (approved):

PR #3121, opened 2018-08-20
cc.Audio.stop logic fix, pause should be invoked before set currentTime = 0

dumganhar:
If `currentTime = 0` is set before `pause`, audio may be played from beginning at a very short time before `pause` since audio API is asynchronous.


pandamicro:
多谢码神

PR #3140, opened 2018-08-23
modify the note  for periodic function

ColinCollins:
Re:
http://forum.cocos.com/t/topic/65134
image

__Engine__:
image

### DIFF
--- a/cocos2d/core/components/CCComponent.js
+++ b/cocos2d/core/components/CCComponent.js
@@ -217,8 +217,10 @@ var Component = cc.Class({
     // We provide Pre methods, which are called right before something happens, and Post methods which are called right after something happens.
 
     /**
-     * !#en Update is called every frame, if the Component is enabled.
-     * !#zh 如果该组件启用，则每帧调用 update。
+     * !#en Update is called every frame, if the Component is enabled.<br/>
+     * This is a lifecycle method. It may not be implemented in the super class. You can only call its super class method inside it. It should not be called manually elsewhere.
+     * !#zh 如果该组件启用，则每帧调用 update。<br/>
+     * 该方法为生命周期方法，父类未必会有实现。并且你只能在该方法内部调用父类的实现，不可在其它地方直接调用该方法。
      * @method update
      * @param {Number} dt - the delta time in seconds it took to complete the last frame
      * @protected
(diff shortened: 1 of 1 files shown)


cocos2d/core/components/CCComponent.js
```diff
@@ -218,7 +218,9 @@ var Component = cc.Class({
 
     /**
      * !#en Update is called every frame, if the Component is enabled.
+     * PS: This is a periodic function, And it can only call the implementation of the parent class inside the method, it can't be called directly in other places.
      * !#zh 如果该组件启用，则每帧调用 update。
+     * 该方法为生命周期方法，父类未必会有实现。并且它只能在该方法内部调用父类的实现，不可在其它地方直接调用该方法。
```
jareguo: 并且它只能在该方法内部调用父类的实现，不可在其它地方直接调用该方法。
最好改成
并且你只能在该方法内部调用它的父类实现，不可在其它地方直接调用该方法。

cocos2d/core/components/CCComponent.js
```diff
@@ -217,17 +217,21 @@ var Component = cc.Class({
     // We provide Pre methods, which are called right before something happens, and Post methods which are called right after something happens.
 
     /**
-     * !#en Update is called every frame, if the Component is enabled.
-     * !#zh 如果该组件启用，则每帧调用 update。
+     * !#en Update is called every frame, if the Component is enabled.<br/>
+     * PS: This is a periodic function, And it can only call the implementation of the parent class inside the method, it can't be called directly in other places.
```
jareguo: 请让育芳帮忙看下英文翻译

xunyi0: The method is a lifecycle method, and the parent class may not be implemented. And you can only call its parent class implementation within the method, and you cannot call the method directly elsewhere.

ColinCollins: 学到了学到了，优秀

cocos2d/core/components/CCComponent.js
```diff
@@ -247,9 +251,11 @@ var Component = cc.Class({
     /**
      * !#en
      * When attaching to an active node or its node first activated.
-     * onLoad is always called before any start functions, this allows you to order initialization of scripts.
+     * onLoad is always called before any start functions, this allows you to order initialization of scripts.<br/>
+     * PS: The method is a lifecycle method, and the base class may not implemente it. You can only call its base class implementation within the method, and you cannot call the method directly elsewhere.
```
jareguo: 去掉 PS
This is a lifecycle method. It may not be implemented in the super class. You can only call its super class method inside it. It should not be called manually elsewhere.

PR #3164, opened 2018-08-30
fix render error when setting 30fps on JSB

drelaptop:
Re: cocos-creator/fireball#8182

关联相关修复的 PR，https://github.com/cocos-creator/engine/pull/2960


drelaptop:
修改后之前 https://github.com/cocos-creator/fireball/issues/7859 出问题的测试例，表现正常。please review and merge @pandamicro

PR #3250, opened 2018-09-14
scene-graph migration

YunHsiao:
Put 2D Node aside for now.
3D Node inherits BaseNode.
Use mixin for EventTarget.
No breaking change for any part of 2D infrastructures.

### TODOs for other parts
* reset the `_hasChanged` flag for all nodes every frame in director.
* adapt the following minor API changes:

| Creator Node 	| engine-3d Entity (deprecated) |
|:------------:	|:----------------:|
|uuid		|id|
|insertChild	|insertAt|
|addChild	|append|
|set active	|activate|
|set active	|deactivate|
|getComponent	|getComp|
|getComponents	|getComps|
|getComponentsInChildren	|getCompsInChildren|
|addComponent	|addComp|

### DIFF
--- a/cocos/scene-graph/base-node.js
+++ b/cocos/scene-graph/base-node.js
@@ -26,7 +26,6 @@
 
 import * as js from '../core/utils/js';
 import IdGenerater from '../core/utils/id-generater';
-import eventManager from '../core/platform/event-manager/CCEventManager';
 import CCObject from '../core/data/object';
 import _decorator from '../core/data/class-decorator';
 const { ccclass, property } = _decorator;
@@ -341,6 +340,9 @@ export default class BaseNode extends CCObject {
     getParent () {
         return this._parent;
     }
+    get parent() {
+        return this._parent;
+    }
 
     /**
      * !#en Set parent of the node.
@@ -365,14 +367,13 @@ export default class BaseNode extends CCObject {
         }
         this._parent = value || null;
 
-        this._onSetParent(value);
+        this._onSetParent(oldParent);
 
         if (value) {
             if (CC_DEBUG && (value._objFlags & Deactivating)) {
                 cc.errorID(3821);
             }
             this._level = value._level + 1;
-            eventManager._setDirtyForNode(this);
             value._children.push(this);
             value.emit && value.emit(CHILD_ADDED, this);
         }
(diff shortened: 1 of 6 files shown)


pandamicro (approved):

cocos/scene-graph/base-node.js
```diff
@@ -1034,7 +1031,8 @@ export default class BaseNode extends CCObject {
         }
     }
 
-    _onSetParent (/*value*/) {}
+    cleanup () {}
+    _onSetParent (/*oldParent*/) {}
```
dabingnn, line 1035: _onSetParent seems confusing

cocos/scene-graph/node.js
```diff
@@ -1,3046 +1,399 @@
...
- * Cocos Creator 场景中的所有节点类。<br/>
- * 支持的节点事件，请参阅 {{#crossLink "Node.EventType"}}{{/crossLink}}。
- * @class Node
- * @extends _BaseNode
- */
- @ccclass("cc.Node")
+@ccclass('cc.Node')
+@mixins(EventTarget)
 export default class Node extends BaseNode {
+    // local transform
```
dabingnn, line 14: seems some more info needs to figure out, property is serialized, a cache, and could be synced to editor or not

cocos/scene-graph/base-node.js
YunHsiao: engine-3d counterpart is `_onParentChanged`, might be a little bit better?

cocos/scene-graph/node.js
YunHsiao: for now all members are preceded with `_`, which makes them private and will not be shown on the editor. more specific details about what and how to expose to the editor could be discuss late after the integration, perhaps?

cocos/3d/assets/audio-clip.js
```diff
@@ -71,7 +71,7 @@ export const AudioSourceType = {
 @mixins(EventTarget)
 export class AudioClip extends Asset {
   /**
-   * @type {AudioSource}
```
shrinktofit: There's a ```@typedef``` annotation of ```AudioSource``` right above defination of AudioClip

PR #3256, opened 2018-09-17
graphics linear gradient basic support

skaraman:
Re: cocos-creator/fireball#

Changelog:
 * added gradient support for 2 colors provided in HEX
 ```
	path.size = {
		x1: 0,
		x2: 10,
		y1: 0,
		y2: 10,
		width: 10,
		height: 10,
		diagonal: 14.142135623730951,
                0: 0
	}
	path.gradientColors = {
		color1: '#ffffff',
		color2: '#000000'
	}
	path.gradientDirection = {
		x: 'x1',
		y: 'y2'
	}
	path.gradient()
```

### DIFF
--- a/cocos2d/core/graphics/graphics.js
+++ b/cocos2d/core/graphics/graphics.js
@@ -6,11 +6,11 @@
 
  Permission is hereby granted, free of charge, to any person obtaining a copy
  of this software and associated engine source code (the "Software"), a limited,
-  worldwide, royalty-free, non-assignable, revocable and non-exclusive license
+	worldwide, royalty-free, non-assignable, revocable and non-exclusive license
  to use Cocos Creator solely to develop games on your target platforms. You shall
-  not use Cocos Creator software for developing other software or tools that's
-  used for developing games. You are not granted to publish, distribute,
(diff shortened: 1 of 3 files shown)


2youyou2:
Hi, @skaraman : 
  Thanks for your contribution. You did a great work.
  But i think the better solution to implement the graphics gradient effect is to create a graphics gradient shader. Since we have not implement the material system yet, maybe we can try again after the material system completed.

skaraman:
Hi thanks!

I tried using shader and material but the shader script I found needed `a_uv0` to be defined as an attribute in the texture and it only had `a_position`. I don't know enough about shaders and materials to fix or work around that.  http://forum.cocos.com/t/2-0-2-cc-graphics/66182/1

also as a side note this implementation would work in canvas vs shaders requiring webGL

pandamicro:
Hi, @skaraman we have already disabled canvas rendering mode in v2.0.
So please stay tuned for the future release of material system APIs, I will close this pr as its current implementation doesn't suit our future plan, thanks for your contribution though, it's deeply appreciated

PR #3341, opened 2018-09-29
add video player source for JSB adapter.

xianyinchen:
need for JSB adapter.


pandamicro:
https://github.com/cocos-creator/2d-tasks/issues/639

PR #3374, opened 2018-10-12
fix camera transform in ortho mode

YunHsiao:
specifically, `screenToWorld` and `screenPointToRay`.

note that in `screenToWorld`,
the `screenPos.z` used to be ill-defined (in the range of something like [0, this._far]).

now it's in the range of [0, 1],
represents exactly the depth into the view frustum(cuboid), i.e. from near to far plane.

### DIFF
--- a/cocos/renderer/scene/camera.js
+++ b/cocos/renderer/scene/camera.js
@@ -1,6 +1,6 @@
 // Copyright (c) 2017-2018 Xiamen Yaji Software Co., Ltd.
 
-import { color4, vec3, mat4 } from '../../core/vmath';
+import { color4, vec3, mat4, lerp } from '../../core/vmath';
 import { ray } from '../../3d/geom-utils';
 import enums from '../enums';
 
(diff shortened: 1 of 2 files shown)


gameall3d:
没有问题~

cocos/renderer/scene/camera.js
```diff
@@ -357,59 +364,33 @@ export default class Camera {
...
 
       // transform to world
       vec3.transformMat4(out, out, _matInvViewProj);
 
-      //
+      // lerp to depth z
       this._node.getWorldPosition(_tmp_v3);
-      vec3.lerp(out, _tmp_v3, out, screenPos.z / this._far);
+
+      vec3.lerp(out, _tmp_v3, out, lerp(this._near, this._far, screenPos.z));
```
dabingnn: seems should be ` vec3.lerp(out, _tmp_v3, out, lerp(0, this._far, screenPos.z));`, coz `_tmp_v3` is the position of camera not a point on the near plane.

YunHsiao:
@2youyou2 这些改动涉及的问题2d应该也有, 可以搬一下~

PR #3382, opened 2018-10-15
Modify the node with opacity 0 to skip rendering.

caryliu1999:
Re: cocos-creator/2d-tasks#381

Changes:
Add opacity judgment in render-flow to make the node with opacity 0 skip rendering.

### DIFF
--- a/cocos2d/core/renderer/render-flow.js
+++ b/cocos2d/core/renderer/render-flow.js
@@ -97,7 +97,7 @@ _proto._children = function (node) {
     let children = node._children;
     for (let i = 0, l = children.length; i < l; i++) {
         let c = children[i];
-        if (!c._activeInHierarchy || c.opacity === 0) continue;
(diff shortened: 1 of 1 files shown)


cocos2d/core/renderer/render-flow.js
```diff
@@ -97,7 +97,7 @@ _proto._children = function (node) {
     let children = node._children;
     for (let i = 0, l = children.length; i < l; i++) {
         let c = children[i];
-        if (!c._activeInHierarchy) continue;
+        if (!c._activeInHierarchy || c.opacity === 0) continue;
```
pandamicro: 用 _opacity

pandamicro (changes requested):

caryliu1999:
Has been modified to access private properties.

pandamicro (approved):

PR #3390, opened 2018-10-16
fixed the pageView can trigger the began event

ColinCollins:
Re: https://github.com/cocos-creator/engine/issues/3212

I have modified part of those codes, but this part of logic still keep a little confusion.
There is the PR of this part.
https://github.com/cocos2d/cocos2d-x/pull/18054
gif


knoxHuang (approved):

jareguo:
修复 PageView 的 began 事件不触发的问题

PR #3401, opened 2018-10-18
gizmos adaptation

YunHsiao:
for cocos-creator/3d-tasks#123.
migrated the gizmos project into 3d branch.

engine specific approaches during the migration:

* a standalone effect for gizmos(using unlit shader), which makes it look nicer and settles the depth problem.
* extended the `ray_triangle` test to work for both single and double sided triangles, depends on the flag being passed in.

also, fixed `worldToScreen` transform.

### DIFF
--- a/cocos/3d/geom-utils/intersect.js
+++ b/cocos/3d/geom-utils/intersect.js
@@ -40,6 +40,7 @@ let line_plane = (function () {
 
 /**
  * ray-triangle intersect
+ * based on http://fileadmin.cs.lth.se/cs/Personal/Tomas_Akenine-Moller/raytri/
  *
  * @param {ray} ray
  * @param {triangle} triangle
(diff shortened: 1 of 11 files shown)


cocos/3d/geom-utils/intersect.js
```diff
@@ -52,32 +53,26 @@ let ray_triangle = (function () {
...
   let qvec = vec3.create(0, 0, 0);
 
-  return function (ray, triangle) {
+  return function (ray, triangle, doubleSided) {
     vec3.sub(ab, triangle.b, triangle.a);
     vec3.sub(ac, triangle.c, triangle.a);
 
     vec3.cross(pvec, ray.d, ac);
     let det = vec3.dot(ab, pvec);
+    if (det < 1e-8 && (!doubleSided || det > -1e-8)) return 0;
```
dabingnn: could we use `EPSILON` instead of magic number?

cocos/renderer/scene/camera.js
```diff
@@ -450,17 +450,10 @@ export default class Camera {
...
-      worldPos.x * _matViewProj.m03 +
-      worldPos.y * _matViewProj.m07 +
-      worldPos.z * _matViewProj.m11 +
-      _matViewProj.m15;
-
     vec3.transformMat4(out, worldPos, _matViewProj);
-    out.x = cx + (out.x / w + 1) * 0.5 * cw;
-    out.y = cy + (out.y / w + 1) * 0.5 * ch;
-    out.z = out.z / w * 0.5 + 0.5;
+    out.x = cx + (out.x + 1) * 0.5 * cw;
```
dabingnn, line 454: is it safe to remove division by w if projection matrix is used to multiplication?

cocos/core/event/event-target.js
```diff
@@ -213,19 +213,6 @@ js.mixin(EventTarget.prototype, {
      * eventTarget.emit('fire', event);
      * eventTarget.emit('fire', message, emitter);
      */
-
```
dabingnn: why remove this?

cocos/scene-graph/base-node.js
```diff
@@ -40,7 +40,7 @@ var idGenerator = new IdGenerator('Node');
 
 function getConstructor(typeOrClassName) {
     if (!typeOrClassName) {
-        cc.errorID(3804);
+        // cc.errorID(3804);
```
dabingnn: why remove this?

cocos/renderer/scene/camera.js
YunHsiao: it's already done in `vec3.transformMat4`

cocos/core/event/event-target.js
YunHsiao: obsolete API

cocos/renderer/scene/camera.js
dabingnn: that is fine, thanks

dabingnn (approved):

PR #3403, opened 2018-10-18
Fix Image Loader on Facebook Playable Ads

hieubuiduc:
Changes:
 * Add platform: Facebook Playable Ads
 * Fix Image Loader when running on Facebook Playable Ads

### DIFF
--- a/cocos2d/core/platform/CCSys.js
+++ b/cocos2d/core/platform/CCSys.js
@@ -667,9 +667,6 @@ else {
         sys.isMobile = false;
         sys.platform = sys.EDITOR_PAGE;
     } 
-    else if (typeof FbPlayableAd !== undefined) { 
-        sys.platform = sys.FB_PLAYABLE_ADS; 
-    }
     else {
         /**
          * Indicate whether system is mobile system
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -136,6 +139,7 @@ defineMacro('CC_JSB', defined('jsb'));
 defineMacro('CC_BUILD', false);
 defineMacro('CC_WECHATGAME', false);
 defineMacro('CC_QQPLAY', false);
+defineMacro('CC_FB_PLAYABLE_ADS', defined('FbPlayableAd'));
```
jareguo: I think it's not necessary to add a macro here. Just use cc.sys.platform is more unified

cocos2d/core/platform/CCSys.js
```diff
@@ -660,6 +666,9 @@ else {
     if (CC_EDITOR) {
         sys.isMobile = false;
         sys.platform = sys.EDITOR_PAGE;
+    } 
+    else if (CC_FB_PLAYABLE_ADS) {
+        sys.platform = sys.FB_PLAYABLE_ADS;
```
jareguo: Can you use `if (typeof FbPlayableAd !== undefined) { sys.platform = sys.FB_PLAYABLE_ADS; }` directly.

cocos2d/core/platform/CCSys.js
```diff
@@ -660,6 +666,9 @@ else {
     if (CC_EDITOR) {
         sys.isMobile = false;
         sys.platform = sys.EDITOR_PAGE;
+    } 
+    else if (typeof FbPlayableAd !== undefined) {
```
jareguo: This test should move to L684, otherwise, sys.isMobile will not initialize.

jareguo (approved):

jareguo:
Thanks for the PR!

PR #3411, opened 2018-10-23
Fix rendering errors for pre-multiplied images using dynamicAtlas.

caryliu1999:
Re: cocos-creator/2d-tasks#437

Changes:
 Fix rendering errors for pre-multiplied images using dynamicAtlas.

### DIFF
--- a/cocos2d/core/assets/CCTexture2D.js
+++ b/cocos2d/core/assets/CCTexture2D.js
@@ -276,6 +276,8 @@ var Texture2D = cc.Class({
         PixelFormat: PixelFormat,
         WrapMode: WrapMode,
         Filter: Filter,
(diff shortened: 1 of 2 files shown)


cocos2d/core/assets/CCTexture2D.js
```diff
@@ -276,6 +276,8 @@ var Texture2D = cc.Class({
         PixelFormat: PixelFormat,
         WrapMode: WrapMode,
         Filter: Filter,
+        FilterIndex: FilterIndex,
```
pandamicro: 加一个下划线吧，不作为公开接口

pandamicro (changes requested):

pandamicro (approved):

PR #3431, opened 2018-10-25
Optimize the render flow of renderComponent when the active state of …

caryliu1999:
…the node is false.

Re: cocos-creator/2d-tasks#

Changes:
When the node is activated, it will execute onEnable and the render-flag will also be reset. So it's not have to update the render-flag when the node is not active.

### DIFF
--- a/cocos2d/core/components/CCRenderComponent.js
+++ b/cocos2d/core/components/CCRenderComponent.js
@@ -130,12 +130,12 @@ let RenderComponent = cc.Class({
     },
     
     _canRender () {
-        return this._enabled;
+        // When the node is activated, it will execute onEnable and the renderflag will also be reset.
+        return this._enabled && this.node._activeInHierarchy;
     },
 
     markForUpdateRenderData (enable) {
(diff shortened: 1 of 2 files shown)


cocos2d/core/components/CCRenderComponent.js
```diff
@@ -134,7 +134,8 @@ let RenderComponent = cc.Class({
     },
 
     markForUpdateRenderData (enable) {
-        if (enable && this._canRender()) {
+        // When the node is activated, it will execute onEnable and the renderflag will also be reset.
+        if (enable && this._canRender() && this.node._activeInHierarchy) {
```
pandamicro: 把这个判断移到 _canRender 中，也就是

canRender = _enabled && node._activeInHierarchy

pandamicro (changes requested):

pandamicro (approved):

caryliu1999:
修复2.0版本中RenderComponent在节点未激活时依然会进行渲染的问题。

PR #3438, opened 2018-10-26
new adaptations for gizmos

YunHsiao:
added run-time VB update support, among other adaptations.

### DIFF
--- a/cocos/3d/assets/texture-cube.js
+++ b/cocos/3d/assets/texture-cube.js
@@ -23,138 +23,58 @@
  THE SOFTWARE.
  ****************************************************************************/
 // @ts-check
-import { _decorator } from "../../core/data/index";
-const { ccclass } = _decorator;
-import Texture from '../../assets/CCTexture2D';
-
+import Texture2D from '../../assets/CCTexture2D';
 import gfx from '../../renderer/gfx';
-import { gfxFilters, gfxWraps, gfxTextureFmts } from '../misc/mappings';
 
-let _opts = {
-  images: [],
-  mipmap: true,
-  width: 2,
-  height: 2,
-  format: gfx.TEXTURE_FMT_RGBA8,
-  anisotropy: 1,
-  wrapS: gfx.WRAP_REPEAT,
-  wrapT: gfx.WRAP_REPEAT,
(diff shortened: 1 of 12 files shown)


cocos/3d/assets/texture-cube.js
```diff
@@ -23,138 +23,58 @@
...
-  height: 2,
-  format: gfx.TEXTURE_FMT_RGBA8,
-  anisotropy: 1,
-  wrapS: gfx.WRAP_REPEAT,
-  wrapT: gfx.WRAP_REPEAT,
-  minFilter: gfx.FILTER_LINEAR,
-  magFilter: gfx.FILTER_LINEAR,
-  mipFilter: gfx.FILTER_LINEAR,
+let isComplete = function(elements) {
+    return elements.reduce((acc, t) => t = t || acc.complete);
```
dabingnn: I love this iterative implementation. but it seems isComplete returns true only if all elements is complete. should we use `t && acc.complete`?

cocos/3d/framework/model-component.js
```diff
@@ -87,7 +87,6 @@ let ModelShadowCastingMode = Enum({
  */
 @ccclass('cc.ModelComponent')
 @executionOrder(100)
-@executeInEditMode
```
dabingnn, line 90: It seems if we do not use `@executeInEditMode`, then `onEnable` `onDisable` will not be called in editor.  model will show wrongly. Am I right @jareguo ?

cocos/3d/builtin/init.js
```diff
@@ -194,24 +167,23 @@ export default function (device) {
...
-    // ].forEach(name => {
-    //     let mat = new Material();
-    //     mat.effectAsset = effects[`builtin-effect-${name}`];
-    //     mat._uuid = `builtin-material-${name}`;
-    //     mat._loaded = true;
-    //     materials[mat._uuid] = mat;
-    // });
-
-    //
+    [
```
dabingnn: given we do not use it, suggest we comment it out now.

cocos/3d/assets/texture-cube.js
YunHsiao: oops :P

dabingnn (approved):

PR #3487, opened 2018-11-09
fix bug #334 spine use premultipliedAlpha,set opacity zero, but still show

sunnylanwanjun:
Re: https://github.com/cocos-creator/2d-tasks/issues/334

### DIFF
--- a/cocos2d/core/renderer/render-flow.js
+++ b/cocos2d/core/renderer/render-flow.js
@@ -118,7 +118,7 @@ _proto._children = function (node) {
     let children = node._children;
     for (let i = 0, l = children.length; i < l; i++) {
         let c = children[i];
-        if (!c._activeInHierarchy) continue;
+        if (!c._activeInHierarchy || c._opacity === 0) continue;
         _cullingMask = c._cullingMask = c.groupIndex === 0 ? cullingMask : 1 << c.groupIndex;
         c._renderFlag |= worldTransformFlag | worldOpacityFlag;
 


pandamicro (approved):

zlpersonal:
If a node wants to fade from 0 to 255, it's child node will ignore the frame with 0 opacity and thus shows 255 before following its parent. Can there be a fix for this? @pandamicro

sunnylanwanjun:
> If a node wants to fade from 0 to 255, it's child node will ignore the frame with 0 opacity and thus shows 255 before following its parent. Can there be a fix for this? @pandamicro

The node has added spine component ？

zlpersonal:
The node has no spine, but the file change applies to all type of node. Maybe it fixed issue with spine, but introduce problem for normal node.

sunnylanwanjun:
> The node has no spine, but the file change applies to all type of node. Maybe it fixed issue with spine, but introduce problem for normal node.

Yeah，the fixed is about if render component use 'add' blend mode, node will show even though opacity is zero. Have you change blend mode from 'normal' to 'add'?

zlpersonal:
screenshot 2019-01-28 at 5 24 05 pm

From this code, the highlighted area should receive opacity from parent, but since opacity === 0 'break' the loop when opacity is 0, children opacity is not changed to 0.

I have not changed the blend mode, this can be easily reproduce on any simple node on a new project.

sunnylanwanjun:
If parent opacity is zero, child won't be rendered, so no need to change child's opacity.

zlpersonal:
Sorry that i mis saw the bottom part, the bottom part is updated on `https://github.com/cocos-creator/engine/commit/a67bfe6753b9d3031fe19a00f8dcf4aeaa43cbc7#diff-5fb4f3b27e379fb8b3d72cd2fe4f3376` by @jareguo where color.a now obtain it's value there. 

Insuch case, if opacity is 0, and break instead of applied to '_color.a', component such as cc.Label which uses '_color.a' for it's opacity will have issue.

PR #3514, opened 2018-11-15
add conversion between euler angles and quat for 2D

2youyou2:
Changes:
 * add conversion between euler angles and quat for 2D


pandamicro:
Node 还是需要拆分开比较干净，这个估计需要多久？

pandamicro (approved):

PR #3519, opened 2018-11-15
 fix motion streak index count error.

xianyinchen:
fix motion streak index count error.


pandamicro:
Fixed in https://github.com/cocos-creator/engine/pull/3520

PR #3568, opened 2018-11-27
set playOnFocus decorate default value to true

benloong:
`@playOnFocus` decorate not set `Component._playOnFocus` to true  as desired.
`@playOnFocus(true)` works correct, but creator.d.ts  type info does not match.
So i think this decorate should has a default `true` value.

Re: cocos-creator/2d-tasks#

Changes:
 * set `@playOnFocus` decorate default value to true


jareguo:
Thanks for your contribution.

PR #3602, opened 2018-12-06
change sprite simple assembler vertex buffer strategy and fix model batch.

xianyinchen:
change sprite simple assembler vertex buffer strategy and fix model batch.


pandamicro:
后续会让 @caryliu1999 处理 bmfont 使用 quad buffer 的问题

PR #3638, opened 2018-12-14
Add cc.sizeTo action for resizing.

unace:
Re: cocos-creator/2d-tasks#

Changes:
 * 

I added cc.sizeTo action for resizing. Chinese documentation comes from cc.scaleTo. I don't know Chinese...

### DIFF
--- a/cocos2d/actions/CCActionInterval.js
+++ b/cocos2d/actions/CCActionInterval.js
@@ -1941,6 +1941,89 @@ cc.scaleBy = function (duration, sx, sy) {
     return new cc.ScaleBy(duration, sx, sy);
 };
 
+/* Resize a Node object to a size factor by modifying it's width/height property.
+ * @warning This action doesn't support "reverse"
+ * @class SizeTo
+ * @extends ActionInterval
+ * @param {Number} duration
+ * @param {Number} width  width parameter
+ * @param {Number} height height parameter
+ * @return {ActionInterval}
+ * @example
+ * // example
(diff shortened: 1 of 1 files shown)


pandamicro:
Thanks for your contribution, can you give a example to show us why sizeTo is necessary while we have scaleTo？

unace:
[test.zip](https://github.com/cocos-creator/engine/files/2691771/test.zip)

Sure, I uploaded a sample project file. (Cocos Creator v.2.0.6)
First, the layout component works correctly with width/height.
Second, I don't want to scale children for some reason.

Please let me know if you have any questions. Thanks for the review!

pandamicro:
I see, thanks for the test case and clarification

pandamicro:
@unace please take a look at this PR https://github.com/cocos-creator/engine/pull/3660, especially the API usage of cc.Animate, do you think it's a good alternative to unsupported actions ?

We are thinking of providing a general purpose tweening action because we don't want to expand the action system anymore. In fact, the reason why we kept Action from the beginning of Creator was only for keeping a certain level of API consistency with Cocos2d-html5. In the near future, when we want to simplify Node APIs, it's possible that we deprecate the current Action system.

For the above reasons, we hesitate to add new actions, please tell us your thoughts about it

unace:
Wow! The cc.animate looks amazing. I totally understand your thought and agree the cc.animate is a better and generic solution for the tweening action. I gonna close this, thanks for the review!

PR #3659, opened 2018-12-20
Change argument of clickEvents for consistency

benloong:
All UI component events should pass `this` to event handler instead of `event` for consistency.

Re: cocos-creator/2d-tasks#

Changes:
 *  pass `this` instead of `event` to `clickEvents` handler

### DIFF
--- a/cocos2d/core/components/CCButton.js
+++ b/cocos2d/core/components/CCButton.js
@@ -635,7 +635,7 @@ let Button = cc.Class({
         if (!this.interactable || !this.enabledInHierarchy) return;
 
         if (this._pressed) {
-            cc.Component.EventHandler.emitEvents(this.clickEvents, this);
+            cc.Component.EventHandler.emitEvents(this.clickEvents, event, this);
             this.node.emit('click', this);
         }
         this._pressed = false;
--- a/cocos2d/core/components/CCButton.js
+++ b/cocos2d/core/components/CCButton.js
@@ -635,7 +635,7 @@ let Button = cc.Class({
         if (!this.interactable || !this.enabledInHierarchy) return;
 
         if (this._pressed) {
-            cc.Component.EventHandler.emitEvents(this.clickEvents, event);
+            cc.Component.EventHandler.emitEvents(this.clickEvents, this);
             this.node.emit('click', this);
         }
         this._pressed = false;


pandamicro:
@benloong Does it break API consistency with older version ?

benloong:
@pandamicro YES, this change will break user code if and only if they used `event` argument in the handler function.

knoxHuang (commented):
I think add ‘this’ to the second parameter.

cc.Component.EventHandler.emitEvents(this.clickEvents, event, this);

This will not break the API consistency of the old version. 
What do you think? @benloong

benloong:
@knoxHuang that sounds good, but i prefer keep our design consistency of all UI Components (Toggle, Button, ScrollView .etc) event Handler API , and breaking a bit of backward compatibility.

pandamicro:
@benloong API consistency is very important to us, we don't want to break user code if not absolutely necessary, please add the desired parameter after the original one. @knoxHuang after this PR get merged, please add callback function example to the API doc of clickEvents, and explain parameters received by callbacks.

pandamicro (approved):

pandamicro:
Thanks for your contribution @benloong

cocos2d/core/components/CCButton.js
```diff
@@ -635,7 +635,7 @@ let Button = cc.Class({
         if (!this.interactable || !this.enabledInHierarchy) return;
 
         if (this._pressed) {
-            cc.Component.EventHandler.emitEvents(this.clickEvents, event);
+            cc.Component.EventHandler.emitEvents(this.clickEvents, event, this);
```
jareguo, line 638: 这里好像不需要改的，因为 event.target 就会是 Button 节点本身了。

benloong: on("click") 的回调参数又是给的button， API不一致，让我难以理解。为何不让他们的回调参数保持一致呢？
在cc.Toggle中 on("toggle") 和checkEvents 的回调参数又保持的是一致的

benloong: 所以我的建议是 compatibility break 就 break嘛 ， clickEvents 也就应该只带一个参数 this

jareguo: @knoxHuang 看下这个问题

PR #3673, opened 2018-12-24
improve languageCode

knoxHuang:
Re: cocos-creator/2d-tasks#939

Changes:
 * 完善 languageCode 统一改成小写的


jareguo (approved):

cocos2d/core/platform/CCSys.js
```diff
@@ -809,9 +810,10 @@ function initSys () {
 
         /**
          * Get current language iso 639-1 code.
+         * Examples of valid language codes include "zh-tw", "en", "en-us", "fr", "fr-fr", "es-es", etc.
```
pandamicro, line 813: 再加一句吧，

Examples of valid language codes include "zh-tw", "en", "en-us", "fr", "fr-fr", "es-es", etc. The actual value totally depends on results provided by destination platform.

PR #3688, opened 2018-12-28
Small fix

shrinktofit:


cocos/assets/texture-2d.js
```diff
@@ -47,6 +47,11 @@ import ImageAsset from './image-asset';
  */
 @ccclass('cc.Texture2D')
 export default class Texture2D extends TextureBase {
+	@property({
+	    override: true
+	})
```
pandamicro: do not use tab, use four spaces

pandamicro (changes requested):

pandamicro (approved):

PR #3690, opened 2018-12-28
do not run phy system in editor;support center of phy shape;fix setWorldPosition() bug

shrinktofit:


dumganhar:
这是把引擎转 ts 的节奏么？

PR #3692, opened 2018-12-28
Port label vertical alignment fix

id-ilych:
Re: cocos-creator/2d-tasks#886

Port of https://github.com/cocos-creator/engine/pull/3664 for 1.10 release

image

Changes:
 * 完善 label 垂直对齐问题

### DIFF
--- a/cocos2d/core/label/CCSGLabelCanvasRenderCmd.js
+++ b/cocos2d/core/label/CCSGLabelCanvasRenderCmd.js
@@ -25,6 +25,10 @@
  THE SOFTWARE.
 
  ****************************************************************************/
+
+var BASELINE_RATIO = 0.25;
+var MIDDLE_RATIO = (BASELINE_RATIO + 1) / 2 - BASELINE_RATIO;
+
 (function () {
     _ccsg.Label.TTFLabelBaker = function () {};
 
@@ -278,9 +282,10 @@
         var paragraphedStrings = node._string.split('\n');
         var i;
         var ctx = this._labelContext;
+        var margin = this._getMargin();
 
         if (_ccsg.Label.Overflow.RESIZE_HEIGHT === node._overFlow) {
-            this._canvasSize.height = this._splitedStrings.length * this._getLineHeight();
+            this._canvasSize.height = (this._splitedStrings.length + BASELINE_RATIO) * this._getLineHeight() + 2 * margin;
             _ccsg.Node.prototype.setContentSize.call(node, this._canvasSize);
         }
         else if(_ccsg.Label.Overflow.NONE === node._overFlow) {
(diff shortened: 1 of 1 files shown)


knoxHuang (approved):

pandamicro:
@id-ilych Thanks, sorry that we don't have recent plan to release any new 1.x version, you will need to use customized engine by yourself. By the way, why haven't you upgraded to v2.0 ？

id-ilych:
@pandamicro I'm planning transition to 2.1 but there are a lot of things to do for that. Just to name few:

1. Adopt API changes (e. g. `rotation` -> `angle`, `cascadeOpacity` -> nothing, sound engine is no longer taking string and is expecting sound clip now)
2. Handle implicit behavior changes (e. g. `emit` no longer wraps the argument in `cc.Event.EventCustom` so subscriber should know whether event would be dispatched or emitted or to wrap all emits with explicit wrappings)
3. CSV asset is handled differently in regards to line endings so I'll replace them with JSON assets.
4. Fix all prefabs as migrations breaks a lot (e. g. spine `premultipliedAlpha` property is reset, `scale` property is reset)
5. Test my app thoroughly as other breaking changes may be unknown.
6. Minor things like fixing `creator.d.ts` (e. g. `contains` returns `void` instead of `boolean`, `getCurrent` accepts argument of type `void` instead of `number`).

So eventually I will do. But it will take some time :( 
Though it probably worth it as performance boost is significant (2x time with some quick benchmark)

jareguo:
@id-ilych Please only transit to 2.0.x because 2.1.0 is not very stable by now. We will keep update 2.0.x until 2.1 is stable enough.

id-ilych:
@jareguo Ok. Is there any estimation when 2.0.x will be released with similar `ttf.js` fix? I only see it in `v2.1-release` branch.

Ronsku:
@id-ilych @jareguo Will this fix be in 2.1.1? Currently I have to build each project with a separate engine and since engine settings in cocos are global, you have to change it every time you swap between projects that use different versions of Cocos Creator. Having this in 2.1.1 would be so nice!

Thank you!

id-ilych:
@Ronsku I am not a part of CC team. Just an occasional contributor :) As jareguo mentioned earlier in this topic 2.1 is not recommended to be used in production yet. So what I can recommend to increase probability of this fix being included in future 2.0.x release is to create new pull request against `v2.0-release` branch. Unfortunatelly I can't do it right now but it seems to be trivial using original PR (https://github.com/cocos-creator/engine/pull/3664). And until project-specific custom engine configuration is fixed (it is in the UI but doesn't work) maybe it's better to use 2.0-based custom version of the engine for all the projects and migrate them synchronously to avoid the issue. I do understand that this could be problematic but it's best idea I have right now and that is what I personally do at work.

jareguo:
> Is there any estimation when 2.0.x will be released with similar `ttf.js` fix? I only see it in `v2.1-release` branch.

No, 2.0.x will not have this larger adjustment.

jareguo:
> Will this fix be in 2.1.1

Sure

PR #3820, opened 2019-01-30
Fix: Spine blinks after instantiation

id-ilych:
If prefab containing spine animation is instantiated from `update` of some other component then it will be rendered in 'setup' pos for one frame. I've experienced problem and verified fix using CocosCreator 2.0.7 but I see no relevant diffs with current state of the fille.

My understanding is that in described case newly created animation will be rendered before `update` on `sp.Skeleton`. Not sure about changes to skeleton data beside instantiation though. 

Is my understanding correct and problem should be fixed like that?

### DIFF
--- a/extensions/spine/Skeleton.js
+++ b/extensions/spine/Skeleton.js
@@ -907,8 +907,8 @@ sp.Skeleton = cc.Class({
                 var res = this._state.setAnimationWith(trackIndex, animation, loop);
                 if (CC_EDITOR && !cc.engine.isPlaying) {
                     this._state.update(0);
-                    this._state.apply(this._skeleton);
                 }
+                this._state.apply(this._skeleton);
                 return res;
             }
         }
@@ -1200,10 +1200,6 @@ sp.Skeleton = cc.Class({
                 this._skeletonCache = SkeletonCache.sharedCache;
             } else if (this.renderMode === RenderModeEnum.PRIVATE_CACHE) {
                 this._skeletonCache = new SkeletonCache;
-                
-                // validate skeleton state
-                // as render may happen before `update`
-                this._state.apply(this._skeleton);
             }
         }
 
--- a/extensions/spine/Skeleton.js
+++ b/extensions/spine/Skeleton.js
@@ -905,9 +905,6 @@ sp.Skeleton = cc.Class({
                     return null;
                 }
                 var res = this._state.setAnimationWith(trackIndex, animation, loop);
-                if (CC_EDITOR && !cc.engine.isPlaying) {
-                    this._state.update(0);
-                }
                 this._state.apply(this._skeleton);
                 return res;
             }
--- a/extensions/spine/Skeleton.js
+++ b/extensions/spine/Skeleton.js
@@ -1200,6 +1200,10 @@ sp.Skeleton = cc.Class({
                 this._skeletonCache = SkeletonCache.sharedCache;
             } else if (this.renderMode === RenderModeEnum.PRIVATE_CACHE) {
                 this._skeletonCache = new SkeletonCache;
+                
+                // validate skeleton state
+                // as render may happen before `update`
+                this._state.apply(this._skeleton);
             }
         }
 


pandamicro:
hi,  @id-ilych I have assigned the reviewer, but we are currently in vacations of spring festival, so your pr will be reviewed in a few days, sorry for the delay

id-ilych:
Follow up question: I've notice that the same problem would be if you'll switch the animation from `update` of component that is further (in terms of `update` invocation order). Example: coins particles (layer on top of everything else) arrived at coins counter/icon and switch icon's animation so it 'reacts' to new coins arrival. In that case icon would be rendered in 'setup pose' for one frame... 

So, maybe `this._state.apply(this._sekeleton)` should be called after animation change as well? Although I do understand that it basically means double work/calculation on that in some cases. What do you think?

sunnylanwanjun:
> Follow up question: I've notice that the same problem would be if you'll switch the animation from `update` of component that is further (in terms of `update` invocation order). Example: coins particles (layer on top of everything else) arrived at coins counter/icon and switch icon's animation so it 'reacts' to new coins arrival. In that case icon would be rendered in 'setup pose' for one frame...
> 
> So, maybe `this._state.apply(this._sekeleton)` should be called after animation change as well? Although I do understand that it basically means double work/calculation on that in some cases. What do you think?

Move the 'this._state.apply(this._sekeleton) ' code into setAnimation function will fix the proplem.
By the way, please change the request merge branch to v2.0-release.

id-ilych:
@sunnylanwanjun I've reverted my initial fix and reimplemented according to your proposal. At the end it's just moving `this._state.apply(this._skeleton);` off the `if (CC_EDITOR && !cc.engine.isPlaying)` clause. No it's just `this._state.update(0);` in this editor-specific code. 

I have tested new fix with my initial cases and it works. Altough I don't know about this new 'cached mode' thing.

Is there a guide for contributors that clarifies flow (e.g. what branch to target)?

sunnylanwanjun:
@id-ilych 
Thanks for your commit.

> ` off the `if (CC_EDITOR && !cc.engine.isPlaying)` clause. No it's just `this._state.update(0);` in this editor-specific code.

If remove this logic what would happen in editor mode?  I'm sorry but I don't understand it.

>I have tested new fix with my initial cases and it works. Altough I don't know about this new 'cached mode' thing.

'cached mode' is a new function about cache animation data instead of realtime calculate skeleton animation data. It can improve performance. The speciality will publish in creator 2.0.9.

> Is there a guide for contributors that clarifies flow (e.g. what branch to target)?

v2.0-release is a branch about fix bug or improve performance. 'master' is a branch about new feature such as 3D.

id-ilych:
@sunnylanwanjun 

> If remove this logic what would happen in editor mode? I'm sorry but I don't understand it.

I think somebody wrote it just in case :) If my fix works in described situations then I suppose that `this._state.apply(this._skeleton)` should be necessary and there is no need for `this._state.update(0)`. I just prefer not to touch things that are not directly relevant to the issue I'm fixing. But if you think this editor-specific `if` should be removed then I can do that.

> 'cached mode' is a new function about cache animation data instead of realtime calculate skeleton animation data. It can improve performance. The speciality will publish in creator 2.0.9.

OK. Just as I guessed. But the point is - I don't know how to properly test that case with regards to the issue I'm fixing with this PR. As you can see in case of 'cached mode' `setAnimation` function will use alternative code-path and will avoid lines I've changed.

> v2.0-release is a branch about fix bug or improve performance. 'master' is a branch about new feature such as 3D.

Good. I will base my PRs on release branch then. Just want to be sure they will not be lost and will somehow be ported to 2.1 and further version.

sunnylanwanjun:
@id-ilych 

> I think somebody wrote it just in case :) If my fix works in described situations then I suppose that `this._state.apply(this._skeleton)` should be necessary and there is no need for `this._state.update(0)`. I just prefer not to touch things that are not directly relevant to the issue I'm fixing. But if you think this editor-specific `if` should be removed then I can do that.

OK,  I think `if (CC_EDITOR && !cc.engine.isPlaying) {...}` is unnecessary, you can remove it.

> OK. Just as I guessed. But the point is - I don't know how to properly test that case with regards to the issue I'm fixing with this PR. As you can see in case of 'cached mode' `setAnimation` function will use alternative code-path and will avoid lines I've changed.

Yeah, `cached mode` is no need to invoke `this._state.apply(this._skeleton)`, it will change animation immediately.

> Good. I will base my PRs on release branch then. Just want to be sure they will not be lost and will somehow be ported to 2.1 and further version.

Don't worry about that, we will merge your pr into master and 2.1 branch when publish 2.1 version.

id-ilych:
@sunnylanwanjun 
> OK, I think `if (CC_EDITOR && !cc.engine.isPlaying) {...}` is unnecessary, you can remove it.

Done

sunnylanwanjun (approved):

pandamicro:
@id-ilych Thanks for the update
@sunnylanwanjun we need a official documentation about cached mode in Spine and DB when it's released, please prepare one

sunnylanwanjun:
>  we need a official documentation about cached mode in Spine and DB when it's released, please prepare one

It has been prepared last week.

PR #3947, opened 2019-03-01
Rename skeleton RenderMode to AnimationCacheMode

sunnylanwanjun:


extensions/dragonbones/ArmatureDisplay.js
```diff
@@ -453,53 +453,53 @@ let ArmatureDisplay = cc.Class({
...
     },
     
     /**
      * !#en Whether in cached mode.
      * !#zh 当前是否处于缓存模式。
-     * @method isCachedMode
+     * @method isAnimationCacheMode
      */
-    isCachedMode () {
+    isAnimationCacheMode () {
```
pandamicro: 改为 isAnimationCached

pandamicro (changes requested):

editor/i18n/en/localization.js
```diff
@@ -260,7 +260,7 @@ module.exports = {
...
+            "animation_cache_mode": "REALTIME model, realtime calculate, support all functions of Spine.\nSHARED_CACHE mode, caching and sharing dragonbones animation data, the equivalent of pre baked skeletal animation, have high performance, does not support the action blend and superposition, only supports the start and end events, as for memory, when creating some same bones and the same action of animation, can present advantages of memory, the greater the amount of skeleton, the more obvious advantages, in conclusion SHARED_CACHE mode is suitable for the scene animation, special effects, monster, NPC and so on, can greatly increase the frame rate and reduce memory.\nPRIVATE_CACHE mode, similar to SHARED_CACHE, but not share animation and texture data, so there is no advantage in memory, there is only a performance advantage, when trying to take advantage of caching pattern of high performance, but there is a change of texture, so you can't share the map data, then PRIVATE_CACHE is suitable for you.",
```
pandamicro: REALTIME model, the animation data is calculated while running, all features supported.
SHARED_CACHE mode, the animation data is pre-calculated and cached globally, the data is shared between skeleton animation instances. Shared cache mode has great performance advantage, but it has some limitation. It doesn't support action blending and addition and it only supports the start and end events. In the memory usage aspect, as same skeleton animation instances are sharing the same animation data, this mode can save a lot of memory. More instances are sharing, more memory saved. In conclusion SHARED_CACHE mode is suitable for the skeletons which could be replicated, such as effects animation, monster, NPC and so on.\nPRIVATE_CACHE mode, similar to SHARED_CACHE, but the animation data and texture data are not shared between skeleton instances. So there is no advantage in memory compare to REALTIME mode. The PRIVATE_CACHE is suitable when there is no animation blending or addition needed, but the skeleton could change textures of sub bone (e.g. weapon).

pandamicro (changes requested):

pandamicro (approved):

PR #4013, opened 2019-03-12
fix the typesetting problem of letter font. (#4007)

caryliu1999:


pandamicro:
Will be merge in https://github.com/cocos-creator/engine/pull/4011/

PR #4056, opened 2019-03-18
Prevents an onLoaded error from causing a callback not to be executed

VisualSJ:
加载资源的时候 onLoaded 报错会导致 callback 用户执行不到


SantyWang:
没问题

PR #4072, opened 2019-03-20
Added <video muted> attribute on video muted or volume = 0

Ronsku:
Changes:
 * Due to playback issues on Apple IOS devices where automatic Video playback needs a user gesture to start playing if it contains sound. If **<video muted> elements will also be allowed to autoplay without a user gesture**. More can be read on https://webkit.org/blog/6784/new-video-policies-for-ios/

* The error you can get if you start playing a video without a user gesture on a IOS device is:
image
It's all about the permissions set by Apple on IOS.

* This PR fixes this autoplay issues when you run videos with Volume = 0 or mute = true. It simply adds an attribute **muted** to the <video> element, like so: **<video muted>**

* I've tested that it works and fixes the permission problem with video playback on IOS.

### DIFF
--- a/cocos2d/videoplayer/video-player-impl.js
+++ b/cocos2d/videoplayer/video-player-impl.js
@@ -166,7 +166,6 @@ let VideoPlayerImpl = cc.Class({
         video.className = "cocosVideo";
         video.setAttribute('preload', 'auto');
         video.setAttribute('webkit-playsinline', '');
-        // This x5-playsinline tag must be added, otherwise the play, pause events will only fire once, in the qq browser.	
         video.setAttribute("x5-playsinline", '');
         video.setAttribute('playsinline', '');
         if (muted) {
--- a/cocos2d/videoplayer/video-player-impl.js
+++ b/cocos2d/videoplayer/video-player-impl.js
@@ -158,7 +158,7 @@ let VideoPlayerImpl = cc.Class({
         video.style.height = height + 'px';
     },
 
-    _createDom () {
+    _createDom (mute, volume) {
         let video = document.createElement('video');
         video.style.position = "absolute";
         video.style.bottom = "0px";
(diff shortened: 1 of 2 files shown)


cocos2d/videoplayer/video-player-impl.js
```diff
@@ -158,23 +158,24 @@ let VideoPlayerImpl = cc.Class({
         video.style.height = height + 'px';
     },
 
-    _createDom () {
+    _createDom (mute, volume) {
```
jareguo: Can we just combine two args into one and named as “muted”? Then pass (mute || volume === 0) as argument directly when calling this function.

cocos2d/videoplayer/video-player-impl.js
```diff
@@ -158,23 +158,24 @@ let VideoPlayerImpl = cc.Class({
...
         let video = document.createElement('video');
         video.style.position = "absolute";
         video.style.bottom = "0px";
         video.style.left = "0px";
         video.className = "cocosVideo";
         video.setAttribute('preload', 'auto');
         video.setAttribute('webkit-playsinline', '');
-        // This x5-playsinline tag must be added, otherwise the play, pause events will only fire once, in the qq browser.
-        video.setAttribute("x5-playsinline", '');
```
jareguo: Why remove the attribute for X5?

cocos2d/videoplayer/video-player-impl.js
```diff
@@ -158,23 +158,24 @@ let VideoPlayerImpl = cc.Class({
...
+        }
 
         this._video = video;
         cc.game.container.appendChild(video);
     },
 
-    createDomElementIfNeeded: function () {
+    createDomElementIfNeeded: function (mute, volume) {
         if (!this._video) {
             this._createDom();
```
jareguo: You forgot to pass the argument to _createDom

cocos2d/videoplayer/video-player-impl.js
Ronsku: Good idea, fixed this!

cocos2d/videoplayer/video-player-impl.js
Ronsku: I'm not sure if I first pulled another version of the engine or if I blindly deleted this. This was not the purpose and it is added back now.

cocos2d/videoplayer/video-player-impl.js
Ronsku: Added here as well 👍

jareguo (approved):

knoxHuang (approved):

jareguo:
@Ronsku @knoxHuang  Should we need to remove the ”muted“ attribute when (this._mute || this._volume === 0) become false?

PR #4139, opened 2019-04-03
Repair collision failure of physical system due to automatic hibernation

JayceLai:
Add triggers

Re: cocos-creator/2d-tasks#

Changes:
 *


jerrywwl:
注意，函数名后面要加空格

PR #4153, opened 2019-04-04
animation

shrinktofit:
Re: cocos-creator/2d-tasks#

Changes:
 *


pandamicro:
cocos-creator/3d-tasks#438

PR #4158, opened 2019-04-08
Fix a problem where the physical system collision event does not trig…

JayceLai:
Fix a problem where the physical system collision event does not trigger because of automatic sleep, adding a trigger property to the collider component


JayceLai:
之前代码格式化没有配置好，现在函数名后面的空格没有去掉了 @jerrywwl

PR #4193, opened 2019-04-11
add huawei platform

wuzhiming:
https://github.com/cocos-creator/2d-tasks/issues/1297
add huawei platform


cocos2d/core/platform/CCSys.js
```diff
@@ -379,6 +380,12 @@ function initSys () {
      * @default 109
      */
     sys.OPPO_GAME = 109;
+    /**
+     * @property {Number} HUAWEI_GAME
+     * @readOnly
+     * @default 109
```
jareguo, line 386: default 值错啦

PR #4201, opened 2019-04-12
Fix mipmap setting issue

pandamicro:
Re: 开发者（鱼）反馈


id-ilych:
Hi. I see there is just one place fixed where `opts.hasMipmap` was assigned instead of `opts.mipmap` but what about other places (`initWithData` and `handleLoadedData`)? I see no places where `hasMipmap` is read from options - only `mipmap` is actually used.

PR #4250, opened 2019-04-18
fix particle life error.

xianyinchen:
when particles was stop，it alway has this value, _renderData.ia._count=6, and when the cc.ParticleSystem node active, it will draw one particle!

issue：
https://github.com/cocos-creator/2d-tasks/issues/1339


xianyinchen:
问题一样
https://github.com/cocos-creator/engine/pull/4243

PR #4289, opened 2019-04-24
put mat4 mul onto gpu,

shrinktofit:


shrinktofit:
depends on https://github.com/cocos-creator/editor-3d/pull/1206

PR #4292, opened 2019-04-24
remove WrapMode from cc, instead make it as a member of AnimationClip

shrinktofit:


shrinktofit:
depends on https://github.com/cocos-creator/editor-3d/pull/1210

PR #4338, opened 2019-04-29
support vtf on webgl2

shrinktofit:
Re: cocos-creator/2d-tasks#

Changes:
 *


shrinktofit:
rely on https://github.com/cocos-creator/editor-3d/pull/1249

PR #4354, opened 2019-04-30
reject ui module

JoneLau:


JoneLau:
https://github.com/cocos-creator/3d-tasks/issues/444

PR #4374, opened 2019-05-06
 fix build error;  re-create animation states when clips are reset

shrinktofit:


shrinktofit:
https://github.com/cocos-creator/3d-tasks/issues/575
https://github.com/cocos-creator/3d-tasks/issues/579

shrinktofit:
可以通过了

PR #4376, opened 2019-05-06
.d.ts bundle

shrinktofit:
Re: cocos-creator/2d-tasks#

Changes:
 *


pandamicro:
https://github.com/cocos-creator/3d-tasks/issues/577

PR #4381, opened 2019-05-06
pause animation on last frame

shrinktofit:
Re: cocos-creator/2d-tasks#

Changes:
 *


shrinktofit:
https://github.com/cocos-creator/3d-tasks/issues/553

PR #4395, opened 2019-05-09
Fixed a problem in TileMap where the renderer could not properly disp…

Y-way:
…lay flipped/rotated tiles

Fixed a problem in TileMap where the renderer could not properly display flipped/rotated tiles

Re: cocos-creator/2d-tasks#

Changes:
 *


sunnylanwanjun:
We have reconstructed tiledmap and fixed  the problem in the meantime. We will publish it in v2.2. 
Here is the pull request(https://github.com/cocos-creator/engine/pull/4245).
But thanks for you to fixed this problem all the same.

PR #4415, opened 2019-05-14
 Improve EventTarget to avoid duplicated implementation

pandamicro:


pandamicro:
FOR https://github.com/cocos-creator/3d-tasks/issues/592

PR #4468, opened 2019-05-21
fix mask does not update bug

JoneLau:


pandamicro (approved):

JoneLau:
https://github.com/cocos-creator/3d-tasks/issues/655

PR #4474, opened 2019-05-22
add canvas don't support button disabled state for 2d-tasks/issues/1451

knoxHuang:
Re: cocos-creator/2d-tasks#1451

Changes:
 * 目前 Canvas 不支持 Gray Effect，先禁用


jareguo:
修复微信小游戏开放域中不支持 Button Gray Effect 导致的报错

PR #4484, opened 2019-05-23
Fix toggle-component

JayceLai:
Re: cocos-creator/3d-tasks#

Changes:
 * fix onDisable function in toggle component


JayceLai:
cocos-creator/3d-tasks#580

pandamicro (approved):

PR #4508, opened 2019-05-27
fix ui  performance

JoneLau:
1. label ui drawcall too much high
2. switch visibility when parent changed


JoneLau:
https://github.com/cocos-creator/3d-tasks/issues/683

PR #4511, opened 2019-05-28
fix ui  performance

JoneLau:
1. label ui drawcall too much high
2. switch visibility when parent changed

### DIFF
--- a/cocos/3d/ui/components/ui-component.ts
+++ b/cocos/3d/ui/components/ui-component.ts
@@ -1,6 +1,8 @@
 import { Component } from '../../../components';
+import { SystemEventType } from '../../../core';
 import { ccclass, executeInEditMode, executionOrder, property } from '../../../core/data/class-decorator';
 import { UI } from '../../../renderer/ui/ui';
+import { Node } from '../../../scene-graph';
 import { CanvasComponent } from './canvas-component';
 
 /**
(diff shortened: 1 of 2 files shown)


cocos/3d/ui/components/ui-component.ts
```diff
@@ -42,7 +44,35 @@ export class UIComponent extends Component {
...
+        this._lastParent = this.node.parent;
+        this._updateVisibility();
+        if (this._lastParent) {
+            this.node.on(SystemEventType.CHILD_REMOVED, this._parentChanged, this);
+        }
+
+    }
+
+    public onDisable () {
+        this._cancelEventFromParent();
```
zxg0622, line 58: 这个地方要把_visibility设为-1

JoneLau: 已修改

zxg0622 (approved):

weekendkoder:
I am on CC v2.11. I also recognized the **cc.Label Performance problem** !!!

You will only recognize it, when you store data into the Label-Component via "`update(dt)`". When values are updated per frame and get read-out by `this.My_Node-A.string = this.My_Node-B.x` It is a really huge performance drop. 

I was working on my foreground / background scrolling. Adding the UI, and then i recognized it. Disabling the **cc.Label UI** stuff from **update(dt)** solves the issue with stuttering scrolling of my background / foreground elements. I think its better recognizable on older low-end Devices. On high-end flagship devices, you will not recognize the issue.

So, it would be really great, if the CC-devs can solve this cc.Label Ui problem. We need a fluent "real time" Data read / write to UI-elements. Thanks.

weekendkoder:
Update:

In CC v.211 you can change the **chaching-mode** of the cc.Label component. Which solves the issue for me :-) ! This means:
> none
> bitmap
> char

... and here, just switch to **char** if you need to update the values every frame.

PR #4587, opened 2019-06-11
API documents for value types

shrinktofit:
Re: cocos-creator/2d-tasks#

Changes:
 *


shrinktofit:
https://github.com/cocos-creator/3d-tasks/issues/483

PR #4667, opened 2019-06-20
fix widget bug and add switch in px to persent

JoneLau:


JoneLau:
https://github.com/cocos-creator/3d-tasks/issues/797

zxg0622 (approved):

PR #4686, opened 2019-06-24
api module declare for gemo-util

JayceLai:
Re: cocos-creator/3d-tasks#819

Changes:
 * api module declare for gemo-util


jerrywwl (changes requested):
请按新规范修改一下

jerrywwl (approved):

PR #4701, opened 2019-06-25
add log2 polyfill

2youyou2:
Re: cocos-creator/2d-tasks#1563

Changes:
 * add log2 polyfill

### DIFF
--- a/polyfill/misc.js
+++ b/polyfill/misc.js
@@ -8,6 +8,12 @@ if (!Math.sign) {
     };
 }
 
+if (!Math.log2) {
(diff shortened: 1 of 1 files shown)


polyfill/misc.js
```diff
@@ -8,6 +8,12 @@ if (!Math.sign) {
     };
 }
 
+if (!Math.log2) {
+    Math.log2 = function (x) {
+        return Math.log(x) / 0.6931471805599453;
```
jareguo: 改成 Math.log(x) * Math.LOG2E 吧？

PR #4743, opened 2019-06-28
fix particle lag (#4718)

zxg0622:
* fix particle lag

* fix model recreate pso on enable

* change value-type lerp interface

* change value-type lerp to static

* fix model not rendered at persistent node

Re: cocos-creator/2d-tasks#

Changes:
 *


jerrywwl (approved):

zxg0622:
cocos-creator/3d-tasks#888

PR #4746, opened 2019-06-28
optimize ubo data copy

jerrywwl:
Re: cocos-creator/2d-tasks#

Changes:
 *


jerrywwl:
cocos-creator/3d-tasks#888

PR #4857, opened 2019-07-10
fix model and particle render are not affected by priority on ui

JoneLau:


JoneLau:
https://github.com/cocos-creator/3d-tasks/issues/922

jerrywwl (approved):

PR #4909, opened 2019-07-18
Fix bug of samepleEvents in animation-state.ts

JayceLai:


JayceLai:
https://github.com/cocos-creator/3d-tasks/issues/1139

jerrywwl (approved):

PR #4921, opened 2019-07-19
avoid reusing scene asset

SantyWang:
Re: cocos-creator/3d-tasks#

Changes:
 * 避免复用场景


jerrywwl (approved):

poetao:
When the scene is loaded after the scene is preloaded, you will download the scene assets twice. Is this reasonable?

SantyWang:
not quite reasonable actually, but some bad issues will occur if reuse scene assets. We will improve it  and thanks for your feedback

PR #4945, opened 2019-07-23
fix pageview inertia invalid

knoxHuang:
Re: https://forum.cocos.com/t/pageview/80553

Changes:
 * 修复 PageView 勾选 inertia 无效的 bug


zilongshanren:
LGTM

PR #5055, opened 2019-08-01
enchance deprecated and add some deprecated in animation module

JayceLai:


JayceLai:
https://github.com/cocos-creator/3d-tasks/issues/1188

jerrywwl (approved):

PR #5129, opened 2019-08-08
add copy func to vec

JayceLai:


JayceLai:
duplicate, vect has 'set'

PR #5233, opened 2019-08-23
add compare API for value-adapter and target-modifier

shrinktofit:
Re: cocos-creator/2d-tasks#

Changes:
 *


shrinktofit:
@gameu3d

PR #5246, opened 2019-08-26
fixed animation

2youyou2:
Re: cocos-creator/2d-tasks#

Changes:
 * fixed stop event not triggered when call stop in lastFrame event callback
 * fixed lastframe event not triggered when register lastframe before play animation

### DIFF
--- a/cocos2d/animation/animation-manager.js
+++ b/cocos2d/animation/animation-manager.js
@@ -46,11 +46,12 @@ var AnimationManager = cc.Class({
         }
 
         var events = this._delayEvents;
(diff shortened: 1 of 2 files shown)


cocos2d/animation/animation-manager.js
```diff
@@ -46,11 +46,12 @@ var AnimationManager = cc.Class({
         }
 
         var events = this._delayEvents;
-        for (let i = 0, l = events.length; i < l; i++) {
+        for (let i = 0; i < events.length; i++) {
```
2youyou2, line 49: 回调中还可能会有事件注册，length 可能会变

PR #5270, opened 2019-08-29
remove default export in director.ts & game.ts

JayceLai:


cocos/core/index.ts
```diff
@@ -47,8 +47,8 @@ export * from './utils';
 export * from './data';
 export * from './event';
 export * from './platform';
-export * from './game';
+export { game, Game } from './game';
 export * from './scheduler';
-export * from './director';
```
pandamicro: 这两处不用改

pandamicro (approved):

PR #5285, opened 2019-08-30
New module division

pandamicro:


pandamicro:
https://github.com/cocos-creator/3d-tasks/issues/1327

PR #5289, opened 2019-08-30
fix memory leak in RenderData

sth2018:
fix memory leak in RenderData
修复RenderData中一处内存泄漏问题


jareguo:
谢谢反馈，该问题我们已经修复，目前代码在 https://github.com/cocos-creator/engine/blob/d00dfbda0f5296bf6fc3dd7a766da0b60cde5ca1/cocos2d/renderer/render-data/render-data.js#L82

PR #5290, opened 2019-08-30
Fix circular dependencies and fix build issues

pandamicro:
https://github.com/cocos-creator/3d-tasks/issues/1327


pandamicro:
对构建影响较大，先合并了

PR #5301, opened 2019-08-31
fix compressed texture filter without mipmap

jerrywwl:
Re: cocos-creator/2d-tasks#

Changes:
 *


pandamicro:
https://github.com/cocos-creator/3d-tasks/issues/1287

PR #5326, opened 2019-09-03
Fixed an issue when want to remove an unregistered event, check exist at first

zty8023ys:
Re: cocos-creator/2d-tasks#

Changes:
 * Fixed an issue when want to remove an unregistered event, check exist at first
 * Fixed an issue when project is too huge to let ipc return in 5000ms and make build failed


cocos2d/core/platform/CCAssetLibrary.js
```diff
@@ -144,7 +144,7 @@ var AssetLibrary = {
                     error.errorCode = 'db.NOTFOUND';
                     callback(error);
                 }
-            });
+            }, -1);
```
jareguo, line 147: 👍

PR #5415, opened 2019-09-18
fix graphics can't show in editor

JoneLau:


cocos/core/components/ui-base/ui-transfrom-component.ts
```diff
@@ -187,7 +187,11 @@ export class UITransformComponent extends Component {
 
     public __preload () {
         this.node.uiTransfromComp = this;
-        this.node.layer = Layers.Enum.UI_2D;
+        if (CC_EDITOR) {
+            this.node.layer = Layers.Enum.UI;
```
pandamicro: 需要修改一下
https://github.com/cocos-creator/engine/pull/5416

pandamicro (approved):

PR #5420, opened 2019-09-19
fix widget target haven't UITransformComponent error

LinYunMo:
Re: https://github.com/cocos-creator/3d-tasks/issues/1276
If widget target haven't UITransformComponent，throw a warn.
add check in _registerTargetEvents() and align()


cocos/ui/components/widget-manager.ts
```diff
@@ -476,6 +480,10 @@ export const widgetManager = cc._widgetManager = {
                 return;
             }
 
+            if (!widgetParent.getComponent(UITransformComponent)) {
+                console.warn(widget.node.name + '\'s widget target must have UITransformComponent,please add it in target');
```
pandamicro: 用 cc.warnID

pandamicro (changes requested):

PR #5507, opened 2019-10-10
optimize size of serialized animation clip

shrinktofit:
Re: cocos-creator/2d-tasks#

Changes:
 *


shrinktofit:
This PR aims to optimize the file size of serialized animation clips, especially for skeletal animation clips. In general, this PR change to store animation curve values and animation keys in binary form in comparison of in JSON form.

To achieve this, a helper class `CompactValueTypeArray`(abbr. CVTA) is added. CVTA denotes a binary form array of value types. It currently supports array of value types `number`, `Vec2`, `Vec3`, `Vec4`, `Quat`. When serializing, it's converted to JavaScript typed array view, i.e. `Uint8Array`, `Float32Array` according precision.

This PR is intend to be the **FIRST STAGE** of animation clip optimization. It reduce even 9/10 space for some animation.

In addition to above. `cc.isValid` is exported to module `"cc"`.

shrinktofit:
This PR is cancelled, see https://github.com/cocos-creator/engine/pull/5559

PR #5537, opened 2019-10-14
fix typing errors

shrinktofit:


shrinktofit:
https://github.com/cocos-creator/engine/pull/5538

PR #5595, opened 2019-10-23
Fix Asset._native property definition

pandamicro:


pandamicro:
https://github.com/cocos-creator/3d-tasks/issues/1418

PR #5750, opened 2019-11-15
node event TRS remove

LinYunMo:
Re: https://github.com/cocos-creator/3d-tasks/issues/1224

### DIFF
--- a/cocos/core/platform/event-manager/event-enum.ts
+++ b/cocos/core/platform/event-manager/event-enum.ts
@@ -120,10 +120,18 @@ export enum SystemEventType {
 
     /**
      * @en
-     * The event type for position, rotation, scale changed.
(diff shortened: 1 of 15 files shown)


LinYunMo:
Don't merge now，see：https://github.com/cocos-creator/3d-tasks/issues/1224

LinYunMo:
finished

cocos/core/platform/event-manager/event-enum.ts
```diff
@@ -120,37 +120,21 @@ export enum SystemEventType {
 
     /**
      * @en
-     * The event type for position, rotation, scale changed.
+     * The event type for position, rotation, scale changed.If need to know which event is,please use callback first parameter type and [[Node.TransformBit]]
```
pandamicro: Use the type parameter as [[Node.TransformBit]] to check which part is changed

PR #5791, opened 2019-11-26
Modify the default value of AlphaThreshold.

caryliu1999:
Re: cocos-creator/2d-tasks#2073


SantyWang (approved):

cocos2d/core/components/CCMask.js
```diff
@@ -190,7 +190,7 @@ let Mask = cc.Class({
          * @default 0
```
jareguo: API 也要改

caryliu1999: > API 也要改

API注释跟悬浮提示都修改了

PR #5818, opened 2019-12-02
Support of partial property animation.

shrinktofit:
## Typo correctness, renaming
In discussion from @shrinktofit  and @gameall3d, we have agreed that term "target modifier" and "value adapter" are so vague to comprehension.  We decided that re-call "target modifier" as "target path" and re-call "value adapter" as a factory to produce "value proxy". 

To fit this change, this PR also introduces new API names and mark those renamed as deprecated:
  - Introduces `import('cc').animation.isPropertyPath()` for both `import('cc').isPropertyModifier()`, `import('cc').isElementModifier()`, the later two are marked as deprecated;
  - `import('cc').isCustomModifier()` -> `isCustomPath()`;
  - `import('cc').UniformValueAdapter` -> `UniformProxyFactory`;
  - `import('cc').ComponentModifier` -> `ComponentPath`;
  - `import('cc').HierachyModifier` -> `HierarchyPath`.

## Adjust exports of animation related API
To avoid essential name conflicts, this PR move some rarely used API to namespace `import('cc').animation`:
- Cubic spline curve value constructors: `CubicSplineNumberValue`, `CubicSplineVec2Value`, `CubicSplineVec3Value`, `CubicSplineVec4Value`, `CubicSplineQuatValue`;
- Target path related classes and types: `TargetPath`, `PropertyPath`, `CustomPath`, `ComponentPath`, `HierarchyPath`, `isPropertyPath()`, `isCustomPath()`;
- Value proxy related classes and types: `IValueProxy`, `IValueProxyFactory`, `UniformProxyFactory`.

### State values
Formerly, curves in an animation clip are run independently to produce sampled values. The sampled values then change the states of objects.
This doesn't help if curves want to contribute to only part(s) of the (sampled) value. For example, it might be convenient to have color curves for material effect only r, g and b channel of colors and leave the alpha channel unchanged.

This PR introduces a new concept "state value"(or, common target). "State" means that the values will affect states of object, just like sampled values do. So that state values also have target paths and value proxy, defining how the value impacts the object state. a curve either modifies state of `Node` or state value. State values might be shared by multi-curves.


shrinktofit:
Request review from @gameall3d

shrinktofit:
Commit https://github.com/cocos-creator/engine/pull/5818/commits/3c8be9339b3fd6965931308961b8587d689571d4 optimizes the storage of `easingMethods`. Request review from @gameall3d

cocos/core/animation/index.ts
```diff
@@ -12,19 +18,82 @@ export * from './animation-manager';
...
+cc.UniformCurveValueAdapter = UniformCurveValueAdapter;
+
+/**
+ * Alias of `isPropertyPath(path) && typeof path === 'string'`.
+ * @deprecated Since v1.1.
+ */
+export function isPropertyModifier (path: TargetPath): path is string {
+    return typeof path === 'string';
+}
+cc.isPropertyModifier = isPropertyModifier;
```
pandamicro, line 81: 有必要作为全局函数吗？不应该放在  animation 模块下吗？

pandamicro: 下面几个也是一样

pandamicro (approved):

PR #5874, opened 2019-12-10
init  canvas Rect on preload

LinYunMo:
Re: https://github.com/cocos-creator/3d-tasks/issues/1997


LinYunMo:
RE:https://github.com/cocos-creator/3d-tasks/issues/2008

pandamicro (approved):

PR #5932, opened 2019-12-20
shader version should be specified inside gfx

YunHsiao:
Re: cocos-creator/3d-tasks#

Changelog:
 *


pandamicro (approved):

pandamicro:
https://github.com/cocos-creator/3d-tasks/issues/1838

PR #5934, opened 2019-12-20
v1.1 remove the isBuffer condition from the deserialize class

GengineJS:
Re: cocos-creator/2d-tasks#

Changes:
 * 
1.remove the isBuffer condition from the cocos/core/data/deserialize class 
2.Convert the buffer to string in editor3d


pandamicro (approved):

pandamicro:
请将 pr base  更新到最新，避免这么多无意义的 commit

PR #5936, opened 2019-12-20
Fix compatibility with old-school `cc.Class()`

shrinktofit:


shrinktofit:
This PR fixes that: an `undefined` class constructor is passed to trigger the event `'on-class-registered'` when registering a cc-class in style of `cc.Class(options)` with `ctor` is not specified in `options`. 

This PR also add warning to invocation of `cc.Class(options)` with `__ES6__` not set.

Request review from @VisualSJ

VisualSJ (approved):

PR #5972, opened 2019-12-26
fix cannot set property color of graphics which has only a getter

JoneLau:
https://github.com/cocos-creator/3d-tasks/issues/2071

### DIFF
--- a/cocos/ui/components/graphics-component.ts
+++ b/cocos/ui/components/graphics-component.ts
@@ -180,6 +180,16 @@ export class GraphicsComponent extends UIRenderComponent {
         return this._color;
     }
 
+    set color (value) {
+        if (this._color === value) {
+            return;
+        }
+
+        this._color.set(value);
+        this._updateColor();
+        this.markForUpdateRenderData();
+    }
+
     public static LineJoin = LineJoin;
     public static LineCap = LineCap;
     public impl: Impl | null = null;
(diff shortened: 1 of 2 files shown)


cocos/ui/components/graphics-component.ts
```diff
@@ -180,6 +180,11 @@ export class GraphicsComponent extends UIRenderComponent {
         return this._color;
     }
 
+    set color (value) {
+        // @ts-ignore
+        super.color = value;
```
pandamicro: 貌似不能这么写

cocos/ui/components/mask-component.ts
```diff
@@ -207,12 +203,8 @@ export class MaskComponent extends UIRenderComponent {
...
 
     set srcBlendFactor (value) {
-        if (this._srcBlendFactor === value) {
-            return;
-        }
-
-        this._srcBlendFactor = value;
-        this._updateBlendFunc();
+        // @ts-ignore
+        super.srcBlendFactor = value;
```
pandamicro: 同上，要复制父类 setter 代码

pandamicro (changes requested):

cocos/ui/components/graphics-component.ts
JoneLau: 那这个统一一下？我是去翻了云潇的代码，他那也有这么写的

pandamicro (approved):

PR #5975, opened 2019-12-26
Special process for implementation of animation of uniform component

shrinktofit:
Because material objects allow set specific channels of a uniform. This PR optimizes `animation.UniformProxyfactory` to benefit this feature.


shrinktofit:
Request review from @YunHsiao  @gameall3d

PR #5977, opened 2019-12-26
Fix animation data decode with ArrayBufferView

pandamicro:


pandamicro:
https://github.com/cocos-creator/3d-tasks/issues/1838

YunHsiao (approved):

PR #6024, opened 2020-01-03
fix body enabled after add or remove shapes

JayceLai:


JayceLai:
https://github.com/cocos-creator/3d-tasks/issues/2203

pandamicro (approved):

PR #6070, opened 2020-01-10
fix trail render error (#6046)

pandamicro:


pandamicro:
cocos-creator/3d-tasks#2159

PR #6100, opened 2020-01-14
fixed calculate define offset

2youyou2:
Re: cocos-creator/2d-tasks#

Changes:
 * fixed calculate define offset

### DIFF
--- a/cocos2d/renderer/core/program-lib.js
+++ b/cocos2d/renderer/core/program-lib.js
@@ -4,22 +4,12 @@ import gfx from '../gfx';
 
 let _shdID = 0;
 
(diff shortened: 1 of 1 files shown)


cocos2d/renderer/core/program-lib.js
```diff
@@ -137,8 +127,8 @@ export default class ProgramLib {
         }.bind(def);
       }
 
-      offset += cnt;
       def._offset = offset;
+      offset += cnt;
```
2youyou2, line 131: 当前间隔应该放到下一个 offset 上

SantyWang (approved):

PR #6139, opened 2020-02-09
fixed zoom ratio when alignWithScreen is false

2youyou2:
Re: cocos-creator/2d-tasks#

Changes:
 * fixed zoom ratio when alignWithScreen is false, orthoSize should be less when zoomRatio is bigger

### DIFF
--- a/cocos2d/core/camera/CCCamera.js
+++ b/cocos2d/core/camera/CCCamera.js
@@ -782,7 +782,7 @@ let Camera = cc.Class({
             fov = Math.atan(Math.tan(fov / 2) / this.zoomRatio) * 2;
             this._camera.setFov(fov);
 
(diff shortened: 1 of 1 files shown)


cocos2d/core/camera/CCCamera.js
```diff
@@ -782,7 +782,7 @@ let Camera = cc.Class({
             fov = Math.atan(Math.tan(fov / 2) / this.zoomRatio) * 2;
             this._camera.setFov(fov);
 
-            this._camera.setOrthoHeight(this._orthoSize * this.zoomRatio);
```
SantyWang, line 785: 这个之前没有测视例测出问题么？

2youyou2: 是的

SantyWang (approved):

PR #6142, opened 2020-02-10
fix preview when physics module not include

JayceLai:


JayceLai:
https://github.com/cocos-creator/3d-tasks/issues/2447

pandamicro (approved):

PR #6203, opened 2020-02-21
Stop handle asset in Terrain restore

MSoft1115:
Re: cocos-creator/2d-tasks#

Changes:
 *


GengineJS:
解决了该bug  
https://github.com/cocos-creator/3d-tasks/issues/2429

PR #6256, opened 2020-03-02
Build time constants

shrinktofit:
This is about how to use build-time constants in code. In former, we use global variables and depend on `uglify` to eliminate the dead code. This PR however introduces a new method: use modules. Instead of polluting the global namespace, it place constants into a single module named "internal:constants". Those constants are declared as `const`. The tree-shaking algorithm which is commonly used by rollup and webpack can eliminate the dead code therefor.


shrinktofit:
Editor: https://github.com/cocos-creator/editor-3d/pull/3153

pandamicro (approved):

PR #6263, opened 2020-03-02
Update engine version

LinYunMo:


LinYunMo:
Re：https://github.com/cocos-creator/3d-tasks/issues/2663

PR #6430, opened 2020-03-26
Fix API regulation breaker in event module

pandamicro:
https://github.com/cocos-creator/3d-tasks/issues/2336


shrinktofit:
Don’t forget to announce this breaking change

pandamicro:
It only contains modifications to APIs started with `_` which means they were "private", and they were designed and only used by InputManager internally, so it's ok to just remove "private" APIs or make them public

PR #6442, opened 2020-03-27
add atob polyfill for ammo

JayceLai:
Re: cocos-creator/3d-tasks#2843

Changes:
 *

### DIFF
--- a/exports/physics-ammo.ts
+++ b/exports/physics-ammo.ts
@@ -2,6 +2,35 @@
  * @hidden
  */
 
+// polyfill
+if (window.atob == null) {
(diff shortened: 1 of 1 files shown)


exports/physics-ammo.ts
```diff
@@ -2,6 +2,35 @@
  * @hidden
  */
 
+// polyfill
```
pandamicro: 需要注明在哪些平台需要适配

pandamicro (changes requested):

pandamicro (approved):

PR #6468, opened 2020-04-01
en api docs of tween

JayceLai:
Re: cocos-creator/3d-tasks#2506

Changes:
 *

### DIFF
--- a/cocos/tween/tween.ts
+++ b/cocos/tween/tween.ts
@@ -40,9 +40,9 @@ export class Tween {
 
     /**
      * @en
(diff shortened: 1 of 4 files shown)


cocos/tween/tween.ts
```diff
@@ -59,9 +59,9 @@ export class Tween {
 
     /**
      * @en
-     * Set tween target
+     * Sets tween target.
      * @zh
-     * 设置 tween 的 target
+     * 设置 tween 的 targ。et
```
2youyou2: 。号打错地方了吧

PR #6478, opened 2020-04-02
Animation state bug fix

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Fix a potential bug that  a sequence of invocations on animation states(with skeletal-liked curves) like:
```ts
.play() // Create blend state writers
.pause() // Set the `IPlayable` state to "Paused"
.initialize() // Blend state writers are reset here
.play() // BUG: `.onResume()` is called instead of `onPlay()`, so the blend state writers would not be set to started anymore
```
causes that animation values are not assigned to target.

Request review from @gameall3d


gameall3d (approved):
LGTM

pandamicro (approved):

PR #6517, opened 2020-04-09
Build engine V3.0.0-beta.0

shrinktofit:
Re: cocos-creator/3d-tasks#

**This is an early drafted change**.

Changelog:
 * Now the result of `build-engine` become a serials of modules and an import-map, rather than a single "cc" module. This change has the purpose to modularize the engine.


shrinktofit:
Please review this change @yanOO1497

PR #6531, opened 2020-04-11
refine d.ts

xduooo:
Re: cocos-creator/2d-tasks#

Changes:
 * Some data types or return values are normalized to prevent error prompts.

### DIFF
--- a/cocos2d/actions/tween.js
+++ b/cocos2d/actions/tween.js
@@ -215,9 +215,9 @@ Tween.stopAllByTarget = function (target) {
  * Insert an action or tween to this sequence
  * !#zh
  * 插入一个 action 或者 tween 到队列中
- * @method then 
- * @param {Action|Tween} other
- * @return {Tween}
+ * @method then
+ * @param {Action|Tween<T>} other
+ * @return {Tween<T>}
  */
 Tween.prototype.then = function (other) {
     if (other instanceof cc.Action) {
@@ -237,7 +237,7 @@ Tween.prototype.then = function (other) {
  * 设置 tween 的 target
  * @method target
  * @param {Object} target
- * @return {Tween}
+ * @return {Tween<T>}
  */
 Tween.prototype.target = function (target) {
     this._target = target;
@@ -250,7 +250,7 @@ Tween.prototype.target = function (target) {
  * !#zh
  * 运行当前 tween
  * @method start
- * @return {Tween}
+ * @return {Tween<T>}
  */
 Tween.prototype.start = function () {
     let target = this._target;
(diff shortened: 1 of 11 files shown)


SantyWang:
Cool, Thank you very much!

xduooo:
There are still mistakes I haven't modified. Please see the error list.

base: branch v2.4.0  [creator_engine_dts_errs.zip](https://github.com/cocos-creator/engine/files/4468145/creator_engine_dts_errs.zip)

cocos2d/core/value-types/color.ts
```diff
@@ -185,7 +185,7 @@ export default class Color extends ValueType {
      * Set the components of a color to the given values.
      * @method set
      * @typescript
-     * set (out: Color, r = 255, g = 255, b = 255, a = 255): Color
+     * set (out: Color, r?: number, g?: number, b?: number, a?: number): Color
```
2youyou2, line 188: Why to remove the default param value

xduooo: When we compile typescript with `tsc`command, we will get this error :

> error TS2371: A parameter initializer is only allowed in a function or constructor implementation.

xduooo: Microsoft is also discussing a better way to express default values ( https://github.com/Microsoft/tsdoc/issues/151 ),  but now it's not the right way:
`set (out: Color, r = 255, g = 255, b = 255, a = 255): Color
`

2youyou2: I see, thanks.

2youyou2 (approved):

cocos2d/actions/tween.js
```diff
@@ -215,9 +215,9 @@ Tween.stopAllByTarget = function (target) {
  * Insert an action or tween to this sequence
  * !#zh
  * 插入一个 action 或者 tween 到队列中
- * @method then 
- * @param {Action|Tween} other
- * @return {Tween}
+ * @method then
+ * @param {Action|Tween<T>} other
+ * @return {Tween<T>}
```
SantyWang, line 220: I don't think api-document-generator can recognize genericity in `param` and `return`. It's better to add a decorator 'typescript' likes below
```
@typescript then (othen: Action|Tween<T>): Tween<T>
```

xduooo: I have tested it with the following tools, it works:
https://github.com/cocos-creator/creator-api-docs and
https://github.com/cocos-creator/firedoc

It is an internal workflow, so you can decide which one is better.

zhefengzhang:
I dealt with this issue on the basis of this PR and submitted a new PR：
image

SantyWang:
Thanks for your contrubution

PR #6550, opened 2020-04-14
fix sampler of splash screen

JayceLai:
Re: cocos-creator/3d-tasks#

Changes:
 * fix sampler of splash screen
 * record command buffer only one times
 * enable retina & resize for splash screen

### DIFF
--- a/cocos/core/splash-screen.ts
+++ b/cocos/core/splash-screen.ts
@@ -111,7 +111,12 @@ export class SplashScreen {
             return;
         } else {
             cc.view.enableRetina(true);
-            cc.view.setDesignResolutionSize(960, 640, 4);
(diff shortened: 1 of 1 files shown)


cocos/core/splash-screen.ts
```diff
@@ -108,6 +110,8 @@ export class SplashScreen {
             this._directCall = true;
             return;
         } else {
+            cc.view.enableRetina(true);
+            cc.view.setDesignResolutionSize(960, 640, 4);
```
pandamicro: This may overwrite user setting

pandamicro (changes requested):

pandamicro (approved):

PR #6564, opened 2020-04-16
Fix cross fade bug

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Fix a bug when cross fade from/to animations having finite playing time.

**Remark**
The bug might be triggered if the animations under cross fading contain an animation state which:
- has finite playing time. For example, when set to `Normal` wrap mode and its duration is less than cross fade duration. Or
- would be manually stopped or paused while the cross fading will not have been finished.

In either case, the stopped/paused state would not be sampled. The sum weight of blending property would not be 1 therefor.

This PR checks if the cross fading states are stopped/paused. If they do, `sample()` is called to manually sample the state.

### DIFF
--- a/cocos/core/animation/cross-fade.ts
+++ b/cocos/core/animation/cross-fade.ts
@@ -27,7 +27,7 @@ export class CrossFade extends Playable {
     }
 
     public update (deltaTime: number) {
-        if (!this.isPlaying || this.isPaused) {
+        if (this.isMotionless) {
             return;
         }
 
@@ -70,6 +70,14 @@ export class CrossFade extends Playable {
                 }
             }
             this._fadings.splice(deadFadingBegin);
+        }
+
+        if (this._managedStates.length !== 0) {
+            this._managedStates.forEach((state) => {
+                if (state.state?.isMotionless) {
+                    state.state.sample();
+                }
+            });
         }
     }
 
--- a/cocos/core/animation/cross-fade.ts
+++ b/cocos/core/animation/cross-fade.ts
@@ -72,12 +72,11 @@ export class CrossFade extends Playable {
             this._fadings.splice(deadFadingBegin);
         }
 
-        if (this._managedStates.length !== 0) {
-            this._managedStates.forEach((state) => {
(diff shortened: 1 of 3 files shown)


gameall3d (approved):
LGTM

cocos/core/animation/cross-fade.ts
```diff
@@ -71,6 +71,14 @@ export class CrossFade extends Playable {
             }
             this._fadings.splice(deadFadingBegin);
         }
+
+        if (this._managedStates.length !== 0) {
+            this._managedStates.forEach((state) => {
```
pandamicro: This is too heavy in an update function, please modify to use simple for loop, so that there won't be any function creation or invocation

cocos/core/animation/playable.ts
```diff
@@ -23,6 +23,15 @@ export class Playable {
     get isPaused () {
         return this._isPaused;
     }
+
+    /**
+     * @en Is paused or stopped?
```
pandamicro: Whether the current playable is stopped or paused

cocos/core/animation/playable.ts
```diff
@@ -23,6 +23,15 @@ export class Playable {
...
+     * @zh 是否已被暂停或停止。
+     */
+    get isMotionless () {
```
pandamicro, line 31: isStill or isStatic seems simpler

shrinktofit: "Still" means not moving clam and quiet,
"Motionless" means not in physicial motion,
"Static" can be ambiguous with "static model".

cocos/core/animation/playable.ts
shrinktofit: Updated.

cocos/core/animation/cross-fade.ts
shrinktofit: Updated.

pandamicro (approved):

pandamicro:
ci is not working ... ignore

PR #6571, opened 2020-04-17
Implements EventTarget in a better way

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * This PR proposes an alternative method so called "real mix-in" to implement integrating event target functionalities into other classes.

**PROS**
- The real mix-in technique make us construct a better and clear code structure;
- It save us from redundant and meaningless code repetition.

In addition, I have replace `AnimationComponent` into this implementation.

Please give it a review on this change @pandamicro  @jareguo .


shrinktofit:
I would submit a performance report and add some unit tests before this can be passed.

pandamicro:
I have gone through the code, concluded the following pros and cons, please comment if I missed anything

Pros:
- No need to use `applyMixins`
- Can use super when EventTarget functions needed to be overwrite in subclass

Cons
- This reimplemented `EventTarget` and `ICallbackTable` with `EventIntegrated` and `EventEntry`
- This adds an inheritance level

The first con is what I'm strongly against, all EventTarget like behavior should use unified implementation. Also, a simple Map as CallbackTable is simpler than EventEntry abstraction, EventEntry class could lead to more function calls and more runtime cost.
The second con also has an important impact, because this means we also need to add inheritance level to Asset which is base class for all assets. At the end of the day, the most important thing is what's running in the runtime, an extra level of inheritance is more expensive than simple property functions.

I think the goal here is to cleanup our Event architecture. Not how beautiful the code should look like, but more importantly, how to solve the pain of using `applyMixins` and `EventTarget.prototype.on.call` etc. Again, it's only our internal pain, it's totally transparent to users, which is the reason why I think this problem is not critical.
I think the ultimate solution should be use composition design instead of inheritance design, like in Node. And the problem is how to plant these key functions and `callbackTable` of EventTarget into the target object.

shrinktofit:
> it's only our internal pain, it's totally transparent to users

It's not only an internal pain. The event APIs are full accessible to developers. They may also use the event entities. It's very weird to teach users how you can implement an event target:

1. You should have a variable member named `_callbackTable` in your class, this member should be created with `js.createMap(true)` or by yourself `Object.create(null)`;
2. You should implement APIs of what an `EventTarget` should be. Note, you should implement all. Even in most cases you just need a `super.x` invocation, you just need to write it down.
3. You should call the `applyMixins` right after your class definition.

---

> The second con also has an important impact, because this means we also need to add inheritance level to Asset which is base class for all assets

This PR do introduce a new level of class inheritance. But what proves that adding a class inheritance break performance seriously?

---

>  EventEntry class could lead to more function calls and more runtime cost.

The `CallbackEntry/EventEntry` should not be cared however, it's just a rename of `CallbackList/CallbackEntry`. **The methods added are just the code that originally lie in methods of `EventTarget`**, which was actually un-conformed to code design pattern --- `EventTarget` shall not directly use private members of `CallbackList/CallbackEntry`.

pandamicro:
> It's not only an internal pain. The event APIs are full accessible to developers. They may also use the event entities.

If they cannot inherit from EventTarget, then they can use it as a composition part like EventHandler in Node

> This PR do introduce a new level of class inheritance. But how can you say that add a class inheritance break performance seriously?

Not seriously, but still it's adding another level for event related functions. I'm especially sensitive here because Event System is a fundamental design which affects assets, components etc, the complexity we add will eventually have impact (even minor most of the time) in runtime. This also brings the point I want to make about EventEntry abstraction:

> **The methods added are just the code that originally lie in methods of `EventTarget`**, which was actually un-conformed to code design pattern --- `EventTarget` shall not directly use private members of `CallbackList/CallbackEntry`.

First of all, `_callbackTable` is a protect property in EventTarget, I think there is no violation. But I do think the current abstraction of `CallbackList`, `CallbacksInvoker` & `EventTarget` can be improved. I just think we cannot add another alternative design of EventTarget which is shown in this PR, I'd like to see EventTarget be directly refactored instead.

About the direction of improvement, I think the most important thing is not design cleanness, but should be the simplicity and performance in runtime. This is the same idea as I sad in previous point, EventTarget as a fundamental design, I want it to be as simple and performant as possible. That's why I said a simple Map as CallbackTable is simpler than EventEntry abstraction.

So there are two things we can discuss and have an agreement later on:

1. How to make `EventTarget` be composed into an existing Class without inheritance
2. How to simplify `EventTarget/CallbacksList` abstraction

shrinktofit:
> About the direction of improvement, I think the most important thing is not design cleanness, but should be the simplicity and performance in runtime.

You claim simplicity is critical. But the 'composition' is not a way to simplicity. It increases the complexity of usage.

------------------------------------

> I'm especially sensitive here because Event System is a fundamental design

This PR respects the idea with no doubt, doesn't it?

------------------------------------

> How to simplify EventTarget/CallbacksList abstraction

No, this PR DO NOT care about how `EventTarget` is using `CallbacksList`. It just optimize by the way. If you have another thoughts about `CallbackList`, let's discuss in another place.

------------------------------------

It seems like you're mainly worrying about the performance.

First, I had been said that I would submit a **performance report**. On the other hand, there is currently **no proof** that "insert a hierarchy" hurts a lot more than "insert manually a serials of prototype methods and insert a member declaration as well as its initialization".

Let's rethink what `EventIntegrated(constructor)` should mean to developers. It was defined as:
**Let the objects constructed by the `constructor` having the ability to process events.**

Yes, no more details about how it implements this function --- Even this PR proposes a heritage way,
you can implement it like:

```js
function EventIntegrated(constructor) {
    // But wait: why do we must do like this? 
    constructor.prototype.on = /* */
}
```

pandamicro:
> You claim simplicity is critical. But the 'composition' is not a way to simplicity. It increases the complexity of usage.

What I believe is "Simpler is faster". I don't want the complexity of usage neither, and I don't have a perfect proposition of composition solution which provide simplicity for both usage and architecture at the moment.

> First, I had been said that I would submit a **performance report**.

I think it can resolve my concern, I prefer the performance report based on mobile native and mobile wechat mini game. As long as it's not worse, I can accept the inheritance way in this PR

> No, this PR DO NOT care about how `EventTarget` is using `CallbacksList`. It just optimize by the way. If you have another thoughts about `CallbackList`, let's discuss in another place.

I will insist on this one. It is related, if this PR get merged, we will have two implementations for event system, I can't let that happen, this PR should resolve the duplicated implementation issue or another PR should be proposed and merged in advance.

jareguo (approved):

jareguo:
Sorry for my Chinglish.
I agree the need to refactor that EventTarget. But I have not found a grace/optimal implementation. (This PR still proprosed a more standard way than before, so I approved)

Let's think about such kind of implementation:

```ts
interface EventManager {
  on (type: string, target: object, callback: Function) {}
  off (type: string, target: object, callback?: Function,) {}
  emit (type: string, ...params: any[]) {}
}
```

After that:

```ts
class MyClass () {
    // NO mixins or inheritance anymore, 
    // Use freely on every object, and no memory overhead if an instance will not have events

    constructor () {
        EventManager.on('mouse-move', this, this.OnMouseMove);
    }
    OnDestroy () {
        EventManager.off('mouse-move', this);
    }
    OnMouseMove (event: Event, ...params: any[]) {
        
    }
}
```

There can be many instances of EventManager, we will provide some managers based on it:
(So event names will not duplicated and performance will be better (but still slower than old implementation))

```ts
cc.systemEvent = new EventManager();
cc.systemEvent.on('mouse-move', this, this.OnMouseMove);

cc.assetManager = new EventManager();
cc.assetManager.on('loaded', this, this.Refresh);
```

Users can create their manager in the same way

```ts
let gameManager = new EventManager();
gameManager.emit('died', enemyID);
```

jareguo:
Further, what if we provide a filter for listeners?

```ts
interface EventManager {
  on (type: string, sender: object, listener: object, callback: Function) {}
  off (type: string, listener: object, callback?: Function,) {}
  emit (type: string, sender: object, ...params: any[]) {}
}

cc.assetManager = new EventManager();
cc.assetManager.once('loaded', spriteFrame.texture, spriteFrame, spriteFrame.refresh);
// Totally the same as
// spriteFrame.texture.on('loaded', spriteFrame.refresh, spriteFrame);
```

jareguo:
Finally, what if the EventManager can be inspected and serialized?
(Every listeners are scene component)

```ts
class GamePlay extends Component {
    this.manager = new EventManager();
}
let { manager } = node.addComponent(GamePlay);
```

Then it will be easy to debug at runtime, and not worry about memory leak, just `node.destroy()`
And designer can use our editor to register events in Properties panel (and save into the scene)

pandamicro:
The need here is a distributed event system instead of a centralized system. In Cocos2dx, and Cocos2d-js we were using such centralized system, and decide to move to distributed mode in Cocos Creator. The benefit are

- Performance in JS is better because we can access directly to the listener list in the node or target.
- APIs are more user friendly, they can user the Node or target API to add listener directly
    ```
    node.on('event', callback);
    eventManager.addListener(node, 'event', callback);
    ```
- Different needs can be satisfied, for example, EventTarget can be customized in inherited classes.

So I think we can agree that distributed mode is better.

In this case, listeners data must be added as property in the EventTarget instance. The functions like `on`, `off` also should be direct functions. I can't figure out a way to avoid presence in the inheritance chain. In this case, I agree with the main idea in this PR.

What's remaining to resolve is how to simplify EventTarget class and CallbacksList

shrinktofit:
Hi @pandamicro  I have written a performance test.

I have been tested it on chrome 81.0.4044.138, with following test cases:
For both instances of class which is implemented either as this PR proposed or in the classic manner:
- construct such instances;
- call event-related methods on these instance;
- call methods on "real" super class.

For each test case, I tested it on both ES5 and ES6.

The following is the report table(with 1000, 000 iterations).

**ES5**
|  | Proposed cost(ms) | Classic cost(ms) |
| -- | --|--|
|Construction | 349.965000001248 | 391.69000000401866|
| Event methods invocation | 369.0599999972619 | 764.5100000081584|
| Base methods invocation | 7.549999994807877 | 6.775000001653098|


**ES6**
|  | Proposed cost(ms) | Classic cost(ms) |
| -- | --|--|
|Construction | 86.53999998932704 | 88.3099999919068|
| Event methods invocation | 357.6699999975972 | 541.1750000057509 |
| Base methods invocation | 7.165000002714805 | 6.330000003799796 |

Look through the table, we can see that:
- When call to event methods, the way this PR proposed is more fast; This should be subjected to that this PR eliminates the indirect calls like `Event.prototype.xx.call(this, ...)`;
- The more fast construction is  still better in proposed way on ES5; but it's indistinguishable on ES6;
- Base methods invocation proves that: even this PR introduce of an extra prototype chain, the cost can be ignored however.

[ES5.zip](https://github.com/cocos-creator/engine/files/4633043/ES5.zip)
[ES6.zip](https://github.com/cocos-creator/engine/files/4633044/ES6.zip)

------ 

To run either case, extract the files and run them use the following HTML:
```html
<html>
	<body>
		<script>
			globalThis.cc = {};
			globalThis.cc.internal = {};
		</script>
		<script src="./cc.js">
		</script>
	</body>
</html>
```

shrinktofit:
@pandamicro  I have now force-pushed:
- I've rolled back all changes applied to the `CallbacksInvoker`(But in the future I may open another PR to track those changes);
- The name `EventIntegrated` is not nice, does it? I've suggested with a `Eventful`.

@JayceLai  Please review the changes made to  `collider-component.ts`. May you need to move the docs into another place?

cocos/core/event/event-target.ts
```diff
@@ -28,170 +28,13 @@
...
 
-const fastRemove = js.array.fastRemove;
+// eslint-disable-next-line @typescript-eslint/no-extraneous-class
+class Empty {};
 
-export interface ITargetImpl extends Object {
-    __eventTargets?: Object[];
-    node?: ITargetImpl;
-}
+export const EventTarget = Eventful(Empty);
```
pandamicro: Suggest to use:

export const EventTarget = Eventful(Object);

cocos/core/event/event-target.ts
```diff
@@ -28,170 +28,13 @@
...
-                    targetImpl.__eventTargets.push(this);
-                } else if (targetImpl.node && targetImpl.node.__eventTargets) {
-                    targetImpl.node.__eventTargets.push(this);
-                }
-            }
-        }
-        return callback;
-    }
-}
+export type EventTarget = IEventful;
```
pandamicro: This type name is the same with previous class name, I don't think it can work properly

cocos/core/event/eventful.ts
```diff
@@ -0,0 +1,198 @@
...
+ * class Base { say() { console.log('Hello!'); } }
+ * class MyClass extends Eventful(Base) { }
+ * function (o: MyClass) {
+ *     o.say(); // Ok: 继承自 `Base`
+ *     o.emit('sing', 'The ghost'); // Ok: `MyClass`类的对象实现了 IEventful
+ * }
+ * ```
+ */
+export function Eventful<TBase> (base: Constructor<TBase>): Constructor<TBase & IEventful> {
+    return class extends (base as unknown as any) implements IEventful {
```
pandamicro: Just want to confirm the implementation is identical to the previous EventTarget ?

cocos/core/event/eventful.ts
```diff
@@ -0,0 +1,198 @@
+import { fastRemove } from '../utils/array';
+import { CallbacksInvoker } from './callbacks-invoker';
+
+type Constructor<T = {}> = new (...args: any[]) => T;
+
+type EventType = string;
+
+/**
+ * 实现该接口的对象具有处理事件的能力。
+ */
```
pandamicro: please be more specific in doc and provide English version too

cocos/core/event/eventful.ts
```diff
@@ -0,0 +1,198 @@
...
+ * ```ts
+ * class Base { say() { console.log('Hello!'); } }
+ * class MyClass extends Eventful(Base) { }
+ * function (o: MyClass) {
+ *     o.say(); // Ok: 继承自 `Base`
+ *     o.emit('sing', 'The ghost'); // Ok: `MyClass`类的对象实现了 IEventful
+ * }
+ * ```
+ */
+export function Eventful<TBase> (base: Constructor<TBase>): Constructor<TBase & IEventful> {
```
pandamicro: I don't think Eventful is ideal, I propose 

IEventful -> IEventTarget
Eventful -> EventTargetExtended
EventTarget = EventTargetExtended(Object)

Make things clearer

cocos/core/event/event-target.ts
shrinktofit: Yep, it works and it's necessary~

cocos/core/event/eventful.ts
shrinktofit: They are all implemented based on `CallbacksInvoker`.

cocos/core/event/eventful.ts
shrinktofit: `EventTargetExtended` is so wordy

cocos/core/event/event-target.ts
shrinktofit: Good idea

shrinktofit: https://github.com/cocos-creator/engine/pull/6571/commits/3d41c70064d74c10a7b5ece2da5593ca9e8ac877

shrinktofit: I would have to revert the changes. `extends Object` causes problem: https://circleci.com/gh/cocos-creator/engine/16178?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
Reason: https://javascript.info/task/class-extend-object

pandamicro (approved):

cocos/core/assets/asset.ts
```diff
@@ -59,7 +55,7 @@ import { Node } from '../scene-graph';
  * @extends RawAsset
  */
 @ccclass('cc.Asset')
-export class Asset extends RawAsset implements IEventTarget {
+export class Asset extends Eventful(RawAsset) {
```
jareguo: image
Eventful has its different meanings. So I suggest to use some more abstract names, such as Eventify

pandamicro: 赞同

jareguo:
This PR looks good to me. It really imporves underlying implementations.
But I still hope we can try to jump out of the panel. 

As we know, all the problems are focused on supporting mixin in TypeScript. But why? Why TS does not support? Why only Creator has to do such complex workaround? (This PR still more better than before)

I think it's because combination is better than inheritance. I often rethink about out inheritance-based design, like EC, like CCObject, and EventTarget.

> So I think we can agree that distributed mode is better.

Even distributed mode can still use the combination way rather than inheritance, so I think we can discuss this problem later.

jareguo (approved):

jareguo:
```ts
If we still like the distributed manner, here is another proposal, the signal-slot type api:

interface IDelegate {
  on (callback: Function, target?: object) {}
  off (callback: Function, target?: object) {}
  emit (...params: any[]) {}
}

interface IEventManager {
  onMouseMove: IDelegate
}

class EventManager implements IEventManager () {
  private _onMouseMove: Delegate<event: MouseEvent, pos: Vec2> = null;
  get onMouseMove () {
    return this._onMouseMove || (this._onMouseMove = new Delegate<event: MouseEvent, pos: Vec2>());
  }

  dispatchMouseEvent (event) {
    this._onMouseMove && this._onMouseMove.emit(event, event.position);
  }
}

class MyClass () {
    start () {
        EventManager.onMouseMove.on(this.OnMouseMove, this);
    }
    OnDestroy () {
        EventManager.onMouseMove.off(this.OnMouseMove, this);
    }
    OnMouseMove (event: MouseEvent, pos: Vec2) {
        
    }
}
```

如果 3.0 开始要改成该方案，需要检查一下 node 定制过的这些实现是否满足需要，还有 once 的实现，事件派发、传递是否会出问题。

PR #6612, opened 2020-04-24
update @cocos/cannon to 1.1.0

JayceLai:
Re: cocos-creator/3d-tasks#3208

Changes:
 * fix collision event


JayceLai:
需要执行 npm install

PR #6620, opened 2020-04-27
Optimize non-morph plain ModelComponent

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Optimizes non-morph plain model components --- a plain `Model` is used to create the model instead of `MorphModel` for those model components.
 * Conditionally display "enableMorph" --- Display "enableMorph" only if there is a valid morph mesh.

This PR should resolve https://github.com/cocos-creator/3d-tasks/issues/2797 @YunHsiao  Please take a look at this.


YunHsiao (commented):
要是能完全不影响 model component 就好了

YunHsiao (approved):

PR #6674, opened 2020-05-11
Fix a language semantic violation

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Fix a language semantic violation: can not call class constructor as a function

### DIFF
--- a/cocos/core/event/callbacks-invoker.ts
+++ b/cocos/core/event/callbacks-invoker.ts
@@ -332,6 +332,13 @@ export class CallbacksInvoker {
             }
         }
     }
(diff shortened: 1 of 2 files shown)


pandamicro:
Adding `clear` to CallbacksInvoker lead IEventTarget to require `clear` function and ci fails because it wasn't implemented in other classes implement IEventTarget

shrinktofit:
> because it wasn't implemented in other classes implement IEventTarget

#6571 should help to this.

pandamicro (approved):

PR #6680, opened 2020-05-12
Rename internal usages of global variable 'cc' into 'legacyGlobalExpo…

shrinktofit:
…rts'

Re: cocos-creator/3d-tasks#

Changes:
 * Rename internal usage of `cc` into `legacyGlobalExports`.
 This is the fundamental before we can unify the `await import('cc')` and `globalThis.cc` in the future.
 * Add import statements to clarify the dependencies on `legacyGlobalExports`, not on the fly.


cocos/audio/assets/clip.ts
```diff
@@ -33,6 +33,7 @@ import { Enum } from '../../core/value-types';
 import { AudioPlayer, PlayingState } from './player';
 import { AudioPlayerDOM } from './player-dom';
 import { AudioPlayerWeb } from './player-web';
+import { legacyGlobalExports } from '../../core/global-exports';
```
pandamicro: The name is too long and not friendly to the package size (consider how frequent it's used), I suggest `legacyCC` instead

pandamicro (commented):
Generally I agree with the direction of modification

cocos/audio/assets/clip.ts
shrinktofit: OK, that's fine. It's now in https://github.com/cocos-creator/engine/pull/6680/commits/4ff2b482504fb56d76be93ac2da2a63b025be778 . By the way, the name of this kind of module variable would be uglified. So please never mind.

pandamicro: Oh, I didn't know that

pandamicro (approved):

PR #6692, opened 2020-05-15
planar shadow instancing support

YunHsiao:
Re: cocos-creator/3d-tasks#

Changelog:
 * joint texture attributes should be updated on material change, otherwise dynamically toggling instancing by changing materials will break baked skeletal animations.
* light buffer detection should check the initial shader properties or the builtin materials won't work (no pipeline when initializing, thus no bindings)
* planar shadow now supports instancing (automatically enabled when the model itself is using instancing)

### DIFF
--- a/cocos/core/gfx/define.ts
+++ b/cocos/core/gfx/define.ts
@@ -794,7 +794,6 @@ export function GFXFormatSurfaceSize (
         size += GFXFormatSize(format, width, height, depth);
         width = Math.max(width >> 1, 1);
         height = Math.max(height >> 1, 1);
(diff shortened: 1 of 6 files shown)


troublemaker52025:
If the transform of the instancing model has been changed, how to notify the planer shadow?

YunHsiao:
> If the transform of the instancing model has been changed, how to notify the planer shadow?

All the local dynamic data are shared between planar shadow & model itself by design so that we can get away with no sychronization hassle.

troublemaker52025 (approved):

pandamicro (approved):

PR #6695, opened 2020-05-15
Prefer array notation for all array types

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Edit lint rule `@typescript-eslint/array-type`

cc @YunHsiao


YunHsiao (approved):

pandamicro:
Prefer using `nodes: Array<Node>` instead of `nodes: Node[]`?

What's the advantage?

shrinktofit:
> Prefer using `nodes: Array<Node>` instead of `nodes: Node[]`?

Array notation means `Node[]`

PR #6705, opened 2020-05-18
update gfx interface

YunHsiao:
Re: cocos-creator/3d-tasks#

Changelog:
 *


pandamicro (approved):

cocos/core/gfx/device.ts
```diff
@@ -505,6 +505,12 @@ export abstract class GFXDevice {
      */
     public abstract createWindow (info: IGFXWindowInfo): GFXWindow;
 
+    /**
+     * @en Begin current frame.
+     * @zh 开始当前帧。
+     */
+    public abstract begin (): void;
```
minggo: Should add end() to make API more symmetrical?

YunHsiao: there's already a `present` serving as end. To be more symmetrical we can use `acquire` instead of begin

minggo: Yep, i think it is better.

minggo (approved):

PR #6708, opened 2020-05-18
not clear ia information every frame

minggo:
Changes: 
- UIDrawBatch
  - not clear ia information every frame
  - remove duplicated information


minggo:
If clear ia information in update, then render stage can not get correct ia data to draw UI elements.

LinYunMo (approved):

pandamicro (approved):

YunHsiao (approved):

PR #6713, opened 2020-05-19
Not record command buffer in submodel

minggo:
This PR remove recording render commands in sub model. After this PR, will remove generating PSO in model.

### DIFF
--- a/cocos/core/renderer/ui/ui-batch-model.ts
+++ b/cocos/core/renderer/ui/ui-batch-model.ts
@@ -46,7 +46,6 @@ export class UIBatchModel extends Model {
     public updateTransform () {
     }
 
-
     public updateUBOs () {
         return false;
     }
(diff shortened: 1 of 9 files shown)


YunHsiao (approved):

caryliu1999 (approved):

LinYunMo (approved):

cocos/core/renderer/ui/ui-batch-model.ts
```diff
@@ -73,8 +73,5 @@ class UISubModel extends SubModel {
     }
 
     public destroy () {
```
pandamicro, line 75: Please delete the destroy function if no logic needed

pandamicro (approved):

cocos/core/renderer/ui/ui-batch-model.ts
minggo: It is a public function, it may be used some where. So i kept this to avoid issues. And can add needed logic in future.

PR #6717, opened 2020-05-20
fix WebView opacity invalid on iOS

knoxHuang:
Re: https://forum.cocos.org/t/2-3-3-bug/92994

Changes:
 * fix webview opacity invalid on iOS

### DIFF
--- a/cocos2d/webview/webview-impl.js
+++ b/cocos2d/webview/webview-impl.js
@@ -122,7 +122,6 @@ let WebViewImpl = cc.Class({
         if (WebViewImpl._polyfill.enableBG)
             this._div.style["background"] = "#FFF";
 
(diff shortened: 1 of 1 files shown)


PPpro (approved):

cocos2d/webview/webview-impl.js
```diff
@@ -122,7 +122,6 @@ let WebViewImpl = cc.Class({
         if (WebViewImpl._polyfill.enableBG)
             this._div.style["background"] = "#FFF";
 
-        this._div.style["background"] = "#FFF";
```
knoxHuang, line 125: The transparency of the iframe setting its child node is invalid because of the background color setting on dev

PR #6721, opened 2020-05-20
fix skinning model pose calc when missing joint nodes

YunHsiao:
Re: cocos-creator/3d-tasks#

Changelog:
 * When joint node is not present, bounding boxes should still be calculated or culling won't work properly

### DIFF
--- a/cocos/core/renderer/models/skeletal-animation-utils.ts
+++ b/cocos/core/renderer/models/skeletal-animation-utils.ts
@@ -56,7 +56,7 @@ export function selectJointsMediumFormat (device: GFXDevice): GFXFormat {
 }
 
 // negative zeros cannot be decoded correctly at GLSL 100 minimum highp float precision, 1/1024
-// and it has a significant effect on the final transformation
+// and it has a significant impact on the final transformation
 function makeStable (n: number) { return n ? n : 0; }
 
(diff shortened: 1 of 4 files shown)


pandamicro (approved):

cocos/core/renderer/models/skeletal-animation-utils.ts
```diff
@@ -224,20 +224,16 @@ export class JointTexturePool {
...
-                aabb.transform(ab_1, bound, m4_1);
+                aabb.transform(ab_1, bound, mat);
                 ab_1.getBoundary(v3_3, v3_4);
                 Vec3.min(v3_min, v3_min, v3_3);
                 Vec3.max(v3_max, v3_max, v3_4);
             }
             if (buildTexture) {
-                Mat4.multiply(m4_1, m4_1, bindposes[i]);
-                uploadJointData(textureBuffer, 12 * i, m4_1, i === 0);
+                if (mat !== Mat4.IDENTITY) { Mat4.multiply(mat, mat, bindposes[i]); }
```
pandamicro, line 235: This could modify the freezed object Mat4.IDENTITY. You should use another output matrix

pandamicro (approved):

PR #6723, opened 2020-05-20
Remove 2d-label material

PatriceJiang:
use 2d-sprite as default Label material


PatriceJiang:
Use dynamic material replacement

SantyWang:
ref https://github.com/cocos-creator/engine/pull/6727

PR #6732, opened 2020-05-22
Create PSO in render stage

minggo:
Will continue to add commits into this PR. **So it is not ready to be merged**. Current implementation includes:
- record PSO in sub model
- Inherited class of Model should use `Model.getMacroPatches()` to add macro patches, and use `Model.updateAttributesAndBinding()` to update instance attributes and local bindings.
- Planar shadow is not handled as @YunHsiao will fix it in v1.1.1

Things left:
- Add a PassManager to manage all passes to avoid redundent invoking Pass.update(). 
- Don't create PSO in Pass.


minggo:
Works are done, things left:
- instance attributes and instance buffer can not work
- planar shadow can not work

@YunHsiao please continue the work based on my PR. Thanks.

cocos/core/gfx/render-pass.ts
```diff
@@ -78,4 +80,21 @@ export abstract class GFXRenderPass extends GFXObject {
     public abstract initialize (info: IGFXRenderPassInfo): boolean;
 
     public abstract destroy (): void;
+    public hash(): number { return this._hash; }
```
pandamicro, line 83: It's better to check whether _hash exists, if not just do computeHash, so that we can ensure hash is  always accessible

cocos/core/gfx/input-assembler.ts
```diff
@@ -52,6 +53,14 @@ export abstract class GFXInputAssembler extends GFXObject {
         return this._attributes;
     }
 
+    /**
+     * @en Get hash of current attributes.
+     * @zh 获取顶点属性数组的哈希值。
+     */
+    get attributesHash(): number {
+        return this._attributesHash;
```
pandamicro, line 61: Same as render pass, it's easier to use if automatically computed

cocos/core/pipeline/forward/forward-stage.ts
```diff
@@ -107,12 +107,13 @@ export class ForwardStage extends RenderStage {
...
                         const pass = passes[p];
-                        const pso = subModel.psos![p];
+                        // const pso = subModel.psos![p];
                         if (pass.instancedBuffer) {
-                            pass.instancedBuffer.merge(subModel, ro.model.instancedAttributes, pso);
+                            // pass.instancedBuffer.merge(subModel, ro.model.instancedAttributes, pso);
                             this._opaqueInstancedQueue.queue.add(pass.instancedBuffer);
                         } else if (pass.batchedBuffer) {
-                            pass.batchedBuffer.merge(subModel, ro, pso);
+                            // pass.batchedBuffer.merge(subModel, ro, pso);
```
pandamicro: Those comments should be removed

cocos/core/pipeline/forward/forward-stage.ts
```diff
@@ -175,15 +176,15 @@ export class ForwardStage extends RenderStage {
         cmdBuff.beginRenderPass(this._framebuffer!, this._renderArea!,
             camera.clearFlag, colors, camera.clearDepth, camera.clearStencil);
 
-        this._renderQueues[0].recordCommandBuffer(cmdBuff);
+        this._renderQueues[0].recordCommandBuffer(this._device!, this._framebuffer!.renderPass!, cmdBuff);
```
pandamicro: TS should not complain about these private properties, please remove `!`

cocos/core/renderer/scene/submodel.ts
```diff
@@ -4,24 +4,38 @@ import { RenderingSubMesh } from '../../assets/mesh';
...
     protected _cmdBuffers: GFXCommandBuffer[] = [];
+    protected _pathces? : IMacroPatch[];
 
-    set psos (val) {
-        this._psos = val;
-    }
-
-    get psos () {
-        return this._psos;
+    get psoCreateInfos() {
```
pandamicro: Could be simplify to `psoInfos`

cocos/core/renderer/core/pass.ts
```diff
@@ -170,17 +151,13 @@ export class Pass {
         Object.assign(target._dss, info.depthStencilState);
     }
 
-    /**
-     * @zh
-     * 根据指定 PSO 信息计算 hash
-     * @param psoInfo 用于计算 PSO hash 的最小必要信息
-     */
```
pandamicro, line 177: Please add API comment to the new version, include English and Chinese, param should use English

cocos/core/renderer/scene/submodel.ts
```diff
@@ -70,9 +84,72 @@ export class SubModel {
...
+        }
+
+        const passes = materail.passes;
+        for (let j = 0; j < passes.length; ++j) {
+            const pass = passes[j];
+            pass.beginChangeStatesSilently();
+            pass.tryCompile(); // force update shaders
+            pass.endChangeStatesSilently();
+        }
+        this._psoCreateInfos = [];
```
pandamicro: `this._psoCreateInfos.length = 0` should be enough

cocos/core/renderer/scene/submodel.ts
```diff
@@ -70,9 +84,72 @@ export class SubModel {
...
+        if (len == 0) {
+            return;
+        }
+
+        for (let i = 0; i < len; ++i) {
+            const { bindingLayout: bl, pipelineLayout: pl } = this._psoCreateInfos[i];
+            bl.destroy();
+            pl.destroy();
+        }
+        this._psoCreateInfos = [];
```
pandamicro: this._psoCreateInfos.length = 0

cocos/core/pipeline/pipeline-state-manager.ts
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+import { GFXPipelineState, IGFXPipelineStateInfo } from '../gfx/pipeline-state';
+import { IPSOCreationInfo } from '../renderer/scene/submodel';
+import { GFXRenderPass } from '../gfx/render-pass';
+import { GFXInputAssembler } from '../gfx/input-assembler';
+import { murmurhash2_32_gc } from '../utils/murmurhash2_gc';
+import { GFXDevice } from '../gfx/device';
+
+export class PipelineStateManager {
```
pandamicro, line 12: Do we need to detect useless PSOs and destroy them if no longer needed ? Like a garbage collect mechanism

cocos/core/gfx/render-pass.ts
minggo: computeHash should be invoked in inherited classes as it know when attributes are set.

cocos/core/pipeline/forward/forward-stage.ts
minggo: It will complain if removed `!`.

cocos/core/pipeline/pipeline-state-manager.ts
minggo: We don't know which PSO is not needed, it may be used later. Of course, we can add timer to record how long a PSO is not used, but it may make things complex. We can destroy PSOs when memory is low if we have the notification. But create PSO is slow, we should cache them as possible. As i know, bgfx doesn't destroy it.

cocos/core/renderer/scene/submodel.ts
minggo: Ok.

cocos/core/renderer/scene/submodel.ts
minggo: Yep, it sounds good.

cocos/core/gfx/input-assembler.ts
minggo: The same reason as render pass.

shrinktofit:
CreationInfo -> CreateInfo, reference to Vulkan API

cocos/core/gfx/pipeline-state.ts
```diff
@@ -151,7 +151,8 @@ export class GFXInputState {
 export interface IGFXPipelineStateInfo {
     primitive: GFXPrimitiveMode;
     shader: GFXShader;
-    inputState: GFXInputState;
+    // inputState: GFXInputState;
```
shrinktofit: If not used, please directly remove it

cocos/core/pipeline/pipeline-state-manager.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    static getOrCreatePipelineState(
+        device: GFXDevice,
+        psoCreateInfo: IPSOCreationInfo,
+        renderPass: GFXRenderPass,
+        ia: GFXInputAssembler
+        ): GFXPipelineState {
+
+            const hash1 = psoCreateInfo.hash;
+            const hash2= renderPass.hash;
```
shrinktofit: space

cocos/core/renderer/models/morph-model.ts
```diff
@@ -7,24 +7,25 @@ import { RenderingSubMesh } from '../../assets/mesh';
...
-
     private _usedMaterials = new Set<Material>();
 
-    protected createPipelineState (pass: Pass, subModelIndex: number, patches?: IMacroPatch[]) {
-        if (!this._morphRenderingInstance) {
-            // @ts-ignore
-            return super.createPipelineState(...arguments);
+    protected getMacroPatches(subModelIndex: number) : any {
+        if (this._morphRenderingInstance) {
+            return this._morphRenderingInstance.requiredPatches(subModelIndex);
```
shrinktofit: In view of `MorphModel`, it's better to conjugate super's result: `return this._morphRenderingInstance.requiredPatches(subModelIndex).concat(super.getMacroPatches() ?? [])`

minggo:
> CreationInfo -> CreateInfo, reference to Vulkan API

Could you please make it more clear? Which one need to be changed?

cocos/core/pipeline/pipeline-state-manager.ts
minggo: Good catch, i don't know why VS code make the space like this.

cocos/core/gfx/pipeline-state.ts
minggo: Sure.

cocos/core/renderer/models/skinning-model.ts
```diff
@@ -244,15 +244,19 @@ export class SkinningModel extends MorphModel {
...
-        if (EDITOR && pass.instancedBuffer) {
-            console.warn('real-time skeletal animation doesn\'t support instancing, expect rendering anomalies');
-        }
-        const pso = super.createPipelineState(pass, subModelIdx, patches?.concat(myPatches) ?? myPatches);
-        const bindingLayout = pso.pipelineLayout.layouts[0];
-        const buffer = this._buffers[this._bufferIndices![subModelIdx]];
-        if (buffer) { bindingLayout.bindBuffer(UBOSkinning.BLOCK.binding, buffer); }
-        return pso;
+    protected getMacroPatches(subModelIndex: number) : any {
+        return myPatches;
```
shrinktofit: As above stated, `myPatches` should conjugate super's patch. Especially for `SkinningModel`, its super class `MorphModel` may require individual patches.

minggo:
@shrinktofit did you mean change `IPSOCreationInfo `  to `IPSOCreateInfo `? Yep, it looks good.

shrinktofit:
Since there is `GFXPipelineState`, let's take to call "pipeline state"(s) instead of "pipeline state object"(s). As a consequence, let's change variable name like `pso**` into `pipelineState**`.

@YunHsiao  What do you think?

shrinktofit:
> @shrinktofit did you mean change IPSOCreationInfo to IPSOCreateInfo ? Yep, it looks good.

Great work.

minggo:
> Since there is `GFXPipelineState`, let's take to call "pipeline state"(s) instead of "pipeline state object"(s). As a consequence, let's change variable name like `pso**` into `pipelineState**`.
> 
> @YunHsiao What do you think?

Yep, it makes things more clear. Let's create another issue for it. This PR is very big, i don't want to add more things into this PR.

cocos/core/gfx/pipeline-state.ts
```diff
@@ -151,7 +151,7 @@ export class GFXInputState {
 export interface IGFXPipelineStateInfo {
     primitive: GFXPrimitiveMode;
     shader: GFXShader;
-    inputState: GFXInputState;
+    inputAssember: GFXInputAssembler,
```
YunHsiao: This is unnecessary since `inputAssembler.attributes` is a valid GFXInputState

cocos/core/gfx/render-pass.ts
pandamicro: Are we accessing the hash before attributes are set ? If so, maybe a dirty flag is needed for the hash, I don't think we can ensure that the attributes won't be modified after the hash is set.

Another concern about this is the maintainability, in this implementation, upper level need to ensure the hash exists before accessing it, or need to handle the null hash by manually trigger the hash calculation. Either way, it's hard to maintain for upper level developers

cocos/core/pipeline/pipeline-state-manager.ts
pandamicro: I see

cocos/core/gfx/render-pass.ts
minggo: This class is an abstract class,  the hash value can be accessed after sub class is created. And the attributes are only set when sub class is created, so that's the correct time to calculate the hash value. And yes, it is set once and can not be modified.

minggo: > And the attributes are only set when sub class is created, so that's the correct time to calculate the hash value.

Sorry, it is not `attributes`. It is `color attachments and depth/stencil attachment`. You can not modify them after it is created. If you want to use different attachments, you should create a new render pass.

cocos/core/gfx/pipeline-state.ts
minggo: Ok.

minggo: Now PipelineStateManager uses `GFXRenderPass`, `GFXInputAssembler ` and `IPSOCreateInfo` to create GFXPipelineState. None of them includes `GFXInputState`.

pandamicro (approved):

cocos/core/gfx/pipeline-state.ts
minggo: Ok, it is fixed.

LinYunMo (approved):

shrinktofit (changes requested):

cocos/core/renderer/models/morph-model.ts
minggo: Ok. But i don't understand why need `?? []`?

PR #6749, opened 2020-05-26
gfx: fence

YunHsiao:
Re: cocos-creator/3d-tasks#

Changelog:
 * add fence object to GFX framework
* fences are not used outside GFX for now, they should only be of concern in a multi-threaded context


cocos/core/pipeline/ui/ui-stage.ts
```diff
@@ -28,6 +28,8 @@ export class UIStage extends RenderStage {
         framebuffer: 'window',
     };
 
+    private _needToWaitForFence = false;
```
pandamicro: Not used anywhere

cocos/core/pipeline/ui/ui-stage.ts
```diff
@@ -91,6 +94,7 @@ export class UIStage extends RenderStage {
         cmdBuff.end();
 
         bufs[0] = cmdBuff;
-        this._device!.queue.submit(bufs);
+        this._device!.queue.submit(bufs, this._fence!);
+        this._fence!.wait(); this._fence!.reset();
```
pandamicro: Why reset directly after wait ?

minggo: The same question. Who notify it?

YunHsiao: Fences need to be reset before next usage. After waiting, it has already served its purpose.

The command queue will signale it after finishing all the work on GPU.

cocos/core/gfx/fence.ts
```diff
@@ -0,0 +1,40 @@
...
+     * @en Wait for this fence.
+     * @zh 等待当前 fence 信号。
+     */
+    public abstract wait (): void;
+
+    /**
+     * @en Reset this fence to unsignaled state.
+     * @zh 重置当前 fence。
+     */
+    public abstract reset (): void;
```
minggo, line 39: What does it mean? All objects that wait this fence will be notified?

cocos/core/gfx/fence.ts
```diff
@@ -0,0 +1,40 @@
+/**
+ * @category gfx
+ */
+
+import { GFXObject, GFXObjectType } from './define';
+import { GFXDevice } from './device';
+
+// tslint:disable-next-line: no-empty-interface
+export interface IGFXFenceInfo {
+}
```
minggo, line 10: Why declare an empty interface?

cocos/core/gfx/fence.ts
YunHsiao: The `wait` function is a blocking command, only returns after the fence has been signaled.
The `reset` function resets the fence to unsignaled state so it can be used again.

cocos/core/gfx/fence.ts
YunHsiao: For future extensibility

cocos/core/pipeline/ui/ui-stage.ts
minggo: > The command queue will signale it after finishing all the work on GPU.

 If so, then i don't understand why `reset` is needed?

cocos/core/gfx/fence.ts
minggo: Did you mean  a fence can `wait` again only after `reset`?

YunHsiao: That is correct

cocos/core/pipeline/ui/ui-stage.ts
YunHsiao: `reset` is needed so that it can be signaled again

minggo: Yep, got it. But it is not good for performance. Indeed CPU only needs to sync with GPU only it wants to submit again.

pandamicro:
@minggo if no further questions, you can merge it directly

minggo:
I just added a comment.

cocos/core/pipeline/ui/ui-stage.ts
YunHsiao: Right so I already removed that. Fences should not be of concern for any module outside GFX for now since they are all single-threaded. I was adding it then for a vulkan-specific validation error, but now that I think of it, we can/should just workarount it inside GFX to avoid the unnecessary synchonization hassle in a single-threaded context.

YunHsiao: But the fence structure will be useful in the future once we start digging around multi-threaded rendering though, so the majority of this work is still valid.

cocos/core/gfx/webgl/webgl-queue.ts
```diff
@@ -23,7 +24,7 @@ export class WebGLGFXQueue extends GFXQueue {
         this._status = GFXStatus.UNREADY;
     }
 
-    public submit (cmdBuffs: GFXCommandBuffer[], fence?) {
+    public submit (cmdBuffs: GFXCommandBuffer[], fence?: GFXFence) {
```
minggo, line 27: Can we remove fence here? gfx internally handle the fence issue.

minggo (approved):

cocos/core/gfx/webgl/webgl-queue.ts
YunHsiao: No this is where the multi-threaded context may take advantage of, by passing the fence here in one thread and wait for it in another.

PR #6751, opened 2020-05-26
Improve typing

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Improve typing on class member enumerations

Related: https://github.com/cocos-creator/engine/pull/6724

### DIFF
--- a/@types/globals.d.ts
+++ b/@types/globals.d.ts
@@ -97,3 +97,5 @@ declare type Getter = () => any;
 declare type Setter = (value: any) => void;
 
 declare const Buffer: any;
+
+declare type EnumAlias<EnumT> = EnumT[keyof EnumT];
--- a/cocos/audio/assets/clip.ts
+++ b/cocos/audio/assets/clip.ts
@@ -132,8 +132,8 @@ export class AudioClip extends Asset {
 }
 
 export namespace AudioClip {
-    export type PlayingState = typeof PlayingState;
-    export type AudioType = typeof AudioType;
+    export type PlayingState = EnumAlias<typeof PlayingState>;
+    export type AudioType = EnumAlias<typeof AudioType>;
 }
 
 cc.AudioClip = AudioClip;
--- a/cocos/core/3d/framework/camera-component.ts
+++ b/cocos/core/3d/framework/camera-component.ts
@@ -651,10 +651,10 @@ export class CameraComponent extends Component {
 }
 
 export namespace CameraComponent {
-    export type ProjectionType = typeof ProjectionType;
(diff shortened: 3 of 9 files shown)


@types/globals.d.ts
```diff
@@ -97,3 +97,5 @@ declare type Getter = () => any;
 declare type Setter = (value: any) => void;
 
 declare const Buffer: any;
+
+declare type EnumAlias<EnumT> = EnumT[keyof EnumT];
```
pandamicro, line 101: It's kind of strange, why a normal enum can't be recognized by the editor ? Why don't we find a proper solution in the editor's interpretation ?

shrinktofit: Because the `export type PixelFormat = typeof PixelFormat;` yields a `typeof PixelFormat` type.

A declaration like `const e: typeof PixelFormat` declares `e` as `PixelFormat` **enumeration itself**. So we can access its enumerators though, for instance, `e.RGB8`. The `typeof e` should be exactly `"object"`.

In the other hand, `const e: PixelFormat`(one `EnumAlias<typeof PixelFormat>` yields) declares `e` as a **enumeration value** of `PixelFormat`, ie, `string` or `number`. The `typeof e` should be `"number"` or `"string"`.

For example:
```ts
function f(e: typeof PixelFormat) {}
f(PixelFormat.RGB8); // Error: PixelFormat.RGB8 is an enum value
f(PixelFormat); // Ok, but meaningless; In most case, this is not what we want
```

cocos/audio/assets/clip.ts
```diff
@@ -132,8 +132,8 @@ export class AudioClip extends Asset {
 }
 
 export namespace AudioClip {
-    export type PlayingState = typeof PlayingState;
-    export type AudioType = typeof AudioType;
+    export type PlayingState = EnumAlias<typeof PlayingState>;
```
pandamicro, line 135: I think typeof is not needed anymore, isn't it ?

@types/globals.d.ts
pandamicro: Will all those declarations be stripped out in the JS file

cocos/audio/assets/clip.ts
shrinktofit: It's necessary. Otherwise tsc will say 'Type alias circular reference itself.'

shrinktofit: > Will all those declarations be stripped out in the JS file

These declarations would not be emitted in output JS.

pandamicro (approved):

PR #6754, opened 2020-05-26
Properly deprecate UI related API in Node

pandamicro:


JayceLai (approved):

pandamicro:
A lot of UI modules are still using those APIs, which cause ci fail, I will try to solve it later

PR #6762, opened 2020-05-27
add protect for event target auto remove

JayceLai:
Re: cocos-creator/3d-tasks#

Changes:
 * add protect for event target auto remove

### DIFF
--- a/cocos/core/event/eventify.ts
+++ b/cocos/core/event/eventify.ts
@@ -130,11 +130,15 @@ export function Eventify<TBase> (base: Constructor<TBase>): Constructor<TBase & 
     return class extends (base as unknown as any) implements IEventified {
         private _callbacksInvoker: CallbacksInvoker = new CallbacksInvoker();
 
+        private get _isVaildInvoker () { return this._callbacksInvoker != null; }
+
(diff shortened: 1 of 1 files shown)


JayceLai:
The new way of `Eventify` let component become a wrapper of `CallbacksInvoker`, but it will bug when event target is component`s node.

Because the component is destroyed first.

image

@shrinktofit

pandamicro:
Do you mean component is destroyed first ? And the component is a eventified class, then its callbacks will be destroyed too, then `component.targetOff` is causing problem ?

JayceLai:
Yes.

The old way is add a `CallbackTable` map, but `for in` usage avoid this bug.
image

pandamicro:
I think runtime check is not the best solution.

What we could do instead:

- In `BaseNode._onPreDestroyBase`: clear `eventTargets` (with `targetOff`) before any destroy
- Fundamentally the problem is the `destroy` of Component haven't really cleanup the data, we could add a `destroy` function in `Eventify` for cleanup data.

```
destroy () {
    if (super.destroy) {
        super.destroy();
    }
    this.removeAll();
}
```

Besides, I find that we have potential issues with current `Eventify.removeAll` it should do `_unregisterThisIntoTarget` for all targets registered.

PR #6768, opened 2020-05-27
physics optimization

JayceLai:
Re: 
cocos-creator/3d-tasks#2610
cocos-creator/3d-tasks#2611

Changes:
 * sync optimization
 * contact data optimization
 * fix cannon center initialize
 * fix body auto sleeping
 * fix time reset & ammo world step
 * set mass default to 1

### DIFF
--- a/cocos/physics/ammo/ammo-const.ts
+++ b/cocos/physics/ammo/ammo-const.ts
@@ -12,7 +12,7 @@ export const CollisionEventObject = {
     selfCollider: null as unknown as ColliderComponent,
     otherCollider: null as unknown as ColliderComponent,
     contacts: [] as IContactEquation[],
-    isA: false,
+    isBodyA: false,
 };
 
 export class AmmoConstant {
--- a/cocos/physics/ammo/ammo-const.ts
+++ b/cocos/physics/ammo/ammo-const.ts
@@ -12,6 +12,7 @@ export const CollisionEventObject = {
     selfCollider: null as unknown as ColliderComponent,
     otherCollider: null as unknown as ColliderComponent,
     contacts: [] as IContactEquation[],
+    isA: false,
 };
 
 export class AmmoConstant {
--- a/cocos/physics/cannon/cannon-rigid-body.ts
+++ b/cocos/physics/cannon/cannon-rigid-body.ts
@@ -205,66 +205,38 @@ export class CannonRigidBody implements IRigidBody {
     }
 
     applyForce (force: Vec3, worldPoint?: Vec3) {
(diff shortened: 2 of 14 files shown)


cocos/physics/ammo/ammo-const.ts
```diff
@@ -12,6 +12,7 @@ export const CollisionEventObject = {
     selfCollider: null as unknown as ColliderComponent,
     otherCollider: null as unknown as ColliderComponent,
     contacts: [] as IContactEquation[],
+    isA: false,
```
pandamicro: The naming is strange,  how about `first` & `second`? or `bodyA` & `bodyB` ?

cocos/physics/cannon/cannon-rigid-body.ts
```diff
@@ -31,23 +31,20 @@ export class CannonRigidBody implements IRigidBody {
...
+        } else {
+            this.impl.type = this._rigidBody.isKinematic ? CANNON.Body.KINEMATIC : CANNON.Body.DYNAMIC;
         }
 
         this.impl.updateMassProperties();
-
-        if (this.impl.isSleeping()) {
-            this.impl.wakeUp();
-        }
+        this.impl.wakeUp();
```
pandamicro: Is the lower level implementation checking state so that it won't run logic repetitively?

cocos/physics/framework/components/collider/collider-component.ts
```diff
@@ -176,6 +176,39 @@ export class ColliderComponent extends Eventify(Component) {
 
     constructor () { super() }
 
+    /**
+     * @zh
+     * 注册触发事件或碰撞事件相关的回调。
+     * @param type - 触发或碰撞事件的类型，可为 'onTriggerEnter'，'onTriggerStay'，'onTriggerExit' 或 'onCollisionEnter'，'onCollisionStay'，'onCollisionExit';
+     * @param callback - 注册的回调函数
+     * @param target - 可选参数，执行回调函数的目标
```
pandamicro: - API description should have English version
- Parameter API comment should all be in English

cocos/physics/cannon/cannon-rigid-body.ts
JayceLai: Great advice, it should check indeed.

cocos/physics/ammo/ammo-const.ts
JayceLai: bullet and cannon used `A B` to identify the body, so I think `A` is not bad.

pandamicro (approved):

PR #6772, opened 2020-05-28
Mark event-related API of AnimationState as deprecated

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * ~Add `'last-frame-arrived'` event(`AnimationComponent.EventType.LAST_FRAME_ARRIVED`);~
* ~Refine the meaning of `on()`, `once`, `off()` on `AnimationState`.~ (For stability, we only mark those API as deprecated)

### The last-frame-arrived event
<del>Users usually expect an animation event which is triggered as long as the last frame is arrived. The `'last-frame'` event is likely but not fully satisfied, it only works on infinite loop case. In compare, the `'finished'` event would only be triggered on non-loop animation. So this PR add a `"last-frame-arrived"` event covering both cases.</del>

### ~Refine the event APIs of `AnimationState`~
<del>Prior to this PR, `AnimationState` provides very similar API to `IEventified`: `on()`, `once()`, `off()`. However their behaviors are not clear: these API forward the invocation to `AnimationComponent.*()`. This means if you call `.on()` on an individual animation state, you may receive events from other animation states:
```ts
animationStateA.on(AnimationComponent.EventType.PLAY, () => {
  console.log('I'm on playing!'); // may be triggered even not from `animationStateA`
});
```
</del>

(We have now undone this change. Because @gameall3d thought it isn't necessary! )

### Little optimization

I changed the `AnimationManager.prototype.pushDelayEvent` to accept `target: any`, `fn: Function` and `args: any[]`. In comparison, the `fn: Function` replace the `funcName: string` as before. I don't understand the idea why not directly pass function but pass function name and use `target[funcName]` to visit the function. @2youyou2 

Please review these change. @2youyou2 @gameall3d

### DIFF
--- a/cocos/core/animation/animation-state.ts
+++ b/cocos/core/animation/animation-state.ts
@@ -75,12 +75,6 @@ export enum EventType {
      * @zh 动画完成播放时触发。
      */
     FINISHED = 'finished',
-
-    /**
-     * @en Triggered when the animation playing to the last frame or when finish its playing.
-     * @zh 当动画完成播放时，或每当动画播放到最后一帧时触发。
-     */
-    ITERATION_END = 'iteration-end',
 }
 ccenum(EventType);
 
@@ -438,7 +432,7 @@ export class AnimationState extends Playable {
      * To process animation events, use `AnimationComponent` instead.
      */
     public emit (...args: any[]) {
-        cc.director.getAnimationManager().pushDelayEvent(this, '_emit', args);
+        cc.director.getAnimationManager().pushDelayEvent(this._emit, this, args);
     }
 
     /**
@@ -635,7 +629,6 @@ export class AnimationState extends Playable {
 
             if (this.repeatCount > 1 && ((info.iterations | 0) > (lastInfo.iterations | 0))) {
                 this.emit(EventType.LASTFRAME, this);
-                this.emit(EventType.ITERATION_END, this);
             }
 
             lastInfo.set(info);
@@ -644,7 +637,6 @@ export class AnimationState extends Playable {
         if (info.stopped) {
             this.stop();
             this.emit(EventType.FINISHED, this);
-            this.emit(EventType.ITERATION_END, this);
         }
     }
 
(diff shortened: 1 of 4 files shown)


cocos/core/animation/animation-state.ts
```diff
@@ -419,57 +436,44 @@ export class AnimationState extends Playable {
...
-            this._target.emit(type, type, state);
+    /**
+     * @zh
+     * 是否允许触发 `LastFrame` 事件。
+     * @en
+     * Whether `LastFrame` should be triggered.
+     * @param allowed True if the last frame events may be triggered.
+     * @param aspect DO NOT pass this argument. It's for internal usage.
+     */
+    public allowLastFrameEvent (allowed: boolean, aspect: number = 1) {
```
gameall3d: 这个aspect是啥含义啊，为啥叫这个名。。

shrinktofit: It indicates who fired the `'allowLastFrameEvent'` request: either from `AnimationState` itself or from `AnimationComponent`. Because both `AnimationState` and `AnimationComponent` can subscribe to the `'last-frame'` event or not. We should distinguish here to keep them isolated.

gameall3d: may be you can explain in the comment, not just a 'internal usage'

cocos/core/animation/animation-state.ts
```diff
@@ -37,22 +37,61 @@ import { WrapMode, WrapModeMask, WrappedInfo } from './types';
...
+     * @en Triggered when finish playing animation.
+     * @zh 动画完成播放时触发。
+     */
+    FINISHED = 'finished',
+
+    /**
+     * @en Triggered when the animation playing to the last frame or when finish its playing.
+     * @zh 当动画完成播放时，或每当动画播放到最后一帧时触发。
+     */
+    LAST_FRAME_ARRIVED = 'last-frame-arrived',
```
2youyou2: better to unify `LASTFRAME` and `LAST_FRAME_ARRIVED ` word format

cocos/core/animation/animation-state.ts
shrinktofit: https://github.com/cocos-creator/engine/pull/6772/commits/82e7153bf59c17540a8bf5c40fbbcae034d14dae

2youyou2 (approved):

PR #6779, opened 2020-05-29
Safe emit game events to avoid editor startup failure

pandamicro:
Resolve: https://github.com/cocos-creator/3d-tasks/issues/3337


gameall3d (approved):
LGTM

PR #6782, opened 2020-05-29
Solve the "once-off" problem

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Solve the "once-off" problem, see below.

### The "once-off" problem
After "once subscription" is triggered the corresponding `CallbacksInvoker.prototype.off` was going to be called as a matter of course. But `Eventify(base).prototype.off` was not called though it should be. This is because `Eventify(base)` use `CallbacksInvoker` in a composition way.

### DIFF
--- a/cocos/core/event/eventify.ts
+++ b/cocos/core/event/eventify.ts
@@ -4,7 +4,7 @@
 
 import { fastRemove } from '../utils/array';
 import { CallbacksInvoker } from './callbacks-invoker';
-import { createMap, mixin } from '../utils/js';
+import { createMap } from '../utils/js';
 
 type Constructor<T = {}> = new (...args: any[]) => T;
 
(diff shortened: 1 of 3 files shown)


pandamicro (approved):

cocos/core/event/eventify.ts
```diff
@@ -158,7 +155,7 @@ export function Eventify<TBase> (base: Constructor<TBase>): Constructor<TBase &
             if (!callback) {
                 this.removeAll(type);
             } else {
-                this._callbacksInvoker.off(type, callback, target);
+                CallbacksInvoker.prototype.off.call(this, type, callback, target);
```
jareguo: Can we cache CallbacksInvoker.prototype.xxx rather than access every time?

shrinktofit: Definitely yes: https://github.com/cocos-creator/engine/pull/6782/commits/d0c59f831310a75ccab575a3b96f6967a2381004

jareguo (approved):

PR #6809, opened 2020-06-02
fix vertex index of ray model

JayceLai:
Re: cocos-creator/3d-tasks#

Changes:
 * fix vertex index of ray model


JayceLai:
https://github.com/cocos-creator/test-cases-3d/pull/242

pandamicro (approved):

PR #6813, opened 2020-06-03
directional light stack

YunHsiao:
Re: cocos-creator/3d-tasks#3397

Changelog:
 *

### DIFF
--- a/cocos/core/renderer/scene/render-scene.ts
+++ b/cocos/core/renderer/scene/render-scene.ts
@@ -176,7 +176,7 @@ export class RenderScene {
             if (dlList.length) {
                 this._mainLight = dlList[dlList.length - 1];
                 if (this._mainLight.node) { // trigger update
-                    this._mainLight.node.hasChangedFlags = TransformBit.ROTATION;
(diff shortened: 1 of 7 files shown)


cocos/core/renderer/scene/render-scene.ts
```diff
@@ -165,13 +166,36 @@ export class RenderScene {
...
 
     public unsetMainLight (dl: DirectionalLight) {
-        dl.detachFromScene();
-        this._mainLight = null;
+        if (this._mainLight === dl) {
+            const dlList = this._directionalLights;
+            if (dlList.length) {
+                this._mainLight = dlList[dlList.length - 1];
+                if (this._mainLight.node) { // trigger update
+                    this._mainLight.node.setRotation(this._mainLight.node.rotation);
```
pandamicro: In the future if we check equality of set value in `setRotation`, the trigger will fail, how it's triggered exactly ? By TRANSFORM_CHANGED event ? if so, you can emit a event here direclty.

pandamicro (approved):

PR #6837, opened 2020-06-07
Improve CallbacksInvoker to remove __eventTargets mechanism

pandamicro:
https://github.com/cocos-creator/3d-tasks/issues/3418
https://github.com/cocos-creator/2d-tasks/issues/2795


pandamicro:
https://github.com/cocos-creator/engine/pull/6828

pandamicro:
Remove `__eventTargets` makes everything much more clear, and removed a lot of code, happy with this solution.

In this PR, I delayed the callbacks destroy in `emit` and `hasListeners`, by check the validity of target object. If a target is destroyed, then `CallbacksInvoker` will no longer execute its callback and remove it.

A passive solution but safe enough.

tests/core/callbacks-invoker.test.ts
```diff
@@ -0,0 +1,519 @@
...
+
+    ci.on('visit', cb1);
+    ci.on('visit', cb2);
+    ci.on('visit', cb3);
+    ci.emit('visit');
+
+    expect(actualSequence).toEqual([cb1, cb2, cb1, cb2, cb3, cb3]);
+});
+
+test('Destroyed node should no longer be invoked as target', () => {
```
pandamicro, line 448: Unit test about the current pr is the following two

cocos/core/event/callbacks-invoker.ts
```diff
@@ -41,16 +42,40 @@ class CallbackInfo {
...
-        this.callback = callback;
+        this.callback = callback || empty;
         this.target = target;
         this.once = !!once;
     }
+
+    public reset () {
+        this.target = undefined;
+        this.callback = empty;
+        this.once = false;
```
pandamicro, line 53: To release all reference asap

cocos/core/event/callbacks-invoker.ts
```diff
@@ -41,16 +42,40 @@ class CallbackInfo {
...
+    }
+
+    public check () {
+        // Validation
+        if (this.target instanceof CCObject && !this.target.isValid) {
```
pandamicro: Only check for CCObject with DESTROYED flag, could be using TO_DESTROY if it's not safe enough. TO_DESTROY flag is set when `destroy()` invoked, DESTROYED flag is set when `deferredDestroy` happened once every frame

cocos/core/event/eventify.ts
```diff
@@ -137,57 +129,14 @@ export interface IEventified {
...
-        public off (type: EventType, callback?: Function, target?: object) {
-            if (!callback) {
-                this.removeAll(type);
-            } else {
-                callbacksInvokerOff.call(this, type, callback, target);
-                if (target) {
-                    this._unregisterThisIntoTarget(target);
-                }
-            }
+            this.on(type, callback, target, true);
```
shrinktofit: `return` is missed

cocos/core/event/callbacks-invoker.ts
jareguo: I guess you can just use `return cc.isValid(this.target);`

> cc.isValid = function (value, strictMode) {
    if (typeof value === 'object') {
        return !!value && !(value._objFlags & (strictMode ? (Destroyed | ToDestroy) : Destroyed));
    }
    else {
        return typeof value !== 'undefined';
    }
};

shrinktofit: It is not suitable to introduce a coupling between event system and `CCObject`. You could define a special property method for that purpose:
```ts
class CCObject { __callback_check__() { return this.isValid; } }
class CallbacksInvoker { check() { return this.target.__callback_check__?.() ?? true; } }
```

cocos/core/event/callbacks-invoker.ts
```diff
@@ -31,6 +31,7 @@
 import { Pool } from '../memop';
 import { array, createMap } from '../utils/js';
 import { TEST } from 'internal:constants';
+import { CCObject } from '../data/object';
```
shrinktofit: See below

shrinktofit (changes requested):

cocos/core/data/object.ts
```diff
@@ -327,7 +327,7 @@ class CCObject {
             this._onPreDestroy();
         }
 
-        if ((TEST ? (/* make EDITOR mockable*/ Function('return !EDITOR'))() : !EDITOR) || cc.engine._isPlaying) {
+        if (!EDITOR || cc.engine._isPlaying) {
```
jareguo, line 330: 这样会导致单元测试中的 EDITOR 变量无法动态修改，也就无法在单元测试中切换不同执行路径了

shrinktofit: jest mock?

cocos/core/event/callbacks-invoker.ts
pandamicro: CCObject is a basic class, I don't think it's inappropriate to use its api here

cocos/core/data/object.ts
pandamicro: 为什么需要动态修改？具体使用的案例在哪里有？

cocos/core/event/eventify.ts
pandamicro: fixed

cocos/core/data/object.ts
jareguo: image
Here will test the behavior in the editor

cocos/core/event/callbacks-invoker.ts
```diff
@@ -276,13 +314,7 @@ export class CallbacksInvoker {
...
                     const info = infos[i];
                     if (info && info.callback === callback && info.target === target) {
-                        if (list.isInvoking) {
-                            list.cancel(i);
-                        }
-                        else {
-                            fastRemoveAt(infos, i);
-                            callbackInfoPool.free(info);
-                        }
+                        list.cancel(i);
```
jareguo, line 308: Why not remove directly? Using cancel will much slower.

cocos/core/event/callbacks-invoker.ts
```diff
@@ -319,11 +351,13 @@ export class CallbacksInvoker {
                     if (info.once) {
                         this.off(key, callback, target);
                     }
-                    if (target) {
-                        callback.call(target, arg0, arg1, arg2, arg3, arg4);
+                    // Lazy check validity of callback target, 
+                    // if target is CCObject and is no longer valid, then remove the callback info directly
+                    if (!info.check()) {
+                        this.off(key, callback, target);
```
caryliu1999, line 348: This still doesn't solve the problem of the event listener not being removed from the EventManager after the once event off.

cocos/core/event/callbacks-invoker.ts
pandamicro: If it's during invocation, we can't change the array, otherwise the for loop may skip some callbacks. The cancel is reimplemented with `isInvoking` check, will remove when it's possible

cocos/core/data/object.ts
pandamicro: I don't think it could be modified at runtime now, we have moved all macros to an internal module, so `_deferredDestroy` is probably using the internal macro EDITOR, not the global one. There is no longer global CC_EDITOR. @shrinktofit is there anyway to simulation such environment dynamically in the unit test ?

cocos/core/event/callbacks-invoker.ts
pandamicro: I see the problem, indeed it doesn't solve the EventListener removal, I updated Node.destroy process to fix it, please review

caryliu1999 (approved):

pandamicro:
@jareguo @shrinktofit ready for a second review

cocos/core/data/object.ts
jareguo: Fine. Then we should export two version's api for test only.
By calling differenct API to test differenct behaviour.

cocos/core/event/callbacks-invoker.ts
jareguo: cool

cocos/core/event/callbacks-invoker.ts
```diff
@@ -174,13 +208,16 @@ export class CallbacksInvoker {
      * @param once - Whether invoke the callback only once (and remove it)
      */
     public on (key: string, callback: Function, target?: Object, once?: boolean) {
-        let list = this._callbackTable[key];
-        if (!list) {
-            list = this._callbackTable[key] = callbackListPool.alloc();
+        if (!this.hasEventListener(key, callback, target)) {
+            let list = this._callbackTable[key];
+            if (!list) {
+                list = this._callbackTable[key] = callbackListPool.alloc();
```
jareguo, line 205: BTW, I saw the renaming of pool's API. I don't think alloc/free are better than get/put. The former seems like memory operation (in C), the latter describes more precisely for pool operations.

cocos/core/event/callbacks-invoker.ts
```diff
@@ -41,16 +42,40 @@ class CallbackInfo {
...
+
+    public reset () {
+        this.target = undefined;
+        this.callback = empty;
+        this.once = false;
+    }
+
+    public check () {
+        // Validation
+        if (this.target instanceof CCObject && !isValid(this.target, true)) {
```
jareguo, line 58: 其实我觉得不太需要判断 instanceof CCObject 了。isValid 对非 CCObject 也会返回正确结果

cocos/core/event/callbacks-invoker.ts
```diff
@@ -41,16 +42,40 @@ class CallbackInfo {
...
+            return false;
+        }
+        else {
+            return true;
+        }
+    }
+
+    public execute (arg0?: any, arg1?: any, arg2?: any, arg3?: any, arg4?: any) {
```
jareguo: I don't think execute API is appropriate for `CallbackInfo`. CallbackInfo is more like a set of data.
If we need to invoke the data, why not invoke in CallbacksInvoker?

cocos/core/event/callbacks-invoker.ts
```diff
@@ -70,6 +95,7 @@ export class CallbackList {
         for (let i = 0; i < this.callbackInfos.length; ++i) {
             const info = this.callbackInfos[i];
             if (info && info.callback === cb) {
+                info.reset();
                 callbackInfoPool.free(info);
```
jareguo, line 90: 新实现的 Pool 对象相比原版移除了原先的 clear 回调，导致每次 free 前都要手动 reset，我认为这不是一个很安全的设计。

image

cocos/core/event/callbacks-invoker.ts
```diff
@@ -341,7 +375,14 @@ export class CallbacksInvoker {
      * 移除所有回调。
      */
     public clear () {
-        this._callbackTable = createMap(true);
+        for (let key in this._callbackTable) {
+            let list = this._callbackTable[key];
+            if (list) {
+                list.clear();
+                callbackListPool.free(list);
+                delete this._callbackTable[key];
```
jareguo, line 379: Is it safe to adjust the dictionary during iterating?

cocos/core/event/callbacks-invoker.ts
pandamicro: Won't be considered in this PR

cocos/core/event/callbacks-invoker.ts
pandamicro: It's needed, because `isValid` returns false for undefined target, but it's a valid input for target

cocos/core/event/callbacks-invoker.ts
pandamicro: Won't be considered in this PR

cocos/core/event/callbacks-invoker.ts
shrinktofit: Even it seems basic in engine, the `CallbacksInvoker` should not rely on it. Just like the _eventTargets you done before...

cocos/core/event/callbacks-invoker.ts
pandamicro: It is, because the for loop use a prefetched local variable.

```
const infos = list.callbackInfos;
for (let i = 0, len = infos.length; i < len; ++i) {
}
```

cocos/core/event/callbacks-invoker.ts
pandamicro: Removed execute

jareguo (approved):

PR #6845, opened 2020-06-09
Eliminate cc.* log functions invocation

shrinktofit:


cocos/core/data/object.ts
```diff
@@ -266,7 +266,7 @@ class CCObject {
      * obj.destroy();
      * ```
      */
-    public destroy (): boolean {
+    public destroy (): boolean | void {
```
pandamicro: Why add void here ? It should return true or false, no exception, all sub class destroy should also call `return super.destroy()` to see whether it's been successfully destroyed or not, it's also a chance to check and correct things

cocos/core/scene-graph/node-activator.ts
```diff
@@ -129,7 +130,7 @@ activateTasksPool.get = function getActivateTask () {
 
 function _componentCorrupted (node, comp, index) {
     if (DEV) {
-        legacyCC.errorID(3817, node.name, index);
+        error(3817, node.name, index);
```
pandamicro: should be errorID

cocos/core/scene-graph/node-activator.ts
```diff
@@ -282,7 +283,7 @@ export default class NodeActivator {
             // zh:
             // 对相同节点而言，无法撤销反激活，防止反激活 - 激活 - 反激活的死循环发生。
             // 这样设计简化了一些引擎的实现，而且对调用者来说能保证反激活操作都能成功。
-            legacyCC.errorID(3816, node.name);
+            error(3816, node.name);
```
pandamicro: same as above

pandamicro (changes requested):

cocos/core/scene-graph/node-activator.ts
shrinktofit: Fixed.

cocos/core/scene-graph/node-activator.ts
shrinktofit: Fixed.

cocos/core/data/object.ts
shrinktofit: That's because in existing codebase there are many derived classes return nothing. Such as one of the biggest customer: `Component`.

pandamicro (approved):

PR #6847, opened 2020-06-10
Type fix for inherited CCObject.destroy()

shrinktofit:
Re: cocos-creator/2d-tasks#

Changes:
 * `CCObject.prototype.destroy()` and its derivations should definitely return boolean values, valueless return is forbidden. cc @pandamicro 

This PR BTW repairs a typo BUG. cc @JoneLau

### DIFF
--- a/cocos/core/components/component.ts
+++ b/cocos/core/components/component.ts
@@ -372,7 +372,7 @@ class Component extends CCObject {
                 legacyCC.director._compScheduler.disableComp(this);
             }
         }
-        return false;
+        return true;
     }
 
     public _onPreDestroy () {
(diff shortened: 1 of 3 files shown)


JoneLau (approved):

cocos/core/components/component.ts
```diff
@@ -362,15 +362,17 @@ class Component extends CCObject {
...
-                return legacyCC.errorID(3626,
+                legacyCC.errorID(3626,
                     getClassName(this), getClassName(depend));
+                return false;
             }
         }
         if (super.destroy()) {
             if (this._enabled && this.node.activeInHierarchy) {
                 legacyCC.director._compScheduler.disableComp(this);
             }
```
pandamicro, line 373: should return true here

pandamicro (changes requested):

cocos/core/components/component.ts
shrinktofit: https://github.com/cocos-creator/engine/pull/6847/commits/0fa4c18192a2dbc7c13dd5b84f6f4887ff1af9c1

pandamicro (approved):

PR #6850, opened 2020-06-10
Treat node:eulerAngle curve as skeletal curve; skeletal blending for common targets

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Curves that would modify `Node.prototype.eulerAngles` are also considered as skeletal curves. So they would be subjected to skeletal animation blending.
 * Common targets that would modify T, R, S, or the `eulerAngles` would be subjected to skeletal animation blending like skeletal curves do.

Request reviews from @YunHsiao  @gameall3d

### DIFF
--- a/cocos/core/animation/animation-state.ts
+++ b/cocos/core/animation/animation-state.ts
@@ -357,24 +357,6 @@ export class AnimationState extends Playable {
             this.repeatCount = 1;
         }
 
-        this._commonTargetStatuses = clip.commonTargets.map((commonTarget, index) => {
-            const target = createBoundTargetOptimized(
-                createBufferedTarget,
-                root,
-                commonTarget.modifiers,
-                commonTarget.valueAdapter,
-                false,
-            );
-            if (target === null) {
-                return null;
-            } else {
-                return {
-                    target,
-                    changed: false,
-                };
-            }
-        });
-
         /**
          * Create the bound target. Especially optimized for skeletal case.
          */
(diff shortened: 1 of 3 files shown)


cocos/core/animation/skeletal-animation-blending.ts
```diff
@@ -63,11 +63,19 @@ export class BlendStateBuffer {
                 s = scale.value;
                 anyChanged = true;
             }
+
+            // Note: rotation and eulerAngles can not co-exist.
```
shrinktofit, line 67: cc @YunHsiao

cocos/core/animation/skeletal-animation-blending.ts
```diff
@@ -165,6 +185,7 @@ interface NodeBlendState {
     properties: {
         position?: PropertyBlendState<Vec3>;
         rotation?: PropertyBlendState<Quat>;
+        eulerAngles?: PropertyBlendState<Vec3>;
         scale?: PropertyBlendState<Vec3>;
```
shrinktofit, line 189: @YunHsiao  Please note that only curves that operating  on these properties are subjected to skeletal blending(eg. would call to `setTRS`).

cocos/core/animation/skeletal-animation-blending.ts
YunHsiao: Should be more specific on this, it is the two curves that shouldn't coexist, not the properties per se.
And we should add in-editor restrictions on this @gameall3d

cocos/core/animation/animation-state.ts
```diff
@@ -889,22 +915,23 @@ export class AnimationState extends Playable {
     }
 }
 
-function isSkeletonCurve (curve: IRuntimeCurve) {
+function isTargetingTrs (path: TargetPath[]) {
```
gameall3d: Trs->TRS?

gameall3d (changes requested):

cocos/core/animation/skeletal-animation-blending.ts
YunHsiao: That's okay for now

YunHsiao (approved):

gameall3d (approved):
LGTM

PR #6855, opened 2020-06-11
check invalid layer uuid

MSoft1115:
Re: cocos-creator/2d-tasks#

Changes:
 *

### DIFF
--- a/cocos/terrain/terrain.ts
+++ b/cocos/terrain/terrain.ts
@@ -984,7 +984,7 @@ export class Terrain extends Component {
 
         for (let i = 0; i < this._layers.length; ++i) {
             const temp = this._layers[i];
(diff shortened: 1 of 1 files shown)


cocos/terrain/terrain.ts
```diff
@@ -984,7 +984,7 @@ export class Terrain extends Component {
 
         for (let i = 0; i < this._layers.length; ++i) {
             const temp = this._layers[i];
-            if (temp != null && temp.detailMap != null) {
+            if (temp && temp.detailMap && temp.detailMap._uuid) {
```
pandamicro: 应该用

import { isValid } from '../core/data/object';
if (temp && isValid(temp.detailMap))

pandamicro (changes requested):

PR #6902, opened 2020-06-18
Fixed CCNode _checkListeners & CCScheduler update

yoki0805:
Changes:
 * CCNode _checkListeners function. The loop index should reset to zero.
 * CCScheduler update function.  When first time, it should call update function once to avoid skip  current frame.


cocos2d/core/CCScheduler.js
```diff
@@ -200,6 +200,7 @@ proto.update = function (dt) {
     if (this._elapsed === -1) {
         this._elapsed = 0;
         this._timesExecuted = 0;
+        this.update(dt);
```
pandamicro: It is the designed behavior, the scheduled tasks will always start after the first time interval ends.

pandamicro (changes requested):
I think the modification to _checkListener is necessary, but the original code seems intended, I want to ask in which scenario we can see problems causing by the current implementation

cocos2d/core/CCScheduler.js
yoki0805: > It is the designed behavior, the scheduled tasks will always start after the first time interval ends.

I coded this, and the task ran after 2 frames. So, the  behavior is correct?

yoki0805:
> I think the modification to _checkListener is necessary, but the original code seems intended, I want to ask in which scenario we can see problems causing by the current implementation

No, just for safety and readability.
When the node._bubblingListeners are check failed, the result of checking node._capturingListeners could be wrong. Or insure that the node._bubblingListeners and the node._capturingListeners are mutex.

cocos2d/core/CCScheduler.js
pandamicro: Yes, actually the second "Component updated" and the first "myFunc updated" is printed in the same frame. The scheduler execution is later than the component update. So the behavior in each frame is like this:

- Frame 0:
    - start -> myFunc scheduled
    - update
- Frame 1:
    - update
    - myFunc call
- repeat Frame 1

pandamicro:
> When the node._bubblingListeners are check failed, the result of checking node._capturingListeners could be wrong. Or insure that the node._bubblingListeners and the node._capturingListeners are mutex.

I agree, all events should be checked in both capturing listener and bubbling listener

cocos2d/core/CCScheduler.js
yoki0805: Okay, I got it.
I misunderstood  the log. Lol.
Thanks for your patience. :)

pandamicro: So please revert this line, and we can merge your pr

yoki0805: I got the 403 problem. Now it reverted.

pandamicro (approved):

caryliu1999 (approved):

SantyWang:
Thanks for your contribution. Please make another pr to v2.4.0-release. By the way, do you have an account of cocos forum, if so, you can leave your account id and we will give you a label

pandamicro:
@holycanvas I rebased the PR to v2.4.0, you can do it too with the "Edit" button at the right of the PR title

Oh, I just notice you have done it, why you change it back ?

SantyWang:
> Oh, I just notice you have done it, why you change it back ?

There will be some useless commits

yoki0805:
Here you are.
[https://github.com/cocos-creator/engine/pull/6905](url)

PR #6903, opened 2020-06-18
splash screen adapt for alipay&cocos play

JayceLai:
Re: cocos-creator/3d-tasks#3567

Changes:
 * splash screen adapt for alipay&cocos play

### DIFF
--- a/cocos/core/splash-screen-webgl.ts
+++ b/cocos/core/splash-screen-webgl.ts
@@ -126,7 +126,7 @@ export class SplashScreenWebgl {
     }
 
     public main (canvas: HTMLCanvasElement) {
-        var ccSetting = globalThis._CCSettings;
+        let ccSetting = globalThis._CCSettings;
         if (ccSetting && ccSetting.splashScreen) {
             this.setting = ccSetting.splashScreen;
             (this.setting.totalTime as number) = this.setting.totalTime != null ? this.setting.totalTime : 3000;
--- a/cocos/core/splash-screen-webgl.ts
+++ b/cocos/core/splash-screen-webgl.ts
@@ -83,13 +83,13 @@ export class SplashScreenWebgl {
     private logoMat33 = new Float32Array([1, 0, 0, 0, 1, 0, 0, 0, 1]);
     private textMat33 = new Float32Array([1, 0, 0, 0, 1, 0, 0, 0, 1]);
 
-
     private setting!: ISplashSetting;
-    private orientation = '';
     private callBack: Function | null = null;
     private cancelAnimate = false;
     private handle = -1;
     private startTime = -1;
+    private orientation = '';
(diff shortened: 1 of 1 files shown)


cocos/core/splash-screen-webgl.ts
```diff
@@ -125,8 +126,9 @@ export class SplashScreenWebgl {
     }
 
     public main (canvas: HTMLCanvasElement) {
-        if (window._CCSettings && window._CCSettings.splashScreen) {
-            this.setting = window._CCSettings.splashScreen;
+        this.ccSetting = window._CCSettings;
```
pandamicro: Please do not save CCSettings, it's too big, only save what you need here. By the way, why you need to save it ?

cocos/core/splash-screen-webgl.ts
```diff
@@ -221,30 +223,45 @@ export class SplashScreenWebgl {
...
-            if (sys.os == cc.sys.OS_OSX || sys.os == cc.sys.OS_IOS) {
+            if (sys.os == sys.OS_OSX || sys.os == sys.OS_IOS) {
                 this.gl.canvas.width = screen.width * devicePixelRatio;
                 this.gl.canvas.height = screen.height * devicePixelRatio;
             }
         }
 
-        // TODO: hack for cocosPlay & XIAOMI cause on landscape canvas value is wrong
+        // adapt for cocos play 1.0.11
+        if (COCOSPLAY) {
```
pandamicro, line 234: There is no version distinguish between this if and the next one, it's not acceptable to run the lame adaptation code  twice for the same platform.

cocos/core/splash-screen-webgl.ts
```diff
@@ -385,13 +402,23 @@ export class SplashScreenWebgl {
 
         gl.bindFramebuffer(gl.FRAMEBUFFER, null);
 
-        // TODO: hack for cocosPlay & XIAOMI cause on landscape canvas value is wrong
+        // adapt for cocos play 1.0.11
+        if (COCOSPLAY) {
```
pandamicro: Again ? Why keep set innerWidth and innerHeight ?

JayceLai: To prevent someone silently modifying

pandamicro: In which case exactly ? I don't understand the scenario where we need to check and adapt window size every frame.

JayceLai: @LinYunMo and I fixed the problem on 1.0.3, https://github.com/cocos-creator/engine/pull/6107

Not sure if it's needed now
I will test the check is needed or not

JayceLai: It is unnecessary on 1.1.1

pandamicro (approved):

PR #6907, opened 2020-06-19
fixed web cannot fullscreen

zhefengzhang:
When the autoFullScreen fails, the game should be able to request full screen again by touch the screen.

Currently, The UC mobile browser cannot support multi-touch after requesting full screen.

I added tips on this in example-case project. https://github.com/cocos-creator/example-cases/pull/842

### DIFF
--- a/cocos2d/core/platform/CCScreen.js
+++ b/cocos2d/core/platform/CCScreen.js
@@ -220,7 +220,6 @@ cc.screen = /** @lends cc.screen# */{
         let self = this;
         let touchTarget = cc.game.canvas || element;
         let fullScreenErrorEventName = this._fn.fullscreenerror;
-        if (typeof document[fullScreenErrorEventName] === "undefined") return;
         let touchEventName = this._touchEvent;
         
         function onFullScreenError () {


PPpro:
奇怪， document[fullScreenErrorEventName] 没有定义，canvas 却可以监听 fullScreenErrorEventName 事件吗
@Jno1995

pandamicro:
I think it's a UC bug, or we haven't find the correct way to register full screen error event, please check on UC environment, what's the correct event name for UC. Simply remove the protection could cause issue in other environment (e.g. if some environment check for null event name in addEventListener).

zhefengzhang:
> 奇怪， document[fullScreenErrorEventName] 没有定义，canvas 却可以监听 fullScreenErrorEventName 事件吗
> @Jno1995

可以的。

zhefengzhang:
> I think it's a UC bug, or we haven't find the correct way to register full screen error event, please check on UC environment, what's the correct event name for UC. Simply remove the protection could cause issue in other environment (e.g. if some environment check for null event name in addEventListener).

element.requestFullscreen is the correct method for uc mobile browser to request full screen.
This is indeed a bug in the uc mobile browser.

pandamicro:
> element.requestFullscreen is the correct method for uc mobile browser to request full screen.
> This is indeed a bug in the uc mobile browser.

So we should not remove a reasonable check just for one browser environment to avoid its own bug, just not worth it.
So I don't agree with this solution, we can just ignore the problem or add a environment check within the if

zhefengzhang:
`if (typeof document[fullScreenErrorEventName] === "undefined") return` is a wrong judgment condition. Because the document does not have the fullscreenerror attribute on the web.

pandamicro (approved):

PR #6920, opened 2020-06-23
Support nested prefab

jareguo:


cocos2d/core/utils/prefab-helper.js
```diff
@@ -125,7 +125,6 @@ module.exports = {
         node._name = _name;
         node._active = _active;
         node._localZOrder = _localZOrder;
-        node._globalZOrder = _globalZOrder;
```
jareguo, line 128: ## Removed `_globalZOrder `, seems not used anymore.

cocos2d/core/components/CCStudioComponent.js
```diff
@@ -288,6 +288,7 @@ var StudioComponent = cc.Class({
     },
 });
 
+// TODO: nest by the built-in manner
```
jareguo, line 291: @knoxHuang We should use synced prefab to implement this component. Or just discard it since the editor now supports nested prefab.

PR #6937, opened 2020-06-28
Morph support on devices that lack texture-float feature

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Support morph target on GFX devices that have no texture-float feature.


cocos/core/assets/morph-rendering.ts
```diff
@@ -48,7 +49,8 @@ export class StdMorphRendering implements MorphRendering {
                 continue;
             }
 
-            if (preferCpuComputing) {
+            const hasFeatureTextureFloat = gfxDevice.hasFeature(GFXFeature.TEXTURE_FLOAT);
+            if (preferCpuComputing || !hasFeatureTextureFloat) {
```
pandamicro: I think `preferCpuComputing` is not necessary, we should only depend on the feature support to decide whether to use GPU computing. If you think there are other conditions to choose cpu computing, then `preferCpuComputing` should be a configurable API.
By the way, in the current logic, even if the device has texture float support, we are still using cpu computing, there is no way to turn on gpu computing.

cocos/core/assets/morph-rendering.ts
```diff
@@ -530,6 +523,53 @@ class MorphUniforms {
...
+    }
+
+    get sampler () {
+        return this._sampler;
+    }
+
+    public destroy () {
+        this._texture.destroy();
+        // TODO: Should we free sampler?
+        // this._sampler.destroy();
```
pandamicro: yes, sampler has a destroy API as a gpu resource, the destroy API implies that we destroy all gpu resources we allocated in the current object

pandamicro (changes requested):

cocos/core/assets/morph-rendering.ts
shrinktofit: But @YunHsiao  says: no

YunHsiao: samplers are globally reused so here it is not necessary

pandamicro: Then a reuse or recycle function is need, no action at all is strange, how we reuse samplers and where can I find the corresponding logic ?

pandamicro: Ok, I see SamplerLib is managing and reusing all samplers with hash as id. @shrinktofit please remove the todo and commented code here

cocos/core/assets/morph-rendering.ts
shrinktofit: I have now turned the `preferCpuComputing` to `false`. Cpu rendering would help to debug.

cocos/core/assets/morph-rendering.ts
```diff
@@ -48,6 +49,7 @@ export class StdMorphRendering implements MorphRendering {
                 continue;
             }
 
+            const hasFeatureTextureFloat = gfxDevice.hasFeature(GFXFeature.TEXTURE_FLOAT);
```
pandamicro: I don't understand, it's never used afterward

cocos/core/assets/morph-rendering.ts
```diff
@@ -530,6 +491,100 @@ class MorphUniforms {
...
+                    return valueView;
+                },
+
+                /**
+                 * Destroy the texture. Release its GPU resources.
+                 */
+                destroy () {
+                    textureAsset.destroy();
+                    // TODO: Should we free sampler?
+                    // this._sampler.destroy();
```
pandamicro, line 578: Please remove the TODO and code

pandamicro (changes requested):

cocos/core/assets/morph-rendering.ts
shrinktofit: Unused code. Has been removed: https://github.com/cocos-creator/engine/pull/6937/commits/9877e50da83a644e8b37f233f22a4bfeb9a4c212

cocos/core/assets/morph-rendering.ts
shrinktofit: https://github.com/cocos-creator/engine/pull/6937/commits/9877e50da83a644e8b37f233f22a4bfeb9a4c212

pandamicro (approved):

pandamicro:
修复下冲突

PR #6940, opened 2020-06-29
Throw the error instead of return null in instantiate and addComponent

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * In usual, developers won't expect and won't handle the error during invocation of  these method as a matter of fact that in most cases these methods do never fail if arguments are correctly passed to them accordingly to the method specification. This PR changes the way how `instantiate()` and `Node.prototype.addComponent()` report their error. Instead of printing the error to console and return a `null` as result, this PR change to throw the error.

Reference about error design:
- [Error Handling in Node.js](https://www.joyent.com/node-js/production/design/errors)

This should fixes https://forum.cocos.org/t/c3d-1-1-1-instantiate/94620/4

### DIFF
--- a/cocos/core/scene-graph/base-node.ts
+++ b/cocos/core/scene-graph/base-node.ts
@@ -31,7 +31,7 @@ import { Component } from '../components/component';
 import { ccclass, property } from '../data/class-decorator';
 import { CCObject } from '../data/object';
 import { Event } from '../event';
-import { errorID, warnID, error, log, assertID } from '../platform/debug';
+import { errorID, warnID, error, log, assertID, getError } from '../platform/debug';
 import { SystemEventType } from '../platform/event-manager/event-enum';
 import { ISchedulable } from '../scheduler';
 import IdGenerator from '../utils/id-generator';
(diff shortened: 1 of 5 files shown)


shrinktofit:
Seems like out-of-memory problem during CI.

cocos/core/scene-graph/base-node.ts
```diff
@@ -916,46 +917,38 @@ export class BaseNode extends CCObject implements ISchedulable {
...
+            throw TypeError(getError(3809));
         }
         if (!js.isChildClassOf(constructor, legacyCC.Component)) {
-            errorID(3810);
-            return null;
+            throw TypeError(getError(3810));
         }
 
         if (EDITOR && constructor._disallowMultiple) {
             if (!this._checkMultipleComp!(constructor)) {
```
jareguo: 这里应该是在 _checkMultipleComp 里 throw 就好

shrinktofit: https://github.com/cocos-creator/engine/pull/6940/commits/58fb8e81cd64ea32ba5d9e4e1b220fd90d8fd52f

jareguo (approved):

PR #6945, opened 2020-06-30
CCAudioEngine bug fixed

endless3000:
There is no need to callback the 'ended' event when the audio is loop, in some case the ended event will trigger immediately and '_id2audio' will lose the audio;
不需要将设置循环音乐的完成事件，并且在web上在某些情况下播放循环音乐会立即触发此事件导致_id2audio被清除


PPpro (approved):

SantyWang:
感谢

PR #6962, opened 2020-07-08
Pipeline refactor

minggo:
把 root 和 device 的访问放到了 PipelineGlobal 里，pipeline 其它模块就不需要自己缓存这个值。同时也隔离了 pipeline 模块对 Root 的依赖。PipelineGlobal 只在 Pipeline 模块内部使用。


shrinktofit:
Maybe `pipelineGlobal` is better? But I would suggest instead pass the "global" device/root to place where they are needed(`class ForwardStage` for example).

caryliu1999 (approved):

PR #6963, opened 2020-07-08
Fix a gl error of particle trail.

caryliu1999:
Re: https://forum.cocos.org/t/3d-1-1-0/95312

Changes:
 
Fix the error of 'range out of bounds for buffer' because the number of bursts is incorrectly calculated when the particle's startLifeTime is higher than duration.


pandamicro:
在 1.1.1 分支修复把，这样可以进入 1.1.2

PR #6964, opened 2020-07-08
Fix a gl error of particle trail.

caryliu1999:
Re: https://forum.cocos.org/t/3d-1-1-0/95312

Changes:

Fix the error of 'range out of bounds for buffer' because the number of bursts is incorrectly calculated when the particle's startLifeTime is higher than duration.

### DIFF
--- a/cocos/particle/renderer/trail.ts
+++ b/cocos/particle/renderer/trail.ts
@@ -354,7 +354,8 @@ export default class TrailModule {
         const psTime = ps.startLifetime.getMax();
         const psRate = ps.rateOverTime.getMax();
         const duration = ps.duration;
-        for (const b of ps.bursts) {
+        for (const k in ps.bursts) {
+            const b = ps.bursts[k];
             burstCount += b.getMaxCount(ps) * Math.ceil(psTime / duration);
         }
         this._trailNum = Math.ceil(psTime * this.lifeTime.getMax() * 60 * (psRate * duration + burstCount));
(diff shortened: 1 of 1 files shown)


cocos/particle/renderer/trail.ts
```diff
@@ -351,11 +351,14 @@ export default class TrailModule {
...
         let burstCount = 0;
-        for (const b of this._particleSystem.bursts) {
-            burstCount += b.getMaxCount(this._particleSystem);
-        }
-        this._trailNum = Math.ceil(this._particleSystem.startLifetime.getMax() * this.lifeTime.getMax() * 60 * (this._particleSystem.rateOverTime.getMax() * this._particleSystem.duration + burstCount));
-        this._trailSegments = new Pool(() => new TrailSegment(10), Math.ceil(this._particleSystem.rateOverTime.getMax() * this._particleSystem.duration));
+        const psTime = ps.startLifetime.getMax();
+        const psRate = ps.rateOverTime.getMax();
+        const duration = ps.duration;
+        for (const b of ps.bursts) {
```
pandamicro: Please do not use for of, use standard for or for in instead

pandamicro (changes requested):

cocos/particle/renderer/trail.ts
caryliu1999: > Please do not use for of, use standard for or for in instead

This has been modified.

pandamicro (approved):

PR #6969, opened 2020-07-10
preventing Camera is null exception, for example recalculating editbox constraints after a resize , or even when rendering screen first time.

murataka:
When resizing the screen, editbox impl raises exception while calculating its position.

Re: cocos-creator/2d-tasks#

Changes:
 *

### DIFF
--- a/cocos2d/core/camera/CCCamera.js
+++ b/cocos2d/core/camera/CCCamera.js
@@ -470,7 +470,7 @@ let Camera = cc.Class({
          * @static
          */
         findCamera (node) {
-            let camera;
+            let camera=null;
             for (let i = 0, l = _cameras.length; i < l; i++) {
                  camera = _cameras[i];
                 if (camera.containsNode(node)) {
--- a/cocos2d/core/camera/CCCamera.js
+++ b/cocos2d/core/camera/CCCamera.js
@@ -470,14 +470,21 @@ let Camera = cc.Class({
          * @static
          */
         findCamera (node) {
+            let camera;
             for (let i = 0, l = _cameras.length; i < l; i++) {
-                let camera = _cameras[i];
+                 camera = _cameras[i];
                 if (camera.containsNode(node)) {
                     return camera;
                 }
             }
-
-            return null;
+                if(!camera){
+                    if(_cameras.length>0) camera=_cameras[0]; 
+                    /// return the first camera instead of returning null. EditBox raises error if no camera found , 
+                        ///this may also prevent some other bugs in some other cases. Generally, gamedev would not want to check if the camera is null.. 
+                    /// return the first camera to precalculate even not active objects, thus preventing exceptions. ...
+                     
+                }
+            return camera;
         },
 
         _findRendererCamera (node) {


murataka:
issue appears when no camera is rendering the editbox , but possibly some other web based components are affected.

jareguo:
> issue appears when no camera is rendering the editbox

Ok, anyway, @PPpro please don't touch the editor box when it's hidden. @murataka Thanks for the suggestion.

cocos2d/core/camera/CCCamera.js
```diff
@@ -470,14 +470,21 @@ let Camera = cc.Class({
...
                 if (camera.containsNode(node)) {
                     return camera;
                 }
             }
-
-            return null;
+                if(!camera){
+                    if(_cameras.length>0) camera=_cameras[0]; 
+                    /// return the first camera instead of returning null. EditBox raises error if no camera found , 
+                        ///this may also prevent some other bugs in some other cases. Generally, gamedev would not want to check if the camera is null..
```
jareguo, line 483: Yes, return the main camera is not bad. But what if the real camera is activated after calling findCamera? It will cause a wrong rendering effect.
So if we still return null, everyone will try to relocate the node until a real camera is on.

cocos2d/core/camera/CCCamera.js
```diff
@@ -470,14 +470,21 @@ let Camera = cc.Class({
...
                     return camera;
                 }
             }
-
-            return null;
+                if(!camera){
+                    if(_cameras.length>0) camera=_cameras[0]; 
+                    /// return the first camera instead of returning null. EditBox raises error if no camera found , 
+                        ///this may also prevent some other bugs in some other cases. Generally, gamedev would not want to check if the camera is null.. 
+                    /// return the first camera to precalculate even not active objects, thus preventing exceptions. ...
```
jareguo, line 484: I am assuming the precalulation is meaningless and will eclipse some real bug. Indeed EditBox raises error, and some other codes in the engine also forget to check it.
@PPpro  **We should improve the api description, add some tips for devs, and search all findCamera usage to ensure no error will result even if no active camera in the scene**

murataka: in fact , the issue appears without developer call. The exact situation is :

There are 2 nodes in stack , same size and same position, one includes a webview component. 

Player wants to switch those nodes as fast as possible, so , instead of activating/deactivating them, we use camera culling and blockInputEvents component. An other  advantage of doing this is web based components are not refreshed ( for example webviev reloads and loses content, when activated/deactivated ).

When user loads the scene or switches between nodes,if engine tries to calculate webviews' position, the node containing web based components are culled by camera , so when engine code tries to calculate the position of that component, it calls position calculating function from the engine , which tries to call the null camera's sub functions, throwing an exception. 

The best solution will of course be to not calculate culled web based components' position by default . My solution is just a workaround , but the bug is still there :)

cocos2d/core/camera/CCCamera.js
murataka: i think you did not get the point, please read my other answer , i need to know if the wontfix label will stay or if issue will be solved. and better cocos team may wait for respond of author, and solve ambiguouties , to not get developers stuck , before labeling "wontfix" or like ...

cocos2d/core/camera/CCCamera.js
murataka: Instead of  "We should improve the api description, add some tips for devs, and search all findCamera usage to ensure no error will result event if no active camera in the scene" , 

better "implement a situation which raises that exception  and then fix it" .
Because this was handled well on all other components , and the behavior should be same. I can try to  implement the situation if needed ...

jareguo: > for example webviev reloads and loses content, when activated/deactivated

So you just use camera to simulate the blink effect? It's so weird.



> engine tries to calculate webviews' position, the node containing web based components are culled by camera , so when engine code tries to calculate the position of that component, it calls position calculating function from the engine , which tries to call the null camera's sub functions, throwing an exception.

It's a bug of the engine, we will fix it to ensure no error will result.

cocos2d/core/camera/CCCamera.js
jareguo: The wontfix label means will solve the issue in a more strict way. I really appreciate your feedback, helping us to polish API usages.
I think I got the point. I will wait for your respond before we close this PR.

PPpro:
don't want to change the behavior of `cc.Camare.findCamera()`
we'll check if the founded camera is `null` where `cc.Camara.findCamera()` is called

thanks again for your feedback

PPpro:
I'll close this commit , and fix the bug  here
https://github.com/cocos-creator/engine/pull/7210
https://github.com/cocos-creator-packages/adapters/pull/162
https://github.com/cocos-creator-packages/jsb-adapter/pull/332

PR #6989, opened 2020-07-15
Fix `AnimationState.prototype.sample` can not sample if animation is motionless

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * 

This should solves:
- https://forum.cocos.org/t/cocoscreator3d/95433
- https://forum.cocos.org/t/creator3d-1-1-1-crossfade-normal-loop-bug/95629


gameall3d (approved):
LGTM

PR #6996, opened 2020-07-17
Collider ext

Y-way:
Re: cocos-creator/2d-tasks#

Changes:
 * Add method containsPoint(p, y) for colliders
 * Add color property for colliders
 * Add intersectsSegment(a1, a2, b1, b2) for Intersection. Check if two line-segments intersect and return intersect-point.

### DIFF
--- a/cocos2d/core/collider/CCBoxCollider.js
+++ b/cocos2d/core/collider/CCBoxCollider.js
@@ -24,7 +24,7 @@
  THE SOFTWARE.
  ****************************************************************************/
 
-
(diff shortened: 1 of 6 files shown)


2youyou2:
@Y-way Thanks for your pr and improvement. But if add containsPoint to colliders, then we also need to think about containsRect, containsLine, containsCircle and others.

2youyou2:
@Y-way We will improve these feature in v3.0 according to your pr, thanks again.

PR #7001, opened 2020-07-20
fixed rigidbody bug that cannot wake up after a long sleep

zhefengzhang:
Re: https://github.com/cocos-creator/2d-tasks/issues/2980

Changes:
 * fixed rigidbody bug that cannot wake up after a long sleep
 * changed cannon.js version to 1.0.4


zhefengzhang:
@Jiaxin.Lai 已经进行了完善，此 pr 已经重复，故关闭。

PR #7002, opened 2020-07-20
Optimize editor decorators

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Optimize editor decorators for package size.

There is an optimization point: arguments passed to editor-only decorators can be eliminated from source code when build engine for runtime.

At first, we use Rollup. We use the tree-shaking functionality of Rollup. But our editor decorator arguments are not culled because there was a BUG at https://github.com/rollup/rollup/issues/3674. Fortunately, Rollup has fixed this BUG and published a new verison.

`createDummyDecorator` was not tree-shakable. I've changed that.

Property decorator `property`'s `tooltip` option carried a lot of string literals. We have to separate it from `property` and export a new decorator `tooltip` which is tree-shakable.

---------------------

#### Benchmark

I've test this PR based on commit 8f1a22812a180f59a7a2560f1aec613fec9d9302.

|  	| Commit 8f1a22 	| Cherry-pick this PR 	|
|-	|-	|-	|
| cc.js 	| 4,380 KB 	| 4,363 KB 	|
| cc.min.js 	| 1,983 KB 	| 1,967 KB 	|

You can see that **about 16 KB** space is saved.

### DIFF
--- a/cocos/audio/audio-source-component.ts
+++ b/cocos/audio/audio-source-component.ts
@@ -28,7 +28,7 @@
  */
 
 import { Component } from '../core/components/component';
-import { ccclass, help, menu, property } from '../core/data/class-decorator';
+import { ccclass, help, menu, property, tooltip } from '../core/data/class-decorator';
 import { clamp } from '../core/math';
 import { AudioClip } from './assets/clip';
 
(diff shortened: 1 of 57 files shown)


shrinktofit:
@pandamicro  @jareguo How do you think it's better to place decorator under namespace `property`?

```ts
import { _decorator } from "cc";
const { property } = _decorator;

class {
  @property.tooltip('')
  p = 0;
}
```

pandamicro (approved):

pandamicro:
Does this PR rely on any editor changes ?

PR #7014, opened 2020-07-22
Update  effects

pandamicro:
https://github.com/cocos-creator/engine/pull/6972


pandamicro:
https://github.com/cocos-creator/engine-design-with-hw/issues/6

PR #7036, opened 2020-07-28
Improve cc.Quat.rotateX, cc.Quat.rotateY, cc.Quat.rotateZ.

MatrixLife:
Improve cc.Quat.rotateX, cc.Quat.rotateY, cc.Quat.rotateZ.
Ensure these 3 methods giving correct result when 'out' and 'a' are same object.

Re: cocos-creator/2d-tasks#

Changes:
 * cocos2d/core/value-types/quat.ts


SantyWang:
Thanks

SantyWang (approved):

PR #7038, opened 2020-07-29
Compatible with old webgl RC

PPpro:
re: https://github.com/cocos-creator/3d-tasks/issues/3986

changLog:
- compatibale with experimental-webgl, webkit-3d and moz-webgl rendering context

sync from  pr https://github.com/cocos-creator/engine/pull/3050

### DIFF
--- a/cocos/core/gfx/webgl/webgl-device.ts
+++ b/cocos/core/gfx/webgl/webgl-device.ts
@@ -224,7 +224,11 @@ export class WebGLGFXDevice extends GFXDevice {
             }
             */
 
-            this._webGLRC = this._canvas.getContext('webgl', webGLCtxAttribs);
+            //@ts-ignore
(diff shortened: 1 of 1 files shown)


pandamicro:
I don't think it's necessary, all modern browsers have already supported standard webgl

PPpro:
there are still some users who use creator in native webview project，
maybe standard webgl is not supported in these webviews.

in this situation, i think it's necessary~

pandamicro:
> maybe standard webgl is not supported in these webviews.

I need a confirmed information, not a guess. Which platform ?

PPpro:
met a user running web-mobile on webkit provided by QT，  
the standard webgl is not supported  

the project focuses on the education sector  

so far all I know is this platform


seems that it's a rarely used case

pandamicro:
> running web-mobile on webkit provided by QT，

I think it's their business to upgrade the QT webkit version, so I'm gonna refuse this PR and close the issue

PR #7050, opened 2020-07-30
lightmap instancing

MSoft1115:
Re: cocos-creator/2d-tasks#

Changes:
 *


YunHsiao (approved):

pandamicro:
@YunHsiao @MSoft1115 CI 报错

cocos/core/renderer/scene/planar-shadows.ts:184:30 - error TS2339: Property 'pso' does not exist on type 'InstancedBuffer'.

184         if (buffer && buffer.pso) {

PR #7067, opened 2020-08-03
Modify pipeline resourceization.

caryliu1999:
Re: https://github.com/cocos-creator/3d-tasks/issues/2395

关联编辑器修改：https://github.com/cocos-creator/editor-3d/pull/3890


cocos/core/assets/asset.ts
```diff
@@ -118,9 +118,6 @@ export class Asset extends Eventify(RawAsset) {
      * @type {String}
      * @readOnly
      */
-    @property({
-        visible: false,
-    })
     get nativeUrl () {
```
pandamicro, line 121: This shouldn't be modified, what's the purpose ?

pandamicro (changes requested):

cocos/core/assets/asset.ts
caryliu1999: > This shouldn't be modified, what's the purpose ?

This property only have a getter. There is no corresponding attribute that needs to be serialized, and the editor encounters an error when modifying the data.

pandamicro (approved):

PR #7077, opened 2020-08-05
BaseNode._checkMultipleComp() default return true

arsen2010:
Re: cocos-creator/2d-tasks#

Changes:
 *


shrinktofit:
@pandamicro  This should be merged into 1.1.2.

PR #7097, opened 2020-08-07
fix scene.dependAssets is empty

SantyWang:
Re: cocos-creator/2d-tasks#

Changes:
 * fix scene.dependAssets is empty array

### DIFF
--- a/cocos/core/load-pipeline/uuid-loader.ts
+++ b/cocos/core/load-pipeline/uuid-loader.ts
@@ -32,6 +32,7 @@ import * as js from '../utils/js';
 import {_getClassById} from '../utils/js';
 import { getError, warnID } from '../platform/debug';
 import { LoadingItems } from './loading-items';
+import { getDependsRecursively } from './auto-release-utils';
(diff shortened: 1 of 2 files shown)


cocos/core/load-pipeline/uuid-loader.ts
```diff
@@ -174,6 +175,16 @@ function loadDepends (pipeline, item, asset, depends, callback) {
                 }
             }
         }
+
+        if (asset instanceof legacyCC.SceneAsset && asset.scene) {
+            const dependAssets = dependKeys.concat();
+            for (const dependUuid of dependKeys) {
```
pandamicro: Please use standard for instead of for of

pandamicro (changes requested):

cocos/core/load-pipeline/uuid-loader.ts
SantyWang: OK, I used a map instead of an array

pandamicro (approved):

PR #7131, opened 2020-08-13
Change type of `Mesh.prototype.data` as non-nullable

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Change the type of the `Mesh.prototype.data` from `Uint8Array | null` to `Uint8Array`

The type of the `Mesh.prototype.data` was set to be `Uint8Array | null`. It can be `null` since at initial we think empty meshes naturally shall not have binary data associated.

However empty meshes are rare. Meshes are usually imported from model file. In most case, no empty meshes are produced except that: during mesh de-serialization or when you are going to create and fill a mesh, the `Mesh` constructor have to be called to produce an empty mesh.

As a consequence, we decide to let it be non-nullable. To avoid uncessary `new Uint8Arrray(0)`, empty meshes created as result of constructor invocation would associate with an global, singleton 0-length binary buffer.

### DIFF
--- a/cocos/core/assets/mesh.ts
+++ b/cocos/core/assets/mesh.ts
@@ -473,8 +473,8 @@ export class Mesh extends Asset {
     set _nativeAsset (value: ArrayBuffer) {
         if (this._data.byteLength === value.byteLength) {
             this._data.set(new Uint8Array(value));
-            if (legacyCC.loader._cache[this.nativeUrl]) {	
-                legacyCC.loader._cache[this.nativeUrl].content = this._data.buffer;	
+            if (legacyCC.loader._cache[this.nativeUrl]) {
+                legacyCC.loader._cache[this.nativeUrl].content = this._data.buffer;
             }
         } else {
             this._data = new Uint8Array(value);
--- a/cocos/core/assets/mesh.ts
+++ b/cocos/core/assets/mesh.ts
@@ -685,6 +685,7 @@ export class Mesh extends Asset {
         this.destroyRenderingMesh();
         this._struct = info.struct;
         this._data = info.data;
+        this._dataLength = this.data.byteLength;
         this._hash = 0;
         this.loaded = true;
         this.emit('load');
--- a/cocos/core/assets/mesh.ts
+++ b/cocos/core/assets/mesh.ts
@@ -559,8 +559,12 @@ export class Mesh extends Asset {
 
         this._initialized = true;
 
-        this._data = new Uint8Array(this._dataLength);
-        postLoadMesh(this);
+        if (this._data.byteLength !== this._dataLength) {
+            // In the case of deferred loading, `this._data` is created before
+            // the actual binary buffer is loaded.
+            this._data = new Uint8Array(this._dataLength);
+            postLoadMesh(this);
+        }
         const buffer = this._data.buffer;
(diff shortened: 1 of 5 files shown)


cocos/core/assets/mesh.ts
```diff
@@ -466,18 +467,18 @@ const v3_2 = new Vec3();
...
     }
 
     set _nativeAsset (value: ArrayBuffer) {
-        if (this._data && this._data.byteLength === value.byteLength) {
-            this._data.set(new Uint8Array(value));
+        const dataU8 = new Uint8Array(value);
+        if (this._data.byteLength === value.byteLength) {
+            this._data.set(dataU8);
             if (legacyCC.loader._cache[this.nativeUrl]) {
                 legacyCC.loader._cache[this.nativeUrl].content = this._data.buffer;
```
shrinktofit, line 477: @holycanvas  I'm uncertain for here. Did you mean that: if the new/old `Uint8Array` have same byte length, in stead of replace the old with new, reassign the content old one and **redirect the loader cache** `legacyCC.loader._cache[this.nativeUrl].content` to the old one? But the type of  old `_data` is `Uint8Array`, which may be only a part of `ArrayBuffer` **not the whole array**.

shrinktofit:
@pandamicro  Could you please request review from @holycanvas

cocos/core/assets/mesh.ts
```diff
@@ -566,10 +567,8 @@ export class Mesh extends Asset {
 
         this._initialized = true;
 
-        if (!this._data) {
```
SantyWang, line 569: I think this `if` statement cannot be removed, it may be rewritten as
```
if (this._data === globalEmptyMeshBuffer) {
     // ...
}
```

cocos/core/assets/mesh.ts
SantyWang: This is for the deferred loading case, An empty data with the correct length will be initialized for use during the delayed load, and then set to this empty data after the real data is loaded. And then the data in the loader need to be updated, otherwise it will take up two copies of memory (one in this._data and one in the the loader)

shrinktofit: `Uint8Array` does not copy memories. It is just a view of the `ArrayBuffer`(of the loader).

shrinktofit: As a result: https://github.com/cocos-creator/engine/pull/7131/commits/323ea58925987c6746903aa1712f0f83c364ccfc

VisualSJ:
@holycanvas Please reconfirm the status of this PR

cocos/core/assets/mesh.ts
```diff
@@ -566,10 +559,8 @@ export class Mesh extends Asset {
...
-            this._data = new Uint8Array(this._dataLength);
-            postLoadMesh(this);
-        }
+        this._data = new Uint8Array(this._dataLength);
```
SantyWang: You don't need to create a new Uint8Array if you don't consider deferred loading, right?

shrinktofit: I did not change current logic. The deferred loading still work now.

cocos/core/assets/mesh.ts
shrinktofit: This conversation is invalidated.

SantyWang (approved):

PR #7143, opened 2020-08-14
remove jsb module declaration in engine repo

PPpro:
related pr: https://github.com/cocos-creator-packages/jsb-adapter/pull/327


pandamicro:
Why should we remove it ? As long as it's a valid global variable in native platform, I do think it's necessary to declare it.

PPpro:
jsb declaration in cc.d.ts conflicts with the declaration in jsb.d.ts.
by the way, i don't think it's good to use jsb in engine repo

pandamicro:
> jsb declaration in cc.d.ts conflicts with the declaration in jsb.d.ts.

ok, if it's separated, I'm ok

PR #7150, opened 2020-08-17
sync #5195

2youyou2:
Re: https://github.com/cocos-creator/engine/pull/5195

Changes:
 * sync #5195


shrinktofit:
Why don't you transform it to `.ts`?

2youyou2:
> Why don't you transform it to `.ts`?

Not sure why the origin file is js.

PR #7166, opened 2020-08-18
Full optimize decorators

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Full optimize decorators.

In PR https://github.com/cocos-creator/engine/pull/7002 , we optimized decorator `@tooltip`. This PR continues that work.

The following decorators are added to replace options of `@property`:
| Decorator 	| Corresponding `@property` option 	| Marked as editor-only 	|
|-	|-	|-	|
| `@serializable(boolean)` 	| `.serializable` 	| ❌ 	|
| `@formerlySerializedAs(boolean)` 	| `.formerlySerializedAs` 	| ❌ 	|
| `@override(boolean)` 	| `.override` 	| ❌ 	|
| ❕`@readOnly(boolean)` 	| `.readonly` 	| ❌	|
| `@animatable(boolean)` 	| `.animatable` 	| ❌ 	|
| `@editorOnly(boolean)` 	| `.editorOnly` 	| ❌ 	|
| `@visible(boolean)` 	| `.visible` 	| ✔	|
| `@displayName(string)` 	| `.displayName` 	| ✔ 	|
| `@range([number, number] \| [number, number, number])` 	| `.range` 	| ✔	|
| ❕`@rangeMin(number)` 	| `.min` 	|✔	|
| ❕`@rangeMax(number)` 	| `.max` 	|✔	|
| ❕`@rangeStep(number)` 	| `.step` 	| ✔ 	|
| `@slide(boolean)` 	| `.slide` 	| ✔ 	|
| `@displayOrder(number)` 	| `.displayOrder` 	|✔ 	|
| `@unit(string)` 	| `.unit` 	| ✔ 	|
| `@radian(boolean)` 	| `.radian` 	| ✔ 	|
| `@multiline(boolean)` 	| `.multiline` 	| ✔ 	|

Decorator names prefixed with ❕ indicates the name is different from the option one:
1. - `readonly` is a TypeScript Keyword and may occur at where decoarators may occur;
2. - Words `min`, `max` are so common, so I advise to rename them; so the `step`.

PR #7178, opened 2020-08-19
Full optimize decorators(2)

shrinktofit:
Re: cocos-creator/3d-tasks#

In the discussion with @jareguo ,we feel that PR #7166 may not be the best result: there are some properties marked with a `@property` and the others doesn't, this looks confuse.

Why is it? Well. We separate "editor decorators" from the others since we want to tree-shake them in non-editor environment, especially when do the building. Look at this example:
```ts
@property({tooltip: "A tooltip"})
field = 0;
```
can not be directly turned into:
```ts
@tooltip("A tooltip")
field = 0;
```
because `@property({tooltip: "A tooltip"})` means not only set the editor tooltip but also means this property to be serializable.
We have to add a `@property` to keep the original semantic:
```ts
@property                    // Oops! Elaborated and confused
@tooltip("A tooltip")
field = 0;
```

However this is not always true. If the field is an accessor instead of a property, we may not need to care about the serialization:
```ts
@tooltip("A tooltip") // No `@property` is needed
get fields() { /* ... */ }
```
This is because `@property` does not perform serialization on accessors.

-----------------------------------------------------------

This PR proposes some entities to replace the `@property`, as follows.

There are two basic decorators:
- `@serializable` - The basic/zero-config/minimal decorator for serialization.
- `@editable` - The basic/zero-config/minimal decorator for editor inter-operation.

The functionalities of those basic decorators are **not overlapped**: either for serialization or for editor inter-operation. These basic decorators can be used directly:
```ts
@serializable
_name = /* ... */;

@editable
get name() { return /* ... */ }
```

To configure a little more on serialization or editor inter-operation. The other decorators can be used.

For serialization:
- `@formerlySerializedAs`

For editor inter-operation:
- `@tooltip`
- `@range`
- .....

Let's call the decorators in the former list **serialization decorators** and the decorators in the later list **editor decorators**. We can definitely specify that:
- Each serialization decorators implies `@serializable`;
- each editor decorators implies `@editable`.

With this kind of grouping, the above example
```ts
@property({tooltip: "A tooltip"})
field = 0;
```
can be converted as:
```ts
@serializable                       // Say: it's subjected to serialization
@tooltip("A tooltip")           // Say: it's subjected to editor inter-operation and, in advance, specify the tooltip
                                            // -----------------------------------------------------------
                                            // It's clear since `@serializable` and `@tooltip` are isolated:
                                            // - a property is serializable if at least one serialization decorator is attached;
                                            // - a property is editable if at least one editor decorator is attached
field = 0;
```
I also document this in [docs-3d](https://github.com/cocos-creator/docs-3d/pull/331).

----------------------------------------------------------------------------------

It's even more hard to convert the current decorators into this PR proposed'.

I convert it according the following rules, reviewers of the PR should concern.

For accessors:
- A single `@property` -> `@editable`  For example: https://github.com/cocos-creator/engine/pull/7178/files#diff-eb6931eacba4603b5111ceb12ca442d6R87
- `@property` + `<editor-decorators>` -> remove `@property`
- A single `@property({ visible: false })` -> **remove all decorators**

For "public" properties(not prefixed with `_`):
- A single `@property` -> `@serializable @editable` For example: https://github.com/cocos-creator/engine/pull/7178/files#diff-11bfc24ef1cf61a1bdab384375ef2941R166-R167
- `@property` + `<editor-decorators>` -> `@serializable` + `<editor-decorators>` For example: https://github.com/cocos-creator/engine/pull/7178/files#diff-eb6931eacba4603b5111ceb12ca442d6R143
- A single `@property({ visible: false })` -> `@serializable` For example: https://github.com/cocos-creator/engine/pull/7178/files#diff-11bfc24ef1cf61a1bdab384375ef2941R86-R87

For "private" properties(prefixed with `_`):
- A single `@property` -> `@serializable` For example: https://github.com/cocos-creator/engine/pull/7178/files#diff-4e6db4eed51a6400de9a2ee5a3b97973R64
- A single `@property({ visible: false})` -> `@serializable` 
- `@property` + `<editor-decorators>` -> Attach `@serializable` onto the property, attach the remain decorators onto the corresponding no-underscore version accessor. For example: https://github.com/cocos-creator/engine/pull/7178/files#diff-e59164d0ed114516e0ae9a715ac1f9bfR63-R81


cocos/core/3d/framework/renderable-component.ts
```diff
@@ -21,10 +21,9 @@ export class RenderableComponent extends Component {
     @type([Material])
     protected _materials: (Material | null)[] = [];
 
-    @property
+    @serializable
     protected _visFlags = Layers.Enum.NONE;
 
-    @visible(false)
     get visibility () {
```
shrinktofit, line 27: @jareguo  Could you please confirm this change? Why does the `visible(false)` need to be attached before?

cocos/core/animation/animation-clip.ts
```diff
@@ -143,47 +143,46 @@ export class AnimationClip extends Asset {
...
     public wrapMode = AnimationWrapMode.Normal;
 
     /**
      * @zh 动画包含的事件数据。
      * @en Associated event data.
      */
-    @property
-    @visible(false)
+    @serializable
     public events: AnimationClip.IEvent[] = [];
```
shrinktofit, line 168: I've just change `@property` into `@serializable` because this properties need not `@editable`.

cocos/core/animation/cubic-spline-value.ts
```diff
@@ -102,13 +102,13 @@ legacyCC.CubicSplineQuatValue = CubicSplineQuatValue;
...
-    @property
+    @serializable
     public dataPoint: number = 0;
 
-    @property
+    @serializable
     public inTangent: number = 0;
 
-    @property
+    @serializable
```
shrinktofit, line 111: Notable changes

cocos/core/assets/asset.ts
```diff
@@ -104,7 +98,7 @@ export class Asset extends Eventify(RawAsset) {
      * 用于本机资产的可序列化URL。供内部使用。
      * @default ""
      */
-    @property
+    @serializable
```
shrinktofit, line 102: I don't think this property need to be shown in editor

cocos/core/assets/render-texture.ts
```diff
@@ -60,24 +60,22 @@ const _windowInfo: IRenderWindowInfo = {
...
 
+    @rangeMin(1)
+    @rangeMax(2048)
     get width () {
         return this._width;
     }
 
+    @rangeMin(1)
+    @rangeMax(2048)
     get height () {
```
shrinktofit, line 79: @JoneLau

cocos/core/components/component.ts
```diff
@@ -67,8 +67,6 @@ const NullNode = null as unknown as Node;
  */
 @ccclass('cc.Component')
 class Component extends CCObject {
-
-    @visible(false)
     get name () {
```
shrinktofit, line 70: @shrinktofit

cocos/core/pipeline/forward/forward-stage.ts
```diff
@@ -37,7 +37,6 @@ export class ForwardStage extends RenderStage {
 
 
     @type([RenderQueueDesc])
-    @visible(true)
```
shrinktofit, line 41: @caryliu1999  Please confirm

cocos/core/pipeline/ui/ui-stage.ts
```diff
@@ -28,7 +28,6 @@ export class UIStage extends RenderStage {
     };
 
     @type([RenderQueueDesc])
-    @visible(true)
```
shrinktofit, line 31: @caryliu1999

cocos/core/scene-graph/scene-globals.ts
```diff
@@ -582,23 +582,27 @@ export class SceneGlobals {
...
      * @en shadow information
      * @zn 普通阴影相关信息
      */
-    @property
+    @editable
+    @serializable
     public shadowMap = new ShadowInfo();
-    @property
+    @serializable
     public _skybox = new SkyboxInfo();
```
shrinktofit, line 602: @caryliu1999  Please confirm: no `@editable`

cocos/particle/renderer/trail.ts
```diff
@@ -229,10 +228,7 @@ export default class TrailModule {
     /**
      * 粒子本身是否存在。
      */
-    @property
-    @visible(false)
-    @displayOrder(7)
-    @tooltip('拖尾是否跟随粒子一起消失')
+    @serializable
```
shrinktofit, line 233: @caryliu1999  Confirm required

cocos/core/3d/framework/renderable-component.ts
jareguo: I don't know, we don't have `visibility` in 2D

cocos/core/data/utils/attribute-defines.ts
```diff
@@ -99,6 +99,19 @@ export interface IExposedAttributes {
...
+     * 此选项是为了在 `@property` 的基础上精确实现 `@serializable`、`@editable`以及所有新增的独立装饰器的行为。
+     * 
+     * 当此字段为 `true` 时。以下规则将不再生效：
+     * - 只要 `@property` 未显式指定选项 `.serializable === false`，就开启序列化；
+     * - 只要 `@property` 未显式指定选项 `.visible === false` 且目标属性的名称不以下划线开头，就开启编辑器交互。
+     * 反之，由以下规则取代：
+     * - 当且仅当 `@property` 显式指定了 `.serializable === true` 时才开启序列化；
+     * - 当且仅当 `@property` 显式指定了 `.visible === true` 时才开启编辑器交互。
+     */
+    __noImplicit?: boolean;
```
shrinktofit, line 114: @jareguo  See https://github.com/cocos-creator/engine/pull/7178/files#diff-afbab3bdaffe775035fa261505d51adeR1162-R1172 and https://github.com/cocos-creator/engine/pull/7178/files#diff-afbab3bdaffe775035fa261505d51adeR1185-R1201

cocos/core/assets/asset.ts
shrinktofit: @VisualSJ Think of dumping, does it matter?

cocos/core/components/component.ts
shrinktofit: Confirmed with @jareguo  and @VisualSJ : `.name`, `.uuid`, `.enabled` of `Component.prototype` are not subjected to serializable or editor-interop.

cocos/core/scene-graph/base-node.ts
```diff
@@ -124,7 +124,7 @@ export class BaseNode extends CCObject implements ISchedulable {
      * @en Name of node.
      * @zh 该节点名称。
      */
-    @property
+    @editable
```
shrinktofit, line 128: Are these decorators needed? @VisualSJ

cocos/core/assets/asset.ts
VisualSJ: Agreed, the editor does not currently use these asset definitions

cocos/particle/renderer/particle-system-renderer-data.ts
```diff
@@ -68,19 +68,16 @@ export default class ParticleSystemRenderer {
...
     private _velocityScale = 1;
 
-    @property
-    @displayOrder(5)
+    @serializable
     private _lengthScale = 1;
 
-    @property
-    @displayOrder(6)
+    @serializable
```
shrinktofit, line 80: @caryliu1999 Notable changes

cocos/core/scene-graph/base-node.ts
VisualSJ: No, the properties on Node are fixed when they are used

cocos/core/data/utils/attribute-defines.ts
jareguo: LGTM

cocos/core/pipeline/ui/ui-stage.ts
caryliu1999: > @caryliu1999

ok

cocos/core/scene-graph/scene-globals.ts
caryliu1999: > @caryliu1999 Please confirm: no `@editable`

This property need edit on inspector, can only serializable be satisfied?

cocos/particle/renderer/trail.ts
caryliu1999: > @caryliu1999 Confirm required

ok

cocos/particle/renderer/particle-system-renderer-data.ts
caryliu1999: > @caryliu1999 Notable changes

Is the order of display on inspector determined by the order of declarations?

caryliu1999 (approved):

cocos/core/scene-graph/scene-globals.ts
shrinktofit: I moved the `@editable` to its accessor

PR #7191, opened 2020-08-20
fixed the bug when the label's cachemode is char

JoneLau:


pandamicro (approved):

LinYunMo:
RE: https://github.com/cocos-creator/3d-tasks/issues/4192

PR #7197, opened 2020-08-21
Add array pool and add more elements into shared memory

minggo:


YunHsiao (approved):

Mee-gu (approved):

cocos/core/director.ts
```diff
@@ -290,7 +291,8 @@ export class Director extends EventTarget {
         this._scene = null;
 
         // FPS
-        this._totalFrames = 0;
+        this._poolHandle = DirectorPool.alloc();
+        DirectorPool.set(this._poolHandle, DirectorView.TOTAL_FRAMES, 0);
```
pandamicro: Why director needs to be stored into Pool ? If total frames is required in renderer, why don't the renderer count by itself

minggo: Ok, can calculate it in native. It is fixed.

PR #7230, opened 2020-08-27
fixed single touch，always keeping only one finger message

JoneLau:
Re: https://github.com/cocos-creator/3d-tasks/issues/4219

review by @LinYunMo @caryliu1999

### DIFF
--- a/cocos/core/platform/event-manager/input-manager.ts
+++ b/cocos/core/platform/event-manager/input-manager.ts
@@ -144,7 +144,7 @@ class InputManager {
         }
         if (handleTouches.length > 0) {
             // this._glView!._convertTouchesWithScale(handleTouches);
-            const touchEvent = new EventTouch(handleTouches, false, EventTouch.BEGAN, this._getUsefulTouches());
+            const touchEvent = new EventTouch(handleTouches, false, EventTouch.BEGAN, macro.ENABLE_MULTI_TOUCH ? this._getUsefulTouches() : handleTouches);
             eventManager.dispatchEvent(touchEvent);
         }
     }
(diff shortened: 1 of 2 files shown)


cocos/core/platform/event-manager/input-manager.ts
```diff
@@ -173,7 +173,7 @@ class InputManager {
         }
         if (handleTouches.length > 0) {
             // this._glView!._convertTouchesWithScale(handleTouches);
-            const touchEvent = new EventTouch(handleTouches, false, EventTouch.MOVED, this._getUsefulTouches());
+            const touchEvent = new EventTouch(handleTouches, false, EventTouch.MOVED, macro.ENABLE_MULTI_TOUCH ? this._getUsefulTouches() : [handleTouches[0]]);
```
pandamicro: Obviously, when you made four times the same logic and they are all using the same function, you should move that logic into the function.

Why not just return correct touch array in `_getUsefulTouches`, by the way, it also avoids another array creation

pandamicro (changes requested):

cocos/core/platform/event-manager/input-manager.ts
JoneLau: I don’t think it’s appropriate to add some extraneous parameters to the function. The modification avoids the creation of another array.

pandamicro: There is no extra parameter needed, just use `macro.ENABLE_MULTI_TOUCH`, it's a global setting, so I think it's appropriate to make the touch collection process obey the behavior

pandamicro (approved):

PR #7236, opened 2020-08-27
Add serialized data synchronization for pipeline.

caryliu1999:


cocos/core/root.ts
```diff
@@ -312,6 +310,12 @@ export class Root {
         return true;
     }
 
+    public createDefaultPipeline () {
+        const rppl = new legacyCC.ForwardPipeline();
```
pandamicro: why use legacyCC.ForwardPipeline ? for native compatibility ?

I think if you import FrowardPipeline from `renderer` module, it can be automatically switched to native ForwardPipeline

pandamicro:
By the way, is there any dependency with https://github.com/cocos-creator/engine/pull/7218

cocos/core/root.ts
caryliu1999: > why use legacyCC.ForwardPipeline ? for native compatibility ?
> 
> I think if you import FrowardPipeline from `renderer` module, it can be automatically switched to native ForwardPipeline

modified

pandamicro (approved):

PR #7258, opened 2020-08-31
remove Don't_Clear in CanvasComponent

JoneLau:
Re: https://github.com/cocos-creator/3d-tasks/issues/4330


pandamicro:
Why remove DONT CLEAR ? It's a necessary option

pandamicro (approved):

PR #7280, opened 2020-09-02
add submodle into array pool

minggo:


LinYunMo (approved):

cocos/core/renderer/ui/ui-batch-model.ts
```diff
@@ -63,6 +69,7 @@ export class UIBatchModel extends Model {
     }
 
     public destroy () {
+        super.destroy();
```
pandamicro: super destroy should be at the end

minggo: Sounds good. It is fixed.

pandamicro (approved):

PR #7291, opened 2020-09-03
Handle WebGL context lost event

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * This PR lets both the `WebGLDevice` and `WebGL2Device` simply handle the WebGL context lost event.

### DIFF
--- a/cocos/core/gfx/webgl/webgl-device.ts
+++ b/cocos/core/gfx/webgl/webgl-device.ts
@@ -1,5 +1,5 @@
 import { ALIPAY, RUNTIME_BASED, BYTEDANCE, WECHAT, DEBUG, VIVO } from 'internal:constants';
-import { macro } from '../../platform';
+import { macro, warnID, warn } from '../../platform';
 import { sys } from '../../platform/sys';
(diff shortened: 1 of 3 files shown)


cocos/core/gfx/webgl/webgl-device.ts
```diff
@@ -789,4 +799,10 @@ export class WebGLDevice extends GFXDevice {
         gl.colorMask(true, true, true, true);
         gl.blendColor(0.0, 0.0, 0.0, 0.0);
     }
+
+    private _onWebGLContextLost (event: Event) {
+        warnID(11000);
+        warn(event);
+        event.preventDefault();
```
shrinktofit: @PPpro  Please verify if the `preventDefault` is implemented on all adatpers.

pandamicro (approved):

cocos/core/gfx/webgl/webgl-device.ts
PPpro: 'webglcontextlost' event not supported in minigame platforms

PR #7300, opened 2020-09-03
fix uniform1iv

YunHsiao:
Re: cocos-creator/3d-tasks#

Changelog:
 * alipay is the first to discover the UB we relied on.


LinYunMo:
Re：https://github.com/cocos-creator/3d-tasks/issues/4341

pandamicro (approved):

PR #7345, opened 2020-09-09
throw an error when no image available

SantyWang:
Re: cocos-creator/2d-tasks#

Changes:
 * 没有可用图片时抛出异常

### DIFF
--- a/cocos2d/core/assets/CCTexture2D.js
+++ b/cocos2d/core/assets/CCTexture2D.js
@@ -975,9 +975,12 @@ var Texture2D = cc.Class({
                 this._setRawAsset(result.bestExt);
                 this._format = result.bestFormat;
             }
-            else {
+            else if (result.defaultExt) {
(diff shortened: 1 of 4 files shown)


cocos2d/core/assets/CCTexture2D.js
```diff
@@ -975,10 +975,13 @@ var Texture2D = cc.Class({
                 this._setRawAsset(result.bestExt);
                 this._format = result.bestFormat;
             }
-            else {
+            else if (result.defaultExt) {
                 this._setRawAsset(result.defaultExt);
                 cc.warnID(3120, result.defaultExt, result.defaultExt);
             }
+            else {
+                throw new Error('No available format of image was found for this platform!');
```
jareguo: 用 throw new Error(debug.getError(xxxx)) 吧
Can't find a texture format supported by the current platform!  Please add a fallback format in the editor.

PR #7410, opened 2020-09-17
Fix the issue of engine restart.

caryliu1999:
Re: https://github.com/cocos-creator/3d-tasks/issues/3360

Fixed engine restart error and memory leaks after restart.

### DIFF
--- a/cocos/core/director.ts
+++ b/cocos/core/director.ts
@@ -540,8 +540,8 @@ export class Director extends EventTarget {
         if (this._root) {
             this._root.resetCumulativeTime();
         }
+        legacyCC.game.resume();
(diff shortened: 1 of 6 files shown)


cocos/core/director.ts
```diff
@@ -540,8 +540,8 @@ export class Director extends EventTarget {
         if (this._root) {
             this._root.resetCumulativeTime();
         }
+        legacyCC.game.resume();
```
pandamicro: Should not resume the game here, it's not the director who paused the game

pandamicro (changes requested):

pandamicro (approved):

PR #7425, opened 2020-09-21
Fix the decorators error in pipeline.

caryliu1999:
装饰器重构时对 pipeline 这里的声明有问题，修复完善

### DIFF
--- a/cocos/core/pipeline/render-flow.ts
+++ b/cocos/core/pipeline/render-flow.ts
@@ -1,7 +1,7 @@
 /**
  * @category pipeline
  */
-import { ccclass, displayOrder, serializable } from 'cc.decorator';
+import { ccclass, displayOrder, serializable, type } from 'cc.decorator';
 import { RenderStage } from './render-stage';
 import { RenderView } from './render-view';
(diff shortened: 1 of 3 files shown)


cocos/core/pipeline/render-flow.ts
```diff
@@ -69,7 +82,7 @@ export abstract class RenderFlow {
     @serializable
     protected _tag: number = 0;
 
-    @serializable
+    @type([RenderStage])
```
pandamicro, line 72: type doesn't include serializable, I think both are needed, please confirm with @shrinktofit

cocos/core/pipeline/render-pipeline.ts
```diff
@@ -59,25 +59,31 @@ export abstract class RenderPipeline extends Asset {
         return this._macros;
     }
 
+    /**
+     * @en The tag of pipeline.
+     * @zh 管线的标签。
+     * @readonly
```
pandamicro, line 64: Remove readonly for all properties with setter

pandamicro (changes requested):

pandamicro (approved):

PR #7427, opened 2020-09-21
init main light info

YunHsiao:
Re: cocos-creator/3d-tasks#

Changelog:
 * or else we cant see sh*t

### DIFF
--- a/cocos/core/renderer/scene/directional-light.ts
+++ b/cocos/core/renderer/scene/directional-light.ts
@@ -102,9 +102,10 @@ export class DirectionalLight extends Light {
         this._type = LightType.DIRECTIONAL;
     }
 
-    public initialize() {
+    public initialize () {
         super.initialize();
         LightPool.set(this._handle, LightView.ILLUMINANCE, Ambient.SUN_ILLUM);
+        LightPool.setVec3(this._handle, LightView.DIRECTION, this._dir);
     }
 
     public update () {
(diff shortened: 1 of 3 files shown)


cocos/core/renderer/scene/directional-light.ts
```diff
@@ -102,9 +102,10 @@ export class DirectionalLight extends Light {
         this._type = LightType.DIRECTIONAL;
     }
 
-    public initialize() {
+    public initialize () {
         super.initialize();
-        LightPool.set(this._handle, LightView.ILLUMINANCE, Ambient.SUN_ILLUM);
+        this.illuminance = Ambient.SUN_ILLUM;
+        this.direction = this._dir;
```
pandamicro: This is not a good idea, if one day someone decide to do parameter check in direction setter, then the initialize process will be broken unintentionally.

Please explicitly set the LightView direction in pool

cocos/core/renderer/scene/light.ts
```diff
@@ -120,6 +120,8 @@ export class Light {
 
     public initialize () {
         this._handle = LightPool.alloc();
+        this.color = this._color;
+        this.colorTemperature = this._colorTemp;
```
pandamicro: Same argument as previous

pandamicro (changes requested):

minggo (approved):

pandamicro (approved):

PR #7455, opened 2020-09-27
fix physics 3d auto sync transform

JayceLai:
Re: https://github.com/cocos-creator/engine/pull/6949

Changes:
 * fix physics 3d auto sync transform


JayceLai:
https://github.com/cocos-creator/2d-tasks/issues/3202

SantyWang (approved):

PR #7462, opened 2020-09-28
fix camera.clearColor

YunHsiao:
Re: cocos-creator/3d-tasks#

Changelog:
 * camera.clearColor should actually be a IVec4Like


cocos/core/renderer/scene/camera.ts
```diff
@@ -381,12 +381,15 @@ export class Camera {
     }
 
     set clearColor (val) {
```
minggo, line 383: val 加类型是否会好点。

minggo (approved):

PR #7472, opened 2020-09-30
Modify the math library to memory pool.

caryliu1999:
https://github.com/cocos-creator/3d-tasks/issues/4111


shrinktofit (commented):
Let's wait for the performance test.

PR #7496, opened 2020-10-10
array pool refactor

YunHsiao:
Re: cocos-creator/3d-tasks#4535

Changelog:
 * we need some kind of raw buffer pool for things like instance buffers. refactored array pool to extends base class `BufferAllocator`, which is designed to allocate exactly that.


cocos/core/renderer/scene/render-scene.ts
```diff
@@ -159,9 +159,11 @@ export class RenderScene {
         this.removeSpotLights();
         this.removeModels();
         if (this._modelArrayHandle) {
-            ModelArrayPool.free(this._modelArrayHandle);
-            ScenePool.free(this._scenePoolHandle);
+            freeHandleArray(this._modelArrayHandle, ModelArrayPool, ModelPool);
```
minggo: `this.removeModels();` 会调用 `Model.destroy()`，从而释放 Model 自己的 handle。为什么需要再次释放呢？

cocos/core/renderer/ui/ui-batch-model.ts
```diff
@@ -67,7 +67,7 @@ export class UIBatchModel extends Model {
 
     public directInitialize (batch: UIDrawBatch) {
         this._subModel.directInitialize(batch.material!.passes, batch.hInputAssembler, batch.hDescriptorSet!);
-        SubModelArrayPool.assign(this._subModelArrayHandle, 0, this._subModel.handle);
+        SubModelArrayPool.assign(ModelPool.get(this._handle, ModelView.SUB_MODEL_ARRAY), 0, this._subModel.handle);
```
minggo: _subModelArrayHandle 直接从父类继承下来，没必要再通过 ModelPool 去获取吧？

cocos/core/renderer/scene/render-scene.ts
YunHsiao: 噢这个不应该改

cocos/core/renderer/ui/ui-batch-model.ts
YunHsiao: 没看到这种高频调用的地方还用到了，我把父类里的成员变量删了，我改下子类吧

PR #7552, opened 2020-10-17
enable babel loose mode

jareguo:
Reduce min size from 2,602 KB -> 2,523 KB

Some differences, the left side is the original, the right side is loosed.
image

image

image

image


shrinktofit:
Could you close this since I fire a more complete https://github.com/cocos-creator/engine/pull/7601 ?

PR #7597, opened 2020-10-26
correct the name of physics material

JayceLai:
Re: cocos-creator/3d-tasks#4714

Changes:
 * correct the name of physics material

### DIFF
--- a/cocos/physics/framework/index.ts
+++ b/cocos/physics/framework/index.ts
@@ -4,7 +4,7 @@
  */
 
 import { PhysicsSystem } from './physics-system';
-import { PhysicMaterial } from './assets/physic-material';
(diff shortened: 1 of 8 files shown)


JayceLai:
关联PR https://github.com/cocos-creator/editor-3d/pull/4399

cocos/physics/framework/index.ts
```diff
@@ -45,15 +45,15 @@ export {
 
     RigidBody,
 
-    PhysicMaterial,
+    PhysicsMaterial as PhysicMaterial,
```
pandamicro: we should export PhysicsMaterial instead of PhysicMaterial

pandamicro (approved):

PR #7610, opened 2020-10-28
Merge TiledMap from 2.x

PatriceJiang:
Re: https://github.com/cocos-creator/3d-tasks/issues/3273

Changes:
 *


cocos/core/scene-graph/node.ts
```diff
@@ -562,11 +562,23 @@ export class Node extends BaseNode {
...
-    public setPosition (val: Vec3 | number, y?: number, z?: number) {
-        if (y === undefined || z === undefined) {
+    
+    /**
+     * @en Set position in local coordinate system
+     * @zh 设置本地坐标
+     * @param x X axis position
+     * @param y Y axis position
+     */
+    public setPosition (x: number, y: number): void;
```
pandamicro, line 572: If position support 2d API, then scale and eulerAngles should too

cocos/core/renderer/ui/ui.ts
```diff
@@ -645,17 +646,25 @@ export class UI {
...
         const len = node.children.length;
 
         const parentOpacity = this._parentOpacity;
         this._parentOpacity *= node._uiProps.opacity;
+
+        // skip if processed
+        const comp = node._uiProps.uiComp as UIRenderable;
+        if (comp) {
+            if (comp._uiWalkGen === this._walkGen) { return /* already run */; }
+            comp._uiWalkGen = this._walkGen;
```
pandamicro: why we need this ? Why a component will be walked more than once

pandamicro (changes requested):
Please put tile map into core/tiled-map module

cocos/core/scene-graph/node.ts
PatriceJiang: fixed in commit https://github.com/cocos-creator/engine/pull/7610/commits/ed35eb2e8e16cbe430e8f77bce7ec3ff5cf7b70d

pandamicro (approved):

PR #7612, opened 2020-10-28
Promisify `game.init` and `game.run`

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Promisify `game.init()` and `game.run()` since we want to dynamicly initialize some resources(shader resources according to device).

### DIFF
--- a/cocos/core/game.ts
+++ b/cocos/core/game.ts
@@ -982,10 +982,11 @@ export class Game extends EventTarget {
     private _setupRenderPipelineAndShowSplashScreen () {
         return Promise.resolve(this._setupRenderPipeline()).then(() => {
             this._safeEmit(Game.EVENT_GAME_INITED);
-            if (this.onStart) {
-                this.onStart();
-            }
-            return this._showSplashScreen();
+            return Promise.resolve(this._showSplashScreen()).then(() => {
+                if (this.onStart) {
+                    this.onStart();
+                }
+            });
         });
     }
 
--- a/cocos/core/game.ts
+++ b/cocos/core/game.ts
@@ -377,7 +377,7 @@ export class Game extends EventTarget {
 
     public _intervalId: number | null = null; // interval target of main
 
-    public declare _lastTime: Date;
+    public declare _lastTime: number;
     public declare _frameTime: number;
 
     // Scenes list
@@ -578,20 +578,20 @@ export class Game extends EventTarget {
         });
     }
 
-    public run (config: IGameConfig, onStart?: Game.OnStart): Promise<void>;
-
-    public run (onStart?: Game.OnStart): Promise<void>;
-
     /**
      * @en Run game with configuration object and onStart function.
      * @zh 运行游戏，并且指定引擎配置和 onStart 的回调。
      * @param onStart - function to be executed after game initialized
      */
+    public run (onStart?: Game.OnStart): Promise<void>;
+
     public run (configOrCallback?: Game.OnStart | IGameConfig, onStart?: Game.OnStart) {
         if (!EDITOR) {
             this._initEvents();
         }
 
+        // To compatible with older version,
+        // we allow the `run(config, onstart?)` form. But it's deprecated.
         let initPromise: Promise<boolean> | undefined;
         if (typeof configOrCallback !== 'function' && configOrCallback) {
             initPromise = this.init(configOrCallback);
@@ -687,7 +687,7 @@ export class Game extends EventTarget {
 
     //  @Time ticker section
     private _setAnimFrame () {
-        this._lastTime = new Date();
+        this._lastTime = new Date().getTime();
         const frameRate = this.config.frameRate;
         this._frameTime = 1000 / frameRate;
         legacyCC.director._maxParticleDeltaTime = this._frameTime / 1000 * 2;
@@ -730,10 +730,10 @@ export class Game extends EventTarget {
     }
     private _stTime (callback) {
         const currTime = new Date().getTime();
-        const elapseTime = Math.max(0, (currTime - (this._lastTime as unknown as number)));
+        const elapseTime = Math.max(0, (currTime - this._lastTime));
         const timeToCall = Math.max(0, this._frameTime - elapseTime);
         const id = window.setTimeout(callback, timeToCall);
-        (this._lastTime as unknown as number) = currTime + timeToCall;
+        this._lastTime = currTime + timeToCall;
         return id;
     }
     private _ctTime (id: number | undefined) {
--- a/cocos/core/game.ts
+++ b/cocos/core/game.ts
@@ -131,7 +131,7 @@ export interface IGameConfig {
      * 5 - `error`，`assert`，`warn` will print on canvas, available only on web.
      * 6 - `error`，`assert` will print on canvas, available only on web.
      */
-    debugMode?: 0 | 1 | 2 | 3 | 4 | 5 | 6;
+    debugMode?: debug.DebugMode;
 
     /**
      * @zh
@@ -778,7 +778,7 @@ export class Game extends EventTarget {
     private _initConfig (config: IGameConfig) {
         // Configs adjustment
         if (typeof config.debugMode !== 'number') {
-            config.debugMode = 0;
+            config.debugMode = debug.DebugMode.NONE;
         }
         config.exposeClassName = !!config.exposeClassName;
         if (typeof config.frameRate !== 'number') {
--- a/cocos/core/game.ts
+++ b/cocos/core/game.ts
@@ -493,7 +493,7 @@ export class Game extends EventTarget {
 
             legacyCC.director.reset();
             this.pause();
-            return this._setupRenderPipelineAndShowSplashScreen().then(() => {
+            return this._setRenderPipelineNShowSplash().then(() => {
                 this.resume();
                 this._safeEmit(Game.EVENT_RESTART);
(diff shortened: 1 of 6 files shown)


shrinktofit:
@caryliu1999  Please review this.

cocos/core/game.ts
```diff
@@ -377,17 +377,14 @@ export class Game extends EventTarget {
 
     public _intervalId: number | null = null; // interval target of main
 
-    public _lastTime: Date | null = null;
-    public _frameTime: number | null = null;
+    public declare _lastTime: Date;
+    public declare _frameTime: number;
```
shrinktofit: This confuses me why `_lastTime` and `_frameTime` can be null. Please confirm this. @pandamicro

cocos/core/game.ts
```diff
@@ -477,22 +474,24 @@ export class Game extends EventTarget {
...
             legacyCC.director.getScene().destroy();
             legacyCC.Object._deferredDestroy();
 
             legacyCC.director.reset();
-            legacyCC.game.pause();
-            legacyCC.game._loadRenderPipeline(() => {
-                legacyCC.game.resume();
-                legacyCC.game._safeEmit(legacyCC.Game.EVENT_RESTART);
+            this.pause();
+            return this._setupRenderPipelineAndShowSplashScreen().then(() => {
```
shrinktofit: @caryliu1999  The `_setupRenderPipelineAndShowSplashScreen` is as its name. You could suggest it with a new name :)

cocos/core/game.ts
```diff
@@ -569,44 +568,49 @@ export class Game extends EventTarget {
...
-        // register system events
-        if (!EDITOR && game.config.registerSystemEvent) {
-            inputManager.registerSystemEvent(game.canvas);
+        let initPromise: Promise<boolean> | undefined;
+        if (typeof configOrCallback !== 'function' && configOrCallback) {
+            initPromise = this.init(configOrCallback);
+            this.onStart = onStart ?? null;
+        } else {
+            this.onStart = configOrCallback ?? null;
         }
```
shrinktofit, line 605: @pandamicro  The logic here should be correct?

cocos/core/game.ts
```diff
@@ -968,67 +972,50 @@ export class Game extends EventTarget {
...
         this.on(Game.EVENT_SHOW, () => {
-            legacyCC.game.resume();
+            this.resume();
         });
     }
 
-    private _loadRenderPipeline (restart: Function | null) {
-        const useSplash = (!EDITOR && !PREVIEW && legacyCC.internal.SplashScreen);
-
-        // Load render pipeline if needed
```
shrinktofit, line 981: @caryliu1999  Please review this.

cocos/core/game.ts
```diff
@@ -968,67 +972,50 @@ export class Game extends EventTarget {
...
-                const splashScreen = legacyCC.internal.SplashScreen.instance as SplashScreen;
-                splashScreen.main(legacyCC.director.root);
-                splashScreen.setOnFinish(() => {
-                    if (this.onStart) { this.onStart(); }
-                    if (restart) { restart(); }
+            if (this.onStart) {
+                this.onStart();
+            }
+            return this._showSplashScreen();
+        });
```
caryliu1999, line 994: onStart should be executed after splashScreen finish.

shrinktofit: Updated.

caryliu1999 (approved):

cocos/core/game.ts
pandamicro: It's ok to only declare here or assign any default value

cocos/core/game.ts
```diff
@@ -377,17 +377,14 @@ export class Game extends EventTarget {
...
+    public declare _frameTime: number;
 
     // Scenes list
     public _sceneInfos: ISceneInfo[] = [];
     public collisionMatrix = [];
     public groupList: any[] = [];
 
-    // @Methods
-
-    //  @Game play control
```
pandamicro, line 390: They are paragraph markers, no need to remove them

cocos/core/game.ts
```diff
@@ -569,44 +568,49 @@ export class Game extends EventTarget {
...
-        if (!EDITOR) {
-            this._initEvents();
-        }
+        return this._initEngine().then(() => {
+            if (!EDITOR) {
+                this._initEvents();
+            }
+    
+            legacyCC.director.root.dataPoolManager.jointTexturePool.registerCustomTextureLayouts(config.customJointTextureLayouts);
+            return this._inited;
```
pandamicro, line 581: How this return value would be used ? Originally it's for the invoker to determine whether the init process succeed, now it seems useless. Instead we should handle fail case in `_initEngine` which could reject.

cocos/core/game.ts
```diff
@@ -569,44 +568,49 @@ export class Game extends EventTarget {
...
-    }
+    public run (onStart?: Game.OnStart): Promise<void>;
 
     /**
      * @en Run game with configuration object and onStart function.
      * @zh 运行游戏，并且指定引擎配置和 onStart 的回调。
      * @param onStart - function to be executed after game initialized
      */
-    public run (onStart: Function | null, legacyOnStart?: Function | null) {
+    public run (configOrCallback?: Game.OnStart | IGameConfig, onStart?: Game.OnStart) {
```
pandamicro, line 592: Please keep the same parameter name, use `onStart` and `legacyOnStart`, because old version in 2.x was

`public run (config, onStart)`

new version is always

`public run (onStart)`

It's just for backward compatibility, not for two kinds of usage

cocos/core/game.ts
```diff
@@ -569,44 +568,49 @@ export class Game extends EventTarget {
...
+                this._initEvents();
+            }
+    
+            legacyCC.director.root.dataPoolManager.jointTexturePool.registerCustomTextureLayouts(config.customJointTextureLayouts);
+            return this._inited;
+        });
+    }
 
-        legacyCC.director.root.dataPoolManager.jointTexturePool.registerCustomTextureLayouts(config.customJointTextureLayouts);
+    public run (config: IGameConfig, onStart?: Game.OnStart): Promise<void>;
```
pandamicro: remove this API, we don't support this any longer

cocos/core/game.ts
```diff
@@ -569,44 +568,49 @@ export class Game extends EventTarget {
...
-        }
-
-        this._setAnimFrame();
-        this._runMainLoop();
 
-        // register system events
-        if (!EDITOR && game.config.registerSystemEvent) {
-            inputManager.registerSystemEvent(game.canvas);
+        let initPromise: Promise<boolean> | undefined;
+        if (typeof configOrCallback !== 'function' && configOrCallback) {
```
pandamicro, line 600: should be 

`if (typeof configOrCallback !== 'function' && onStart) {`

When the first parameter isn't a function and the second parameter exists

cocos/core/game.ts
pandamicro: The first parameter is required, not optional

cocos/core/game.ts
```diff
@@ -569,44 +568,49 @@ export class Game extends EventTarget {
...
+            initPromise = this.init(configOrCallback);
+            this.onStart = onStart ?? null;
+        } else {
+            this.onStart = configOrCallback ?? null;
         }
 
-        this._loadRenderPipeline(null);
+        return Promise.resolve(initPromise).then(() => {
```
pandamicro, line 607: can resolve undefined ?

pandamicro: should handle reject case

cocos/core/game.ts
pandamicro: Try a shorter name _setRenderPipelineNShowSplash

cocos/core/game.ts
shrinktofit: `Promise.resolve(undefined)` is OK. It yields an empty promise. Promise rejection is handled by the caller of `game.run()` or `game.init()`.

cocos/core/game.ts
shrinktofit: Resolved in https://github.com/cocos-creator/engine/pull/7612/commits/04a8e362f817b805fde203a67a70293315b9c7bd

cocos/core/game.ts
shrinktofit: Resolved in https://github.com/cocos-creator/engine/pull/7612/commits/04a8e362f817b805fde203a67a70293315b9c7bd

cocos/core/game.ts
shrinktofit: This branch should handle the `run(/* no-args*/)` so may it should by `typeof configOrCallback !== 'function' && configOrCallback && onStart`

cocos/core/game.ts
shrinktofit: The returned value will be exactly `theValue` in `game.run().then((theValue) => {})`

cocos/core/game.ts
shrinktofit: Resolved in https://github.com/cocos-creator/engine/pull/7612/commits/15094fec2c0315092c61a774e05fdf320b8faff3

cocos/core/game.ts
pandamicro: I think twice, it shouldn't create big issue

pandamicro (approved):

PR #7632, opened 2020-10-30
Make ENABLE_WEBGL_ANTIALIAS as true by default

pandamicro:


pandamicro:
https://github.com/cocos-creator/3d-tasks/issues/4743

PR #7664, opened 2020-11-01
Fix an error of tween action when the property is deprecated.

caryliu1999:


JayceLai (approved):

pandamicro:
@jareguo mentioned the fix isn't right, I will merge this one, but need an updated PR later

cocos/tween/tween-action.ts
```diff
@@ -104,7 +104,7 @@ export class TweenAction extends ActionInterval {
         this._props = Object.create(null);
         for (let name in props) {
             let value = props[name];
-
+            if (!value || typeof value === 'function') continue;
```
jareguo, line 107: 这里应该用 hasOwnProperty 进行判断。改成 if (!props.hasOwnProperty(name)) continue;

JayceLai:
@caryliu1999 I will fix that later, thanks

PR #7707, opened 2020-11-05
Fix the display error of particle system 2d.

caryliu1999:


cocos/particle-2d/particle-simulator-2d.ts
```diff
@@ -111,6 +111,7 @@ export class Simulator {
     private emitCounter = 0;
     private _worldRotation = 0;
     private declare sys;
+    public declare renderData;
```
pandamicro: public shouldn't be placed after private

pandamicro (changes requested):

cocos/particle-2d/particle-simulator-2d.ts
caryliu1999: > public shouldn't be placed after private

Fixed

pandamicro (approved):

PR #7725, opened 2020-11-06
Fix the error of getter override.

caryliu1999:
Re: https://github.com/cocos-creator/3d-tasks/issues/4826

### DIFF
--- a/cocos/particle/animator/gradient.ts
+++ b/cocos/particle/animator/gradient.ts
@@ -16,11 +16,20 @@ const Mode = Enum({
 
 @ccclass('cc.ColorKey')
 export class ColorKey {
-
+    /**
+     * @en Color value.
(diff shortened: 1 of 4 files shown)


cocos/particle/animator/gradient.ts
```diff
@@ -42,15 +59,27 @@ export class AlphaKey {
 export default class Gradient {
 
     public static Mode = Mode;
-
+    /**
+     * @en Array of color key.
+     * @zh 颜色关键帧列表。
+     * @property {[ColorKey]} colorKeys
```
pandamicro: No need to add property in ts api doc

caryliu1999: > No need to add property in ts api doc

Fixed

pandamicro (approved):

PR #7737, opened 2020-11-09
reduce native getPhase jsb calls

Mee-gu:
ref: [cocos2d-x-lite/pull/2995](https://github.com/cocos-creator/cocos2d-x-lite/pull/2995)

Changes:
 * reduce native getPhase jsb calls

### DIFF
--- a/cocos/core/pipeline/index.jsb.ts
+++ b/cocos/core/pipeline/index.jsb.ts
@@ -1,6 +1,9 @@
 
 declare const nr: any;
 
+import { getPhaseID } from './pass-phase'
+nr.getPhaseID = getPhaseID;
+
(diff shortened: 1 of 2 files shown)


cocos/core/pipeline/index.jsb.ts
```diff
@@ -1,6 +1,9 @@
 
 declare const nr: any;
 
+import{getPhaseID} from './pass-phase'
```
pandamicro: Space between import and {

pandamicro (changes requested):

cocos/core/pipeline/index.jsb.ts
Mee-gu: done

pandamicro (approved):

PR #7778, opened 2020-11-12
Suggested lint changes

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 *

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -64,9 +64,6 @@ rules:
     lines-between-class-members: off # be more lenient on member declarations
     max-classes-per-file: off # helper classes are common
     max-len: [warn, 150] # more lenient on max length per line
-    no-underscore-dangle: [warn, {
-      allowFunctionParams: true # designate the unused param
-    }] # allow underscores
     no-plusplus: off # allow increment/decrement operators
     no-continue: off # allow unlabeled continues
     no-mixed-operators: off # this is just cumbersome
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -65,7 +65,7 @@ rules:
     max-classes-per-file: off # helper classes are common
     max-len: [warn, 150] # more lenient on max length per line
     no-underscore-dangle: [warn, {
-      allowAfterThis: true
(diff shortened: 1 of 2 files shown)


.eslintrc.yaml
```diff
@@ -64,10 +64,12 @@ rules:
     lines-between-class-members: off # be more lenient on member declarations
     max-classes-per-file: off # helper classes are common
     max-len: [warn, 150] # more lenient on max length per line
-    new-cap: off # some class still doesn't follow this yet
-    no-underscore-dangle: off # allow underscores
+    new-cap: error
```
shrinktofit: If there are exceptions, just `/* eslint-disable-next-line new-cap */`

.eslintrc.yaml
```diff
@@ -64,10 +64,12 @@ rules:
...
+    no-underscore-dangle: [warn, {
```
shrinktofit, line 67: I don't know why we have underscored variable names except those for private members? The variables are indeedly invisible if they are not exported.

.eslintrc.yaml
YunHsiao: Every `off` rule specified here is to shut down the default enabled behavior. So if you need some rule to be kept around, just remove the line.

.eslintrc.yaml
YunHsiao: Some temp variables are named like `v3_1`, which I deemed reasonable. But it may be a good idea to keep a more consistent rule anyways, so I agree to re-enable this

.eslintrc.yaml
```diff
@@ -64,10 +64,12 @@ rules:
...
+      allowFunctionParams: true # designate the unused param
```
YunHsiao: But I don't know about this though. I think the default behavior(just remove the `off` rule) would suffice.

shrinktofit: For unused parameter, tsc will report error unless you prefixed it with an underscore.

.eslintrc.yaml
shrinktofit: Got it

YunHsiao (approved):

PR #7785, opened 2020-11-13
Revert "Add USE_LOCAL processing in ui component batch."

minggo:
as it slow down performance heavily, so revert it


minggo:
等 @caryliu1999 的修复，如果无法修复就合并该 PR。

PR #7815, opened 2020-11-18
fix iOS 14 performance issue

SantyWang:
Re: cocos-creator/2d-tasks#

Changes:
 *


LinYunMo:
RE: https://github.com/cocos-creator/3d-tasks/issues/5252

PR #7835, opened 2020-11-20
[3.0] bugfix: dragonbone cache mode: sync frame info

PatriceJiang:
Re: cocos-creator/2d-tasks#

Changes:
 *


PatriceJiang:
@minggo 麻烦合并下

PR #7838, opened 2020-11-20
[3.0] dragonbones  adapter _nativeAsset type: Uint8Array

PatriceJiang:
Re: cocos-creator/2d-tasks#

Changes:
 *


PatriceJiang:
@minggo 麻烦合并下

minggo (approved):

PR #7856, opened 2020-11-24
fix device resolution

minggo:
lite: https://github.com/cocos-creator/cocos2d-x-lite/pull/3066

### DIFF
--- a/cocos/core/platform/view.ts
+++ b/cocos/core/platform/view.ts
@@ -38,6 +38,7 @@ import visibleRect from './visible-rect';
 import { EDITOR, MINIGAME, JSB, RUNTIME_BASED } from 'internal:constants';
 import { legacyCC } from '../global-exports';
 import { logID, errorID } from './debug';
+import { sys } from './sys';
 
 class BrowserGetter {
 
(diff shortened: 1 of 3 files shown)


cocos/core/platform/view.ts
```diff
@@ -933,13 +934,13 @@ class ContainerStrategy {
         locContainer.style.width = locCanvas.style.width = w + 'px';
         locContainer.style.height = locCanvas.style.height = h + 'px';
         // Setup pixel ratio for retina display
-        let devicePixelRatio = _view._devicePixelRatio = 1;
+        _view._devicePixelRatio = 1;
         if (_view.isRetinaEnabled()) {
-            devicePixelRatio = _view._devicePixelRatio = Math.min(_view._maxPixelRatio, window.devicePixelRatio || 1);
+            _view._devicePixelRatio = Math.min(_view._maxPixelRatio, window.devicePixelRatio || 1);
```
pandamicro, line 939: This is intend to be freely controlled by user, actually user can set view.maxPixelRatio to a fixed scale, so that the view's pixel ratio will never exceed that.
The benefit is to limit the size of viewport, and pixel fill rate can be controlled.
e.g. we have a device with 3x device pixel ratio, and we can set view.maxPixelRatio to 2, then our viewport will be limited.

minggo: Good to know. So only use device resolution on native.

pandamicro (approved):

PR #7857, opened 2020-11-24
[3.0] [bugfix] TiledLayer should watch parent node position

PatriceJiang:
Re: cocos-creator/2d-tasks#

Changes:
 *


PatriceJiang:
@minggo  麻烦review 下

PR #7894, opened 2020-12-02
[3.0] revert spine version

PatriceJiang:
Re: https://github.com/cocos-creator/3d-tasks/issues/5527

Changes:
 * 回退 spine-runtime 版本


PatriceJiang:
@minggo 麻烦合并一下

PR #7917, opened 2020-12-04
Suppress no-initializer warns if the property is declared to have object type

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Properties specified with object type(i.e non-primitive/non-enum type) are usually have no initializer. This PR allows to write:
```ts
class C {
  @property
  p: Node;
}
```
without warnings.

Otherwise, users have to write:
```ts
class C {
  @property
  p: Node | null = null;
}
```

### DIFF
--- a/cocos/core/data/class.ts
+++ b/cocos/core/data/class.ts
@@ -433,11 +433,9 @@ export function CCClass<TFunction> (options: {
  * @return {Boolean}
  * @private
  */
-CCClass._isCCClass = function isCCClass (constructor): boolean {
-    // Does not support fastDefined class (ValueType).
-    // Use `instanceof ValueType` if necessary.
-    // eslint-disable-next-line no-prototype-builtins, @typescript-eslint/no-unsafe-return
-    return constructor?.hasOwnProperty?.('__ctors__');     // __ctors__ is not inherited
+CCClass._isCCClass = function (constructor) {
+    return constructor && constructor.hasOwnProperty &&
+        constructor.hasOwnProperty('__ctors__');     // is not inherited __ctors__
 };
 
 //
@@ -571,7 +569,7 @@ function parseAttributes (constructor: Function, attributes: IAcceptableAttribut
 
     if ('default' in attributes) {
         (attrs || initAttrs())[`${propertyNamePrefix}default`] = attributes.default;
-    } else if (DEV && warnOnNoDefault && !(attributes.get || attributes.set)) {
+    } else if (((EDITOR && !window.Build) || TEST) && warnOnNoDefault && !(attributes.get || attributes.set)) {
         warnID(3654, className, propertyName);
     }
 
--- a/cocos/core/data/class.ts
+++ b/cocos/core/data/class.ts
@@ -569,7 +569,7 @@ function parseAttributes (constructor: Function, attributes: IAcceptableAttribut
         }
     }
 
-    if (Object.prototype.hasOwnProperty.call(attributes, 'default')) {
+    if ('default' in attributes) {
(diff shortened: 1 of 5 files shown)


cocos/core/data/decorators/property.ts
```diff
@@ -176,35 +176,21 @@ function genProperty (
...
-        let defaultValue: any;
-        let isDefaultValueSpecified = false;
         if (descriptor) {
             // In case of Babel, if an initializer is given for class field.
             // That initializer is passed to `descriptor.initializer`.
             // babel
             if (descriptor.initializer) {
-                defaultValue = getDefaultFromInitializer(descriptor.initializer);
-                isDefaultValueSpecified = true;
+                propertyRecord.default = getDefaultFromInitializer(descriptor.initializer);
```
shrinktofit, line 184: "Whether a default value is specified" is decided by `hasOwnProperty()` instead of `undefined`.

cocos/core/data/class.ts
```diff
@@ -562,6 +567,12 @@ function parseAttributes (constructor: Function, attributes: IAcceptableAttribut
         }
     }
 
+    if (Object.prototype.hasOwnProperty.call(attributes, 'default')) {
```
jareguo: attributes.hasOwnProperty 就行了吧…… 不会有人传入一个 { foo: bar, hasOwnProperty: function () {} }
甚至直接 'default' in attributes 就好了

cocos/core/data/decorators/property.ts
```diff
@@ -176,35 +176,21 @@ function genProperty (
...
             if (actualDefaultValues.hasOwnProperty(propertyKey)) {
-                defaultValue = actualDefaultValues[propertyKey];
-                isDefaultValueSpecified = true;
+                propertyRecord.default = actualDefaultValues[propertyKey];
             }
         }
-
-        if ((EDITOR && !window.Build) || TEST) {
-            if (!fullOptions && options && options.hasOwnProperty('default')) {
-                warnID(3653, propertyKey, js.getClassName(ctor));
```
jareguo, line 201: 这个提示不要了吗？如果开发者写了 default 理论上应该去掉

cocos/core/data/class.ts
```diff
@@ -562,6 +567,12 @@ function parseAttributes (constructor: Function, attributes: IAcceptableAttribut
...
+        (attrs || initAttrs())[`${propertyNamePrefix}default`] = attributes.default;
+    } else if (DEV && warnOnNoDefault && !(attributes.get || attributes.set)) {
```
jareguo: 用 ((EDITOR && !window.Build) || TEST) 代替 DEV 吧，缩小警告范围，这个问题并不严重。

cocos/core/data/class.ts
shrinktofit: eslint warns on `.hasOwnProperty `

jareguo: How about just 'default' in attributes? I think it's safe enough, and faster.

shrinktofit: https://github.com/cocos-creator/engine/pull/7917/commits/dbdc083040d62045e413caeea1eecdcd05acf6cf

cocos/core/data/decorators/property.ts
shrinktofit: 这个警告还在哈，就是移到你下面评论的那块了

cocos/core/data/class.ts
shrinktofit: OK

PR #7934, opened 2020-12-08
Effect array property type fix. Aligned to Mat4 content.

ymiroshnyk:
Re: cocos-creator/2d-tasks#

Changes:
 * Fixed effect property type to satisfy engine-wide float array type (which is used in Mat4 as well).

### DIFF
--- a/cocos2d/core/assets/material/effect-base.ts
+++ b/cocos2d/core/assets/material/effect-base.ts
@@ -1,4 +1,5 @@
 import Pass from '../../../renderer/core/pass';
+import { FLOAT_ARRAY_TYPE } from '../../value-types/utils';
 
 const gfx = cc.gfx;
 
@@ -28,8 +29,8 @@ export default class EffectBase {
         let uniform = Object.create(null);
         uniform.name = name;
         uniform.type = prop.type;
-        if (prop.value instanceof Float32Array) {
-            uniform.value = new Float32Array(prop.value);
+        if (prop.value instanceof FLOAT_ARRAY_TYPE) {
+            uniform.value = new FLOAT_ARRAY_TYPE(prop.value);
         }
         else {
             uniform.value = prop.value;


SantyWang:
Thanks, I wonder what your purpose

ymiroshnyk:
> Thanks, I wonder what your purpose

The purpose is to have working MaterialVariant when you have Mat4 parameter overridden :) 

I made several materials which have Mat4 as additional parameter. Mat4 has Float64Array under it's hood (which is chosen by the engine). When I try to set overridden matrix in MaterialVariant, it becomes to set up in source Material as well. That's because of type inconsistency fixed in the pull request.

2youyou2 (approved):

SantyWang:
Thanks again

ymiroshnyk:
Thank you, @holycanvas. Will this fix be in 2.4.4 release?

Ronsku:
Why has this change not gone into 2.4.4?

Ronsku:
@holycanvas @pandamicro @2youyou2 
I see that this is merged, but is not released in 2.4.4 and I cannot see it in the Dev version of 2.4.5 either? Where is this merged to and why can't we get it in the releases?
We have to use custom engines just to support this.

Ronsku:
@holycanvas @pandamicro @2youyou2 @
Sorry for pinging you again, but two weeks has passed again. I just want to understand where this PR has been merged to and when it will appear in the CC 2 release?
I do understand that you focus mostly on CC3, which is probably a good thing but there are a few hickups in CC 2 like this one that would be good to fix, especially since most of the effort has already been done by @ymiroshnyk.
For @ymiroshnyk this is very critical thing for his extension to work properly with the stock engine. You can't expect extension users to always create custome engines just to have an extension working. It is also important to understand that this fix isn't because the extension is bad, but that there are issues with the CC 2 renderer itself.

This PR was merged on Dec 12, 2020 and there has been at least 2 versions released of CC 2 since then without this fix. When can we expect this hotfix to be released?
@ymiroshnyk is also currently working on another hickup in the CC2 renderer that will also be created as a PR for you, but there are no point creating fixes and PRs that will be merged and never end up final product.

Hope you understand my concern.

SantyWang:
After merging this pr, we found a problem in the editor, so we changed it to the following form and it has been released in v2.4.4, what's the problem now?

```
if (prop.value instanceof Float32Array) {
      uniform.value = new Float32Array(prop.value);
}
else if (prop.value instanceof Float64Array) {
      uniform.value = new Float64Array(prop.value);
}
else {
      uniform.value = prop.value;
}
```
ref: #8073

ymiroshnyk:
@holycanvas This is even better as it covers both types. Thank you.

SantyWang:
So what's the problem now? I noticed the problem you met seems not to be resolved

ymiroshnyk:
Yes, everything is just fine now with this pull request. Thank you.

However, there is [subsequent one](https://github.com/cocos-creator/engine/pull/8400), related to a type of a property value. Would you be so kind to check it as well? It's more crucial, as fixes broken rendering on some Chrome versions.

Ronsku:
Thanks for the quick response @holycanvas! Looks like the new PR solves also this PR, works well for me, i tested multiple scenarios.

PR #7937, opened 2020-12-09
fix bug : when destory node, 'this._siblingIndex' is not update

yeshao2069:
when destory node, 'this._siblingIndex' is not update

Re:  https://github.com/cocos-creator/3d-tasks/issues/5290


cocos/core/scene-graph/base-node.ts
```diff
@@ -1352,6 +1352,7 @@ export class BaseNode extends CCObject implements ISchedulable {
                 if (parent.emit) {
                     parent.emit(SystemEventType.CHILD_REMOVED, this);
                 }
+                parent._updateSiblingIndex();
```
SantyWang: 这句感觉放到 emit 之前比较好

SantyWang (approved):

PR #7948, opened 2020-12-11
Disable auto lint when save

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * We should do this manually

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -15,7 +15,4 @@
         "javascript",
         "typescript"
     ],
-    "editor.codeActionsOnSave": {
-        "source.fixAll": true,
-    },
 }


shrinktofit:
@pandamicro  Could this be safely merged in favor of https://github.com/cocos-creator/engine/pull/7964 ?

pandamicro:
Yes, I think so, our developers can enable auto save in their local settings of VSCode

SantyWang:
It seems that most people prefer auto-fix, Let's give it a little more time

YunHsiao:
The only way I see this could become reasonable is after we build an auto-fix robot that submits fix PRs automatically after every merge operation. Until then I suggest keeping this as-is.

PR #7968, opened 2020-12-15
Restructure Module

SantyWang:
Re: https://github.com/cocos-creator/3d-tasks/issues/5509

Changes:
 * Separate out more modules
image

* with all modules unselected: 

| Before | After | 
| :-- | :-- |
|1053K | ~~843K~~ 866k |

* with simple 2d renderer selected: 

| Before | After | 
| :-- | :-- |
|1289K | 1040k |

* hierarchy
image

related: https://github.com/cocos-creator/editor-3d/pull/4831


shrinktofit:
My questions:

- Are light, skinned mesh rendering worth isolating from 3d? As I known, these are frequently being used in 3d development.
- Animation should not be isolated, it should be in core. Even skeletal animation can be.

I suggest the following(expressed in module level):
## Core
- `cc/core` the core module, shared by all aspects of engine. Include: math, component-related, serialization, renderer, animation, even.

## 3D features
- `cc/3d` include  3D-specific features which are commonly used in 3D development: 3D renderer, mesh, lighting, skinned mesh...
- `cc/3d/primitives`
- `cc/3d/particle`
- `cc/3d/physics`

## 2D features
- `cc/2d` include 2D-specific features which are commonly used in 2D development.
- `cc/2d/particle`
- `cc/2d/physics`
- `cc/2d/intersection`

## Other features
- `cc/audio`
- `cc/video`
- .....

As for the UI, I must admit it's a common features can used in either 2D/3D development. We can, as we have been done, isolates the `cc/ui`.

SantyWang:
> Animation should not be isolated, it should be in core. Even skeletal animation can be.

I think `animation` should not be necessary, if only develop 2d games, spine and dragonbones animation system is more often used by users

> Are light, skinned mesh rendering worth isolating from 3d?

It's a matter of granularity here, if you want to minimize the package, it's best to split it, light is not needed for all 3d games in my opinion (especially for some unlit based games). skinned mesh renderer is not needed either

> cc/3d/primitives

agreed and wished, but do we currently support a sub-module under modules?

shrinktofit:
>  if you want to minimize the package, it's best to split it, light is not needed for all 3d games in my opinion

But.. the scene created by Editor contains defaultly a directional light, right?

shrinktofit:
> but do we currently support a sub-module under modules

Really easy. 👍

SantyWang:
> But.. the scene created by Editor contains defaultly a directional light, right?

Yeah, it may be possible to let developers choose a different project template(for example. 2d and 3d) when creating a new project in the future, and for some template, it may not create a new directional light by default

SantyWang:
image

PR #7980, opened 2020-12-17
improve video player code

knoxHuang:
Re: cocos-creator/2d-tasks#

Changes:
 * improve video player code

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -2,10 +2,10 @@ root: true
 
 parser: '@typescript-eslint/parser'
 
-parserOptions: 
+parserOptions:
     project: ./tsconfig.json
 
-extends: 
+extends:
     - eslint:recommended
     - airbnb-base
     - plugin:@typescript-eslint/recommended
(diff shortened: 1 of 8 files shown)


.eslintrc.yaml
```diff
@@ -51,11 +51,12 @@ rules:
     prefer-rest-params: off # we need ES5 to be fast too
     prefer-spread: off # we need ES5 to be fast too
     space-before-function-paren: [warn, always] # we require this explicitly
+    radix: off # we sometimes do not need to pass a second parameter
```
shrinktofit, line 54: Otherwise, we need to `parseInt("10", 10)`, it's verbose to be honest.

cocos/video/video-player.ts
```diff
@@ -247,7 +248,9 @@ export class VideoPlayer extends Component {
     @tooltip('i18n:videoplayer.fullScreenOnAwake')
     get fullScreenOnAwake () {
         if (!EDITOR) {
-            this._fullScreenOnAwake = this._impl && this._impl.fullScreenOnAwake;
```
shrinktofit, line 250: WARN: Semantic is changed.

PR #8023, opened 2020-12-25
fix mask state change bug

xuewuli:
Changes:
 * fix StencilManager state change bug 

when nesting cc.mask，stencil.ref value not update correctly


SantyWang:
We have refactored the stencil manager, and have fixed this issue, THANKS anyway.

PR #8027, opened 2020-12-26
Optimize WASM binary path emit

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Refiles https://github.com/cocos-creator/engine/pull/7979


shrinktofit:
@holycanvas  Lint CI should be skipped since I modified `@cocos/build-engine`.

JayceLai (approved):

PR #8098, opened 2021-01-07
support kinematic contact

JayceLai:
Re: cocos-creator/3d-tasks#

Changes:
 * support kinematic contact


JayceLai:
https://github.com/cocos-creator/3d-tasks/issues/5855

PR #8102, opened 2021-01-08
Fix the problem of ui model render to render texture.

caryliu1999:
https://github.com/cocos-creator/3d-tasks/issues/5872


LinYunMo:
https://github.com/cocos-creator/3d-tasks/issues/5934 需要用此方案解决此问题

PR #8105, opened 2021-01-08
Remove current() and ratio() setters

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * In https://github.com/cocos-creator/engine/pull/8037 we added `.current` and `.ratio` accessors to provide convenient accesses to "current time progress"(which lays in `[0, duration]`). Now we decide to remove the setters and keep only the getters since the need of reverse calculation of "accumulated time" from 'current time'. https://github.com/cocos-creator/engine/pull/8037 is not a released change so we do not need migration here.

### DIFF
--- a/cocos/core/animation/animation-state.ts
+++ b/cocos/core/animation/animation-state.ts
@@ -291,27 +291,11 @@ export class AnimationState extends Playable {
     }
 
     /**
-     * @en Sets the time progress, in seconds.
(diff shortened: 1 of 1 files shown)


2youyou2 (approved):

gameall3d (approved):
LGTM

PR #8114, opened 2021-01-09
binarySearch series functions: document and unit test

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Note this did not change the original semantic defined in 2D.


cocos/core/algorithm/binary-search.ts
```diff
@@ -0,0 +1,89 @@
...
+ * Searches the sorted number array for an element and returns the index of that element.
+ * @param array The array to sort.
+ * @param value The value to search.
+ * @param EPSILON The epsilon to compare the numbers. Default to `1e-6`.
+ * @return The index of the searched element in the sorted array, if one is found;
+ * otherwise, a negative number that is the bitwise complement of the index of the next element that is large than the searched value or,
+ * if there is no larger element(include the case that the array is empty), the bitwise complement of array's length.
+ */
+export function binarySearchEpsilon (array: number[], value: number, EPSILON = 1e-6) {
+    let low = 0;
```
shrinktofit, line 49: Code repetition with `binarySearchBy()` is on purpose here out of the performance reason. Forward to invoke `binarySearchBy((lhs, rhs) => lhs < rhs)` if one could prove that the performance issue won't  matter.

shrinktofit:
The two lint errors are caused by mis-configuration of eslint, please skip them for now. @holycanvas

jareguo (approved):

PR #8341, opened 2021-03-09
avoid native crash when animation include an error frame.

xianyinchen:
when texture = 0 and call material.setProperty('texture', texture), native will crash.


xianyinchen:
https://github.com/cocos-creator/2d-tasks/issues/3288

SantyWang (approved):

PR #8352, opened 2021-03-12
change vertex format for mesh particle.

xianyinchen:
test demo.
[NewProject_4.zip](https://github.com/cocos-creator/engine/files/6146395/NewProject_4.zip)

the right result.
image


pandamicro:
The correct way to do this is to provide either of the following
1. Provide `Mesh.copyAttributes(vfmt, buffer)`
2. Provide `elements` in VertexFormat

Then we should remove any assumption in particle's `_createParticleData`.

SantyWang:
v3.x will optimize the interface of mesh, then particle system can get stride of color chanel and restructure vertex buffer

PR #8363, opened 2021-03-16
Repair of the same frame to play audio, registered callback end, pause and resume play, leading to the end of the pullback in advance trigger.

OnlyKoei:
Repair of the same frame to play audio, registered callback end, pause and resume play, leading to the end of the pullback in advance trigger.

Re: https://github.com/cocos-creator/engine/issues/7701

### DIFF
--- a/cocos2d/audio/CCAudio.js
+++ b/cocos2d/audio/CCAudio.js
@@ -66,7 +66,7 @@ Audio.State = {
     /**
      * @property {Number} ERROR
      */
-    ERROR : -1,
+    ERROR: -1,
     /**
      * @property {Number} INITIALZING
      */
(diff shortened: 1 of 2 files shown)


cocos2d/audio/CCAudio.js
```diff
@@ -456,10 +456,13 @@ let WebAudioElement = function (buffer, audio) {
         // If more than the duration of the audio, Need to take the remainder
         this.playedLength %= this._buffer.duration;
         let audio = this._currentSource;
+        if (audio) {
+            audio.onended = null;
```
PPpro: 这里 可以直接调用 this._unbindEnded() ， 里边还有 binded 状态位的 更新，相对 更严谨一些

PPpro (approved):

PR #8368, opened 2021-03-17
Mark an attribute to be shallow readonly

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Sometimes we expect the array's elements to be editable but its length to be immutable.

### DIFF
--- a/cocos/core/data/utils/attribute-defines.ts
+++ b/cocos/core/data/utils/attribute-defines.ts
@@ -56,8 +56,18 @@ export interface IExposedAttributes {
 
     /**
      * 指定该属性是否为可读的。
+     * 将 `readonly` 指定为 `true` 或选项对象时都将标记为该属性是可读的；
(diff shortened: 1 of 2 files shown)


cocos/core/data/utils/attribute-defines.ts
```diff
@@ -59,6 +59,12 @@ export interface IExposedAttributes {
      */
     readonly?: boolean;
 
+    /**
+     * 如果该属性是数组，则指定该数组的长度是否是固定的。
+     * 和 `readonly` 不同，`readonly` 为 `true` 时不仅不可以改变数组长度，也将不可改变
```
jareguo: 也将不可改变（什么？）

shrinktofit: Outdated.

jareguo (approved):

PR #8398, opened 2021-03-25
feat: tween props support function

lllyin:
Re: cocos-creator/3d-tasks#

Changelog:
 * tween props support function


cocos/tween/tween-action.ts
```diff
@@ -137,7 +137,10 @@ export class TweenAction extends ActionInterval {
             // eslint-disable-next-line no-prototype-builtins
             if (!props.hasOwnProperty(name)) continue;
             let value = props[name];
-            if (value == null || typeof value === 'string' || typeof value === 'function') continue;
+            if (typeof value === 'function') {
+                value = value();
+            }
+            if (value == null || typeof value === 'string') continue;
```
pandamicro, line 143: This is a type pre-check, should be placed before all actual assigns.

cocos/tween/tween-action.ts
```diff
@@ -137,7 +137,10 @@ export class TweenAction extends ActionInterval {
             // eslint-disable-next-line no-prototype-builtins
```
JayceLai, line 137: Please also modify the comment above

JayceLai (changes requested):

pandamicro (approved):

PR #8400, opened 2021-03-25
mat4 uniform setting up fix

ymiroshnyk:
This pull request fixes type of mat4 uniform property to Float32Array. It ensures that WebGLRenderingContext.uniformMatrix4fv() always gets Float32Array as a value parameter as required by specification: https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/uniformMatrix . 

Particularly it fixes setting up mat4 parameters on Chrome 66.0.3359.181.

### DIFF
--- a/cocos2d/renderer/types.js
+++ b/cocos2d/renderer/types.js
@@ -69,7 +69,7 @@ export let enums2default = {
     [enums.PARAM_FLOAT2]: new Float32Array([0, 0]),
     [enums.PARAM_FLOAT3]: new Float32Array([0, 0, 0]),
     [enums.PARAM_FLOAT4]: new Float32Array([0, 0, 0, 0]),
-    [enums.PARAM_MAT4]: cc.mat4().m,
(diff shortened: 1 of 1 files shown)


SantyWang:
Thanks, In which case you met the problem? Did `Float64Array` cause some errors on Chrome 66.0.3359.181.

ymiroshnyk:
@holycanvas Yes, on that specific version it produced following error:
`Uncaught TypeError: Failed to execute 'uniformMatrix4fv' on 'WebGLRenderingContext': No function was found that matched the signature provided.`

I faced the problem when was trying to set mat4 shader uniform through the Cocos material property.

SantyWang:
> I faced the problem when was trying to set mat4 shader uniform through the Cocos material property.

I see, but I don't understand why this PR can fix it up with changing the default value of Mat4-Typed uniform to Float32Array. I think it's just default value, and after setting a Mat4 uniform to material property, it will turn to Float64Array again

PR #8402, opened 2021-03-27
Optimize the document on pal/audio

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Optimizes the document on pal/audio.

Note the following:
- The document should be elaborate if possible. You'd better to illustrate the details and notes and side effects of API.

- Some return types are missed.

- Keep lines between declarations.

- Textual descriptions should be formal(caption, punctuators, syntax).

- If symbol names(class names, function names, var names) are referenced, one shall enclose them like `this`.

- `get duration()` did not specify the time unit. In seconds or milliseconds?

- The `volumn` parameter is not elaborate. What if it's not specified?

### DIFF
--- a/@types/pal/audio.d.ts
+++ b/@types/pal/audio.d.ts
@@ -44,7 +44,7 @@ declare module 'pal/audio' {
          * Asynchronously load a native audio for playing one shot.
          * @param url URL to the audio.
          * @param opts Load options.
-         * @returns The one shot player.
+         * @returns The native audio such as HTMLAudioElement or AudioBuffer.
          */
         static loadNative (url: string, opts?: import('pal/audio/type').AudioLoadOptions): Promise<unknown>;
 
--- a/@types/pal/audio.d.ts
+++ b/@types/pal/audio.d.ts
@@ -44,7 +44,7 @@ declare module 'pal/audio' {
          * Asynchronously load a native audio for playing one shot.
          * @param url URL to the audio.
          * @param opts Load options.
-         * @returns The native audio such as HTMLAudioElement or AudioBuffer.
+         * @returns The native audio such as `HTMLAudioElement` or `AudioBuffer`.
          */
         static loadNative (url: string, opts?: import('pal/audio/type').AudioLoadOptions): Promise<unknown>;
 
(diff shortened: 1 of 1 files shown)


@types/pal/audio.d.ts
```diff
@@ -1,126 +1,157 @@
...
         static load (url: string, opts?: import('pal/audio/type').AudioLoadOptions): Promise<AudioPlayer>;
+
         /**
-         * load native audio for playing one shot
-         * @param url
-         * @param opts
+         * Asynchronously load a native audio for playing one shot.
+         * @param url URL to the audio.
+         * @param opts Load options.
+         * @returns The one shot player.
```
PPpro:
```suggestion
         * @returns The native audio such as HTMLAudioElement or AudioBuffer.
```

PPpro (changes requested):

@types/pal/audio.d.ts
```diff
@@ -1,130 +1,162 @@
...
         static load (url: string, opts?: import('pal/audio/type').AudioLoadOptions): Promise<AudioPlayer>;
+
         /**
-         * load native audio for playing one shot
-         * @param url
-         * @param opts
+         * Asynchronously load a native audio for playing one shot.
+         * @param url URL to the audio.
+         * @param opts Load options.
+         * @returns The native audio such as HTMLAudioElement or AudioBuffer.
```
PPpro:
```suggestion
         * @returns The native audio such as `HTMLAudioElement` or `AudioBuffer`.
```

PPpro (changes requested):

PPpro (approved):

@types/pal/audio.d.ts
```diff
@@ -1,130 +1,162 @@
...
          */
         static loadNative (url: string, opts?: import('pal/audio/type').AudioLoadOptions): Promise<unknown>;
+
         /**
-         * max audio channel, if the amount of playing audios exceeds maxAudioChannel, some audio instances should be discarded by audio manager
+         * Max audio channel count allowed on current platform.
+         * if the amount of playing audios exceeds the limit,
```
PPpro:
```suggestion
         * If the amount of playing audios exceeds the limit,
```

PR #8418, opened 2021-04-01
Add custom material processing and pipeline type for pipeline.

caryliu1999:
https://github.com/cocos-creator/engine-native/pull/3716


cocos/core/pipeline/pipeline-scene-data.ts
```diff
@@ -61,10 +62,30 @@ export class PipelineSceneData {
...
+        if (this._deferredLightingMat === mat) return;
+        this._deferredLightingMat = (mat && mat.effectAsset) ? mat : builtinResMgr.get<Material>('builtin-deferred-material');
+        this.updateDeferredPassInfo();
+    }
+
+    public get deferredPostMat () {
+        return this._deferredPostMat;
+    }
+
+    public set deferredPostMat (mat: Material) {
```
pandamicro: The above api naming don't follow our rules, Mat should be Material instead

cocos/core/pipeline/define.ts
```diff
@@ -63,6 +63,15 @@ export enum RenderPriority {
     DEFAULT = 0x80,
 }
 
+/**
+ * @en The pipeline render type.
+ * @zh 渲染管线类型。
+ */
+export enum PipelineType {
```
pandamicro: The pipeline is designed for customization, this enum just make it static. I think the pipeline scene data is a future obstacle for customized pipeline, the different operations should be placed into the pipeline logic. If it's hard to modify currently, I suggest to at least make some APIs private:

- This enum
- pipelineType in Game
- deferred related properties in pipeline-scene-data

I think we can extend PipelineSceneData in each pipeline

cocos/core/pipeline/deferred/postprocess-stage.ts
```diff
@@ -167,25 +158,18 @@ export class PostprocessStage extends RenderStage {
...
-        if (builtinPostProcess) {
-            pass = builtinPostProcess.passes[0];
-            shader = ShaderPool.get(pass.getShaderVariant());
-        } else {
-            pass = this._postprocessMaterial!.passes[POSTPROCESSPASS_INDEX];
-            shader = ShaderPool.get(this._postprocessMaterial!.passes[POSTPROCESSPASS_INDEX].getShaderVariant());
-        }
+        const builtinPostProcess = sceneData.deferredPostMat;
+        const pass = builtinPostProcess.passes[0];
+        const shader = ShaderPool.get(pass.getShaderVariant());
```
2youyou2: Postprocess stage should bind material descriptorSet too.

```js
cmdBuff.bindDescriptorSet(SetIndex.MATERIAL, pass.descriptorSet);
```

caryliu1999: Already added it

cocos/core/pipeline/pipeline-scene-data.ts
caryliu1999: > The above api naming don't follow our rules, Mat should be Material instead

Already rename all the Mat.

cocos/core/pipeline/define.ts
caryliu1999: > The pipeline is designed for customization, this enum just make it static. I think the pipeline scene data is a future obstacle for customized pipeline, the different operations should be placed into the pipeline logic. If it's hard to modify currently, I suggest to at least make some APIs private:
> 
> * This enum
> * pipelineType in Game
> * deferred related properties in pipeline-scene-data
> 
> I think we can extend PipelineSceneData in each pipeline

Already delete enum PipelineType and add deferred-pipeline-scene-data.

pandamicro:
@caryliu1999 please delay this PR to v3.3

pandamicro (approved):

minggo (approved):

PR #8424, opened 2021-04-04
Fix cc.ImageAsset memory leaks (new legacyCC.Texture2D())

wyb10a10:
Re: cocos-creator/3d-tasks#

unload spine asset will be leaked,  the `GFX Texture Mem(M)` on the web mobile platform will be increased by loaded and released repeatedly! because this code will new a Texture, but not released on destroy().

```ts
    get _texture () {
        if (!this._tex) {
            const tex = new legacyCC.Texture2D();
            tex.name = this.nativeUrl;
            tex.image = this;
            this._tex = tex;
        }
        return this._tex;
    }
```

Changelog:
 * Fix cc.ImageAsset memory leaks

### DIFF
--- a/cocos/core/assets/image-asset.ts
+++ b/cocos/core/assets/image-asset.ts
@@ -233,7 +233,7 @@ export class ImageAsset extends Asset {
     }
 
     public destroy () {
-        if (this._tex && this._tex.image == this && this._tex.destroy) {
+        if (this._tex && this._tex.image === this) {
             this._tex.destroy()
         } 
         if (this.data && this.data instanceof HTMLImageElement) {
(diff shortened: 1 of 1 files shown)


pandamicro:
Hi, @wyb10a10 we are aware of the issue because of your PR, thank you very much !
However we have fixed the issue in v3.3 with another approach, we removed `ImageAsset._texture` which is much more proper. Please take a look at

https://github.com/cocos-creator/engine/pull/8707
https://github.com/cocos-creator/engine/pull/8742

So I'm sorry I have to close your PR, but I want you to know your contribution notified us about the problem and is very much appreciated.

PR #8434, opened 2021-04-09
native engine msaa

hana-alice:
Re: cocos-creator/3d-tasks#

Changelog:
 * 
jsb update, native-engine at least update to :https://github.com/cocos-creator/engine-native/pull/3383


hana-alice:
@minggo coresponding commit https://github.com/cocos-creator/engine-native/pull/3383. these two are paired.

PR #8462, opened 2021-04-16
Optimize the performance of the memory pool.

pandamicro:


pandamicro:
#8456 
cocos-creator/engine-native#3414

pandamicro:
https://github.com/cocos-creator/3d-tasks/issues/6760

PR #8486, opened 2021-04-19
temporarily bypass splash screen

hana-alice:
Re: cocos-creator/3d-tasks#

Changelog:
 * 
temp solution, issue here : https://github.com/cocos-creator/3d-tasks/issues/6874


pandamicro:
What would happen otherwise. Honestly it's not an acceptable solution, at least we can only skip it in native environment

hana-alice:
> What would happen otherwise. Honestly it's not an acceptable solution, at least we can only skip it in native environment

stuck white screen when launched. cannot do that because this happens before getting into native engine, parse png failed in jsb

pandamicro (approved):

PR #8500, opened 2021-04-20
fix auto Atlas PersistRootNode bug

LinYunMo:
Re: https://github.com/cocos-creator/3d-tasks/issues/6822


pandamicro (approved):

cocos/2d/assets/sprite-frame.ts
```diff
@@ -1105,6 +1105,7 @@ export class SpriteFrame extends Asset {
         this._rect.x = frame.x;
         this._rect.y = frame.y;
         this._calculateUV();
+        this.emit('updateUV');
```
pandamicro: change event name to uv-updated

pandamicro (changes requested):

pandamicro (approved):

PR #8509, opened 2021-04-21
fixed material set texture property

2youyou2:
Re: cocos-creator/3d-tasks#

Changelog:
 * fixed material set texture property


pandamicro (approved):

pandamicro:
https://github.com/cocos-creator/3d-tasks/issues/6870

PR #8541, opened 2021-04-25
fix terrain component sample weight error, add get max weight layer f…

MSoft1115:
…unction

Re: cocos-creator/3d-tasks#

Changelog:
 *

### DIFF
--- a/cocos/terrain/terrain.ts
+++ b/cocos/terrain/terrain.ts
@@ -231,17 +231,23 @@ class TerrainRenderable extends RenderableComponent {
                 defines: block._getMaterialDefines(nLayers),
             });
 
-            if (this._brushMaterial !== null) {
+            if (this._brushMaterial !== null && this._brushMaterial.passes !== null && this._brushMaterial.passes.length > 0) {
                 const passes = this._currentMaterial.passes;
-
                 passes.push(this._brushMaterial.passes[0]);
+
+                if (init) {
+                    this._model.initSubModel(0, this._meshData, this._currentMaterial);
+                }
+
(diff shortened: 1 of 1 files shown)


cocos/terrain/terrain.ts
```diff
@@ -1535,6 +1547,47 @@ export class Terrain extends Component {
         return n;
     }
 
+    /**
+     * @en get max weight layer by point
+     * @zh 根据点的坐标获得权重最大的纹理层
+     */
+    public GetMaxWeightLayerAt (x: number, y: number) {
```
pandamicro: Rename to `getMaxWeightLayerAt` instead

cocos/terrain/terrain.ts
```diff
@@ -231,17 +231,23 @@ class TerrainRenderable extends RenderableComponent {
...
+                    this._model.initSubModel(0, this._meshData, this._currentMaterial);
+                }
+
+                this.setMaterial(this._currentMaterial, 0);
+            } else {
+                if (init) {
+                    this._model.initSubModel(0, this._meshData, this._currentMaterial);
+                }
+
+                this.setMaterial(this._currentMaterial, 0);
```
pandamicro: seems exactly identical in if and else, I think initSubModel and setMaterial can remain the same as before, unless there is some modification missing

pandamicro (changes requested):

cocos/terrain/terrain.ts
MSoft1115: 这个是为了方便调试改的

pandamicro (approved):

PR #8553, opened 2021-04-25
fix some deprecated value warn

LinYunMo:
Re: https://github.com/cocos-creator/3d-tasks/issues/6950


SantyWang (approved):

LinYunMo:
RE: https://github.com/cocos-creator/3d-tasks/issues/7034

PR #8555, opened 2021-04-26
[Fix] Deferred rt adaptor

GengineJS:


GengineJS:
https://github.com/cocos-creator/engine-native/pull/3474

PR #8601, opened 2021-04-29
delete useless code

SantyWang:
Re: cocos-creator/3d-tasks#

Changelog:
 *

### DIFF
--- a/cocos/2d/assets/sprite-frame.ts
+++ b/cocos/2d/assets/sprite-frame.ts
@@ -549,9 +549,11 @@ export class SpriteFrame extends Asset {
      * Returns whether the texture have been loaded.
      * @zh
      * 返回是否已加载精灵帧。
+     *
+     * @deprecated since v3.3
      */
     public textureLoaded () {
-        return this.texture && this.texture.loaded;
+        return !!this.texture;
     }
 
     /**
@@ -701,7 +703,6 @@ export class SpriteFrame extends Asset {
 
         if (info) {
             if (info.texture) {
-                this.loaded = false;
                 this._rect.x = this._rect.y = 0;
                 this._rect.width = info.texture.width;
                 this._rect.height = info.texture.height;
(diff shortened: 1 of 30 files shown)


cocos/2d/components/sprite.ts
```diff
@@ -589,23 +581,12 @@ export class Sprite extends Renderable2D {
     }
 
     protected _canRender () {
-        // if (cc.game.renderType === cc.game.RENDER_TYPE_CANVAS) {
```
SantyWang, line 592: 大段大段的注释，这种习惯真不好

cocos/2d/assets/sprite-frame.ts
```diff
@@ -549,9 +549,11 @@ export class SpriteFrame extends Asset {
      * Returns whether the texture have been loaded.
      * @zh
      * 返回是否已加载精灵帧。
+     *
+     * @deprecated since v3.3
```
pandamicro, line 553: Why deprecate this API, even if sprite frame have default texture, user may still need to know when the actual asset is loaded.

pandamicro (approved):

cocos/2d/assets/sprite-frame.ts
SantyWang: I think they can use ```spriteframe.texture``` to check it instead

cocos/core/assets/texture-2d.ts
```diff
@@ -270,21 +270,6 @@ export class Texture2D extends SimpleTexture {
         return Object.assign(texInfo, presumed);
     }
 
-    protected _checkTextureLoaded () {
```
pandamicro, line 273: If we ignore the ImageAsset upload and make texture directly usable, then its size will be 0 before ImageAsset downloaded. That could cause potential issue.

A way to resolve this is to serialize and deserialize size for ImageAsset and Texture2D, which we haven't done yet

pandamicro: The reason currently you haven't seen any issue, is because we have serialized size in SpriteFrame.
But it's a structural issue in the Asset design, in native asset manager for future version, we will have multithread asset loading process, that means asynchronous loading.

PR #8606, opened 2021-04-30
Fix animation BUG

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * This should be a BUG introduced from #8540


SantyWang (approved):

gameall3d (approved):
LGTM

PR #8624, opened 2021-04-30
Clarify the handling of non-CCClass types by dynamic deserialization

jareguo:
Uniform with serialization rules. Not support serializing named class unless they're declared by `@ccclass` or `fastDefine`.

In https://github.com/cocos-creator/engine/pull/8600, the editor normally serialized a ccclass with js engine. But the simulator dynamically deserialized it with jsb wrapper, which resulting in inconsistent types.

This PR is used to prevent this from happening. It disables deserialization of all native objects that inject serialization information directly using `setClassName`. **Please adjust the corresponding jsb wrapper after merging.**

### DIFF
--- a/cocos/core/data/deserialize-dynamic.ts
+++ b/cocos/core/data/deserialize-dynamic.ts
@@ -28,9 +28,8 @@
  * @hidden
  */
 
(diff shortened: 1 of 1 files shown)


pandamicro:
Please fix the linter issue

shrinktofit (approved):

PR #8636, opened 2021-05-06
fix stopping audio won't seek to 0  when audio is paused on Xiaomi

PPpro:
fix https://github.com/cocos-creator/3d-tasks/issues/7222

Changelog:
 * 修复小米平台音频 pause 之后，无法 stop 的问题


PPpro:
等待合并到 3.1.x

SantyWang (approved):

PR #8692, opened 2021-05-18
update sphere light effect

troublemaker52025:
Re: cocos-creator/3d-tasks#

Changelog:
 *


troublemaker52025:
https://github.com/cocos-creator/3d-tasks/issues/4020

PR #8694, opened 2021-05-20
Fix repeatForever description in tween.js

KimJuHeok:
Change tween.js's repeatForever notation same as cc.repeat does


pandamicro (approved):

pandamicro:
@holycanvas  Please confirm whether we can merge it

PR #8699, opened 2021-05-21
move build:effect

VisualSJ:
Re: cocos-creator/3d-tasks#

Cannot run yet, need to provide cc/location

Changelog:
 *


SantyWang:
build-effect will be remove in 3.6

PR #8703, opened 2021-05-22
change AnimationClip IEvent interface params support (number string boolean)

kittizz:
Re: cocos-creator/3d-tasks#

Changelog:
 change the AnimationClip IEvent interface params string[] to any[] for support (number string boolean)

```
defaultClip.events.push({
                frame: 0.5, // trigger event on the 0.5 second
                func: "onflowAnimToYEnd", // name of event to be called
                params: [0], // parameters passed to `func`
})
(params: [0] error): Type 'number' is not assignable to type 'string'.ts(2322)
```

### DIFF
--- a/cocos/core/animation/animation-clip.ts
+++ b/cocos/core/animation/animation-clip.ts
@@ -96,7 +96,7 @@ export declare namespace AnimationClip {
     export interface IEvent {
         frame: number;
         func: string;
-        params: string[];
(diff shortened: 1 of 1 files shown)


pandamicro (approved):

pandamicro:
Thanks for your contribution, this will be released with v3.1.2

PR #8718, opened 2021-05-26
Add for consistency between Webgl and WebGl2

Jacksparrow2525:
Small inconsistency in the GFX layer


minggo:
Oh, v3.1-release is a released version. You should send to v3.3.

Jacksparrow2525:
> Oh, v3.1-release is a released version. You should send to v3.3.

OK!

PR #8752, opened 2021-06-02
Forbid dynamic atlas for texture in sub context view

finscn:
Re: cocos-creator/3d-tasks#

Changelog:
 *


cocos2d/core/components/SubContextView.js
```diff
@@ -80,6 +80,7 @@ let SubContextView = cc.Class({
     ctor () {
         this._sprite = null;
         this._tex = new cc.Texture2D();
+        this._tex._packable = false;
```
SantyWang, line 83: 为啥要赋值私有变量

jareguo: image

PPpro (approved):

PR #8761, opened 2021-06-03
fix morph instance reuse

YunHsiao:
`Models` are reused after destroy and `MorphModel` has a fragile dependency on the existence of the member `_morphRenderingInstance` at initialization. This will incorrectly enable morph for models that instead have none after some rounds of destroy & init loop.


troublemaker52025:
https://github.com/cocos-creator/3d-tasks/issues/7721

shrinktofit (approved):

shrinktofit:
Great catch.

PR #8778, opened 2021-06-07
Add editor extendable

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Created a tag name, exposed as `editorExtrasTag`, to store editor specific property. Editor may use this tag to visit those properties. Classes declared with such property are expressed via interface `EditorExtendableObject`.

 * **For convinient**, **internal** class `EditorExtendableMixin` and **internal** function `EditorExtendable` are created to simplify the creation of `EditorExtendableObject`. I have documented that these utility function should not been exposed to users currently.

### DIFF
--- a/cocos/core/data/editor-extendable.ts
+++ b/cocos/core/data/editor-extendable.ts
@@ -1,0 +1,60 @@
+import { EDITOR } from 'internal:constants';
+import { js } from '../utils/js';
+import { CCClass } from './class';
+import { EditorExtendableObject, editorExtrasTag } from './editor-extras-tag';
+
+// Functions and classes exposed from this module are useful to
+// make a class to be `EditorExtendableObject`.
+//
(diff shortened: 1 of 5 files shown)


gameall3d (approved):
LGTM

cocos/core/data/editor-extendable.ts
```diff
@@ -0,0 +1,60 @@
+import { EDITOR } from 'internal:constants';
+import { js } from '../utils/js';
+import { CCClass } from './class';
+import { EditorExtendableObject, editorExtrasTag } from './editor-extras-tag';
+
+// Functions and classes exposed from this module are useful to
+// make a class to be `EditorExtendableObject`.
```
shrinktofit, line 7: Mark my words.

jareguo (approved):

PR #8784, opened 2021-06-08
V3.1.2 zxx43 particle

zxx43:
Re: cocos-creator/3d-tasks#

Changelog:
 *


cocos/particle/renderer/particle-system-renderer-data.ts
```diff
@@ -203,11 +203,24 @@ export default class ParticleSystemRenderer {
 
     private _particleSystem: any = null!; // ParticleSystem
 
+    create (ps: any) {
+        if (this._particleSystem === null) {
+            this._particleSystem = ps;
+            const useGPU = this._useGPU && isSupportGPUParticle();
+            this._particleSystem.processor = useGPU ? new ParticleSystemRendererGPU(this) : new ParticleSystemRendererCPU(this);
+            this._particleSystem.processor.onInit(this._particleSystem);
```
pandamicro: If particle is not enabled or node not activated in editor, this process will create more useless memory.
We need to control things more strictly, I think it's enough to assign `this._particleSystem = ps;`.

cocos/particle/renderer/particle-system-renderer-data.ts
```diff
@@ -203,11 +203,24 @@ export default class ParticleSystemRenderer {
 
     private _particleSystem: any = null!; // ParticleSystem
 
+    create (ps: any) {
+        if (this._particleSystem === null) {
```
pandamicro, line 210: Need to emit error when create is invoked and `_particleSystem` is not null, notify user when they are not invoking APIs correctly, not allowing create to be invoked twice with different particle system.

cocos/particle/renderer/particle-system-renderer-data.ts
pandamicro: create (ps: any) {
        if (this._particleSystem === null) {
            this._particleSystem = ps;
        }
        else if (this._particleSystem !== ps) {
            errorID(...); // Message: ParticleSystem cannot be created with different particle system
        }
    }

cocos/particle/renderer/particle-system-renderer-data.ts
```diff
@@ -203,11 +203,24 @@ export default class ParticleSystemRenderer {
...
-        this._particleSystem.processor = useGPU ? new ParticleSystemRendererGPU(this) : new ParticleSystemRendererCPU(this);
-        this._particleSystem.processor.onInit(ps);
+        // if particle system is null we run the old routine
+        // else if particle system is not null we do nothing
+        if (this._particleSystem === null) {
+            this._particleSystem = ps;
+            const useGPU = this._useGPU && isSupportGPUParticle();
+            this._particleSystem.processor = useGPU ? new ParticleSystemRendererGPU(this) : new ParticleSystemRendererCPU(this);
+            this._particleSystem.processor.onInit(ps);
+        }
```
pandamicro, line 225:
```
onInit (ps: any) {
    // No repetitive code
    this.create(ps);
    if (!this._particleSystem.processor) {
        const useGPU = this._useGPU && isSupportGPUParticle();
        this._particleSystem.processor = useGPU ? new ParticleSystemRendererGPU(this) : new ParticleSystemRendererCPU(this);
        this._particleSystem.processor.onInit(ps);
    }
    else {
        // message: Particle system shouldn't be initialized repetitively, otherwise there will be potential leak.
        errorID(...);
    }
}
```

pandamicro (changes requested):

cocos/particle/renderer/particle-system-renderer-data.ts
```diff
@@ -204,18 +204,23 @@ export default class ParticleSystemRenderer {
     private _particleSystem: any = null!; // ParticleSystem
```
shrinktofit, line 205: Could you supply the type info instead of any?

pandamicro (approved):

cocos/particle/renderer/particle-system-renderer-data.ts
zxx43: > 
> 
> Could you supply the type info instead of any?

ParticleSystemRenderer will import ParticleSystem and ParticleSystem will import ParticleSystemRenderer too, so if we use type ParticleSystem here will cause endless import and the editor will crash.

PR #8838, opened 2021-06-17
Optimize the synchronization of Pass.

caryliu1999:
Native: https://github.com/cocos-creator/engine-native/pull/3628


cocos/core/renderer/core/pass.ts
```diff
@@ -398,6 +418,12 @@ export class Pass {
             this._rootBuffer = null;
         }
 
+        if (JSB) {
+            if (this._passHandle) {
+                PassPool.free(this._passHandle);
+            }
+        }
```
minggo: 这放到 _destroy 吧

PR #8861, opened 2021-06-22
3.3 bugfix instance attr buffer sync

PatriceJiang:
Re: cocos-creator/3d-tasks#

Changelog:
 * fix instanced buffer error

### DIFF
--- a/cocos/core/renderer/scene/model.ts
+++ b/cocos/core/renderer/scene/model.ts
@@ -517,7 +517,6 @@ export class Model {
         for (let j = 0; j < attributes.length; j++) {
             const attribute = attributes[j];
             if (!attribute.isInstanced) { continue; }
-            // const attr = attrs.attributes[j];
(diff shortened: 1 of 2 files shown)


cocos/core/renderer/scene/model.ts
```diff
@@ -517,15 +517,16 @@ export class Model {
         for (let j = 0; j < attributes.length; j++) {
             const attribute = attributes[j];
             if (!attribute.isInstanced) { continue; }
-            const attr = attrs.attributes[j];
+            // const attr = attrs.attributes[j];
```
minggo: 删掉吧

PatriceJiang:
```suggestion
```

PatriceJiang: ok

caryliu1999 (approved):

PR #8863, opened 2021-06-22
[Add] native renderScene update

GengineJS:


GengineJS:
https://github.com/cocos-creator/engine-native/pull/3638

PR #8866, opened 2021-06-23
add spine&db system && optimize system priority

novain1102:
Re: 
https://github.com/cocos-creator/3d-tasks/issues/6583
https://github.com/cocos-creator/3d-tasks/issues/7389
- 添加spine/ dragonbones system管理. 统一在postUpdate更新原生和web的数据. 修复和widget并用位置更新错误

另外: 
- 为System添加Priority枚举类. 用于优化注册系统时的参数
- 修复PRIORITY_SYSTEM: number = 1 << 31;会溢出的错误. 并将其废弃, 转调System.Priority.SCHDULED
- 移除没有被使用的PRIORITY_NON_SYSTEM

### DIFF
--- a/cocos/core/components/system.ts
+++ b/cocos/core/components/system.ts
@@ -39,7 +39,7 @@ export default class System implements ISchedulable {
     static Priority = Enum({
         LOW: 0,
         MEDIUM: 100,
-        HIGH: 1000,
+        HIGH: 200,
         SCHEDULER: (1 << 31) >>> 0,
     });
 
--- a/cocos/core/components/system.ts
+++ b/cocos/core/components/system.ts
@@ -40,7 +40,7 @@ export default class System implements ISchedulable {
         LOW: 0,
         MEDIUM: 100,
         HIGH: 1000,
-        SCHEDULED: (1 << 31) >>> 0,
+        SCHEDULER: (1 << 31) >>> 0,
     });
 
     protected _id = '';
--- a/cocos/core/components/system.ts
+++ b/cocos/core/components/system.ts
@@ -29,12 +29,20 @@
  */
 
 import { ISchedulable } from '../scheduler';
+import { Enum } from '../value-types/enum';
 
 /**
  * @en Base class for all functional system managed by [[Director]].
  * @zh 功能系统的基类，由 [[Director]] 管理。
  */
 export default class System implements ISchedulable {
+    static Priority = Enum({
+        LOW: 0,
+        MEDIUM: 100,
+        HIGH: 1000,
+        SCHEDULED: (1 << 31) >>> 0,
+    });
+
     protected _id = '';
     protected _priority = 0;
     protected _executeInEditMode = false;
--- a/cocos/dragon-bones/ArmatureSystem.ts
+++ b/cocos/dragon-bones/ArmatureSystem.ts
@@ -13,13 +13,10 @@ export class ArmatureSystem extends System {
      */
     static readonly ID = 'ARMATURE';
 
-    private static instance: ArmatureSystem;
+    private static _instance: ArmatureSystem;
 
     private constructor () {
         super();
-        if (ArmatureSystem.instance) {
-            errorID(12101);
-        }
     }
 
     /**
@@ -29,34 +26,34 @@ export class ArmatureSystem extends System {
      * 获取 Dragonbones Armature系统的单例。
      */
     public static getInstance () {
-        if (!ArmatureSystem.instance) {
-            ArmatureSystem.instance = new ArmatureSystem();
-            director.registerSystem(ArmatureSystem.ID, ArmatureSystem.instance, System.Priority.HIGH);
+        if (!ArmatureSystem._instance) {
+            ArmatureSystem._instance = new ArmatureSystem();
+            director.registerSystem(ArmatureSystem.ID, ArmatureSystem._instance, System.Priority.HIGH);
         }
-        return ArmatureSystem.instance;
+        return ArmatureSystem._instance;
     }
 
-    private armatures = new Set<ArmatureDisplay>();
+    private _armatures = new Set<ArmatureDisplay>();
 
     public add (armature: ArmatureDisplay | null) {
         if (!armature) return;
-        if (!this.armatures.has(armature)) {
-            this.armatures.add(armature);
+        if (!this._armatures.has(armature)) {
+            this._armatures.add(armature);
         }
     }
 
     public remove (armature: ArmatureDisplay | null) {
         if (!armature) return;
-        if (this.armatures.has(armature)) {
-            this.armatures.delete(armature);
+        if (this._armatures.has(armature)) {
+            this._armatures.delete(armature);
         }
     }
 
     postUpdate (dt: number) {
-        if (!this.armatures) {
+        if (!this._armatures) {
             return;
         }
-        this.armatures.forEach((armature) => {
+        this._armatures.forEach((armature) => {
             armature.updateAnimation(dt);
         });
     }
--- a/cocos/dragon-bones/ArmatureSystem.ts
+++ b/cocos/dragon-bones/ArmatureSystem.ts
@@ -2,6 +2,7 @@ import { director } from '../core/director';
 import { System } from '../core/components';
 import { ArmatureDisplay } from './ArmatureDisplay';
 import { legacyCC } from '../core/global-exports';
+import { errorID } from '../core';
 
 export class ArmatureSystem extends System {
     /**
@@ -16,6 +17,9 @@ export class ArmatureSystem extends System {
 
     private constructor () {
         super();
+        if (ArmatureSystem.instance) {
+            errorID(12101);
+        }
     }
 
     /**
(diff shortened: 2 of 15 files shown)


cocos/core/components/system.ts
```diff
@@ -29,12 +29,20 @@
...
 /**
  * @en Base class for all functional system managed by [[Director]].
  * @zh 功能系统的基类，由 [[Director]] 管理。
  */
 export default class System implements ISchedulable {
+    static Priority = Enum({
+        LOW: 0,
+        MEDIUM: 100,
+        HIGH: 1000,
+        SCHEDULED: (1 << 31) >>> 0,
```
pandamicro: SCHEDULER

cocos/dragon-bones/ArmatureSystem.ts
```diff
@@ -0,0 +1,71 @@
...
+     * Gets the instance of the tween system.
+     * @zh
+     * 获取Spine骨骼系统的实例。
+     */
+
+    public static get instance (): ArmatureSystem {
+        return ArmatureSystem._instance;
+    }
+
+    public static set instance (sys:ArmatureSystem) {
```
pandamicro: Not necessary, singleton instance don't need to be overridden. You need to check whether `_instance` exist, if not, create it automatically

cocos/dragon-bones/ArmatureSystem.ts
```diff
@@ -0,0 +1,71 @@
...
+director.once(Director.EVENT_INIT, () => {
+    initArmatureSystem();
+});
+
+function initArmatureSystem () {
+    const oldIns = ArmatureSystem.instance;
+    if (oldIns) {
+        director.unregisterSystem(oldIns);
+    }
+    const sys = new ArmatureSystem();
```
pandamicro: Just use `ArmatureSystem.instance`, you can't instantiate a singleton object with `new`, if you want to avoid that, you can create a constructor, and check for `ArmatureSystem.instance` existence, if it exists already, just report an error.

cocos/dragon-bones/ArmatureSystem.ts
```diff
@@ -0,0 +1,71 @@
...
+    (ArmatureSystem.instance as any) = sys;
+    director.registerSystem(ArmatureSystem.ID, sys, System.Priority.SCHEDULED);
```
pandamicro: Suggest to put registration into `constructor`, then you can be sure the singleton instance won't be created twice or registered twice

cocos/dragon-bones/ArmatureSystem.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+director.once(Director.EVENT_INIT, () => {
+    initArmatureSystem();
+});
+
+function initArmatureSystem () {
+    const oldIns = ArmatureSystem.instance;
+    if (oldIns) {
+        director.unregisterSystem(oldIns);
```
pandamicro: As the following suggestion, don't need to do check and unregistration here

cocos/spine/skeleton-system.ts
```diff
@@ -0,0 +1,69 @@
+import { director, Director } from '../core/director';
+import { System } from '../core/components';
+import { Skeleton } from './skeleton';
+
+export class SkeletonSystem extends System {
```
pandamicro, line 6: Basically the same implementation, you can merge `SkeletonSystem` with `ArmatureSystem` to `SkeletonAnimation2D` (as we also have 3d skeleton animation). Reuse the same implementation

cocos/spine/skeleton.ts
```diff
@@ -707,7 +708,7 @@ export class Skeleton extends Renderable2D {
         return this._cacheMode !== AnimationCacheMode.REALTIME;
     }
 
-    public update (dt: number) {
+    public updateAnimation (dt: number) {
```
pandamicro, line 711: To be fully comply type check, you can have an interface

```
interface SkeletonAnimation2D {
    updateAnimation (dt: number): void;
}
```

Implement this interface in  both spine skeleton and db armature

pandamicro (changes requested):

cocos/spine/skeleton-system.ts
novain1102: Skeleton refers to Spine, Armature refers to Dragonbones, it's unreasonable to unify them as SkeletonAnimation. What's more, there are some slightly differences between their implementation. And the difference might be bigger and bigger in the future , although them are roughly similar now.

cocos/spine/skeleton.ts
novain1102: Same as above. I think that there is no need to use a unified interface to increase coupling

cocos/dragon-bones/ArmatureSystem.ts
```diff
@@ -0,0 +1,61 @@
...
+     */
+    public static getInstance () {
+        if (!ArmatureSystem.instance) {
+            ArmatureSystem.instance = new ArmatureSystem();
+            director.registerSystem(ArmatureSystem.ID, ArmatureSystem.instance, System.Priority.SCHEDULER);
+        }
+        return ArmatureSystem.instance;
+    }
+
+    private static armatures = new Set<ArmatureDisplay>();
```
pandamicro: ArmatureSystem is a singleton, don't need to share `armatures` as static property, just use class property

cocos/dragon-bones/ArmatureSystem.ts
```diff
@@ -0,0 +1,61 @@
...
+     * The ID flag of the system.
+     * @zh
+     * 此系统的 ID 标记。
+     */
+    static readonly ID = 'ARMATURE';
+
+    private static instance: ArmatureSystem;
+
+    private constructor () {
+        super();
```
pandamicro, line 19: You can check whether instance exist here, if exist report error to avoid duplicated singleton

```
if (ArmatureSystem.instance) {
    errorID(...);
}
```

cocos/dragon-bones/ArmatureSystem.ts
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * @en
+     * Gets the instance of the ArmatureSystem system.
+     * @zh
+     * 获取 Dragonbones Armature系统的单例。
+     */
+    public static getInstance () {
+        if (!ArmatureSystem.instance) {
+            ArmatureSystem.instance = new ArmatureSystem();
+            director.registerSystem(ArmatureSystem.ID, ArmatureSystem.instance, System.Priority.SCHEDULER);
```
pandamicro: I noticed we are actually using 200 for scheduler system, code can be found in director. All current system priorities:

- AnimationManager: Scheduler.PRIORITY_SYSTEM
- Physics2D & 3D: 0
- Tween: 100
- Scheduler: 200

Please update them with the following

1. Scheduler: System.Priority.SCHEDULER
2. AnimationManager & Tween: System.Priority.HIGH
3. SpineSystem & ArmatureSystem: System.Priority.HIGH
4. Physics2D & 3D: System.Priority.LOW

That means Scheduler should be executed first, then all animation system, at last physics update

cocos/spine/skeleton-system.ts
pandamicro: You are right, separate them into spine and db module can also help to cull system code when module is desactivated.

cocos/spine/skeleton-system.ts
```diff
@@ -0,0 +1,58 @@
...
+     * The ID flag of the system.
+     * @zh
+     * 此系统的 ID 标记。
+     */
+    static readonly ID = 'SKELETON';
+
+    private static instance: SkeletonSystem;
+
+    private constructor () {
+        super();
```
pandamicro, line 18: Same as ArmatureSystem

cocos/spine/skeleton-system.ts
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * @en
+     * Gets the instance of the Spine Skeleton system.
+     * @zh
+     * 获取 Spine 骨骼系统的单例。
+     */
+    public static getInstance () {
+        if (!SkeletonSystem.instance) {
+            SkeletonSystem.instance = new SkeletonSystem();
+            director.registerSystem(SkeletonSystem.ID, SkeletonSystem.instance, System.Priority.SCHEDULER);
```
pandamicro: Same as ArmatureSystem

cocos/spine/skeleton-system.ts
```diff
@@ -0,0 +1,58 @@
...
+        }
+        return SkeletonSystem.instance;
+    }
+
+    private static skeletons = new Set<Skeleton>();
```
pandamicro: Same as ArmatureSystem

pandamicro (changes requested):

cocos/dragon-bones/ArmatureSystem.ts
```diff
@@ -0,0 +1,65 @@
...
+     */
+    static readonly ID = 'ARMATURE';
+
+    private static instance: ArmatureSystem;
+
+    private constructor () {
+        super();
+        if (ArmatureSystem.instance) {
+            errorID(12101);
+        }
```
pandamicro: Just found another issue, if use new multiple times, it won't trigger any issue, 

```
let singleton1 = new ArmatureSystem();
let singleton2 = new ArmatureSystem();
let singleton3 = new ArmatureSystem();
// Still work
```

we should assign instance here instead of in `getInstance`.

cocos/dragon-bones/ArmatureSystem.ts
```diff
@@ -0,0 +1,65 @@
...
+     */
+    public static getInstance () {
+        if (!ArmatureSystem.instance) {
+            ArmatureSystem.instance = new ArmatureSystem();
+            director.registerSystem(ArmatureSystem.ID, ArmatureSystem.instance, System.Priority.HIGH);
+        }
+        return ArmatureSystem.instance;
+    }
+
+    private armatures = new Set<ArmatureDisplay>();
```
pandamicro: add `_` in the beginning of private properties `_armatures`

pandamicro (changes requested):

pandamicro (approved):

PR #8876, opened 2021-06-24
sprite fillbuffer cache & sprite misc change when set spf

LinYunMo:
Re: https://github.com/cocos-creator/3d-tasks/issues/7806

### DIFF
--- a/cocos/2d/assembler/sprite/simple.ts
+++ b/cocos/2d/assembler/sprite/simple.ts
@@ -82,39 +82,11 @@ export const simple: IAssembler = {
         }
     },
 
-    fillBuffers (sprite: Sprite, renderer: Batcher2D) {
-        if (sprite === null) {
-            return;
-        }
-
-        // const buffer: MeshBuffer = renderer.createBuffer(
-        //     sprite.renderData!.vertexCount,
-        //     sprite.renderData!.indicesCount,
-        // );
-        // const commitBuffer: IUIRenderData = renderer.createUIRenderData();
+    updateWorldVerts (sprite: Sprite, vData: Float32Array) {
         const renderData = sprite.renderData;
 
         const dataList: IRenderData[] = renderData!.data;
         const node = sprite.node;
-
-        let buffer = renderer.acquireBufferBatch()!;
-
-        let vertexOffset = buffer.byteOffset >> 2;
-        let indicesOffset = buffer.indicesOffset;
-        let vertexId = buffer.vertexOffset;
-
-        const isRecreate = buffer.request();
-        if (!isRecreate) {
-            buffer = renderer.currBufferBatch!;
-            vertexOffset = 0;
-            indicesOffset = 0;
-            vertexId = 0;
-        }
-
-        // buffer data may be reallocated, need get reference after request.
-        const vBuf = buffer.vData!;
-        const iBuf = buffer.iData!;
-        const vData = renderData!.vData!;
 
         const data0 = dataList[0];
         const data3 = dataList[3];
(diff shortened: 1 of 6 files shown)


pandamicro:
https://github.com/cocos-creator/3d-tasks/issues/7487

cocos/2d/assembler/sprite/simple.ts
```diff
@@ -170,6 +142,42 @@ export const simple: IAssembler = {
...
+        // const commitBuffer: IUIRenderData = renderer.createUIRenderData();
+
+        let buffer = renderer.acquireBufferBatch()!;
+
+        let vertexOffset = buffer.byteOffset >> 2;
+        let indicesOffset = buffer.indicesOffset;
+        let vertexId = buffer.vertexOffset;
+
+        const isRecreate = buffer.request();
+        if (!isRecreate) {
```
pandamicro: The naming  here  is confusing, should only update buffer if it is recreated, so I suggest name this variable `bufferUnchanged`

cocos/2d/components/sprite.ts
```diff
@@ -763,8 +763,12 @@ export class Sprite extends Renderable2D {
 
         if (spriteFrame) {
             if (!oldFrame || spriteFrame !== oldFrame) {
-                if  (spriteFrame.loaded) {
-                    this._onTextureLoaded();
+                if (spriteFrame.loaded) {
+                    if (oldFrame && oldFrame.texture.getHash() !== spriteFrame.texture.getHash()) {
```
pandamicro: If old frame not exist, we should also invoke `_onTextureLoaded`, the logic here is too complicated

```
let textureChanged = false;
if (spriteFrame) {
    if (!oldFrame || oldFrame.texture !== spriteFrame.texture) {
         textureChanged = true;
    }
    if (spriteFrame.loaded) {
        if (textureChanged) {
            this._onTextureLoaded();
        }
        else {
            this._applySpriteSize();
        }
    }
    else {
        spriteFrame.once('load', this._onTextureLoaded, this);
    }
}
```

cocos/2d/framework/ui-transform.ts
```diff
@@ -92,6 +92,7 @@ export class UITransform extends Component {
         } else {
             this.node.emit(SystemEventType.SIZE_CHANGED);
         }
+        this._rectDirty = true;
```
pandamicro: should use the same mechanism to mark render data vertex dirty:

```
this.node.uiProps.uiComp.markForUpdateRenderData(true);
```

cocos/2d/assembler/sprite/simple.ts
```diff
@@ -170,6 +142,42 @@ export const simple: IAssembler = {
...
         }
+    },
+
+    fillBuffers (sprite: Sprite, renderer: Batcher2D) {
+        if (sprite === null) {
+            return;
+        }
+
+        const vData = sprite.renderData!.vData!;
+        if (sprite.node.hasChangedFlags || (this.uiTransCache && this.uiTransCache._rectDirty)) {
```
pandamicro: We can only check for `sprite.node.hasChangedFlags` here. UITransform size and anchor dirty should be updated directly in `updateVertexData`

cocos/2d/assembler/sprite/simple.ts
```diff
@@ -170,6 +142,42 @@ export const simple: IAssembler = {
...
+            this.updateWorldVerts(sprite, vData);
+            this.uiTransCache._rectDirty = false;
+        }
```
pandamicro, line 156:
```
if (sprite.node.hasChangedFlags) {
    this.updateWorldVerts(sprite, vData);
}
```

cocos/2d/assembler/sprite/simple.ts
```diff
@@ -229,13 +237,12 @@ export const simple: IAssembler = {
...
         dataList[0].x = l;
         dataList[0].y = b;
-        dataList[0].z = 0;
 
         dataList[3].x = r;
         dataList[3].y = t;
-        dataList[3].z = 0;
 
         renderData.vertDirty = false;
+        this.uiTransCache = uiTrans;
```
pandamicro: No need to save uiTransCache, just update world vertex here

```
this.updateWorldVerts(sprite, renderData.vData);
```

pandamicro (commented):
I don't feel right about `rectDirty` design and the checks in fillBuffers, it feels like adding up complexity and easy to trigger issues

cocos/2d/assembler/sprite/simple.ts
LinYunMo: will be update twice，node.hasChangedFlags always true until frame finish

pandamicro (approved):

PR #8890, opened 2021-06-26
[BREAKING CHANGE] remove returning callback on systemEvent.on interface

PPpro:
Re: cocos-creator/3d-tasks#

Changelog:
 *


PPpro:
https://github.com/cocos-creator/3d-tasks/issues/7928

PR #8894, opened 2021-06-28
Class decorator uniquelyReferenced

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Introduces the class decorator unique to optimize serialization of "value type"s.

Engine side only register the option info. Implementation now resides in serialization in Editor: https://github.com/cocos-creator/editor-3d/pull/5739

### DIFF
--- a/cocos/core/data/decorators/serializable.ts
+++ b/cocos/core/data/decorators/serializable.ts
@@ -28,7 +28,7 @@
  * @module decorator
  */
 
-import { EDITOR } from 'internal:constants';
+import { EDITOR, TEST } from 'internal:constants';
 import { emptyDecorator, LegacyPropertyDecorator } from './utils';
 import { property, IPropertyOptions } from './property';
 import { getOrCreateSerializationMetadata } from '../serialization-metadata';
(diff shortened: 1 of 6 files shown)


shrinktofit:
## Naming bikeshed

The name should express what it does:

- `unique`(verus "shared")

- `nonShared`(more compresshensive but verbose)

- `valueType`(easy to misunderstand: should it inherits from `ValueType`?)

----------------

### Update

I think `unique` is misunderstood too. It can mean: "This class has only one instance".

So I proposed a more intuitional name: `uniquelyReferenced`.

> We're not alone: https://developer.apple.com/documentation/swift/2429905-isknownuniquelyreferenced

cocos/core/data/decorators/serializable.ts
```diff
@@ -56,3 +57,59 @@ function makeSerializable (options: IPropertyOptions) {
...
+ * const bar1 = new Bar();
+ * const bar2 = new Bar();
+ * bar2.foo = bar1.foo; // 由程序逻辑让它们引用同一个对象
+ * ```
+ * `bar1` 和 `bar2` 引用同一个 `Foo` 对象。
+ * 但在反序列化之后，`bar1.foo === bar2.foo` 永不成立。
+ */
+// eslint-disable-next-line func-names, @typescript-eslint/ban-types
+export function unique<TFunction extends Function> (constructor: TFunction) {
+    const metadata = getOrCreateSerializationMetadata(constructor);
```
jareguo: 加上 EDITOR 的判断吧，一般游戏没有序列化的必要，所以 unique 整个装饰器都能剔除掉

shrinktofit: 加了

SantyWang (approved):

PR #8896, opened 2021-06-28
Allow specifying delta time in `Game.prototype.step`

shrinktofit:
Re: cocos-creator/3d-tasks#7930

Changelog:
 * The usage is, for example, to step in fixed FPS.

### DIFF
--- a/cocos/core/director.ts
+++ b/cocos/core/director.ts
@@ -927,6 +927,21 @@ export class Director extends EventTarget {
         }
     }
 
+    /**
+     * @en
+     * Run once main loop.
+     * @zh
+     * 运行一次主循环。
+     * @param deltaTime 此次循环与上次循环间隔的时间，单位为毫秒。默认为自然流逝的时间。
+     */
+    public tick (deltaTime?: number) {
+        this.mainLoop(
+            typeof deltaTime === 'undefined'
+                ? undefined
+                : this._lastUpdate + deltaTime,
+        );
+    }
+
     private _initOnRendererInitialized () {
         this._totalFrames = 0;
         this._lastUpdate = performance.now();
--- a/cocos/core/director.ts
+++ b/cocos/core/director.ts
@@ -301,7 +301,7 @@ export class Director extends EventTarget {
      * @en Calculates delta time since last time it was called, the result is saved to an internal property.
      * @zh 计算从上一帧到现在的时间间隔，结果保存在私有属性中
      */
-    public calculateDeltaTime (now) {
+    public calculateDeltaTime (now?: number) {
         if (!now) now = performance.now();
 
         this._deltaTime = now > this._lastUpdate ? (now - this._lastUpdate) / 1000 : 0;
@@ -877,14 +877,14 @@ export class Director extends EventTarget {
      * @en Run main loop of director
      * @zh 运行主循环
      */
-    public mainLoop (time: number) {
+    public mainLoop (time?: number) {
         if (this._purgeDirectorInNextLoop) {
             this._purgeDirectorInNextLoop = false;
             this.purgeDirector();
         } else if (!this._invalid) {
             // calculate "global" dt
             if (EDITOR && !legacyCC.GAME_VIEW) {
-                this._deltaTime = time;
+                this._deltaTime = time!;
             } else {
                 this.calculateDeltaTime(time);
             }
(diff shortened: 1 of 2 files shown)


cocos/core/game.ts
```diff
@@ -368,8 +368,8 @@ export class Game extends EventTarget {
      * @en Run the game frame by frame.
      * @zh 执行一帧游戏循环。
      */
-    public step () {
-        legacyCC.director.mainLoop();
+    public step (now?: number) {
+        legacyCC.director.mainLoop(now);
```
shrinktofit: Another choice is to define the parameter as delta time so the implementation becomes:
```ts
public step (deltaTime?: number) {
    // deltaTime is in seconds?
    legacyCC.director.mainLoop(typeof deltaTime === 'number'
        ? (legacyCC.director.getCurrentTime() + deltaTime * 1000.0) // Convert to now
        : undefined,
    );
}
```

pandamicro: We should use deltaTime instead of now, but that means we need to move out director's calculation to game, so a little more work that this version.
Remove `_lastUpdate`, `_startTime` in director, and implement deltaTime calculation in game

pandamicro: To preserve mainLoop API usage, we can add `director.tick`

shrinktofit: Updated. Please review.

shrinktofit:
@PPpro  Could you please review this to see if it'd affect PAL/adapter?

cocos/core/director.ts
```diff
@@ -927,6 +927,21 @@ export class Director extends EventTarget {
...
+     * 运行一次主循环。
+     * @param deltaTime 此次循环与上次循环间隔的时间，单位为毫秒。默认为自然流逝的时间。
+     */
+    public tick (deltaTime?: number) {
+        this.mainLoop(
+            typeof deltaTime === 'undefined'
+                ? undefined
+                : this._lastUpdate + deltaTime,
+        );
+    }
```
PPpro, line 943: is it necessary to add a new interface for developer ? @pandamicro

PPpro: I suggest to change this interface to a private method, because it's not  a useful interface for developper

PPpro: I suggest to change this interface to a private method, because it's not a useful interface for developper

PPpro (changes requested):

cocos/core/director.ts
PPpro: I suggest to change this interface to a private method, because it's not a useful interface for developper

PPpro (changes requested):

cocos/core/director.ts
```diff
@@ -877,14 +877,14 @@ export class Director extends EventTarget {
      * @en Run main loop of director
      * @zh 运行主循环
      */
-    public mainLoop (time: number) {
+    public mainLoop (time?: number) {
```
pandamicro, line 880: `mainLoop` is the function that we want to deprecated in the future, because it's not using a regular tick parameter. So I think we should move actual implementation to new standard function `tick` and use `tick` in mainLoop

    public mainLoop (time?: number) {
        if (!this._invalid) {
            // calculate "global" dt
            this.calculateDeltaTime(time);
            const dt = this._deltaTime;

            this.tick(dt);
        }
    }

cocos/core/director.ts
pandamicro: It is a useful method, and we should use the new `tick` function instead of `mainLoop`.
`tick` accept dt, which means developer can easier control speed of game globally.

cocos/core/game.ts
```diff
@@ -365,11 +365,14 @@ export class Game extends EventTarget {
...
+     * @en
+     * Run once game loop.
+     * @zh
+     * 执行一帧游戏循环。
+     * @param deltaTime 此次循环与上次循环间隔的时间，单位为毫秒。默认为自然流逝的时间。
      */
-    public step () {
-        legacyCC.director.mainLoop();
+    public step (deltaTime?: number) {
+        legacyCC.director.tick(deltaTime);
```
pandamicro, line 375: We should also change `mainLoop` usage to `tick` in `runMainLoop` and calculate delta time in game directly.

pandamicro (changes requested):

PPpro:
AGREE

PR #8902, opened 2021-06-28
Fix testcase raycast scene

GengineJS:


GengineJS:
native: https://github.com/cocos-creator/engine-native/pull/3665

PR #8913, opened 2021-06-29
Move director delta time to game and remove deprecated APIs

pandamicro:
To replace https://github.com/cocos-creator/engine/pull/8896

https://github.com/cocos-creator/3d-tasks/issues/7930


cocos/core/deprecated.ts
```diff
@@ -265,3 +267,62 @@ replaceProperty(Root.prototype, 'Root.prototype', [
...
+        name: 'setDepthTest',
+        suggest: 'please use camera API instead',
+    },
+    {
+        name: 'setClearColor',
+        suggest: 'please use camera API instead',
+    },
+    {
+        name: 'getWinSize',
+        suggest: 'please use cc.winSize instead',
```
PPpro: cc.winSize will be deprecated too, use cc.view.getVisibleSize instead

cocos/core/director.ts
```diff
@@ -297,21 +288,6 @@ export class Director extends EventTarget {
...
-        if (!now) now = performance.now();
-
-        this._deltaTime = now > this._lastUpdate ? (now - this._lastUpdate) / 1000 : 0;
-        if (DEBUG && (this._deltaTime > 1)) {
-            this._deltaTime = 1 / 60.0;
-        }
-
-        this._lastUpdate = now;
-    }
-
```
PPpro: I think the public method can't be removed, should be marked as deprecation instead

or it would be a breaking change in v3.3
maybe can call `game._calculateDT()` here

cocos/core/director.ts
```diff
@@ -361,33 +337,9 @@ export class Director extends EventTarget {
...
-     * On Mac winSize and winSizeInPixels return the same value.
-     * (The pixel here refers to the resource resolution. If you want to get the physics resolution of device, you need to use `view.getFrameSize()`)
-     * @zh
-     * 获取视图大小，以像素为单位（这里的像素指的是资源分辨率。
-     * 如果要获取屏幕物理分辨率，需要用 `view.getFrameSize()`）
-     * @deprecated since v2.0
-     */
-    public getWinSizeInPixels () {
-        return size(legacyCC.winSize);
+        this.on(Director.EVENT_END_FRAME, () => {
```
PPpro:
```suggestion
        this.once(Director.EVENT_END_FRAME, () => {
```

cocos/core/director.ts
```diff
@@ -740,51 +637,6 @@ export class Director extends EventTarget {
...
-    }
-
-    /**
-     * @en Returns the current time.
-     * @zh 获取当前帧的时间。
-     */
-    public getCurrentTime () {
-        return this._lastUpdate;
-    }
-
```
PPpro: should we use 
```
/**
     * @en Returns the delta time since last frame.
     * @zh 获取上一帧的增量时间。
     * @depreacated since v3.3, please use game.getDeltaTime() instead
     */
    public getDeltaTime () { }

    /**
     * @en Returns the total passed time since game start, unit: ms
     * @zh 获取从游戏开始到现在总共经过的时间，单位为 ms
     * @depreacated since v3.3, please use game.getTotalTime() instead
     */
    public getTotalTime () { }

    /**
     * @en Returns the current time.
     * @zh 获取当前帧的时间。
     * @depreacated since v3.3, please use game.getCurrentTime() instead
     */
    public getCurrentTime () { }
```

so that developers can see the deprecation info when they are coding in VSCode

cocos/core/director.ts
pandamicro: It can still be invoked, I have added removed warning through `removeProperty` in deprecated

cocos/core/director.ts
pandamicro: That's a valid suggestion, thanks

SantyWang:
I think delta time calculation can be moved to another module named `Time Manager` or `Timer`, and cc.game holds a timer, and developer can use their custom timer

pandamicro:
~~Current implementation have some issues on setting frame rate, will update solution~~

Fixed

pandamicro:
Ready for review

cocos/core/deprecated.ts
```diff
@@ -265,3 +267,66 @@ replaceProperty(Root.prototype, 'Root.prototype', [
...
+        name: 'calculateDeltaTime',
+    },
+    {
+        name: 'getDeltaTime',
+        suggest: 'Use game.timer.deltaTime instead',
+    },
+    {
+        name: 'getTotalTime',
+        suggest: 'Use game.timer.deltaTime instead',
+    },
```
PPpro:
```suggestion
    {
        name: 'getTotalTime',
        suggest: 'Use game.timer.totalTime instead',
    },
```

cocos/core/index.ts
```diff
@@ -50,6 +50,7 @@ export * from './data';
 export * from './event';
 export * from './assets';
 export * from './platform';
+export * from './timer';
```
PPpro: This means we expose the Timer class to the developers, should we ? Or it's just a private class for engine ?

cocos/core/timer.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    /**
+     * @en Check whether the timer is paused.
+     * @zh 判断计时器是否暂停。
+     */
+    public get paused () {
+        return this._paused;
+    }
+
+    public declare frameTime: number;
```
PPpro: why  just declare the `frameTime`, instead of define it ? the same question in these properties `_sys _initTime _startTime _callback`

cocos/core/timer.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    private declare _sys: TickByTimer;
```
PPpro: the name of `_sys` is kind of strange, I thought it was defined in `sys.ts` at the first sight.
Maybe it should be named as `_ticker: ITicker` ?

PPpro (changes requested):

cocos/core/index.ts
pandamicro: It's exposed for developers to customize their own Timer and TickByTimer class

cocos/core/timer.ts
pandamicro: No need to initialize them,  they will be initialized during boot process

pandamicro:
@PPpro updated

PPpro (approved):

SantyWang (approved):

PPpro (approved):

PR #8925, opened 2021-07-01
Change the timing of node destroy

GengineJS:


pandamicro (commented):
You can delete the destroy function

pandamicro (approved):

PR #8962, opened 2021-07-13
bug fix: when the node is destory, node._activeInHierarchy may be error

yeshao2069:
Re: https://github.com/cocos-creator/engine/issues/4982

when the node is destory ,  to visit node._activeInHierarchy may be error


yeshao2069:
3.x  have done this.    

https://github.com/cocos-creator/engine/blob/v3.3.0/cocos/core/components/component.ts#L137

pandamicro (approved):

PR #8976, opened 2021-07-14
remove dependant info when release asset

SantyWang:
Re: cocos-creator/3d-tasks#

Changelog:
 *


cocos/core/asset-manager/release-manager.ts
```diff
@@ -251,7 +252,20 @@ class ReleaseManager {
         }
         dependUtil.remove(uuid);
         if (EDITOR) {
-            references!.remove(uuid);
+            const dependant = references!.get(uuid);
```
pandamicro, line 255: Where is the `references` from ?

pandamicro (approved):

PR #8978, opened 2021-07-14
fix bug remove all spine animations could cause a memory leak on native iOS

hailua54:
Steps to reproduce the bug: 
1. create 2 buttons: one for adding a new spine, one for remove all spines on the scene.
2. keep doing the process several times: using the 2 buttons to create some spine animations and remove all.
Result on iOS native: the memory keep increasing very fast and never be released.

The issue comes from using recycle pool for MeshBuffer in batcher-2d.ts. Also need to modify some code in mesh-buffer.ts to make sure the process of using recycle pool is correct.

### DIFF
--- a/cocos/2d/renderer/mesh-buffer.ts
+++ b/cocos/2d/renderer/mesh-buffer.ts
@@ -105,8 +105,9 @@ export class MeshBuffer {
         this._iaInfo = new InputAssemblerInfo(this.attributes, this.vertexBuffers, this.indexBuffer);
 
         // for recycle pool using purpose --
-        if (!this.vData || !this.iData)
+        if (!this.vData || !this.iData) {
             this._reallocBuffer();
+        }
         // ----------
     }
 
--- a/cocos/2d/renderer/mesh-buffer.ts
+++ b/cocos/2d/renderer/mesh-buffer.ts
@@ -104,7 +104,10 @@ export class MeshBuffer {
         this._attributes = attrs;
         this._iaInfo = new InputAssemblerInfo(this.attributes, this.vertexBuffers, this.indexBuffer);
 
-        this._reallocBuffer();
+        // for recycle pool using purpose --
+        if (!this.vData || !this.iData)
+            this._reallocBuffer();
+        // ----------
     }
 
     public request (vertexCount = 4, indicesCount = 6) {
@@ -152,6 +155,9 @@ export class MeshBuffer {
         this._nextFreeIAHandle = 0;
 
         this._dirty = false;
+
+        this._initIDataCount = 256 * 6;
+        this._initVDataCount = 256 * this._vertexFormatBytes;
     }
 
     public destroy () {
--- a/cocos/2d/renderer/mesh-buffer.ts
+++ b/cocos/2d/renderer/mesh-buffer.ts
@@ -156,9 +156,6 @@ export class MeshBuffer {
         this._nextFreeIAHandle = 0;
 
         this._dirty = false;
-
-        this._initIDataCount = 256 * 6;
-        this._initVDataCount = 256 * this._vertexFormatBytes;
     }
 
     public destroy () {
--- a/cocos/2d/renderer/batcher-2d.ts
+++ b/cocos/2d/renderer/batcher-2d.ts
@@ -114,6 +114,14 @@ export class Batcher2D {
                 }
             }
         }
+
+        // release the buffer to recycle pool --
+        const idx = this._bufferBatchPool.data.indexOf(buffer);
+        if (idx != -1) {
+            buffer.reset();
+            this._bufferBatchPool.removeAt(idx);
+        }
+        // ---
     }
 
     set currStaticRoot (value: UIStaticBatch | null) {
--- a/cocos/2d/renderer/batcher-2d.ts
+++ b/cocos/2d/renderer/batcher-2d.ts
@@ -117,7 +117,7 @@ export class Batcher2D {
 
         // release the buffer to recycle pool --
         const idx = this._bufferBatchPool.data.indexOf(buffer);
-        if (idx != -1) {
+        if (idx !== -1) {
             buffer.reset();
             this._bufferBatchPool.removeAt(idx);
         }


novain1102:
Could you please send a demo for us to test it? We may repair this issue in other relative pr in 3.3

hailua54:
Hi @mmyduckx ! You can check the demo in https://github.com/hailua54/test_mem_v3_spine
Repeating create and clear all spine in a native iOS device to see the memory leak issue.
some result screenshot also added in https://github.com/hailua54/test_mem_v3_spine/tree/master/docs

I tested the demo on cocos creator 3.1.1 which using a native-engine already merge fix for iOS memory leak in https://github.com/cocos-creator/engine-native/commit/b998435cca79c2c76ae40283c5fad98cec61284d.

novain1102 (approved):

LinYunMo (approved):

cocos/2d/renderer/mesh-buffer.ts
```diff
@@ -104,7 +104,10 @@ export class MeshBuffer {
         this._attributes = attrs;
         this._iaInfo = new InputAssemblerInfo(this.attributes, this.vertexBuffers, this.indexBuffer);
 
-        this._reallocBuffer();
+        // for recycle pool using purpose --
+        if (!this.vData || !this.iData)
```
pandamicro: I understand repetitive allocation could be improved.
But if we only remove allocation, with `_initVDataCount` reset, it could cause mismatch and due to unnecessary reallocation. Please consider reset `_initVDataCount` accordingly.

```
let count = this._initVDataCount;
this._initVDataCount = Math.floor(count / this._vertexFormatBytes) * this._vertexFormatBytes;
```

cocos/2d/renderer/mesh-buffer.ts
```diff
@@ -152,6 +155,9 @@ export class MeshBuffer {
         this._nextFreeIAHandle = 0;
 
         this._dirty = false;
+
+        this._initIDataCount = 256 * 6;
+        this._initVDataCount = 256 * this._vertexFormatBytes;
```
pandamicro: If we want to keep the same `vData` and `iData` buffer, we shouldn't reset them like this.

cocos/2d/renderer/mesh-buffer.ts
pandamicro: More seriously, if the new `_vertexFormatBytes` is bigger than the previous one, and we re-initialize it, the current implementation could cause out of bound access.

pandamicro (changes requested):

cocos/2d/renderer/mesh-buffer.ts
tomkt54: Dear! if you don't reset this._initIDataCount and this._initVData, any time you call 
`public request (vertexCount = 4, indicesCount = 6)`
...
The code will always double this._initVDataCount from previous value of using the buffer in Recycle Pool so it can lead to an infinity value => max length create buffer BUG.
```
if (byteOffset > byteLength || indicesOffset > indicesLength) {
  while (byteLength < byteOffset || indicesLength < indicesOffset) {
      this._initVDataCount *= 2;
      this._initIDataCount *= 2;
      byteLength = this._initVDataCount * 4;
      indicesLength = this._initIDataCount;
  }
  this._reallocBuffer();
}
```
Yes we can reset it to an initialize/smallest value so anytime you call buffer.request(...) it can be recalculated from a valid value.

cocos/2d/renderer/mesh-buffer.ts
tomkt54: _initIDataCount and _initVDataCount is not a value to determine the length of the buffer you created. It is just a variable for you to calculate a new valid buffer size incase you request the buffer with bigger this.vData/this.iData's length. Please check the code in cocos/2d/renderer/mesh-buffer.ts
`public request (vertexCount = 4, indicesCount = 6)`
...
```
let byteLength = this.vData!.byteLength;
let indicesLength = this.iData!.length;
if (byteOffset > byteLength || indicesOffset > indicesLength) {
    while (byteLength < byteOffset || indicesLength < indicesOffset) {
        this._initVDataCount *= 2;
        this._initIDataCount *= 2;

        byteLength = this._initVDataCount * 4;
        indicesLength = this._initIDataCount;
    }

    this._reallocBuffer();
}
```

also check the change request in https://github.com/cocos-creator/engine/blob/a6e996d2377ad38f9b6c2b6bf4437956bde9873a/cocos/2d/renderer/mesh-buffer.ts
it should be
```
// for recycle pool using purpose --
if (!this.vData || !this.iData)
    this._reallocBuffer();
// ----------
```

tomkt54: Note that we reuse this.vData and this.iData, the buffers's length will not be changed until you request a bigger length (need a _reallocBuffer()). 
Also please take a look into my commit for the issue
https://github.com/hailua54/engine/commit/a6e996d2377ad38f9b6c2b6bf4437956bde9873a

pandamicro: Ok, indeed we are using byteOffset and actual ArrayBuffer byteLength to check availability, so won't cause any out of bound issue. But the `_initVDataCount` value should be matching vData byteLength, that's just an implication of their relationship.

Imaging the following case
1. The buffer was initialized to contain 256 vertices
2. The buffer have been requested multiple times, till it grow to 1024 vertices length
3. The recycle pool reseted
4. Then the same buffer is retrieved and  re-initialized
5. The `_initVDataCount` reset to 256 while the buffer still have 1024 vertices long
6. While requested a buffer bigger than 1024 vertices, the while loop to calculate correct `_initVDataCount` will be running 2 more times than needed.

cocos/2d/renderer/mesh-buffer.ts
pandamicro: As the `byteOffset` is being reset to 0, the request would only double the buffer size while the `byteLength` and indices length is consumed out again, as `byteLength` equals to `_initVDataCount  * 4` and `indicesLength` equals to `_initIDataCount`, it's totally reasonable. Just as the implication I mentioned before.

pandamicro:
@hailua54 I think except not meeting the implication of `_initVDataCount` and `_initIDataCount` matching the `vData` and `iData` size, there is no problem with the PR. My comments are also based on my understanding of the code, I think it's not the best way to determine whether this PR is valid. As MeshBuffer is a very important infrastructure, @LinYunMo  will build some unit tests for it next week. We can accept this PR if unit tests passed, what do you think about that ?

cocos/2d/renderer/mesh-buffer.ts
hailua54: in the case above `_initVDataCount` will be double in while loop from a beginning value (`256 * this._vertexFormatBytes`) until the byteLength (`this._initVDataCount * 4`) bigger or equal the new required byteLength (`this.byteOffset + vertexCount * this._vertexFormatBytes`). Not sure it is the best way to calculate the new allocated byte length but it is from the original code already.

hailua54:
@pandamicro Yes since the issue increase memory size very fast so please create the unit test for it. You al so can check my test on https://github.com/hailua54/test_mem_v3_spine.

tomkt54 (approved):

cocos/2d/renderer/mesh-buffer.ts
hailua54: @pandamicro you still want to remove the code? It is not mandatory to keep the code since I already check condition to relocate this.iData and this.vData in `if (!this.vData || !this.iData)`
```
this._initIDataCount = 256 * 6;
this._initVDataCount = 256 * this._vertexFormatBytes;
```

hailua54:
@pandamicro you still want to remove the code? 
```
this._initIDataCount = 256 * 6;
this._initVDataCount = 256 * this._vertexFormatBytes;
```
It is not mandatory to keep the code since I already check condition to relocate this.iData and this.vData in `if (!this.vData || !this.iData)`
Keep the code is just a guaranty to make sure the while loop
```
while (byteLength < byteOffset || indicesLength < indicesOffset) {
    this._initVDataCount *= 2;
    this._initIDataCount *= 2;

    byteLength = this._initVDataCount * 4;
    indicesLength = this._initIDataCount;
}
```
always be executed from a beginning value in case the mesh-buffer just getting from pool. Incase remove the code, If someone need to reset this.vData or this .iData, they need to reset the values of this._initVDataCount and this._initIDataCount manually too before putting back the buffer to the pool.

pandamicro:
> you still want to remove the code?

Actually no, I asked @LinYunMo to add unit test cases for mesh buffer, to make sure its stability, so please wait for the result after that, just need a little more patient

pandamicro (approved):

pandamicro:
Merged to be tested with https://github.com/cocos-creator/engine/pull/9095

PR #8985, opened 2021-07-15
fix rich Text opacity bug

LinYunMo:
Re: https://github.com/cocos-creator/3d-tasks/issues/8034

### DIFF
--- a/cocos/2d/renderer/batcher-2d.ts
+++ b/cocos/2d/renderer/batcher-2d.ts
@@ -615,7 +615,7 @@ export class Batcher2D {
 
     private _preProcess (node: Node) {
         const render = node._uiProps.uiComp;
-        if (!render) { // hack for opacity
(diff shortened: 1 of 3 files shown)


cocos/2d/renderer/batcher-2d.ts
```diff
@@ -615,7 +615,7 @@ export class Batcher2D {
 
     private _preProcess (node: Node) {
         const render = node._uiProps.uiComp;
-        if (!render) { // hack for opacity
+        if (!render || !render.isRender2D) { // hack for opacity
```
pandamicro: Please find another way to implement this feature. Opacity could be stored in batcher2d and keep inheritance valid through the hierarchy

pandamicro (changes requested):

pandamicro (approved):

PR #8988, opened 2021-07-15
Allow custom effects to customize the inspector displayed in the material

chichinohaha:
Re: cocos-creator/3d-tasks#

Changelog:
 * Now we can customize the presentation of the effect in the material's inspector like this.
  image
Suppose we have an enabled extension named 'ss'
and there is a hack.js file in that plugin directory
This file is similar to the inspector,
but will get the currently used material in the first argument of the update method
example:
```js
// ss/hack.js
"use strict";
module.exports = Editor.Panel.define({
    template:
    /* html*/ `
    <ui-prop>
        <ui-label slot="label"></ui-label>
        <ui-checkbox slot="content"></ui-checkbox>
    </ui-prop>
    `,
    $: {
        checkbox: 'ui-checkbox',
        label: 'ui-label',
    },
    ready() {
        var _a;
        (_a = this.$.checkbox) === null || _a === void 0 ? void 0 : _a.addEventListener('change', (event) => {
            this.material.data[0].passes[0].defines[0].value = event.target.value;
            this.dispatch('change');
        });
    },
    update(material, assetList, metaList) {
        this.material = material;
        this.$.label.value = this.material.data[0].passes[0].defines[0].name;
        this.$.checkbox.value = this.material.data[0].passes[0].defines[0].value;
    },
});

```
 
image

### DIFF
--- a/editor/i18n/zh/assets.js
+++ b/editor/i18n/zh/assets.js
@@ -299,7 +299,8 @@ module.exports = {
             '警告：WebGL 1.0 平台不支持非 2 次幂贴图的 repeat 过滤模式，运行时会自动改为 clamp-to-edge 模式，这会使材质的 tilingOffset 等属性完全失效。',
         },
         material: {
-            'fail-to-load-custom-inspector': 'material: 自定义材质 {effect} 的 inspector 加载失败',
+            'fail-to-load-custom-inspector': 'material: 自定义 effect {effect} 的 inspector 加载失败',
+            'illegal-inspector-url': "Inspector的路径不合法",
         },
     },
 };
--- a/editor/i18n/zh/assets.js
+++ b/editor/i18n/zh/assets.js
@@ -298,5 +298,8 @@ module.exports = {
             modeWarn:
             '警告：WebGL 1.0 平台不支持非 2 次幂贴图的 repeat 过滤模式，运行时会自动改为 clamp-to-edge 模式，这会使材质的 tilingOffset 等属性完全失效。',
         },
+        material: {
+            'fail-to-load-custom-inspector': 'material: 自定义材质 {effect} 的 inspector 加载失败',
+        },
     },
 };
--- a/editor/inspector/assets/material.js
+++ b/editor/inspector/assets/material.js
@@ -69,20 +69,27 @@ exports.methods = {
         this.dirtyData.origin = this.dirtyData.realtime;
         this.dirtyData.uuid = '';
     },
-
-    async updateCustomInspector() {
+    /**
+     * 
+     * @param {string} inspector 
+     */
+    async updateCustomInspector(inspector) {
         this.$.customPanel.hidden = false;
         this.$.section.hidden = true;
         this.$.materialDump.hidden = true;
         try {
-            const relatePath = normalize((await this.getCustomInspector()).replace('packages://', ''));
-            const name = relatePath.split(sep)[0];
-            const packagePath = Editor.Package.getPackages({ name, enable: true })[0].path;
-            const path = join(packagePath, relatePath.split(name)[1]);
-            if (this.$.customPanel.getAttribute('src') !== join(path)) {
-                this.$.customPanel.setAttribute('src', path);
+            if (inspector.startsWith('packages://')) {
+                const relatePath = normalize(inspector.replace('packages://', ''));
+                const name = relatePath.split(sep)[0];
+                const packagePath = Editor.Package.getPackages({ name, enable: true })[0].path;
+                const path = join(packagePath, relatePath.split(name)[1]);
+                if (this.$.customPanel.getAttribute('src') !== join(path)) {
+                    this.$.customPanel.setAttribute('src', path);
+                }
+                this.$.customPanel.update(this.material, this.assetList, this.metaList);
+            } else {
+                throw Editor.I18n.t('ENGINE.assets.material.illegal-inspector-url');
             }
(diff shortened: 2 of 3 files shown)


editor/i18n/zh/assets.js
```diff
@@ -298,5 +298,8 @@ module.exports = {
             modeWarn:
             '警告：WebGL 1.0 平台不支持非 2 次幂贴图的 repeat 过滤模式，运行时会自动改为 clamp-to-edge 模式，这会使材质的 tilingOffset 等属性完全失效。',
         },
+        material: {
+            'fail-to-load-custom-inspector': 'material: 自定义材质 {effect} 的 inspector 加载失败',
```
arsen2010: “自定义材质” change to “自定义 Effect”

editor/inspector/assets/material.js
```diff
@@ -40,9 +44,19 @@ exports.$ = {
     materialDump: '.material-dump',
     useInstancing: '.useInstancing',
     useBatching: '.useBatching',
+    customSection: '.customSection',
+    customPanel: '.customPanel',
 };
 
 exports.methods = {
+    async getCustomInspector() {
```
arsen2010, line 52: advise use 'hasCustomInspector' as function name

editor/inspector/assets/material.js
```diff
@@ -40,9 +44,19 @@ exports.$ = {
...
+        const currentEffectInfo = this._effects.find(effect => { return effect.name === this.material.effect; });
+        if (currentEffectInfo && currentEffectInfo.uuid) {
+            const meta = await Editor.Message.request('asset-db', 'query-asset-meta', currentEffectInfo.uuid);
+            return meta && meta.userData && meta.userData.editor && meta.userData.editor.inspector;
```
arsen2010, line 56: return true or false

editor/inspector/assets/material.js
```diff
@@ -56,6 +70,25 @@ exports.methods = {
         this.dirtyData.uuid = '';
     },
 
+    async updateCustomInspector() {
+        this.$.customPanel.hidden = false;
+        this.$.section.hidden = true;
+        this.$.materialDump.hidden = true;
+        try {
+            const relatePath = normalize((await this.getCustomInspector()).replace('packages://', ''));
```
arsen2010: need to use 'switch case‘ deel with 'packages://' data

editor/inspector/assets/material.js
```diff
@@ -268,24 +301,25 @@ exports.update = async function(assetList, metaList) {
...
 
     this.updateTechniqueOptions();
     this.setDirtyData();
-
-    // optimize calculate speed when edit multiple materials in node mode
-    requestIdleCallback(() => {
-        this.updatePasses();
-    });
+    if (await this.getCustomInspector()) {
+        this.updateCustomInspector();
```
arsen2010: if it is in a 'if case' , use function name ’hasCustomInspector‘  would be better

editor/inspector/assets/material.js
```diff
@@ -56,6 +70,25 @@ exports.methods = {
...
 
+    async updateCustomInspector() {
+        this.$.customPanel.hidden = false;
+        this.$.section.hidden = true;
+        this.$.materialDump.hidden = true;
+        try {
+            const relatePath = normalize((await this.getCustomInspector()).replace('packages://', ''));
+            const name = relatePath.split(sep)[0];
+            const packagePath = Editor.Package.getPackages({ name, enable: true })[0].path;
+            const path = join(packagePath, relatePath.split(name)[1]);
```
arsen2010: before join path, need to judge the package is exist or not.

editor/inspector/assets/material.js
```diff
@@ -80,13 +113,14 @@ exports.methods = {
      * Update the pass data that is finally displayed in the panel
      */
     updatePasses() {
+        this.$.customPanel.hidden = true;
```
arsen2010, line 126: if $.customPanel is exist and its ’src‘ is not empty, need to remove attr ’src‘

editor/inspector/assets/material.js
```diff
@@ -56,6 +70,25 @@ exports.methods = {
...
+            const name = relatePath.split(sep)[0];
+            const packagePath = Editor.Package.getPackages({ name, enable: true })[0].path;
+            const path = join(packagePath, relatePath.split(name)[1]);
+            if (this.$.customPanel.getAttribute('src') !== join(path)) {
+                this.$.customPanel.setAttribute('src', path);
+            }
+            this.$.customPanel.update(this.material, this.assetList, this.metaList);
+        } catch (error) {
+            console.error(error);
+            console.error(Editor.I18n.t('ENGINE.assets.material.fail-to-load-custom-inspector', { effect: this.material.effect }));
```
arsen2010, line 95: error case should inclued packagePath is not exist.

editor/inspector/assets/material.js
chichinohaha: Its return value is now used

editor/inspector/assets/material.js
chichinohaha: There is no need. A try catch is used

editor/inspector/assets/material.js
chichinohaha: The error was expected

arsen2010 (approved):

VisualSJ (approved):

PR #8989, opened 2021-07-15
[Fix] the problem that the pipeline scene data switch scene ia is null

GengineJS:


GengineJS:
native code: https://github.com/cocos-creator/engine-native/pull/3724

PR #9016, opened 2021-07-20
Modify the sharing method of node changeFlag

GengineJS:


GengineJS:
native code: https://github.com/cocos-creator/engine-native/pull/3746

minggo (approved):

PR #9042, opened 2021-07-22
TextureCube: Add isRGBE

VisualSJ:
Re: cocos-creator/3d-tasks#8172

Changelog:
 * TextureCube: Add isRGBE


cocos/core/assets/texture-cube.ts
```diff
@@ -305,6 +309,7 @@ legacyCC.TextureCube = TextureCube;
 
 interface ITextureCubeSerializeData {
     base: string;
+    isRGBE: boolean;
```
pandamicro: rgbe

PR #9066, opened 2021-07-26
Support native nodes to set parent to null

GengineJS:


GengineJS:
native code: https://github.com/cocos-creator/engine-native/pull/3764

PR #9073, opened 2021-07-27
add CCSpriteFrame  public interface   refreshTexture

yeshao2069:
Re:  https://github.com/cocos-creator/engine/issues/6667

https://github.com/cocos-creator/2d-tasks/issues/3436


pandamicro:
This is not appropriate, from the discussion and feedback, we should check whether the Texture is a RenderTexture in `_refreshTexture` and listen to its content update event to refresh the texture (if it does't exist, the event need to be added).

PR #9138, opened 2021-08-05
fix mouse wheel can cause the phone number to negative

wanghaha1991:
Re: cocos-creator/3d-tasks#

Changelog:
 * fix mouse wheel can cause the phone number to negative

### DIFF
--- a/cocos2d/core/components/editbox/WebEditBoxImpl.js
+++ b/cocos2d/core/components/editbox/WebEditBoxImpl.js
@@ -434,6 +434,7 @@ Object.assign(WebEditBoxImpl.prototype, {
         } else if(inputMode === InputMode.PHONE_NUMBER) {
             type = 'number';
             elem.pattern = '[0-9]*';
+            elem.onmousewheel = 'return false;';
         } else if(inputMode === InputMode.URL) {
             type = 'url';
         } else {
--- a/cocos2d/core/components/editbox/WebEditBoxImpl.js
+++ b/cocos2d/core/components/editbox/WebEditBoxImpl.js
@@ -434,7 +434,7 @@ Object.assign(WebEditBoxImpl.prototype, {
         } else if(inputMode === InputMode.PHONE_NUMBER) {
             type = 'number';
             elem.pattern = '[0-9]*';
-            elem.onmousewheel = 'return false;';
+            elem.onmousewheel = function () { return false; };
         } else if(inputMode === InputMode.URL) {
             type = 'url';
         } else {


cocos2d/core/components/editbox/WebEditBoxImpl.js
```diff
@@ -434,6 +434,7 @@ Object.assign(WebEditBoxImpl.prototype, {
         } else if(inputMode === InputMode.PHONE_NUMBER) {
             type = 'number';
             elem.pattern = '[0-9]*';
+            elem.onmousewheel = 'return false;';
```
PPpro:
```suggestion
            elem.onmousewheel = function () { return false; };
```

PPpro: Thanks for feedback
we thought that the eval string is not a good practice, and it doesn't work well on the Chrome on MacOS

maybe it is better to change like this.

wanghaha1991: I have no local compilation environment, so I directly get the input object from the implementation of Cocos Creator 2.4.2:
I'll create a new Editbox on a blank scene and test it as follows:
```
   onLoad: function () {
        this.editBox._impl._elem.setAttribute("onmousewheel","function () { return false; };");
        console.log( this.editBox._impl._elem);
    },
```
chrome report error:
image


Using native HTML5 tests, error as above (missing function name):
```
 <form>
        name: <input type="number" name="number" onmousewheel="function a(){ return false; };"><br>
</form>
```
After add a function name：
```
<form>
        name: <input type="number" name="number" onmousewheel="function a(){ return false; };"><br>
 </form>

 this.editBox._impl._elem.setAttribute("onmousewheel","function a() { return false; };");
```
The mouse wheel effect does not fail

PPpro: ## the difference is that
```
elem.onmousewheel = "function () { return false; };"
```
means we **declare** a nameless function at the mousewheel callback, which has never been called


```
elem.onmousewheel = function () { return false; };
```
means we **call** the namelesss function  at the mousewheel callback, it's kind of like:
```
elem.onmousewheel = "(function () { return false; })();"
```

wanghaha1991: Thank you for your answer. I learned a lot. I am a back-end developer，I just started working on cocos. 😸
I have updated this PR

PPpro (approved):

SantyWang:
Thanks for your contribution again

PR #9154, opened 2021-08-09
fix:scene node cause function transformInverseRTS make an error

zhakesi:
scene node has no position info.

issue link [https://github.com/cocos-creator/3d-tasks/issues/8506](https://github.com/cocos-creator/3d-tasks/issues/8506)
https://github.com/cocos-creator/3d-tasks/issues/7837


mikecoker:
Why was this PR closed? It's still a bug. If the scene node doesn't have _lrot/_lpos/_lscale then don't traverse up to the scene. It makes this method useless.

PR #9179, opened 2021-08-13
Add easing method to quat curve

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * Add easing method to quat curve keyframe to compatible with legacy easing methods


cocos/core/data/utils/asserts.ts
```diff
@@ -41,6 +41,7 @@ export function assertIsNonNullable<T> (expr: T, message?: string): asserts expr
  */
 export function assertIsTrue (expr: unknown, message?: string): asserts expr {
     if (DEBUG && !expr) {
+        debugger;
```
gameall3d, line 45: 这个不需要吧？

gameall3d (changes requested):

cocos/core/data/utils/asserts.ts
shrinktofit: 需要，调试的时候方便点

gameall3d (approved):
LGTM

PR #9181, opened 2021-08-14
Add dt threshold

pandamicro:


pandamicro:
cocos-creator/3d-tasks#7930

PR #9190, opened 2021-08-16
MADD Optimized hemishperic environment lighting shader

Jacksparrow2525:
Simple shader optimization for hemispheric lighting. Quick fix.
Now done in one GPU instruction


pandamicro:
@jk20012001 is this included in #9439

PR #9195, opened 2021-08-17
Diffuse Reflection Irradiance Convolution Map

Jacksparrow2525:


pandamicro:
Duplicated with #9439

PR #9218, opened 2021-08-20
do not shrink array when alloc and fix some logic error

minggo:


YunHsiao (approved):

minggo:
@YunHsiao 再帮忙 review 一下，有改了些逻辑。

YunHsiao (approved):

PR #9263, opened 2021-08-31
Mesh dynamic operation API

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
  <See below>

- Drop the concept that `Mesh` is composed as "structure" + "binary".

- Deprecate the API: `Mesh.struct`, `Mesh.data`, `Mesh._nativeAsset`, `Mesh.reset`, the alternatives of old stuffs are:

  - `Mesh.struct.minPosition: Vec3 | undefined`/`Mesh.struct.maxPosition: Vec3 | undefined` -> `Mesh.minPosition: Vec3`/`Mesh.maxPosition: Vec3`. Min position is initialized to positive infinities, max position is intialized as negative infinities.

  - `Mesh.struct.primitives` and other stuffs -> Take a look at `Mesh.subMeshes`

- Introduce the `SubMesh` class. `SubMesh`s store polygons. A mesh contains zero or more sub meshes. Sub meshes are isolated from each-other. They do not share nothing.(As a comparison, legacy mesh data do).

- To manipulate any vertex vertex, you need to visit the sub mesh and call to the `viewAttribute()` which returns a `VertexAttributeView`. Vertex attribute view objects can proceed with any component of any vertex. If the vertex format is well-known to caller, you may wrap it in `VertexAttributeVec3View` or `VertexAttributeVec4View` to get bettern vector operation experience. It is worth mentioning that vertex attribute view are designed so it feel like JavaScript's typed array buffer view. You might also reference to the unit test in this PR to inspect the usage.

- Sub mesh also supplied the following methods to proceed with polygons:

  - `resize()` - Resize the number of vertices, without touch the vertex layout. Both shrinking and expansion are considered and having definitely behaviours.

  - `rearange()` - Adjust the vertex layout. After this method call, old mesh data is erased.

  - `clone()` - Clone the mesh.

- All attribute operations(for example `Mesh.readAttribute`) are implemented through sub mesh API. I did not mark them as deprecated but I'd suggest to do so.

- `Mesh.flush()` flushes all sub mesh data for rendering:

  - If only vertex attribute data is changed, it just siliently update the backend buffers.

  - Otherwise, for example, the vertex layout or the vertex count is changed, it dispatches a `Mesh.EventType.RENDERING_SUB_MESH_RECONSTITUTED`. `MeshRenderer` listened to this event and then refresh the model info.

### DIFF
--- a/cocos/3d/assets/mesh.ts
+++ b/cocos/3d/assets/mesh.ts
@@ -40,10 +40,13 @@ import { warnID } from '../../core/platform/debug';
 import { RenderingSubMesh } from '../../core/assets';
 import {
     Attribute, Device, Buffer, BufferInfo, AttributeName, BufferUsageBit, Feature, Format,
-    FormatInfos, FormatType, MemoryUsageBit, PrimitiveMode, getTypedArrayConstructor,
+    FormatInfos, FormatType, MemoryUsageBit, PrimitiveMode, getTypedArrayConstructor, FormatInfo,
 } from '../../core/gfx';
-import { Mat4, Quat, Vec3 } from '../../core/math';
-import { Morph, MorphRendering, createMorphRendering } from './morph';
+import { IVec3Like, IVec4Like, Mat4, Quat, Vec3, Vec4 } from '../../core/math';
+import { MorphRendering, createMorphRendering } from './morph';
+import { gfx } from '../../core';
+import { assertIsTrue } from '../../core/data/utils/asserts';
+import { removeAt } from '../../core/utils/array';
 
 function getIndexStrideCtor (stride: number) {
     switch (stride) {
@@ -155,7 +158,43 @@ export declare namespace Mesh {
          * @en The morph information of the mesh
          * @zh 网格的形变数据
          */
-        morph?: Morph;
+        morph?: {
+            /**
+             * Morph data of each sub-mesh.
+             */
+            subMeshMorphs: ({
+                /**
+                 * Attributes to morph.
+                 */
+                attributes: AttributeName[];
+
+                /**
+                 * Targets.
+                 */
+                targets: {
+                    /**
+                     * Displacement of each target attribute.
+                     */
+                    displacements: Mesh.IBufferView[];
+                }[];
+
+                /**
+                 * Initial weights of each target.
+                 */
+                weights?: number[];
+            } | null)[];
+
+            /**
+             * Common initial weights of each sub-mesh.
+             */
+            weights?: number[];
+
+            /**
+             * Name of each target of each sub-mesh morph.
+             * This field is only meaningful if every sub-mesh has the same number of targets.
+             */
+            targetNames?: string[];
+        };
     }
 
     export interface ICreateInfo {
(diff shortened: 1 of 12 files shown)


cocos/3d/assets/mesh.ts
```diff
@@ -155,7 +158,43 @@ export declare namespace Mesh {
          * @en The morph information of the mesh
          * @zh 网格的形变数据
          */
-        morph?: Morph;
+        morph?: {
```
minggo, line 161: It is better to define a class or interface.

cocos/3d/assets/mesh.ts
```diff
@@ -177,12 +216,789 @@ const v3_1 = new Vec3();
...
+        );
+
+        return renderingSubMesh;
+    }
+
+    /**
+     * Flush vertices for rendering.
+     * You should ensure the vertex's layout and size are not changed your own.
+     */
+    public [flushTag] (renderingSubMesh: RenderingSubMesh) {
```
minggo, line 600: What's `[flushTag]` ？

cocos/3d/assets/mesh.ts
```diff
@@ -177,12 +216,789 @@ const v3_1 = new Vec3();
...
+    }
+
+    private _primitiveMode: gfx.PrimitiveMode = gfx.PrimitiveMode.TRIANGLE_LIST;
+    private _indices: Uint8Array | Uint16Array | Uint32Array | null = null;
+    private _streams: Array<{
+        data: Uint8Array;
+        stride: number;
+    }> = [];
+    private _vertexCount = 0;
+    private _attributes: Record<string, {
```
minggo, line 641: It is better to define a class or interface.

zxx43 (approved):

cocos/3d/assets/mesh.ts
shrinktofit: It denotes a private name

SantyWang:
完整看了一遍，现有 API 设计还是对开发者太不友好了。如果是引擎内部使用，我觉得 OK，但是对外面的开发者学习成本太高了。他必须理解 sub mesh, attribute, format 这些概念，学习成本太高。目前的用法，除非高级用户，否则使用起来还是困难

Unreal 中的操作方式，https://docs.unrealengine.com/4.27/en-US/API/Plugins/ProceduralMeshComponent/UProceduralMeshComponent/

Unity 中的操作方式
https://docs.unity3d.com/cn/current/Manual/Example-CreatingaBillboardPlane.html

shrinktofit:
> 完整看了一遍，现有 API 设计还是对开发者太不友好了。如果是引擎内部使用，我觉得 OK，但是对外面的开发者学习成本太高了。他必须理解 sub mesh, attribute, format 这些概念，学习成本太高。目前的用法，除非高级用户，否则使用起来还是困难
> 
> Unreal 中的操作方式，https://docs.unrealengine.com/4.27/en-US/API/Plugins/ProceduralMeshComponent/UProceduralMeshComponent/
> 
> Unity 中的操作方式 https://docs.unity3d.com/cn/current/Manual/Example-CreatingaBillboardPlane.html

用户既然要操作网格，肯定是要理解这些东西的？

SantyWang:
dynamic operation of mesh already supported

PR #9314, opened 2021-09-10
Fix trail create size 0 buffer crash

zxx43:
Re: cocos-creator/3d-tasks#

Changelog:
 * 

fix issue https://github.com/cocos-creator/3d-tasks/issues/9092


pandamicro:
It's too hack to do such wasted calculation here, if trail vertex count is zero, please skip the trail calculation and rendering, just like trail module is disabled or any up level control of  it.

PR #9318, opened 2021-09-11
2 subtle improvements

jcyuan:
2 changes, and all they are subtle improvements:

1, add an `out` argument for `generateRenderArea` method, because this method is called a lot of times in a single frame tick (for example in `Render` method), this wastes memory, so i think an `out` parameter is supposed to be here..

2, do not create temp function when set `camera.targetTexture` as null, in this situation this temp function is redundant here.

### DIFF
--- a/cocos/core/pipeline/render-pipeline.ts
+++ b/cocos/core/pipeline/render-pipeline.ts
@@ -171,8 +171,8 @@ export abstract class RenderPipeline extends Asset {
      * @param camera the camera
      * @returns
      */
-    public generateRenderArea (camera: Camera): Rect {
-        const res = new Rect();
+    public generateRenderArea (camera: Camera, out?: Rect): Rect {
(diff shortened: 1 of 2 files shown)


cocos/core/pipeline/render-pipeline.ts
```diff
@@ -171,8 +171,8 @@ export abstract class RenderPipeline extends Asset {
      * @param camera the camera
      * @returns
      */
-    public generateRenderArea (camera: Camera): Rect {
-        const res = new Rect();
+    public generateRenderArea (camera: Camera, out?: Rect): Rect {
+        const res = out || new Rect();
```
minggo, line 175: 为什么这么改呢？

minggo (approved):

pandamicro (approved):

pandamicro:
@jcyuan Thank you for your contribution

PR #9324, opened 2021-09-13
- fix spine render is not batching

Jackie0110:
Re: cocos-creator/3d-tasks#

Changelog:
 *


Jackie0110:
- fix spine render is not batching in web due to comparing two instances of material is always false

cocos/2d/renderer/batcher-2d.ts
```diff
@@ -389,7 +389,7 @@ export class Batcher2D {
         const blendTargetHash = renderComp.blendHash;
         const depthStencilStateStage = renderComp.stencilStage;
 
-        if (this._currScene !== renderScene || this._currLayer !== comp.node.layer || this._currMaterial !== mat
+        if (this._currScene !== renderScene || this._currLayer !== comp.node.layer || this._currMaterial.hash !== mat?.hash
```
YunHsiao, line 383: hash 只关心 layout，不关心数据（UBO，贴图等）确定这样改能满足需求吗？

Jackie0110: Sorry but I do not know Chinese, I had to google translate: Hash only cares about the layout, not the data (UBO, textures, etc.). Is this change to meet the demand?
=> Because each spine use material instances, not using shared material as sprite, so compare 2 instances of material always return false, it break batching. UBO and data is still in another condition ||

Jackie0110: here is the full condition

Jackie0110: @YunHsiao  any suggestion to check problem or to improve this ?

YunHsiao: Sorry I mistook you for one of the engine team guys, thanks for the contribution btw.

As for the suggested change, here's the thing: right now the material hash does not account for any data stored inside it (textures, UBOs, which would require substantial hashing calculations per matetial update)

So in the case of the spine system, the problem is not with this branching condition here, but rather the instantiation behavior inside spine component. This is a known issue internally and we will be addressing this properly in the coming refactorings. 

However if you need to optimize this particular behavior right now for you own project, the suggested change is a valid approach, with one catch: custom materials with additional data (i.e. textures, UBOs) will not be accounted for here and will lead to incorrect rendering results.

So for this PR I'd have to say no, we can't make this change as is for now, but I'll let the spine guy know and work on the problem asap. Would that work for you?

Jackie0110: OK, thank you

PR #9423, opened 2021-09-29
V340 fix node with UI opacity zero still render

Jackie0110:
Re: cocos-creator/3d-tasks#

Changelog:
 *

### DIFF
--- a/cocos/2d/framework/renderable-2d.ts
+++ b/cocos/2d/framework/renderable-2d.ts
@@ -453,14 +453,10 @@ export class Renderable2D extends RenderableComponent {
     protected _postCanRender () {}
 
     protected _updateColor () {
-        // Need update rendFlag when opacity changes from 0 to !0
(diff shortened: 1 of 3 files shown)


Jackie0110:
Step to reproduce bug:
1. Make an empty scene
2. Add a sprite
3. Add UIOpacity component = 0
4. Notice that draw call is still 2 with empty black scene: ( can use spector.js and see that the image is rendered)

This usually happen when we set parent node UIOpacity to 0 to disable its children from rendering but draw call is still count.

LinYunMo:
Thank you for your PR, I noticed that you closed it, so I submitted a PR to fix this problem, and of course your PR is also valid, thanks a lot @Jackie0110

Jackie0110:
@LinYunMo Yes, thank you, I noticed that my source code was out-dated so I just closed PR to check with new code. Luckily you saw it :)

PR #9425, opened 2021-09-29
replace utils function ,update "sortProp" function

chichinohaha:
Re: cocos-creator/3d-tasks#

Changelog:

replace utils function
-  "getCustomPropElements" -> "updateCustomPropElements"
this function is use to render custom node

update  "sortProp" function
- item may be null
 *

### DIFF
--- a/editor/inspector/utils/prop.js
+++ b/editor/inspector/utils/prop.js
@@ -10,16 +10,18 @@ exports.sortProp = function(propMap) {
 
     Object.keys(propMap).forEach((key) => {
         const item = propMap[key];
-        if ('displayOrder' in item) {
-            orderList.push({
-                key,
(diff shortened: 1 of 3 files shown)


editor/inspector/utils/prop.js
```diff
@@ -10,16 +10,18 @@ exports.sortProp = function(propMap) {
...
-                key,
-                dump: item,
-            });
-        } else {
-            normalList.push({
-                key,
-                dump: item,
-            });
+        if (item != null) {
+            if ('displayOrder' in item) {
```
arsen2010, line 14: how about `if(!item)`

arsen2010 (approved):

VisualSJ (approved):

PR #9426, opened 2021-09-29
Change the content of the panel, fix the multi-selection situation, remove the inputflag

chichinohaha:
Change panel content
-The ui-prop that uses this tag sets the disabled of the content-slot to the marked dump.value
- add particle-system show bounds button
- add Regenerate bounding box button
-When culling is not turned on, disable the elements in the enableCulling
catergory
    image
    image
Fix multiple selection
-showBounds adapts to the case of multiple selections, when changing, it sends message for each component
-The Regenerate bounding box button will be set to disabled according to the selected component enableCulling
-Fix the ui-checkbox missing invalid judgement in ui-section header
-Fix the problem that ui-prop is not set dump.values to each component when multiple-selection

Remove inputflag
-No panel uses this mark, remove the related code

Re: cocos-creator/3d-tasks#

Changelog:
 *

### DIFF
--- a/editor/inspector/components/particle-system.js
+++ b/editor/inspector/components/particle-system.js
@@ -644,4 +644,3 @@ exports.style = /* css */`
         margin-left: 10px;
     }
 `;
-;
--- a/editor/inspector/components/particle-system.js
+++ b/editor/inspector/components/particle-system.js
@@ -643,5 +643,5 @@ exports.style = /* css */`
     .particle-system-component > .content >.indent {
         margin-left: 10px;
     }
-`
+`;
 ;
--- a/editor/inspector/components/particle-system.js
+++ b/editor/inspector/components/particle-system.js
@@ -2,7 +2,7 @@
 
 const propUtils = require('../utils/prop');
 
-exports.template = `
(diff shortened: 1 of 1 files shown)


editor/inspector/components/particle-system.js
```diff
@@ -545,3 +639,9 @@ exports.update = function(dump) {
         }
     }
 };
+exports.style = /* css */`
+    .particle-system-component > .content >.indent {
+        margin-left: 10px;
+    }
+`
+;
```
VisualSJ:
```suggestion

```

editor/inspector/components/particle-system.js
```diff
@@ -545,3 +639,9 @@ exports.update = function(dump) {
         }
     }
 };
+exports.style = /* css */`
+    .particle-system-component > .content >.indent {
+        margin-left: 10px;
+    }
+`
```
VisualSJ:
```suggestion
`;
```

VisualSJ (approved):

arsen2010 (approved):
the code logic is complex, you need to make the testing succeful.

PR #9469, opened 2021-10-12
Fix possible box2d shape internal state error when it's node is not active

biliskener:
Fix possible box2d shape internal state error when it's node is not active


cocos/physics-2d/builtin/shapes/shape-2d.ts
```diff
@@ -64,6 +68,7 @@ export class BuiltinShape2D implements IBaseShape {
     }
 
     onGroupChanged () {
+        if(!this._activated) return;
```
pandamicro, line 71: I think It will break group sync between the component and physics shape, so I don't understand the problem here

biliskener:
I test it again, no problems found.

PR #9534, opened 2021-10-25
fix camera pre-rotation adaptation

YunHsiao:
Re: cocos-creator/3d-tasks#9570

`camera.width/height` is oriented.
`window.width/height` is pre-rotated. (i.e. not oriented, always in identity/portrait mode)

Changelog:
 * fix up camera pre-rotation adaptations

### DIFF
--- a/cocos/core/renderer/core/render-window.ts
+++ b/cocos/core/renderer/core/render-window.ts
@@ -53,16 +53,16 @@ const orientationMap: Record<Orientation, SurfaceTransform> = {
  */
 export class RenderWindow {
     /**
-     * @en Get window width.
-     * @zh 窗口宽度。
+     * @en Get window width. Pre-rotated (i.e. always in identity/portrait mode) if supported.
+     * @zh 窗口宽度。如果支持交换链预变换，返回值始终处于单位旋转（竖屏）坐标系下。
      */
     get width (): number {
         return this._width;
     }
 
     /**
-     * @en Get window height.
-     * @zh 窗口高度。
+     * @en Get window height. Pre-rotated (i.e. always in identity/portrait mode) if supported.
+     * @zh 窗口高度。如果支持交换链预变换，返回值始终处于单位旋转（竖屏）坐标系下。
      */
     get height (): number {
         return this._height;
@@ -219,9 +219,7 @@ export class RenderWindow {
         }
 
         for (const camera of this._cameras) {
-            if (camera.isWindowSize) {
-                camera.resize(width, height);
-            }
+            camera.resize(width, height);
         }
     }
 
--- a/cocos/core/renderer/core/render-window.ts
+++ b/cocos/core/renderer/core/render-window.ts
@@ -53,16 +53,18 @@ const orientationMap: Record<Orientation, SurfaceTransform> = {
  */
 export class RenderWindow {
     /**
-     * @en Get window width. Pre-rotated (i.e. always in identity/portrait mode) if supported.
-     * @zh 窗口宽度。如果支持交换链预变换，返回值始终处于单位旋转（竖屏）坐标系下。
+     * @en Get window width. Pre-rotated (i.e. rotationally invariant, always in identity/portrait mode) if possible.
+     * If you want to get oriented size instead, you should use [[Camera.width]] which corresponds to the current screen rotation.
+     * @zh 获取窗口宽度。如果支持交换链预变换，返回值将始终处于单位旋转（竖屏）坐标系下。如果需要获取旋转后的尺寸，请使用 [[Camera.width]]。
      */
     get width (): number {
         return this._width;
     }
 
     /**
-     * @en Get window height. Pre-rotated (i.e. always in identity/portrait mode) if supported.
-     * @zh 窗口高度。如果支持交换链预变换，返回值始终处于单位旋转（竖屏）坐标系下。
+     * @en Get window height. Pre-rotated (i.e. rotationally invariant, always in identity/portrait mode) if possible.
+     * If you want to get oriented size instead, you should use [[Camera.width]] which corresponds to the current screen rotation.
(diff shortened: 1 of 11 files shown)


PPpro:
Do we have any interface to query the window size when the camera is rotated ?

cocos/core/renderer/scene/camera.ts
```diff
@@ -203,6 +203,18 @@ export class Camera {
         }
     }
 
+    private _updateAspect (oriented = true) {
+        this._aspect = (this.window.width * this._viewport.width) / (this.window.height * this._viewport.height);
+        if (oriented) {
+            const swapchain = this.window.swapchain;
+            const orientation = swapchain && swapchain.surfaceTransform || SurfaceTransform.IDENTITY;
+            if (orientation % 2) this._aspect = 1 / this._aspect;
```
pandamicro, line 213: Should put more explanation here, or make the check more obvious, it may take a while to understand what's happening here. We should make our engine more understandable, especially for external contributors

cocos/core/renderer/scene/camera.ts
```diff
@@ -456,10 +461,12 @@ export class Camera {
         return this._clearColor as IVec4Like;
     }
 
+    // Pre-rotated (i.e. always in identity/portrait mode) if supported.
```
pandamicro: I think it's only valid in vulkan backend ?

YunHsiao:
As mentioned, both `window.width/height` and `camera.width/height` could be used to query the size, depending on the specific scenarios.

cocos/core/renderer/core/render-window.ts
```diff
@@ -53,16 +53,16 @@ const orientationMap: Record<Orientation, SurfaceTransform> = {
...
      */
     get width (): number {
         return this._width;
     }
 
     /**
-     * @en Get window height.
-     * @zh 窗口高度。
+     * @en Get window height. Pre-rotated (i.e. always in identity/portrait mode) if supported.
+     * @zh 窗口高度。如果支持交换链预变换，返回值始终处于单位旋转（竖屏）坐标系下。
```
pandamicro: If the render window's size could be inversed, where we can get a stable size without inversion in any circumstances

cocos/core/renderer/scene/camera.ts
```diff
@@ -456,10 +461,12 @@ export class Camera {
...
     get viewport () {
         return this._viewport;
     }
 
+    // Oriented
```
pandamicro: The getter and setter of viewport seems different, on pre-rotated, one oriented ?

cocos/core/renderer/scene/camera.ts
YunHsiao: that's the 'if supported' part

cocos/core/renderer/core/render-window.ts
YunHsiao: Again let me clear some terms here:
`camera.width/height` is **Oriented**, means they are rotationally variant. (local to the current screen rotation)
`window.width/height` is **Pre-rotated**, means they are rotationally invariant. (always in portrait mode)

cocos/core/renderer/scene/camera.ts
YunHsiao: yes, the setter is used in camera component to pass in the user-specified viewport info.
the getter is used in pipeline to calculate render area, viewport, etc. which needs to be pre-rotated

cocos/core/renderer/scene/camera.ts
pandamicro: I guess it should be `if necessary`

cocos/core/renderer/core/render-window.ts
pandamicro: Missed that comment, but can you put it somewhere in the engine code comment or api comment ?
It also need to be declared in usage doc

cocos/core/renderer/scene/camera.ts
pandamicro: It's a very strange behavior, getter setter is the same property, we should keep the meaning and data consistent, if render pipeline need another data form, we should provide via another API

cocos/core/renderer/core/render-window.ts
YunHsiao: I thought the comments on render window here are enough since the oriented size is what people assume by default, so in that regard the render window here is the only special part, and need mentioning of the pre-rotation stuff. IMHO everywhere else should not even mention this to avoid further confusion over it.

cocos/core/renderer/scene/camera.ts
YunHsiao: done

cocos/core/renderer/scene/camera.ts
YunHsiao: It's not a 'necessary' situation since pre-rotation is completely optional, I think `if possible` would be better.

cocos/core/renderer/scene/camera.ts
YunHsiao: This pre-rotation stuff is supposed to be transparent, so I would prefer we don't expose this detail at the interface level. But anyway, open for discussion.

cocos/core/renderer/core/render-window.ts
pandamicro: How about this

```
@en Get window width. It's pre-rotated if possible, which means it's rotationally invariant and always in identity/portrait mode. 
If you want to get oriented size instead, you should use [[Camera.width]] which corresponds to the current screen rotation.
```

I hope it can be as clear as possible

cocos/core/renderer/scene/camera.ts
pandamicro: Exactly, the inconsistence between getter and setter makes the pre rotation stuff exposed to user, once they get viewport they will notice it in Vulkan. I suggest two solutions

1. `viewport` is readonly with a getter for pre-rotated data, and provide `setViewport` which accepts oriented parameter, mark the data difference clearly in API doc
2. `viewport` has getter setter with consistent rotated data, we provide `getDeviceViewport` for internal usage with pre-rotated data.

pandamicro (approved):

pandamicro (approved):

PR #9544, opened 2021-10-26
cc.find() does not return scene, and be compatible with path which consists of empty string

Zhmz:
Re: cocos-creator/3d-tasks#9475

Changelog:
 * 处理了getChildByPath里不兼容空路径的问题
 * 增加了find.test.ts里的单元测试用例

### DIFF
--- a/cocos/core/scene-graph/base-node.ts
+++ b/cocos/core/scene-graph/base-node.ts
@@ -518,7 +518,7 @@ export class BaseNode extends CCObject implements ISchedulable {
         let lastNode: this = this;
         for (let i = 0; i < segments.length; ++i) {
             const segment = segments[i];
-            if (segment.length === 0) {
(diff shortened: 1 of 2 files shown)


cocos/core/scene-graph/base-node.ts
```diff
@@ -518,7 +518,7 @@ export class BaseNode extends CCObject implements ISchedulable {
         let lastNode: this = this;
         for (let i = 0; i < segments.length; ++i) {
             const segment = segments[i];
-            if (segment.length === 0) {
+            if (i === 0 && segment.length === 0) {
```
pandamicro, line 521: Why this could happen ? If segments.length equals 0, then the for loop won't execute at all

Zhmz: 这里是segment.length，即当前一段字符串的长度，不是字符串数组segments的长度

PR #9554, opened 2021-10-28
Added _isDeserializing property in legacyCC.game to determine object is deserializing.

biliskener:
Added _isDeserializing property in legacyCC.game to determineobject is deserializing.
I really need this to determine whether the object is in deserializing state for singleton class.

### DIFF
--- a/cocos/core/data/deserialize-dynamic.ts
+++ b/cocos/core/data/deserialize-dynamic.ts
@@ -822,8 +822,10 @@ export function deserializeDynamic (data: SerializedData | CCON, details: Detail
     const deserializer = _Deserializer.pool.get(details, classFinder, reportMissingClass, customEnv, ignoreEditorOnly);
 
     legacyCC.game._isCloning = true;
+    legacyCC.game._isDeserializing = true;
     const res = deserializer.deserialize(data);
     legacyCC.game._isCloning = false;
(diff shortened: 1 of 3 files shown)


SantyWang:
Thanks for your contribution to Creator, I wonder why you need this state?

biliskener:
> Thanks for your contribution to Creator, I wonder why you need this state?

When I use a component instance as a singleton, I need to check the instance whether is in deserializing state or instantiating state.
The component is exists in a prefab. So the instance objects should be two after instantiated.

See below codes:

```
@ccclass('CountDownWidget')
export class CountDownWidget extends Component {
    private static gInstance: CountDownWidget = null as any as CountDownWidget;
    public static get INSTANCE(): CountDownWidget {
        return this.gInstance;
    }

    constructor() {
        super();
        if (!game._isDeserializing) {
            assert(CountDownWidget.gInstance == null);
            CountDownWidget.gInstance = this;
        }
    }

    public _onPreDestroy(): void {
        super._onPreDestroy();
        assert(CountDownWidget.gInstance == this);
        CountDownWidget.gInstance = null as any as CountDownWidget;
    }
}

```

SantyWang:
Thanks for your contribution again. but sorry，I don't think this flag is a common requirement. maybe you can custom engine in your own repo.

PR #9558, opened 2021-10-28
material: linear space input

YunHsiao:
Re: cocos-creator/3d-tasks#

Changelog:
 * support parameter auto conversion to linear space

### DIFF
--- a/editor/assets/effects/builtin-standard.effect
+++ b/editor/assets/effects/builtin-standard.effect
@@ -7,26 +7,23 @@ CCEffect %{
     - vert: standard-vs
       frag: standard-fs
       properties: &props
-        tilingOffset:   { value: [1.0, 1.0, 0.0, 0.0] }
-        mainColor:      { value: [1.0, 1.0, 1.0, 1.0], target: albedo, editor: { displayName: Albedo, type: color } }
-        albedoScale:    { value: [1.0, 1.0, 1.0], target: albedoScaleAndCutoff.xyz }
-        alphaThreshold: { value: 0.5, target: albedoScaleAndCutoff.w, editor: { parent: USE_ALPHA_TEST, slide: true, range: [0, 1.0], step: 0.001 } }
-        occlusion:      { value: 1.0, target: pbrParams.x, editor: { slide: true, range: [0, 1.0], step: 0.001 }  }
-        roughness:      { value: 0.8, target: pbrParams.y, editor: { slide: true, range: [0, 1.0], step: 0.001 }  }
-        metallic:       { value: 0.6, target: pbrParams.z, editor: { slide: true, range: [0, 1.0], step: 0.001 } }
-        SpecularIntensity:  { value: 0.5, target: pbrParams.w, editor: {  slide: true, range: [0.0, 1.0], step: 0.001 } }
-        normalStrenth:  { value: 1.0, target: miscParams.x, editor: { parent: USE_NORMAL_MAP, slide: true, range: [0, 1.0], step: 0.001 } }
-        emissive:       { value: [0.0, 0.0, 0.0, 1.0], editor: { type: color } }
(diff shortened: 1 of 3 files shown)


jk20012001 (approved):

editor/assets/effects/builtin-standard.effect
```diff
@@ -7,26 +7,23 @@ CCEffect %{
...
-        emissiveScale:  { value: [1.0, 1.0, 1.0], target: emissiveScaleParam.xyz }
-        mainTexture:    { value: grey, target: albedoMap, editor: { displayName: AlbedoMap } }
-        normalMap:      { value: normal       }
-        pbrMap:         { value: grey         }
-        metallicRoughnessMap: { value: grey   }
-        occlusionMap:   { value: white        }
-        emissiveMap:    { value: grey         }
-      migrations:
-        properties:
-          pbrParams: { removeImmediately: true }
```
pandamicro, line 29: Why remove this migration ? User could upgrade from a very low version

editor/assets/effects/builtin-standard.effect
```diff
@@ -220,11 +202,10 @@ CCProgram standard-fs %{
     #pragma define EMISSIVE_UV options([v_uv, v_uv1])
   #endif
 
-  #define OCCLUSION_CHANNEL r
-  #define ROUGHNESS_CHANNEL g
-  #define METALLIC_CHANNEL  b
-  #define SPECULARINTENSITY_CHANNEL w
```
pandamicro, line 226: Will this modification break user effects ?

editor/assets/effects/builtin-standard.effect
```diff
@@ -7,26 +7,23 @@ CCEffect %{
...
+        albedoScale:          { value: [1.0, 1.0, 1.0], target: albedoScaleAndCutoff.xyz }
+        alphaThreshold:       { value: 0.5, target: albedoScaleAndCutoff.w, editor: { parent: USE_ALPHA_TEST, slide: true, range: [0, 1.0], step: 0.001 } }
+        occlusion:            { value: 1.0, target: pbrParams.x, editor: { slide: true, range: [0, 1.0], step: 0.001 } }
+        roughness:            { value: 0.8, target: pbrParams.y, editor: { slide: true, range: [0, 1.0], step: 0.001 } }
+        metallic:             { value: 0.6, target: pbrParams.z, editor: { slide: true, range: [0, 1.0], step: 0.001 } }
+        SpecularIntensity:    { value: 0.5, target: pbrParams.w, editor: { slide: true, range: [0.0, 1.0], step: 0.001 } }
+        emissive:             { value: [0.0, 0.0, 0.0, 1.0], linear: true, editor: { type: color } }
+        emissiveScale:        { value: [1.0, 1.0, 1.0], target: emissiveScaleParam.xyz }
+        normalStrenth:        { value: 1.0, target: emissiveScaleParam.w, editor: { parent: USE_NORMAL_MAP, slide: true, range: [0, 1.0], step: 0.001 } }
```
pandamicro, line 20: Changes to normal strength could break user effect setups, if it's been modified by user, right ?

YunHsiao: nope, this is completely an internal change, and shouldn't cause any compatibility problems.

editor/assets/effects/builtin-standard.effect
YunHsiao: no, regarding compatibility we should only worry about interface changes in the header files, never the effect file changes, since everything will be copy-pasted into their own instances.

editor/assets/effects/builtin-standard.effect
YunHsiao: I thought this has been around for long enough, for something that is only briefly introduced in one or two ancient versions.

pandamicro (approved):

PR #9559, opened 2021-10-28
Solve the problem that the properties of the Object below the ObjectGroup in TiledMap are incorrectly mixed into the property list of the ObjectGroup.

biliskener:
Solve the problem that the properties of the Object below the ObjectGroup in TiledMap are incorrectly mixed into the property list of the ObjectGroup. The function getElementsByTagName() searches for child nodes in a deep way.


SantyWang (approved):

SantyWang:
Thanks for contribution.

PR #9595, opened 2021-11-02
gfx: import from pal/system-info directly

YunHsiao:
Re: cocos-creator/3d-tasks#

Changelog:
 * remove dependency on legacyCC

### DIFF
--- a/@types/pal/system-info.d.ts
+++ b/@types/pal/system-info.d.ts
@@ -1,11 +1,19 @@
 // TODO: some interface need to be designed in a module called 'Application' and 'Network'
 declare module 'pal/system-info' {
+    type Feature = import('../../pal/system-info/enum-type').Feature;
+    type NetworkType = import('../../pal/system-info/enum-type').NetworkType;
+    type Platform = import('../../pal/system-info/enum-type').Platform;
+    type Language = import('../../pal/system-info/enum-type').Language;
+    type OS = import('../../pal/system-info/enum-type').OS;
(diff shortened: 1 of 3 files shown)


@types/pal/system-info.d.ts
```diff
@@ -1,11 +1,19 @@
 // TODO: some interface need to be designed in a module called 'Application' and 'Network'
 declare module 'pal/system-info' {
+    type Feature = import('../../pal/system-info/enum-type').Feature;
+    type NetworkType = import('../../pal/system-info/enum-type').NetworkType;
+    type Platform = import('../../pal/system-info/enum-type').Platform;
+    type Language = import('../../pal/system-info/enum-type').Language;
+    type OS = import('../../pal/system-info/enum-type').OS;
+    type BrowserType = import('../../pal/system-info/enum-type').BrowserType;
+    type PalSystemEvent = import('../../pal/system-info/enum-type').PalSystemEvent;
```
PPpro, line 9: please import from `'pal/system-info/enum-type'` instead of `'../../pal/system-info/enum-type'`
This may block `npm install`  workflow in editor like https://github.com/cocos-creator/engine/pull/9315

YunHsiao: The gfx-test-case module's compilation relies on this import-by-relative-path behavior, and the enum-type module does not depend on any other modules so I guess it's fine to do this anyways.

PPpro (approved):

SantyWang (approved):

PR #9605, opened 2021-11-03
pass owns instanced/batched buffer

YunHsiao:
Re: cocos-creator/3d-tasks#9793

https://github.com/cocos-creator/3d-tasks/issues/9885

https://github.com/cocos-creator/engine/issues/9497

Changelog:
 * pass now owns instanced/batched buffers

### DIFF
--- a/cocos/core/renderer/core/pass.ts
+++ b/cocos/core/renderer/core/pass.ts
@@ -47,6 +47,8 @@ import { RenderPassStage, RenderPriority } from '../../pipeline/define';
 import { NativePass } from '../scene/native-scene';
 import { errorID } from '../../platform/debug';
 import { PassHandle, PassView, NULL_HANDLE, PassPool } from './memory-pools';
+import { InstancedBuffer } from '../../pipeline/instanced-buffer';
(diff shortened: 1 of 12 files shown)


cocos/core/renderer/core/pass.ts
```diff
@@ -381,6 +385,14 @@ export class Pass {
         }
     }
 
+    public getInstancedBuffer (extraKey = 0) {
```
pandamicro, line 388: getInstancedBuffer(0) and getInstancedBuffer() will collide, are you sure it's ok ?

editor/assets/primitives.fbx.meta
```diff
@@ -1,11 +1,9 @@
 {
-  "ver": "2.1.9",
```
pandamicro, line 2: Shouldn't mix in primitives change

pandamicro (changes requested):

cocos/core/renderer/core/pass.ts
YunHsiao: It's the same behavior as before

pandamicro (approved):

PR #9606, opened 2021-11-03
[temp hack] use simple flag to pause splashScreen

timlyeee:
Build with
* https://github.com/cocos-creator/engine-native/pull/4008 
To support windows platform if window goes to background, and the surface should be destroyed, while the render frame should pause.

Just a temp solution

### DIFF
--- a/cocos/core/splash-screen.ts
+++ b/cocos/core/splash-screen.ts
@@ -98,11 +98,11 @@ export class SplashScreen {
     private watermarkMat!: Material;
     private watermarkTexture!: Texture;
 
-    public pause(){
+    public pauseRendering(){
         this.isPause = true;
     }
 
-    public resume(){
+    public resumeRendering(){
         this.isPause = false;
     }
 
--- a/cocos/core/splash-screen.ts
+++ b/cocos/core/splash-screen.ts
@@ -73,7 +73,7 @@ export class SplashScreen {
     private callBack: (() => void) | null = null;
     private cancelAnimate = false;
     private startTime = -1;
-    private needRender = true;
+    private isPause = false;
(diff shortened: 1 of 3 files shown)


timlyeee:
temp hack for splash screen, which will pause the procedure if it's minimized on Windows, otherwise it will resume game as the tick will start to run, which is inpredictable.
Splash Screen will restructure in 3.5.

cocos/core/splash-screen.ts
```diff
@@ -73,6 +73,7 @@ export class SplashScreen {
     private callBack: (() => void) | null = null;
     private cancelAnimate = false;
     private startTime = -1;
+    private isPause = false;
```
YunHsiao, line 76: pauseRendering 更准确一些

cocos/core/splash-screen.ts
```diff
@@ -97,6 +98,14 @@ export class SplashScreen {
     private watermarkMat!: Material;
     private watermarkTexture!: Texture;
 
+    public pause(){
+        this.isPause = true;
+    }
+
+    public resume(){
```
YunHsiao: 同上，resumeRendering

cocos/core/splash-screen.ts
timlyeee: 1

cocos/core/splash-screen.ts
timlyeee: 1

timlyeee:
该部分修改将在3.5统一插屏界面后被重构
* https://github.com/cocos-creator/engine/pull/9606

minggo (approved):

PR #9613, opened 2021-11-03
Refactor particle system render culling

chichinohaha:
Re: cocos-creator/3d-tasks#
企业微信截图_16359400206356
Changelog:
 * enableCulling to render Culling

reply https://github.com/cocos-creator/engine/pull/9611


pandamicro (approved):
Still need data culling

PR #9614, opened 2021-11-03
Fix color.fromHex function  can't set alpha  to zero

yanjifa:
Re: cocos-creator/3d-tasks#

Changelog:
 * Fix color.fromHex function  can't set alpha  to zero
 * 修复 color.fromHex 函数不能设置透明度为 0
 * 修复 color.toHex 函数注释错误


pandamicro (approved):

pandamicro:
Thank you for your contribution

PR #9643, opened 2021-11-09
fix #9639

### LINKED ISSUE #9639: 【3.4.0预览版】PhysicsGroup枚举值未在编辑器中显示，但运行时正常。
版本：v3.4.0预览版

在代码中写以下代码（2D box2d）：
image
但在编辑器中不会显示枚举值：
image
但在运行时值是正确的：
image

使用3D模块的`PhysicsSystem.PhysicsGroup`能正确显示：
image

所以我去看了一下3D物理初始化的源码，发现是在`director`的`INIT`事件时初始化的`PhysicsGroup`枚举值。
2D的`PhysicsSystem`的源码看上去感觉又没有什么问题，不过为了测试，所以我还是在本地使用自定义引擎将2D的`PhysicsGroup`以3D同样的方式进行初始化测试，但是发现IDE上依然不会显示，因为涉及到IDE的初始化逻辑，所以具体原因就找不出来了，也就没有能力提交PR修复了。
希望官方能修正一下这个问题，十分感谢。

谢谢。

jcyuan:
修正2D物理组选项在IDE中不显示的问题
#9639

### DIFF
--- a/cocos/physics-2d/framework/physics-internal-types.ts
+++ b/cocos/physics-2d/framework/physics-internal-types.ts
@@ -1,4 +1,4 @@
 export declare class DelayEvent {
-    target: Object;
-    func: Function;
+    target: unknown;
+    func: () => void;
 }
--- a/cocos/physics-2d/framework/physics-system.ts
+++ b/cocos/physics-2d/framework/physics-system.ts
@@ -3,20 +3,23 @@
  * @module physics2d
  */
 
-import { EDITOR, DEBUG } from 'internal:constants';
-import { System, Vec2, director, Director, game, error, IVec2Like, Rect, Eventify, Enum } from '../../core';
-import { IPhysicsWorld } from '../spec/i-physics-world';
-import { createPhysicsWorld } from './instance';
-import { physicsEngineId } from './physics-selector';
-import { DelayEvent } from './physics-internal-types';
-import { IPhysicsConfig, ICollisionMatrix } from '../../physics/framework/physics-config';
-import { CollisionMatrix } from '../../physics/framework/collision-matrix';
-import { ERaycast2DType, RaycastResult2D, PHYSICS_2D_PTM_RATIO, PhysicsGroup } from './physics-types';
-import { Collider2D } from './components/colliders/collider-2d';
-
-let instance: PhysicsSystem2D | null = null;
-
-export class PhysicsSystem2D extends Eventify(System) {
+ import { EDITOR, DEBUG } from 'internal:constants';
+ import { System, Vec2, director, Director, game, error, IVec2Like, Rect, Eventify, Enum } from '../../core';
+ import { IPhysicsWorld } from '../spec/i-physics-world';
+ import { createPhysicsWorld } from './instance';
+ import { physicsEngineId } from './physics-selector';
+ import { DelayEvent } from './physics-internal-types';
+ import { IPhysicsConfig, ICollisionMatrix } from '../../physics/framework/physics-config';
+ import { CollisionMatrix } from '../../physics/framework/collision-matrix';
+ import { ERaycast2DType, RaycastResult2D, PHYSICS_2D_PTM_RATIO, PhysicsGroup } from './physics-types';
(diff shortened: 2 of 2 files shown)


cocos/physics-2d/framework/physics-internal-types.ts
```diff
@@ -1,4 +1,4 @@
 export declare class DelayEvent {
-    target: Object;
-    func: Function;
+    target: unknown;
```
pandamicro, line 2: Can you explain the reason to modify this to unknown

cocos/physics-2d/framework/physics-system.ts
```diff
@@ -354,6 +357,7 @@ director.once(Director.EVENT_INIT, () => {
 
 function initPhysicsSystem () {
     if (!PhysicsSystem2D.PHYSICS_NONE && !EDITOR) {
-        director.registerSystem(PhysicsSystem2D.ID, PhysicsSystem2D.instance, System.Priority.LOW);
+        const sys = PhysicsSystem2D.instance;
+        director.registerSystem(PhysicsSystem2D.ID, sys, sys.priority);
```
pandamicro, line 361: The default priority doesn't necessarily be LOW priority, please keep using `System.Priority.LOW`

cocos/physics-2d/framework/physics-system.ts
```diff
@@ -13,6 +13,9 @@ import { IPhysicsConfig, ICollisionMatrix } from '../../physics/framework/physic
 import { CollisionMatrix } from '../../physics/framework/collision-matrix';
 import { ERaycast2DType, RaycastResult2D, PHYSICS_2D_PTM_RATIO, PhysicsGroup } from './physics-types';
 import { Collider2D } from './components/colliders/collider-2d';
+import { legacyCC } from '../../core/global-exports';
+
+legacyCC.internal.PhysicsGroup = PhysicsGroup;
```
pandamicro, line 18: Override will lead to undefined behavior, so we need to sort this problem out with other approach. I will talk to editor team

cocos/physics-2d/framework/physics-internal-types.ts
jcyuan: 2 reasons:

1, because `DelayEvent` is used for pending execution for `PhysicsSystem` only, the `target` will be the function scope to be applied.
for example:

    PhysicsSystem2D.instance._callAfter(func, scope);
    // and _callAfter:
    _callAfterStep (target: unknown, func: () => void) {
        if (this._steping) {
            this._delayEvents.push({
                target,
                func,
            });
        } else {
            func.call(target);
        }
    }

2, the main reason: ESLint does not allow `Object`, `Function` to be a type. so the ESLint test on this PR can't be passed....
    according to the scope parameter type the `Function.prototype.apply / call` need, they are defined as this:

    apply<T, A extends any[], R>(this: (this: T, ...args: A) => R, thisArg: T, args: A): R;
    call<T, A extends any[], R>(this: (this: T, ...args: A) => R, thisArg: T, ...args: A): R;

so, if not use `unknown`, better to modify this `DelayEvent` as `DelayEvent<TScope>` etc solution, in my opinion this `DelayEvent` should not be a class. an `interface` is enough.

for the reason i modified `Function` as `() => void`, actually `Function` type is acceptable, but ESLint does not allow it...

cocos/physics-2d/framework/physics-system.ts
jcyuan: emmm, alright, because i read the source code from 3D `PhysicsSystem`, it uses this way, so i modified it.

cocos/physics-2d/framework/physics-system.ts
jcyuan: yes, agree! last night i found there were empty dropdown lists for RigidBody2D's `group` member, but user custom `PhysicsGroup` works fine.
so i wanted to comment out "Please don't merge, i found something strange and i need to test it thorough etc blahblah", and just found your comment here.
thanks, Panda.

jcyuan:
close as the approch to solve this problem is not right.

PR #9710, opened 2021-11-23
AnimationController.variables()

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 *


cocos/core/animation/marionette/animation-controller.ts
```diff
@@ -31,12 +31,24 @@ export class AnimationController extends Component {
         this._graphEval?.update(deltaTime);
     }
 
+    public variables () {
```
SantyWang: function name should be getVariables ?

shrinktofit: Solved.

SantyWang (approved):

PR #9721, opened 2021-11-24
initialize mesh when asset is loaded

SantyWang:
Re: https://github.com/cocos-creator/3d-tasks/issues/8950

Changelog:
 *

### DIFF
--- a/cocos/3d/assets/mesh.ts
+++ b/cocos/3d/assets/mesh.ts
@@ -287,6 +287,10 @@ export class Mesh extends Asset {
         super();
     }
 
+    public onLoaded () {
+        this.initialize();
+    }
+
     public initialize () {
         if (this._initialized) {
             return;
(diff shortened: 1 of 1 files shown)


cocos/3d/assets/mesh.ts
```diff
@@ -1012,10 +1016,6 @@ export class Mesh extends Asset {
             data: globalEmptyMeshBuffer,
         });
     }
-
-    public validate () {
-        return this.renderingSubMeshes.length > 0 && this.data.byteLength > 0;
-    }
```
pandamicro, line 1018: Why remove `validate` function ? And which problem this PR wants to solve

shrinktofit (approved):

cocos/3d/assets/mesh.ts
SantyWang: Can not validate mesh correctly, so move initialize function to `onloaded` and use `try-catch` to validate it

shrinktofit: This validation is not accurate and groundless, it broke empty meshes.

pandamicro (approved):

cocos/3d/assets/mesh.ts
dumganhar: If using JS try-catch to validate, how to do it in C++?

PR #9723, opened 2021-11-24
console exception if comp'node is null

Zhmz:
Re: cocos-creator/3d-tasks#9214

Changelog:
 * 在debug模式下，如果遇到comp的node为空的情况，报错提示异常组件

### DIFF
--- a/cocos/core/components/component.ts
+++ b/cocos/core/components/component.ts
@@ -74,7 +74,6 @@ class Component extends CCObject {
         if (this.node) {
             return `${this.node.name}<${className}>`;
         } else {
-            error(`id = ${this.uuid}, className = ${className}, component's node is null or undefined.`);
             return className;
         }
     }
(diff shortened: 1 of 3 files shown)


cocos/core/components/component.ts
```diff
@@ -70,7 +70,13 @@ class Component extends CCObject {
         if (trimLeft >= 0) {
             className = className.slice(trimLeft + 1);
         }
-        return `${this.node.name}<${className}>`;
+
+        if (this.node) {
+            return `${this.node.name}<${className}>`;
+        } else {
+            error(`id = ${this.uuid}, className = ${className}, component's node is null or undefined.`);
```
pandamicro: No need to report error here, this API shouldn't judge whether node should present or not. Just give the class name when node is absent

Zhmz:
image
测试场景

pandamicro (approved):

PR #9753, opened 2021-12-02
when the pointer is locked, don't move it.

mikecoker:
Bug fix with locked pointer. Issue [https://github.com/cocos-creator/engine/issues/9749](url)

Changelog:
 * Remove the adjustment to the pointer when the cursor is locked.

### DIFF
--- a/pal/input/web/mouse-input.ts
+++ b/pal/input/web/mouse-input.ts
@@ -37,8 +37,8 @@ export class MouseInputSource {
     private _getLocation (mouseEvent: MouseEvent): Vec2 {
         const canvasRect = this._getCanvasRect();
         const dpr = screenAdapter.devicePixelRatio;
-        let x = this._pointLocked ? (this._preMousePos.x / dpr + mouseEvent.movementX) : (mouseEvent.clientX - canvasRect.x);
-        let y = this._pointLocked ? (this._preMousePos.y / dpr - mouseEvent.movementY) : (canvasRect.y + canvasRect.height - mouseEvent.clientY);
+        let x = this._pointLocked ? (this._preMousePos.x / dpr) : (mouseEvent.clientX - canvasRect.x);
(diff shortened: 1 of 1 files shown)


mikecoker:
looks like the lint process did not run correctly?

SantyWang:
Thanks for contribution

pandamicro:
> looks like the lint process did not run correctly?

I relaunched it and passed, thanks for the contribution

PPpro:
Thanks for contribution.

I'm afraid this Pull Request can't be merged for now, because the engine haven't officially support PointerLock feature yet .  
For now, the `location` property in `EventMouse`  is designed to display the virtual location of pointer even if it's locked.  
And we should not change this interface behavior, which causes __Breaking Change__

we'll fix this bug after we offially support  PointerLock , here is the issue: https://github.com/cocos-creator/engine/issues/9766

mikecoker:
Sure, I understand this is not an engine feature yet, but the way it is implemented for the web is currently causing unexpected results. While the pointer is "locked" you can move the pointer over a UI control and that control will get an event that was not meant for it. Specifically in my case I right click to move the camera, where I lock/hide the pointer but if I release the mouse button while over a control, the camera gets stuck in move mode. I've fixed this in my fork of the engine as a work around so any solution to the problem is welcome. Thank you.

PPpro:
Yes, you can fix the issue this way in your project.
But this bug fix changes the behavior when you call `eventMouse.getLocation()`. Any breaking change on interface can introduce other bugs in projects.  We should avoid the compatibility issue like this.

I think this is an issue about __interface design__.
For developers, it's better to provide a new property like `input.isPointerLocked` to avoid UI interaction when pointer is locked，insteadof changing the behavior of `eventMouse.getLocation()`

PPpro:
sorry, we have to close this Pull Request
please track this issue https://github.com/cocos-creator/engine/issues/9766 

thanks again for your contribution !

PR #9760, opened 2021-12-04
v3.4 reset renderable component need to setMaterial

arsen2010:
Re: cocos-creator/3d-tasks#9907

Changelog:
 *


cocos/core/components/renderable-component.ts
```diff
@@ -72,10 +72,11 @@ export class RenderableComponent extends Component {
     }
 
     set sharedMaterials (val) {
-        for (let i = 0; i < val.length; i++) {
```
SantyWang, line 75: 这样改不对吧，为啥要改成 <=

PR #9803, opened 2021-12-13
Update instantiate-jit.ts

JasonWmj:
check obj.node is null

Re: cocos-creator/3d-tasks#

Changelog:
 *

### DIFF
--- a/cocos/core/data/instantiate-jit.ts
+++ b/cocos/core/data/instantiate-jit.ts
@@ -476,7 +476,7 @@ class Parser {
                             return this.getObjRef(obj);
                         }
                     } else if (obj instanceof legacyCC.Component) {
-                        if (!obj.node.isChildOf(this.parent)) {
+                        if (obj.node && !obj.node.isChildOf(this.parent)) {
                             // should not clone other component if not descendant
                             return this.getObjRef(obj);
                         }


JasonWmj:
check obj.node is null

cocos/core/data/instantiate-jit.ts
```diff
@@ -476,7 +476,7 @@ class Parser {
                             return this.getObjRef(obj);
                         }
                     } else if (obj instanceof legacyCC.Component) {
-                        if (!obj.node.isChildOf(this.parent)) {
+                        if (obj.node && !obj.node.isChildOf(this.parent)) {
```
jareguo, line 479: 建议是
```suggestion
                        if (!(obj.node && obj.node.isChildOf(this.parent)) {
```
或者
```suggestion
                        if (!obj.node?.isChildOf(this.parent)) {
```

jareguo:
感谢提交 PR，看上去像是异常情况下出现的容错，似乎是组件被 destroy 了才会导致 node 为 null？
但是我们不太理解为什么要在这么细的一个分支进行判断。这类型的容错应该是更系统化或者从使用角度去避免（比如校验传入 cc.instantiate 的 obj.isValid）。 

@JasonWmj 由于前面一直没有得到回复，先 close 本 PR，有进一步反馈欢迎随时提交或者 reopen。

----

Thank you for submitting the PR. It looks like a fault tolerance that occurs under exceptional circumstances. It seems that the node becomes null only when the component is destroyed. However, we do not understand why such judgment needs to be made in such a fine-grained branch. This type of fault tolerance should be avoided more systematically or from a user perspective (such as verifying obj.isValid passed to cc.instantiate).

Since there has been no response before, let's close this PR first. Please feel free to submit further feedback or reopen it if necessary.

PR #9879, opened 2021-12-27
Jsb-event-handler for event dispatch

timlyeee:
* Define property jsb-event-handler for Jsb 
* Use with engine-native pr


minggo (approved):

minggo:
代码多很多格式相关的修改，请确认一下。

PR #9882, opened 2021-12-27
 fix fuzzy text on high resolution displays

ankye:
Re: cocos-creator/3d-tasks#

Changelog:
 * 增加文字dpr功能，支持none，bitmap和char模式


pandamicro:
Duplicated with https://github.com/cocos-creator/engine/pull/9748

PR #9885, opened 2021-12-28
Fix particle index count 0

zxx43:
Re: cocos-creator/3d-tasks#

Changelog:
 * 

issue: https://github.com/cocos-creator/3d-tasks/issues/10673


cocos/particle/models/particle-batch-model.ts
```diff
@@ -308,6 +308,9 @@ export default class ParticleBatchModel extends scene.Model {
     }
 
     public updateIA (count: number) {
+        if (count <= 0) {
+           return;
```
pandamicro: Please fix indent

pandamicro (approved):

PR #9948, opened 2022-01-07
fix some spine bug

LinYunMo:
https://github.com/cocos-creator/engine/pull/9802

### DIFF
--- a/cocos/2d/renderer/batcher-2d.ts
+++ b/cocos/2d/renderer/batcher-2d.ts
@@ -700,13 +700,14 @@ class LocalDescriptorSet  {
         this.uploadLocalData();
     }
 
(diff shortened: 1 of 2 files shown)


cocos/2d/renderer/batcher-2d.ts
```diff
@@ -702,6 +702,10 @@ class LocalDescriptorSet  {
 
     public updateLocal () {
         if (!this._transform) return;
+        if (!this._transform.isValid) {
+            this.reset();
```
pandamicro: Also remove from the global cache

pandamicro (approved):

pandamicro (approved):

PR #9971, opened 2022-01-12
fix: keep childs with same layer in editbox

dhb118:


dhb118:
`const node = new Node("editbnox");`
`node.layer = Layers.Enum.UI_2D;`
`node.addComponent(EditBox);`

pandamicro (approved):

PR #9978, opened 2022-01-12
Separate ia range and buffer range in MeshRenderData

pandamicro:


pandamicro:
Replacing https://github.com/cocos-creator/engine/pull/9977

SantyWang (approved):

PR #9980, opened 2022-01-12
Fix tiled map

pandamicro:


pandamicro:
@cocos-robot run test cases

PR #10006, opened 2022-01-17
Fix skeletal animation sample() does not work as intended

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 * This PR fixes:
   - The first frame of skeletal animation is not played.

   - If `SkeletalAnimationState.play()` is not called, the `SkeletalAnimationState.sample()` takes no effect. (https://forum.cocos.org/t/topic/127457/198?u=shrinktofit)

 * Also optimized:
 
  - The mesh renderer should be child of skeletal animation.

### DIFF
--- a/cocos/3d/skeletal-animation/skeletal-animation-state.ts
+++ b/cocos/3d/skeletal-animation/skeletal-animation-state.ts
@@ -164,7 +164,7 @@ export class SkeletalAnimationState extends AnimationState {
         const info = this._animInfo!;
 
         // Ensure I'm the one on which the anim info is sampling.
-        if (!this._animInfoMgr.isCurrentlySampling(info, this.clip)) {
+        if (!this._animInfoMgr.isSampling(info, this.clip)) {
             // If not, switch to me.
             this._animInfoMgr.switchClip(this._animInfo!, this.clip);
 
--- a/cocos/3d/skeletal-animation/skeletal-animation-state.ts
+++ b/cocos/3d/skeletal-animation/skeletal-animation-state.ts
@@ -179,11 +179,8 @@ export class SkeletalAnimationState extends AnimationState {
         }
 
         const curFrame = (ratio * this._frames + 0.5) | 0;
-        if (info.hasSampled && curFrame === info.data[0]) {
-            return;
-        }
+        if (curFrame === info.data[0]) { return; }
 
-        info.hasSampled = true;
         info.data[0] = curFrame;
         this._setAnimInfoDirty(info, true);
         for (let i = 0; i < this._sockets.length; ++i) {
--- a/cocos/3d/skeletal-animation/skeletal-animation-state.ts
+++ b/cocos/3d/skeletal-animation/skeletal-animation-state.ts
@@ -162,15 +162,16 @@ export class SkeletalAnimationState extends AnimationState {
     private _sampleCurvesBaked (time: number) {
         const ratio = time / this.duration;
         const info = this._animInfo!;
+        const clip = this.clip;
 
         // Ensure I'm the one on which the anim info is sampling.
-        if (!this._animInfoMgr.isSampling(info, this.clip)) {
+        if (info.currentClip !== clip) {
             // If not, switch to me.
-            this._animInfoMgr.switchClip(this._animInfo!, this.clip);
+            this._animInfoMgr.switchClip(this._animInfo!, clip);
 
             const users = this._parent!.getUsers();
             users.forEach((user) => {
-                user.uploadAnimation(this.clip);
+                user.uploadAnimation(clip);
             });
         }
 
--- a/cocos/3d/skeletal-animation/skeletal-animation-state.ts
+++ b/cocos/3d/skeletal-animation/skeletal-animation-state.ts
@@ -83,7 +83,7 @@ export class SkeletalAnimationState extends AnimationState {
         this._frames = frames - 1;
         this._animInfo = this._animInfoMgr.getData(root.uuid);
         this._bakedDuration = this._frames / samples; // last key
-        this.setUseBaked(this._parent.useBakedAnimation);
+        this.setUseBaked(baked);
     }
 
     public setUseBaked (useBaked: boolean) {
--- a/cocos/3d/skeletal-animation/skeletal-animation-state.ts
+++ b/cocos/3d/skeletal-animation/skeletal-animation-state.ts
@@ -83,19 +83,13 @@ export class SkeletalAnimationState extends AnimationState {
         this._frames = frames - 1;
         this._animInfo = this._animInfoMgr.getData(root.uuid);
         this._bakedDuration = this._frames / samples; // last key
+        this.setUseBaked(this._parent.useBakedAnimation);
     }
 
-    public onPlay () {
-        super.onPlay();
-        const baked = this._parent!.useBakedAnimation;
-        if (baked) {
+    public setUseBaked (useBaked: boolean) {
+        if (useBaked) {
             this._sampleCurves = this._sampleCurvesBaked;
             this.duration = this._bakedDuration;
-            this._animInfoMgr.switchClip(this._animInfo!, this.clip);
-            const users = this._parent!.getUsers();
-            users.forEach((user) => {
-                user.uploadAnimation(this.clip);
-            });
         } else {
             this._sampleCurves = super._sampleCurves;
             this.duration = this.clip.duration;
@@ -168,8 +162,24 @@ export class SkeletalAnimationState extends AnimationState {
     private _sampleCurvesBaked (time: number) {
         const ratio = time / this.duration;
         const info = this._animInfo!;
+
+        // Ensure I'm the one on which the anim info is sampling.
+        if (!this._animInfoMgr.isCurrentlySampling(info, this.clip)) {
+            // If not, switch to me.
+            this._animInfoMgr.switchClip(this._animInfo!, this.clip);
+
+            const users = this._parent!.getUsers();
+            users.forEach((user) => {
+                user.uploadAnimation(this.clip);
+            });
+        }
+
         const curFrame = (ratio * this._frames + 0.5) | 0;
-        if (curFrame === info.data[0]) { return; }
+        if (info.hasSampled && curFrame === info.data[0]) {
+            return;
+        }
+
+        info.hasSampled = true;
         info.data[0] = curFrame;
         this._setAnimInfoDirty(info, true);
         for (let i = 0; i < this._sockets.length; ++i) {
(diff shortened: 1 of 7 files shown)


cocos/3d/skeletal-animation/skeletal-animation-state.ts
```diff
@@ -83,19 +83,13 @@ export class SkeletalAnimationState extends AnimationState {
         this._frames = frames - 1;
         this._animInfo = this._animInfoMgr.getData(root.uuid);
         this._bakedDuration = this._frames / samples; // last key
+        this.setUseBaked(baked);
```
shrinktofit, line 86: This ensure the bake option(`true` vs `false`) of animation state is up-to-date with skinned mesh renderer, whenever.

cocos/3d/skeletal-animation/skeletal-animation-state.ts
```diff
@@ -83,19 +83,13 @@ export class SkeletalAnimationState extends AnimationState {
...
     }
 
-    public onPlay () {
```
shrinktofit, line 88: Before this PR: do "sync bake option" on playing.

This PR: Sync at `initialize()`, sync bake option each time it changes.

cocos/3d/skeletal-animation/skeletal-animation-state.ts
```diff
@@ -83,19 +83,13 @@ export class SkeletalAnimationState extends AnimationState {
...
-        super.onPlay();
-        const baked = this._parent!.useBakedAnimation;
-        if (baked) {
+    public setUseBaked (useBaked: boolean) {
+        if (useBaked) {
             this._sampleCurves = this._sampleCurvesBaked;
             this.duration = this._bakedDuration;
-            this._animInfoMgr.switchClip(this._animInfo!, this.clip);
```
shrinktofit, line 94: Moved to below.

cocos/3d/skeletal-animation/skeletal-animation-state.ts
```diff
@@ -168,8 +162,24 @@ export class SkeletalAnimationState extends AnimationState {
     private _sampleCurvesBaked (time: number) {
         const ratio = time / this.duration;
         const info = this._animInfo!;
+
+        // Ensure I'm the one on which the anim info is sampling.
+        if (!this._animInfoMgr.isCurrentlySampling(info, this.clip)) {
```
shrinktofit: @YunHsiao Please review the main fix here. I'll do the check-switch each time the animation is sampled.

cocos/3d/skeletal-animation/skeletal-animation-state.ts
```diff
@@ -168,8 +162,24 @@ export class SkeletalAnimationState extends AnimationState {
...
+            this._animInfoMgr.switchClip(this._animInfo!, this.clip);
+
+            const users = this._parent!.getUsers();
+            users.forEach((user) => {
+                user.uploadAnimation(this.clip);
+            });
+        }
+
         const curFrame = (ratio * this._frames + 0.5) | 0;
-        if (curFrame === info.data[0]) { return; }
```
shrinktofit, line 172: @YunHsiao Please review the fix here. This cause the first frame always be skipped.

cocos/3d/skeletal-animation/skeletal-animation-state.ts
```diff
@@ -168,8 +162,24 @@ export class SkeletalAnimationState extends AnimationState {
...
+        if (info.hasSampled && curFrame === info.data[0]) {
```
shrinktofit: And the fix.

cocos/3d/skeletal-animation/skeletal-animation-utils.ts
```diff
@@ -457,6 +457,11 @@ export interface IAnimInfo {
     buffer: Buffer;
     data: Float32Array;
     dirty: boolean;
+    currentClipRef: WeakSet<AnimationClip>;
```
shrinktofit: I want to keep a record of `AnimationClip`, something like `WeakRef`. If you have better opinion, please.

cocos/3d/skeletal-animation/skeletal-animation-utils.ts
```diff
@@ -457,6 +457,11 @@ export interface IAnimInfo {
...
+    /**
+     * Has sampled at least once since last switching.
+     */
+    hasSampled: boolean;
```
shrinktofit: Fix about the first frame BUG.

cocos/3d/skeletal-animation/skeletal-animation-utils.ts
```diff
@@ -506,7 +511,16 @@ export class JointAnimationInfo {
         }
     }
 
+    public isCurrentlySampling (info: IAnimInfo, clip: AnimationClip) {
+        return info.currentClipRef.has(clip);
+    }
+
     public switchClip (info: IAnimInfo, clip: AnimationClip | null) {
+        info.hasSampled = false;
+        info.currentClipRef = new WeakSet();
```
shrinktofit: `WeakSet` doesn not provide a `clear()`

cocos/3d/skinned-mesh-renderer/skinned-mesh-renderer.ts
```diff
@@ -146,43 +154,31 @@ export class SkinnedMeshRenderer extends MeshRenderer {
...
     }
 
-    private _animation: SkeletalAnimation | null = null;
-
     private _tryBindAnimation () {
         const { _skinningRoot: skinningRoot } = this;
         if (!skinningRoot) {
             return;
         }
+
```
shrinktofit, line 167: Corresponding to https://github.com/cocos-creator/engine/pull/10006/files#diff-6b7762c60bea81084e63372b54ddae7766cb7e91ecc85de1977d3a51664bcd25R170-R181 . They should take same effect.

cocos/3d/skeletal-animation/skeletal-animation-state.ts
YunHsiao: I think initializing the anim data to a special illegal value would do, saves the ad-hoc flag test

cocos/3d/skeletal-animation/skeletal-animation-utils.ts
```diff
@@ -506,7 +511,16 @@ export class JointAnimationInfo {
         }
     }
 
+    public isCurrentlySampling (info: IAnimInfo, clip: AnimationClip) {
```
pandamicro: `isSampling` is enough, the present continuous tense is the same as current

cocos/3d/skeletal-animation/skeletal-animation-utils.ts
pandamicro: It has been created previously, why replace it here

cocos/3d/skeletal-animation/skeletal-animation-utils.ts
YunHsiao: Just store the clip itself. The whole anim info object will be GCed once the component is destroyed.

cocos/3d/skeletal-animation/skeletal-animation-state.ts
YunHsiao: Are you sure this works for both baked & sampled animations? I think the sampled branch has some init code that still needs to be executed on each play event.

cocos/3d/skeletal-animation/skeletal-animation-state.ts
pandamicro: This looks very strange, previously `onPlay` is only invoked in the beginning, but `_sampleCurvesBaked` is invoked in every animation frame, it just doesn't seem elegant enough

cocos/3d/skeletal-animation/skeletal-animation-state.ts
shrinktofit: > 

The only illegal is `NaN`, could I?

cocos/3d/skeletal-animation/skeletal-animation-utils.ts
shrinktofit: Got it .

cocos/3d/skeletal-animation/skeletal-animation-utils.ts
shrinktofit: Fixed.

cocos/3d/skeletal-animation/skeletal-animation-utils.ts
shrinktofit: Fixed.

cocos/3d/skeletal-animation/skeletal-animation-state.ts
shrinktofit: > Are you sure this works for both baked & sampled animations? I think the sampled branch has some init code that still needs to be executed on each play event.

The init code of baked branch was moved to `setUseBaked`, which is invoked at `SkeletalAnimationState.initialize()` and `set SkeletalAnimation.useBakedAnimation()`. Is there something I missed?

cocos/3d/skeletal-animation/skeletal-animation-state.ts
shrinktofit: There are two methods that trigger the sampling process:

- `SkeletalAnimationState.play()`/`SkeletalAnimationState.resume()` - Called once, do sampling every frame

- `SkeletalAnimationState.sample()` - Called once, do once sampling immediately.

So I captured the sampling process.

cocos/3d/skeletal-animation/skeletal-animation-state.ts
YunHsiao: I think any negative number would do?

pandamicro (approved):

cocos/3d/skeletal-animation/skeletal-animation-state.ts
YunHsiao: I mean what about the `else` branch, under real-time sampled animation mode?

cocos/3d/skeletal-animation/skeletal-animation-utils.ts
```diff
@@ -507,7 +508,8 @@ export class JointAnimationInfo {
     }
 
     public switchClip (info: IAnimInfo, clip: AnimationClip | null) {
-        info.data[0] = 0;
+        info.currentClip = clip;
+        info.data[0] = -1;
```
shrinktofit, line 512: @YunHsiao

cocos/3d/skeletal-animation/skeletal-animation-state.ts
shrinktofit: OK

YunHsiao (approved):

PR #10057, opened 2022-01-27
3.4 wgpu rebase

hana-alice:
Re: cocos-creator/3d-tasks#

Changelog:
 * wgpu abailable!
 * This guy is new to typescript, review as strict as possible pls.


YunHsiao:
The bulk of this PR is all glue code, and can be optimized away entirely using a custom template for binding generator, with the `gfx.ini` config doesn't even need to be changed. I personally do not recommend hastily merging this at this point.

YunHsiao (changes requested):
The bulk of this PR is all hand-written glue code, and can be optimized away entirely using a custom template for binding generator, with the `gfx.ini` config doesn't even need to be changed. I personally do not recommend hastily merging it at this point.

PR #10058, opened 2022-01-27
fix binding mappings

YunHsiao:
Re: cocos-creator/3d-tasks#

Changelog:
 * binding mappings are not specified correctly for local set's storage image count


fish9167:
@cocos-robot run test cases

minggo (approved):

PR #10078, opened 2022-02-11
Fix potential crash after changing v8 weakcallback type from kFinalizer to kParameter.

dumganhar:
Fix potential crash after https://github.com/cocos-creator/engine-native/pull/4232

Re: cocos-creator/3d-tasks#

Changelog:
 *


Bush16:
我已收到你的邮件，谢谢。

PatriceJiang (approved):

minggo (approved):

PR #10088, opened 2022-02-13
reset input assembler when particle is reset

caogtaa:
Re: cocos-creator/2d-tasks#

### Bug Description
I would like to re-use a pool of `cc.ParticleSystem` for performance reason.
But when I reuse a particle system, its previous state is rendered to screen in the 1st frame after I call `resetSystem()`.

This bug is similar to https://github.com/cocos-creator/engine/pull/7060

### How to Reproduce
1. If particle system is no longer used, stop it and also call **`active = false`** on its node.
2. To reuse a particle system, call **`active = true`** and `resetSystem()` **in tween callback**

### How to Fix
Reset input assembler when particle is reset.

### Screen
粒子残留_v248

### Demo
[testParticleReuseBug248.zip](https://github.com/cocos-creator/engine/files/8055603/testParticleReuseBug248.zip)

### DIFF
--- a/cocos2d/particle/particle-simulator.js
+++ b/cocos2d/particle/particle-simulator.js
@@ -112,6 +112,9 @@ Simulator.prototype.reset = function () {
     for (let id = 0; id < particles.length; ++id)
         pool.put(particles[id]);
     particles.length = 0;
+    let assembler = this.sys._assembler;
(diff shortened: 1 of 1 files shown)


SantyWang (approved):

SantyWang:
感谢贡献

PR #10094, opened 2022-02-14
Fix loop audio would stop for ios system

timlyeee:
* IOS system has an uncertain state between foreground and background, and our condition sometimes miss the point to load buffers.
* Queue for 3 buffers might be too short to live, use 4 or even more instead.
* Issue could be solved: 
* https://github.com/cocos2d/cocos2d-x/issues/19480, 
* https://github.com/cocos-creator/3d-tasks/issues/11135

### DIFF
--- a/native/cocos/audio/apple/AudioMacros.h
+++ b/native/cocos/audio/apple/AudioMacros.h
@@ -26,7 +26,7 @@
 
 #pragma once
 
-#define QUEUEBUFFER_NUM       (3)
+#define QUEUEBUFFER_NUM       (4)
 #define QUEUEBUFFER_TIME_STEP (0.05f)
 
 #define QUOTEME_(x) #x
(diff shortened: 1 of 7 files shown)


native/cocos/audio/apple/AudioMacros.h
```diff
@@ -26,7 +26,7 @@
 
 #pragma once
 
-#define QUEUEBUFFER_NUM       (3)
+#define QUEUEBUFFER_NUM       (4)
```
dumganhar, line 29: 这里有必要改为 4 才能修复这个问题么？

timlyeee: 不必，但是可以保险

dumganhar (approved):

minggo (approved):

dumganhar (approved):

PR #10181, opened 2022-02-25
fix cc local batch

YunHsiao:
Re: cocos-creator/3d-tasks#

Changelog:
 * fix `cc-local-batch` does not include what it uses introduced by #10123


YunHsiao:
closed in favor of https://github.com/cocos-creator/engine-extensions/pull/47

PR #10184, opened 2022-02-25
Fix ios device display effect 

timlyeee:


minggo (approved):

timlyeee:
Same as * https://github.com/cocos-creator/engine-native/pull/4265
issue related * https://github.com/cocos-creator/3d-tasks/issues/11159

PR #10195, opened 2022-02-26
fix misuse of drawAPI on metal

hana-alice:
Re: cocos-creator/3d-tasks#

Changelog:
 *

PR #10202, opened 2022-02-28
fix misuse of drawAPI on metal (#10195)

hana-alice:
Re: cocos-creator/3d-tasks#

Changelog:
 *


hana-alice:
@cocos-robot run test cases

minggo (approved):

PR #10310, opened 2022-03-11
fix egl & gl error while enter backgroun and reenter forground

bofeng-song:


native/cocos/renderer/gfx-base/GFXDevice.h
```diff
@@ -165,6 +169,7 @@ class CC_DLL Device : public Object {
     std::array<bool, static_cast<size_t>(Feature::COUNT)>         _features;
     std::array<FormatFeature, static_cast<size_t>(Format::COUNT)> _formatFeatures;
 
+    bool _isSurfaceReady{false};
```
minggo: device 感觉已经变成了 gfx module 了。我觉得应该需要有个 gfx 模块的信息，模块是否初始化好，是否能绘制等信息应该是放在模块类里。

bofeng-song: 已更新

native/cocos/platform/android/AndroidPlatform.cpp
```diff
@@ -64,7 +65,9 @@ void AndroidPlatform::waitWindowInitialized() {
 int32_t AndroidPlatform::loop() {
     while (_jniNativeGlue->isRunning()) {
         pollEvent();
-        runTask();
+        if (cc::gfx::Device::getInstance()->isDeviceAvailable()) {
+            runTask();
+        }
```
minggo: It will block game logic, right? As we said, we can not just block game logic, may be we should just cancel rendering.

minggo: It will block game logic, right? As we said, we can not just block game logic, may be we should just cancel rendering.

native/cocos/platform/android/AndroidPlatform.h
```diff
@@ -41,5 +41,6 @@ class AndroidPlatform : public UniversalPlatform {
     void waitWindowInitialized();
 
     JniNativeGlue* _jniNativeGlue;
+    bool _renderAvailable{false};
```
minggo: I think it is not needed.

minggo: I think it is not needed.

native/cocos/renderer/gfx-base/GFXDevice.h
```diff
@@ -48,6 +48,13 @@
 namespace cc {
 namespace gfx {
 
+enum class DeviceState : char {
+    STATE_UNKNOWN = 0,
+    STATE_INITIALIZED,
```
minggo: It seems STATE_INITIALIZED is not needed. Just need a bool value to check if device is ready for rendering or not?

minggo: It seems STATE_INITIALIZED is not needed. Just need a bool value to check if device is ready for rendering or not?

native/cocos/platform/android/AndroidPlatform.cpp
bofeng-song: do we need to block rendering or just discard render command?

bofeng-song: if block rendering, if may make game thread's state change to block

native/cocos/renderer/gfx-agent/DeviceAgent.cpp
```diff
@@ -141,6 +141,9 @@ void DeviceAgent::acquire(Swapchain *const *swapchains, uint32_t count) {
 }
 
 void DeviceAgent::present() {
+    if (!cc::gfx::Device::getInstance()->isRendererAvailable()) {
```
minggo, line 144: I think should disable rendering in render pipeline, may be in root.cpp?

bofeng-song: version of v3.5.0, root is not native

PR #10314, opened 2022-03-11
avoid spine and dragon crash.

xianyinchen:
without call Object::clearPrivateData, the c++ object maybe bind to two JS object, when one was release, other one will case crash.

Some existent crash stacks.
image
image
image


PatriceJiang (approved):

minggo:
@xianyinchen does this PR sync from other PR in v2.x?

minggo (approved):

xianyinchen:
yes.
https://github.com/cocos/engine-native/pull/4271

PR #10344, opened 2022-03-15
Windows platform handles multiple events at once

qiuguohua:
[issue](https://github.com/cocos/3d-tasks/issues/11391)


minggo:
Why modify like this?

qiuguohua:
> Why modify like this?

Because events are very fast and rendering is relatively slow, doing so can handle events in a timely manner

minggo (approved):

minggo:
Any test case to show the difference? Or did you meet any problem?

qiuguohua:
> Any test case to show the difference? Or did you meet any problem?

Not yet, this is what the previous implementation did, and what was missed in the previous adaptation, I will restore it

minggo:
Got it.

dumganhar (approved):

PR #10365, opened 2022-03-16
Add destroy window event and recreate window event on windows platform

qiuguohua:
[issue](https://github.com/cocos/3d-tasks/issues/8798)

It is available on the android platform and also on the windows platform, but there may be some differences. Apple does not send this event. Currently, these events are added according to the implementation of 3.4.

### DIFF
--- a/native/cocos/engine/Engine.cpp
+++ b/native/cocos/engine/Engine.cpp
@@ -331,7 +331,9 @@ void Engine::onPause() {
     AppEvent appEv;
     appEv.type = AppEvent::Type::PAUSE;
     dispatchEventToApp(OSEventType::APP_OSEVENT, appEv);
-
+#if CC_PLATFORM == CC_PLATFORM_WINDOWS
+    cc::EventDispatcher::dispatchDestroyWindowEvent();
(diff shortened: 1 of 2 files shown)


native/cocos/engine/Engine.cpp
```diff
@@ -331,15 +331,19 @@ void Engine::onPause() {
     AppEvent appEv;
     appEv.type = AppEvent::Type::PAUSE;
     dispatchEventToApp(OSEventType::APP_OSEVENT, appEv);
-
+#if CC_PLATFORM == CC_PLATFORM_WINDOWS
+    cc::EventDispatcher::dispatchDestroyWindowEvent();
+#endif
```
minggo, line 336: Is pause event the same as destroy window event?

qiuguohua: > Is pause event the same as destroy window event?


Unlike the destroy event, it is only used in the render module. This was done in 3.4, I just reused the call

minggo (approved):

timlyeee (approved):

PR #10382, opened 2022-03-17
fix director.ts typo

novain1102:
developer feedback: https://forum.cocos.org/t/topic/131605/230?u=308303735

### DIFF
--- a/cocos/core/director.ts
+++ b/cocos/core/director.ts
@@ -345,7 +345,7 @@ export class Director extends EventTarget {
         const persistNodeList = Object.keys(game._persistRootNodes).map((x) => game._persistRootNodes[x] as Node);
         for (let i = 0; i < persistNodeList.length; i++) {
             const node = persistNodeList[i];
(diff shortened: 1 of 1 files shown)


cocos/core/director.ts
```diff
@@ -762,9 +762,7 @@ export declare namespace Director {
     export type OnLoadSceneProgress = (completedCount: number, totalCount: number, item: any) => void;
 }
 
-legacyCC.Director = Director;
-
 /**
  * 导演类。
  */
-export const director: Director = Director.instance = legacyCC.director = new Director();
```
pandamicro: This shouldn't be changed, it's for backward compatibility

pandamicro (changes requested):

cocos/core/director.ts
novain1102: already fallbacked and tested

pandamicro (approved):

PR #10414, opened 2022-03-21
remove hidden

SantyWang:
Re: cocos-creator/3d-tasks#

Changelog:
 *

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -16,6 +16,6 @@
         "typescript"
     ],
     "editor.codeActionsOnSave": {
-        "source.fixAll": true,
+        "source.fixAll": false,
     },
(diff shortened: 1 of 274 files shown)


shrinktofit (approved):

.vscode/settings.json
```diff
@@ -16,6 +16,6 @@
         "typescript"
     ],
     "editor.codeActionsOnSave": {
-        "source.fixAll": true,
```
pandamicro: This would affect the current using experience, suggest to add an option about this elsewhere and keep the current behavior

pandamicro (changes requested):

pandamicro (approved):

PR #10443, opened 2022-03-22
V3.5 rename assets

PatriceJiang:
Solve: https://github.com/cocos/3d-tasks/issues/11484

Depends: https://github.com/cocos/creator-runtime-extensions/pull/140

Changelog:
 * rename `assets/` folder to `data/`
 * Xcode: link `data/` to `Resources/data/` instead of sub-folders.


native/cocos/engine/Engine.cpp
```diff
@@ -82,6 +82,8 @@ namespace cc {
 Engine::Engine() {
     _scheduler = std::make_shared<Scheduler>();
     FileUtils::getInstance()->addSearchPath("Resources", true);
+    FileUtils::getInstance()->addSearchPath("data", true);
```
minggo, line 85: Is it needed?

PatriceJiang: `Resources/data` is not needed and removed

SantyWang (approved):

Issue #10465, opened 2022-03-23
Link error when running virtualizer on ios


qiuguohua:
![26c9802b966633061ae4c302f449c2f9](https://user-images.githubusercontent.com/5205637/159663720-4d87e784-ac0e-4c33-a868-2aae120f3efb.jpg)

timlyeee:
Did it once, but not merged. dum....https://github.com/cocos/cocos-engine-external/pull/200

minggo:
oops, i forgot to merge it? @timlyeee could you please notice me if i forgot  in future.

PR #10477, opened 2022-03-24
fix ios m1 simulator tbb link error

novain1102:
Re: cocos-creator/3d-tasks#
https://github.com/cocos/cocos-engine/issues/10465


novain1102:
已更新到tag105, close

PR #10485, opened 2022-03-24
Optimize the way how animation component update its animation content

shrinktofit:
Re: cocos/cocos-engine#

Changelog:
 * Optimize the way how animation component update its animation content

This address the issue that even an animation is not played, it will still get update when main loop tick.

Before this PR, all animation components share the same "animation blend buffer" and that buffer is updated per tick. This PR however create individual animation blender buffer for every animation component instance, and control the updating of the buffer by itself.

By the way, the cross fade is optimized to be controled by animation component itself -- even it should be by design at the outset.

### DIFF
--- a/cocos/core/animation/animation-component.ts
+++ b/cocos/core/animation/animation-component.ts
@@ -190,6 +190,7 @@ export class Animation extends Eventify(Component) {
     public onDisable () {
         assertIsTrue(typeof this._animationUpdateTaskHandle !== 'undefined');
         getGlobalAnimationManager().removeUpdateTask(this._animationUpdateTaskHandle);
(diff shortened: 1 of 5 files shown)


cocos/core/animation/animation-component.ts
```diff
@@ -163,7 +169,7 @@ export class Animation extends Eventify(Component) {
         this.clips = this._clips;
         for (const stateName in this._nameToState) {
             const state = this._nameToState[stateName];
-            state.initialize(this.node);
+            this._initializeState(state);
```
shrinktofit, line 172: There are 3 points to do the work, so I extracted it.

SantyWang (approved):

PR #10509, opened 2022-03-25
V3.5.0 appdelegate

qiuguohua:
Move appdelegate into template


native/templates/ios/View.mm
```diff
@@ -24,16 +24,15 @@ of this software and associated engine source code (the "Software"), a limited,
 ****************************************************************************/
 
 #import "View.h"
-#import "platform/ios/AppDelegate.h"
```
minggo: View.mm don't have to be in template. It can be inside engine, as it is added by engine. It is not a concept of iOS. And it is related to iOS rendering, the logic may change.

native/templates/ios/AppDelegate.mm
```diff
@@ -25,24 +25,25 @@ of this software and associated engine source code (the "Software"), a limited,
...
-
 @synthesize window;
+@synthesize appDelegateBridge;
 
 #pragma mark -
 #pragma mark Application lifecycle
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
-    [[SDKWrapper shared] application:application didFinishLaunchingWithOptions:launchOptions];
+    appDelegateBridge = [[AppDelegateBridge alloc] init];
```
minggo, line 45: Where to destroy the bridge?

native/templates/ios/AppDelegate.mm
```diff
@@ -25,24 +25,25 @@ of this software and associated engine source code (the "Software"), a limited,
...
+    [[SDKWrapper shared] addSdkDelegate:appDelegateBridge];
```
minggo: Why need it?

qiuguohua: Because I also treat cocos as an sdk, so it don't need to send messages twice with SDKWrapper.

native/templates/ios/AppDelegate.mm
qiuguohua: Because sdkwrapper itself is a static class that contains other users' listening events. When destroying an event, it is also necessary to handle the event before destroying it. I'm not sure where a release would be better, so kept the previous design

native/templates/ios/View.mm
qiuguohua: ok, i'll change it

native/cocos/platform/ios/AppDelegateBridge.h
```diff
@@ -0,0 +1,36 @@
+/****************************************************************************
+ Copyright (c) 2021-2022 Xiamen Yaji Software Co., Ltd.
```
dumganhar: The license's time information of new files should start from current year (2022) .

dumganhar: Copyright (c) 2022 Xiamen Yaji Software Co., Ltd.

qiuguohua: OK

native/CMakeLists.txt
```diff
@@ -376,10 +376,6 @@ elseif(MACOSX)
     cocos_source_files(
         cocos/main/main-mac.mm
     )
-elseif(IOS)
-    cocos_source_files(
-        cocos/main/main-ios.mm
```
PatriceJiang, line 381: This file should be removed too?

native/templates/ios/AppDelegate.mm
```diff
@@ -55,11 +56,8 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
     [self.window setRootViewController:_viewController];
 
     [self.window makeKeyAndVisible];
-
-    _platform = dynamic_cast<cc::IOSPlatform *>(cc::BasePlatform::getPlatform());
-    CCASSERT(_platform != nullptr, "Platform pointer can't be null");
-    _platform->loop();
-
+    
+    [[SDKWrapper shared] application:application didFinishLaunchingWithOptions:launchOptions];
```
dumganhar: Do we have to re-use the logic of SDKWrapper? It seems that SDKWrapper and SDKDelegate should only take care of SDK things, however, now we use it for notifying app lifecycle event to engine too.

native/CMakeLists.txt
qiuguohua: Yes, I plan to put all the main in the template later. This allows users to extend the functionality of the platform。
[issue](https://github.com/cocos/cocos-engine/issues/10510)

native/templates/ios/AppDelegate.mm
minggo: Got it, should think about it in future.

native/templates/ios/AppDelegate.mm
qiuguohua: Yes, but think in another way, cocos is also an sdk for users. If the mechanism of SDKWrapper is not used, the notification event will be sent twice

native/templates/ios/AppDelegate.mm
minggo: SDK here means 3rd party libraries, such as advertisement sdk, pay sdk and such on. Engine is different from it.

native/templates/ios/AppDelegate.mm
minggo: I think we can keep the same logic as before before we have a good design.

qiuguohua: OK

dumganhar: Yes, I agree with what @minggo said.
We need to treat SDKWrapper as a standalone 3rd SDK.

https://github.com/cocos/cocos-engine/blob/v3.5.0/native/cocos/platform/ios/AppDelegate.mm#L72

I think it's ok to keep the old logic , just move AppDelegate.mm out of engine.

dumganhar: Or
Is it better to add an `objective-c` adaptor for synchronize system event to engine rather than use `_platform` C++ object directly?

Using `objective-c` code will be more easily to transfer event to engine. It will be similar with what SDKWrapper does but simpler.

// AppDelegate.mm (in template)
```objc
- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
    [[SDKWrapper shared] application:application didFinishLaunchingWithOptions:launchOptions];
    // Add the view controller's view to the window and display.
    CGRect bounds = [[UIScreen mainScreen] bounds];
    self.window   = [[UIWindow alloc] initWithFrame:bounds];

    // Should create view controller first, cc::Application will use it.
    _viewController                           = [[ViewController alloc] init];
    _viewController.view                      = [[View alloc] initWithFrame:bounds];
    _viewController.view.contentScaleFactor   = UIScreen.mainScreen.scale;
    _viewController.view.multipleTouchEnabled = true;
    [self.window setRootViewController:_viewController];

    [self.window makeKeyAndVisible];

    [[PlatformEventAdapter sharedAdapter] application:application didFinishLaunchingWithOptions:launchOptions];
    return YES;
}

- (void)applicationWillResignActive:(UIApplication *)application {
    [[SDKWrapper shared] applicationWillResignActive:application];
    [[PlatformEventAdapter sharedAdapter] applicationWillResignActive:application];
}

- (void)applicationDidBecomeActive:(UIApplication *)application {
    [[SDKWrapper shared] applicationDidBecomeActive:application];
    [[PlatformEventAdapter sharedAdapter] applicationDidBecomeActive:application];
}
```

// ViewController.mm (in template)
```objc
- (void)viewWillTransitionToSize:(CGSize)size withTransitionCoordinator:(id<UIViewControllerTransitionCoordinator>)coordinator {
    [[PlatformEventAdapter sharedAdapter] viewWillTransitionToSize:size withTransitionCoordinator:coordinator];
}
```

// PlatformEventAdapter.mm (`in engine`)
```objc
- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {

    _platform = dynamic_cast<cc::IOSPlatform *>(cc::BasePlatform::getPlatform());
    CCASSERT(_platform != nullptr, "Platform pointer can't be null");
    _platform->loop();

    return YES;
}

- (void)applicationWillResignActive:(UIApplication *)application {
    _platform->onPause();
}

- (void)applicationDidBecomeActive:(UIApplication *)application {
    _platform->onResume();
}

- (void)viewWillTransitionToSize:(CGSize)size withTransitionCoordinator:(id<UIViewControllerTransitionCoordinator>)coordinator {
    cc::IScreen::Orientation orientation;
    // reference: https://developer.apple.com/documentation/uikit/uiinterfaceorientation?language=objc
    // UIInterfaceOrientationLandscapeRight = UIDeviceOrientationLandscapeLeft
    // UIInterfaceOrientationLandscapeLeft = UIDeviceOrientationLandscapeRight
    switch ([UIDevice currentDevice].orientation) {
        case UIDeviceOrientationPortrait:
            orientation = cc::IScreen::Orientation::PORTRAIT;
            break;
        case UIDeviceOrientationLandscapeRight:
            orientation = cc::IScreen::Orientation::LANDSCAPE_LEFT;
            break;
        case UIDeviceOrientationPortraitUpsideDown:
            orientation = cc::IScreen::Orientation::PORTRAIT_UPSIDE_DOWN;
            break;
        case UIDeviceOrientationLandscapeLeft:
            orientation = cc::IScreen::Orientation::LANDSCAPE_RIGHT;
            break;
        default:
            break;
    }
    cc::DeviceEvent ev;
    cc::BasePlatform* platform = cc::BasePlatform::getPlatform();
    cc::IScreen* screenIntf = platform->getInterface<cc::IScreen>();
    ev.type           = cc::DeviceEvent::Type::DEVICE_ORIENTATION;
    ev.args[0].intVal = static_cast<int>(screenIntf->getDeviceOrientation());
    AppDelegate* delegate = [[UIApplication sharedApplication] delegate];
    [delegate dispatchEvent:ev];

    float pixelRatio = screenIntf->getDevicePixelRatio();
    cc::WindowEvent resizeEv;
    resizeEv.type = cc::WindowEvent::Type::RESIZED;
    resizeEv.width = size.width * pixelRatio;
    resizeEv.height = size.height * pixelRatio;
    [delegate dispatchEvent:resizeEv];
    //CAMetalLayer is available on ios8.0, ios-simulator13.0.
    CAMetalLayer *layer = (CAMetalLayer *)self.view.layer;
    CGSize tsize             = CGSizeMake(static_cast<int>(size.width * pixelRatio),
                                         static_cast<int>(size.height * pixelRatio));
    layer.drawableSize = tsize;
}
```

qiuguohua: If developers integrate cocos into their own apps as a so, isn't it like an SDK?

If a SDKWrapper cannot meet the needs of the SDK, will the code be directly extended to the appdelegate? Using SDKWrapper to dynamically monitor this method allows users to better connect and extend their own special methods

I will restore the old one first, but reserve my opinion

minggo: -  iOS does not support .so. 
- SDKWrapper is only served as builtin services, it is not a universal sdk wrapper. Developers will not use it to integrate other sdk.

minggo (approved):

PR #10515, opened 2022-03-26
profiler with surface transform

stanleyljl:
Re: cocos-creator/3d-tasks#

Changelog:
 * profiler with surface transform
 * dynamic mesh optimization for mobile vulkan


stanleyljl:
@cocos-robot run test cases

native/cocos/scene/Camera.h
```diff
@@ -310,6 +310,8 @@ class Camera : public RefCounted {
     inline float getScreenScale() const { return _screenScale; }
     inline void  setScreenScale(float val) { _screenScale = val; }
 
+    inline gfx::SurfaceTransform getSurfaceTransform() const { return _curTransform; }
```
minggo, line 313: Should skip it if do want to export to JS.

minggo: Oh, i found it is declared as a property in ts, so should modify corresponding .ini file to bind it as a property too. Or it is different from ts.

minggo (approved):

star-e (approved):

PR #10533, opened 2022-03-28
Adjust the default linewidth of cc.graphics

zhefengzhang:
Re: cocos-creator/3d-tasks#

Changelog:
 * Graphics to 4, because with a lineWidth of 2, the drawing effect is not observed at some resolutions. If this is not changed, the user may mistakenly think it is a bug.
 
 Before modification:
图片
Modified:
图片

### DIFF
--- a/cocos2d/core/graphics/graphics.js
+++ b/cocos2d/core/graphics/graphics.js
@@ -48,7 +48,7 @@ let Graphics = cc.Class({
     },
 
     properties: {
-        _lineWidth: 2,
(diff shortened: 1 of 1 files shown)


SantyWang:
This will have an impact on the compatibility

zhefengzhang:
> This will have an impact on the compatibility

Why?This just adjusts the default lineWidth.

zhefengzhang:
This change will affect user upgrades and is not recommended for long term maintenance releases.

PR #10541, opened 2022-03-28
remove CoreStd.h

minggo:


native/cocos/base/job-system/job-system-taskflow/TFJobSystem.h
```diff
@@ -28,6 +28,7 @@
 #include <algorithm>
 #include <thread>
 #include "taskflow/taskflow.hpp"
+#include "base/memory/Memory.h"
```
star-e, line 31: Is it better to include files from cocos?

star-e (approved):

native/cocos/base/job-system/JobSystem.h
```diff
@@ -23,6 +23,12 @@
  THE SOFTWARE.
 ****************************************************************************/
 
+#if CC_USE_JOB_SYSTEM_TBB
```
dumganhar: Doesn't need 
```c++
#define CC_JOB_SYSTEM_TASKFLOW 1
#define CC_JOB_SYSTEM_TBB      2
```
?

native/cocos/base/job-system/job-system-taskflow/TFJobSystem.h
minggo: Yep, it is better to include from cocos/ to avoid conflict.

hana-alice (approved):

native/cocos/base/job-system/JobSystem.h
minggo: It is defined in CMakeLists.txt, and it set through editor build setting.

dumganhar (approved):

SantyWang (approved):

PR #10551, opened 2022-03-28
Replacing the Observer Pattern with the Event Pattern

qiuguohua:
ref:[issue](https://github.com/cocos/3d-tasks/issues/11551)

### DIFF
--- a/native/cocos/application/CocosApplication.cpp
+++ b/native/cocos/application/CocosApplication.cpp
@@ -52,8 +52,16 @@ int CocosApplication::init() {
     if (_engine->init()) {
         return -1;
     }
-    auto callback = std::bind(&CocosApplication::handleAppEvent, this, std::placeholders::_1); // NOLINT(modernize-avoid-bind)
-    _engine->addEventCallback(OSEventType::APP_OSEVENT, callback);
+
+    _engine->on(BaseEngine::ON_RESUME, [this]() {
+        this->onResume();
+    });
+    _engine->on(BaseEngine::ON_PAUSE, [this]() {
+        this->onPause();
+    });
+    _engine->on(BaseEngine::ON_CLOSE, [this]() {
+        this->onClose();
+    });
 
     se::ScriptEngine *se = se::ScriptEngine::getInstance();
 
--- a/native/cocos/application/CocosApplication.cpp
+++ b/native/cocos/application/CocosApplication.cpp
@@ -46,7 +46,11 @@ CocosApplication::CocosApplication() {
     CCASSERT(_systemWidow != nullptr, "Invalid interface pointer");
 }
 
-CocosApplication::~CocosApplication() = default;
+CocosApplication::~CocosApplication() {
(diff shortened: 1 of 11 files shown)


native/cocos/core/event/CallbacksInvoker.h
```diff
@@ -0,0 +1,475 @@
...
+        _isCCObject = false;
+#if CC_DEBUG
+        _argTypes.clear();
+#endif
+    }
+
+    bool check() const override {
+        // Validation
+        //if (_isCCObject) {
+        //    if (!isObjectValid(reinterpret_cast<CCObject *>(_target), true)) {
```
qiuguohua: Comment here, because it will check whether to inherit CCObject

native/cocos/core/event/CallbacksInvoker.h
```diff
@@ -0,0 +1,475 @@
...
+#include <typeinfo>
+#include <unordered_map>
+#include <vector>
+
+#include "base/Log.h"
+#include "base/Macros.h"
+#include "base/Utils.h"
+
+namespace cc {
+class CCObject {
```
qiuguohua: An implementation of an empty CCObject is defined because the code depends on this class

minggo (approved):

native/cocos/application/CocosApplication.cpp
```diff
@@ -52,8 +52,16 @@ int CocosApplication::init() {
...
-    _engine->addEventCallback(OSEventType::APP_OSEVENT, callback);
+
+    _engine->on(BaseEngine::ON_RESUME, [this]() {
+        this->onResume();
+    });
+    _engine->on(BaseEngine::ON_PAUSE, [this]() {
+        this->onPause();
+    });
+    _engine->on(BaseEngine::ON_CLOSE, [this]() {
+        this->onClose();
```
dumganhar, line 67: Do we need to `off` the subscribing somewhere?

native/cocos/application/CocosApplication.cpp
```diff
@@ -52,8 +52,16 @@ int CocosApplication::init() {
     if (_engine->init()) {
         return -1;
     }
-    auto callback = std::bind(&CocosApplication::handleAppEvent, this, std::placeholders::_1); // NOLINT(modernize-avoid-bind)
-    _engine->addEventCallback(OSEventType::APP_OSEVENT, callback);
```
dumganhar, line 56: Could remove addEventCallback now?

qiuguohua: It cannot be removed, because it contains not only the state of the engine, but also some other window events that the user can handle. AppEvent can be removed

native/cocos/application/CocosApplication.cpp
qiuguohua: Thanks for pointing out the mistake, I forgot to think about it. This also considers the restart part.

dumganhar (approved):

PR #10552, opened 2022-03-28
v3.6.0 fix cc.ClickEvent target type error

arsen2010:
Re: cocos-creator/3d-tasks#

Changelog:
 *


VisualSJ (approved):

cocos/core/components/component-event-handler.ts
```diff
@@ -30,7 +30,7 @@
  */
 
 import { ccclass, type, serializable, editable, tooltip } from 'cc.decorator';
-import type { Node } from '../scene-graph';
+import { Node } from '../scene-graph';
```
minggo: Can not modify it, or it will meet error in native.
@dumganhar i remembered you modify it for native.

dumganhar: Yes, I remember that @shrinktofit suggested to add type keyword here. It seems to resolve cycle reference issues.

What errors did you get if the type keyword is not removed?

dumganhar (approved):

cocos/core/components/component-event-handler.ts
arsen2010: Error is  `@type(Node)`  would not work

VisualSJ (approved):

minggo (approved):

cocos/core/components/component-event-handler.ts
minggo: If there is not cycle reference issue. Then it is not problem.

shrinktofit (approved):

minggo (approved):

cocos/core/components/component-event-handler.ts
```diff
@@ -101,7 +101,6 @@ export class EventHandler {
      * 事件响应组件和函数所在节点
      */
     @serializable
-    @type(legacyCC.Node)
```
dumganhar, line 104: How about adding a comment to describe why we should not use `@type` decorator here and where the `type` information is registered ?

arsen2010: ok, i would add comments

dumganhar (approved):

cocos/core/components/component-event-handler.ts
shrinktofit: `@type(Node)` will introduce circular reference(_node_ -> _component_ -> _component-event-handler_ -> _node_) that's why our hack module _./component-event-handler.schema_ comes.

PR #10556, opened 2022-03-29
Modify the application interface name and remove the js keyword

qiuguohua:


native/cocos/application/ApplicationManager.h
```diff
@@ -87,7 +87,7 @@ class CC_DLL ApplicationManager {
         return app->run(argc, argv);                                         \
     } while (0)
 
-#define CC_REGISTER_GAME(className)               \
+#define CC_REGISTER_APPLICATION(className)               \
```
minggo, line 90: There is a `STARG_PLATFORM` in main.cpp, and this macro will `START_APPLICATION`, and unify them?

minggo: I mean can unify `STARG_PLATFORM ` and `CC_REGISTER_APPLICATION `?
We can require that, the class Name should be Game and we require there needs a `cocos_main`.

minggo: Then can only invoke one macro in main.cpp to setup.

qiuguohua: I've thought about this, and I tried it yesterday, and there will be a lot of problems. Because the current design is that the platform and the application are isolated.
(1) The platform layer needs to execute the main loop, in this case, the game needs to be registered in advance
(2) Some platforms do not have main (Android, ohos), so it is difficult to unify
(3) When running in server mode, an empty implementation of main is required to be built into the engine
So I still keep the original design，I will think about the design of this place sometime.

minggo: Ok. But we should keep compatible as possible.

dumganhar: @qiuguohua why revert to CC_REGISTER_APPLICATION? I remember you changed it to `CC_REGISTER_GAME`  yesterday.

minggo (approved):

native/cocos/application/ApplicationManager.h
qiuguohua: This is my mistake, because it is not appropriate to use the start application to move the main into the template layer, it is more appropriate to use the concept of registration. Yesterday, I brought the game as usual, but the inside is the application.

dumganhar (approved):

PR #10576, opened 2022-03-29
Fix cross fade bug introduced from #10485

shrinktofit:
Re: cocos-creator/3d-tasks#11781

Changelog:
 *

To reproduce the issue:
- Switch the animation 3 times(Idle -> Walk -> Run -> Idle), by default duration.
- Change duration to `0` and switch, the bug occurs.

### DIFF
--- a/cocos/core/animation/cross-fade.ts
+++ b/cocos/core/animation/cross-fade.ts
@@ -81,8 +81,6 @@ export class CrossFade {
      * @param duration 切换时间。
      */
     public crossFade (state: AnimationState | null, duration: number) {
-        this._finished = false;
-
         if (this._managedStates.length === 0) {
             // If we are cross fade from a "initial" pose,
             // we do not use the duration.
@@ -110,6 +108,9 @@ export class CrossFade {
             easeTime: 0,
             target,
         });
+
+        // Note: clear() set this to true
+        this._finished = false;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


cocos/core/animation/cross-fade.ts
```diff
@@ -81,8 +81,6 @@ export class CrossFade {
      * @param duration 切换时间。
      */
     public crossFade (state: AnimationState | null, duration: number) {
-        this._finished = false;
```
shrinktofit, line 84: BEFORE: Since `this.clear()` set `this._finished = true`, the `state` get no chance to update its weight.(`this.clear()` would be called if `duration === 0` or `this._managedStates.length === 0`, just like this issue does).

tests/animation/cross-fade.test.ts
```diff
@@ -98,5 +98,43 @@ describe('Cross fade', () => {
...
+        clipBar.duration = 0.4;
+        const foo = new AnimationState(clipFoo, 'foo');
+        foo.weight = 0.64;
+        const fooInitialWeight = foo.weight;
+        const bar = new AnimationState(clipBar, 'bar');
+        bar.weight = 0.37;
+        const barInitialWeight = bar.weight;
+
+        crossFade.crossFade(foo, 0.3);
+        expect(foo.weight).toBeCloseTo(fooInitialWeight);
```
shrinktofit, line 119: Still until `update()`

SantyWang (approved):

PR #10580, opened 2022-03-29
fix ios enter forground crash by useless appDelegateBridge

novain1102:
https://github.com/cocos/3d-tasks/issues/11793


qiuguohua (approved):
Sorry, it was my mistake, I missed it. 
There is currently no need to handle this event notification

minggo (approved):

PR #10596, opened 2022-03-31
3.6 Apple replace URLConnection with URLSession

PatriceJiang:
Resolve: https://github.com/cocos/3d-tasks/issues/4687

Changelog:
 * Replace `NSURLConnection` with `NSURLSession`


minggo:
Please make sure it works on iOS 10 as cocos supports iOS 10+.

minggo (approved):

PatriceJiang:
It has been verified that it works on iOS 10, since Downloader was implemented with NSURLSession.

novain1102 (approved):

PR #10624, opened 2022-04-01
change rich text split Long String warn

LinYunMo:
Re: https://github.com/cocos/3d-tasks/issues/11782


cocos/2d/components/rich-text.ts
```diff
@@ -695,7 +696,7 @@ export class RichText extends Component {
             }
         }
 
-        console.error(`try times = ${1000 - leftTryTimes}`);
+        log(`RichText split try times = ${1000 - leftTryTimes}`);
```
SantyWang: 用 logId 吧

SantyWang (approved):

PR #10643, opened 2022-04-02
Minor module refactor in core/renderer/

pandamicro:


cocos/core/renderer/scene/ambient.ts
```diff
@@ -24,10 +24,10 @@
  */
 
 import { JSB } from 'internal:constants';
-import { Color, Vec3, Vec4 } from '../../math';
+import { Vec4 } from '../../math';
 import { legacyCC } from '../../global-exports';
-import { AmbientInfo } from '../../scene-graph/scene-globals';
-import { NativeAmbient } from './native-scene';
+import type { AmbientInfo } from '../../scene-graph/scene-globals';
```
star-e, line 29: why type is used before { AmbientInfo }?

star-e (approved):

cocos/core/renderer/scene/ambient.ts
pandamicro: It will force the imported type to be used only as type indication, no other usage is allowed, it's better for module or type striping

SantyWang (approved):

PR #10645, opened 2022-04-02
Remove deprecated API in animation module

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 *


cocos/core/animation/deprecated.ts
```diff
@@ -1,63 +0,0 @@
...
-    },
-]);
-
-/**
- * Alias of [[Animation]]
- * @deprecated Since v1.2
- */
-export { Animation as AnimationComponent };
-legacyCC.AnimationComponent = Animation;
-js.setClassAlias(Animation, 'cc.AnimationComponent');
```
pandamicro, line 63: This kind of deprecation wasn't notifying user clearly, please first make the deprecation warning very clear in this version, then remove it in v3.6

pandamicro (approved):

PR #10633, opened 2022-04-01
fix memory leak

minggo:

PR #10649, opened 2022-04-02
Revert "fix memory leak"

minggo:
Reverts cocos/cocos-engine#10633


native/cocos/renderer/GFXDeviceManager.h
```diff
@@ -142,17 +142,15 @@ class CC_DLL DeviceManager final {
         Device *device = CC_NEW(DeviceCtor);
 
         if (DETACH_DEVICE_THREAD) {
-            CC_DELETE(device);
             device = CC_NEW(gfx::DeviceAgent(device));
```
minggo, line 145: device will be passed as a parameter

SantyWang (approved):

Issue #10663, opened 2022-04-06
iOS Simulator support for M1 Mac


timlyeee:
Right now if one would like to build one Xcode project that supports the M1 iOS Simulator, he should modify the CMakelist.txt in the external folder, which means that one cannot use the same project for building the M1 iOS Simulator App and Device App.

What produces this problem is that M1 Mac requires a different static lib for iOS Simulator(arm64), and lipo cannot create a fat lib for both arm64(device) and arm64(simulator). 

We do want to use the xcframework to support it, but the newest version of CMake is still working on it: https://gitlab.kitware.com/cmake/cmake/-/issues/21752

One reason is that `xcframework` is basically a folder containing different static-libs, so CMake should still target the library by using `-f` flag.

## Why should we use different static library for arm64(device) and arm64(simulator)? 

https://developer.apple.com/forums/thread/673387
So basically, Apple is starting to treat iOS Simulator and iOS as two different platforms.

timlyeee:
Even more, [Mac Catalyst](https://developer.apple.com/mac-catalyst/) requires a different static lib(arm64).

PR #10679, opened 2022-04-07
exclude arm64 arch and specify x86_64 for iosimulator by default

timlyeee:
This will ignore the link error for m1 mac when it tries to compile the application for the arm64 simulator

### Added:

It's actually not a solution, but a way to skip the problem that we might meet on M1 Mac, can keep watch with this [issue](https://github.com/cocos/cocos-engine/issues/10663) for more details.

Once we specify the architecture for the project, Xcode can only build apps for the x86_64 iPhone simulator and arm64 iPhone(up to now that's what we should support). 

Normally we should make it work by using xcframework, but since we build the project by CMake, it's hard to make it work.

Also, this modification will apply to cocos2d and app, otherwise link error will be produced.

### DIFF
--- a/native/templates/cmake/apple.cmake
+++ b/native/templates/cmake/apple.cmake
@@ -55,6 +55,13 @@ macro(cc_ios_after_target target_name)
         XCODE_ATTRIBUTE_SKIP_INSTALL NO
         XCODE_ATTRIBUTE_INSTALL_PATH "$(LOCAL_APPS_DIR)"
     )
+    ## exclude arm64 arch and specify x86_64 for iphonesimulator by default, this will apply to both target.
+    set(CMAKE_XCODE_ATTRIBUTE_EXCLUDED_ARCHS[sdk=iphonesimulator*] "arm64")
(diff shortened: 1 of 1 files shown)


native/templates/cmake/apple.cmake
```diff
@@ -55,6 +55,13 @@ macro(cc_ios_after_target target_name)
         XCODE_ATTRIBUTE_SKIP_INSTALL NO
         XCODE_ATTRIBUTE_INSTALL_PATH "$(LOCAL_APPS_DIR)"
     )
+    ## exclude arm64 arch and specify x86_64 for iphonesimulator by default, this will apply to both target.
```
minggo, line 58: please add the reason

minggo (approved):

native/templates/cmake/apple.cmake
timlyeee: Edited in the description of PR

PatriceJiang (approved):

novain1102 (approved):

PR #10682, opened 2022-04-07
Replace CCASSERT with CC_ASSERT

minggo:
CCASSERT is added like NSAssert, but not the same:
- all platforms but Android will not print the log, just breaks the the code line
- Android will not breaks, just print the log, which i will fix in other PR

I deleted most message as they are not need, but keep some meaningful message as comments.

### DIFF
--- a/native/cocos/math/Vec2.cpp
+++ b/native/cocos/math/Vec2.cpp
@@ -27,7 +27,7 @@
 NS_CC_MATH_BEGIN
 
 // returns true if segment A-B intersects with segment C-D. S->E is the overlap part
-bool isOneDimensionSegmentOverlap(float A, float B, float C, float D, float *S, float *E) {
+bool isOneDimensionSegmentOverlap(float A, float B, float C, float D, float *S, float *E) { //NOLINT(readability-identifier-naming)
     float ABmin = std::min(A, B);
     float ABmax = std::max(A, B);
     float CDmin = std::min(C, D);
@@ -55,7 +55,7 @@ bool isOneDimensionSegmentOverlap(float A, float B, float C, float D, float *S, 
 }
 
 // cross product of 2 vector. A->B X C->D
-float crossProduct2Vector(const Vec2 &A, const Vec2 &B, const Vec2 &C, const Vec2 &D) {
+float crossProduct2Vector(const Vec2 &A, const Vec2 &B, const Vec2 &C, const Vec2 &D) { // //NOLINT(readability-identifier-naming)
     return (D.y - C.y) * (B.x - A.x) - (D.x - C.x) * (B.y - A.y);
 }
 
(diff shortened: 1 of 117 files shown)


dumganhar (approved):

native/cocos/math/Vec2.cpp
```diff
@@ -26,37 +26,37 @@
...
-    float CDmin = std::min(C, D);
-    float CDmax = std::max(C, D);
-
-    if (ABmax < CDmin || CDmax < ABmin) {
+// returns true if segment a-b intersects with segment c-d. s->e is the overlap part
+bool isOneDimensionSegmentOverlap(float a, float b, float c, float d, float *s, float *e) {
+    float abmin = std::min(a, b);
+    float abmax = std::max(a, b);
+    float cdmin = std::min(c, d);
+    float cdmax = std::max(d, d);
```
SantyWang: seems logic error

native/cocos/math/Vec2.cpp
```diff
@@ -26,37 +26,37 @@
...
-            if (S != nullptr) *S = ABmin;
-            if (E != nullptr) *E = CDmax < ABmax ? CDmax : ABmax;
-        } else if (ABmax >= CDmin && ABmax <= CDmax) {
+            if (s != nullptr) *s = abmin;
+            if (e != nullptr) *e = cdmax < abmax ? cdmax : abmax;
+        } else if (abmax >= cdmin && abmax <= cdmax) {
             // ABmin->CDmin->ABmax->CDmax
-            if (S != nullptr) *S = CDmin;
-            if (E != nullptr) *E = ABmax;
+            if (s != nullptr) s = &cdmin;
```
SantyWang: why change pointer?

native/cocos/renderer/core/PassUtils.cpp
```diff
@@ -262,4 +262,45 @@ bool overrideMacros(MacroRecord &target, const MacroRecord &source) {
     return isDifferent;
 }
 
+MaterialProperty toMaterialProperty(gfx::Type type, const ccstd::vector<float> &vec) {
```
SantyWang, line 265: why change this?

minggo: Because the logic of https://github.com/cocos/cocos-engine/blob/v3.6.0/native/cocos/scene/Pass.cpp#L310 is error. It implicitly convert to Vec2. The data type may be Vec2, Vec3, Vec4 and mat4, just add a new function to convert it.

The error is trigged after i changed 
```c++
Vec2(const float *array);
```
to
```c++
explicit Vec2(const float *array);
```

native/cocos/math/Vec2.cpp
minggo: good catch, i fixed it

native/cocos/math/Vec2.cpp
minggo: good catch, i fixed it

SantyWang (approved):

PR #10684, opened 2022-04-07
add spline

stanleyljl:


cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,292 @@
...
+ * @zh
+ * 基础几何 Spline
+ */
+
+export class Spline {
+    private readonly _type: number;
+    private _mode: SplineMode = SplineMode.CATMULL_ROM;
+    private _knots: Vec3[] = [];
+
+    public constructor(mode: SplineMode = SplineMode.CATMULL_ROM, knots: Vec3[] = []) {
```
minggo: As there is a create method, should make the constructor public?

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,292 @@
...
+        }
+
+        return out;
+    }
+
+    get type () {
+        return this._type;
+    }
+
+    public getMode () {
```
minggo: Why not treat mode as a property?

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,292 @@
...
+        return new Spline(mode, knots);
+    }
+
+    public static clone (s: Spline) {
+        return new Spline(s.getMode(), s.getKnots());
+    }
+
+    public static copy (out: Spline, s: Spline) {
+        out._mode = s.getMode();
+        out._knots = [];
```
minggo: out._knots already exists, create a new one is not needed. May be can just do `out._knots.length = 0` to reuse the array.

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,292 @@
...
+
+        for (let i = 0; i < s.getKnots().length; i++) {
```
minggo: I always cache length like this, it seems is more efficient, but i am not sure
```TypeScript
for (let i = 0, len = s.getKnots().length;  i < len; i++)
```

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,292 @@
...
+            out._knots[i] = new Vec3(s.getKnots()[i]);
```
minggo: It is better to cache s.getKnots() outside for loop to be more efficient.

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,292 @@
...
+
+    public getMode () {
+        return this._mode;
+    }
+
+    public setMode (value) {
+        this._mode = value;
+    }
+
+    public getKnots () {
```
minggo: Why not treat knots as a property?

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,292 @@
...
+        return this._knots;
+    }
+
+    public setKnots (value) {
+        this._knots = [];
```
minggo: out._knots already exists, create a new one is not needed. May be can just do out._knots.length = 0 to reuse the array.

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,292 @@
...
+
+        for (let i = 0; i < value.length; i++) {
+            this._knots[i] = new Vec3(value[i]);
+        }
+    }
+
+    public clearKnots() { 
+        this._knots = [];
```
minggo: this._knots.length = 0 is better.

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,292 @@
...
+    public removeKnot(index: number) {
+        assertIsTrue(index < this._knots.length, "Spline: invalid index");
+
+        this._knots.splice(index, 1);
+    }
+
+    public setKnot(index: number, knot: Vec3) {
+        assertIsTrue(index < this._knots.length, "Spline: invalid index");
+
+        this._knots[index].set(knot);
```
minggo, line 164: `this._knots[index] = knot ` is better?

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    public getKnot(index: number) {
```
minggo: It is better to return a readonly Vec3, or the value may be modified outside.

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,292 @@
...
+        assertIsTrue(index < this._knots.length, "Spline: invalid index");
+
+        return this._knots[index];
+    }
+
+    // get a point at t with repect to the `index` segment of curve or the whole curve.
+    public getPoint(t: number, index: number = SPLINE_WHOLE_INDEX) {
```
minggo, line 174: It is better to return a readonly Vec3, or the value may be modified outside.

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,292 @@
...
+        t = clamp(t, 0.0, 1.0);
+
+        const segments = this.getSegments();
+        if (segments == 0) {
+            return new Vec3(0.0, 0.0, 0.0);
```
minggo, line 179: In TS, it is better to use global Vec3 defined in the file.

cocos/core/geometry/spline.ts
minggo: And what does t mean?

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,292 @@
...
+            return this._knots[this._knots.length - 1];
+        }
+
+        switch (this._mode) {
+            case SplineMode.LINEAR:
+                return Spline.calcLinear(this._knots[index], this._knots[index + 1], t);
+            case SplineMode.BEZIER:
+                return Spline.calcBezier(this._knots[index * 4], this._knots[index * 4 + 1], this._knots[index * 4 + 2], this._knots[index * 4 + 3], t);
+            case SplineMode.CATMULL_ROM: {
+                const v0 = index > 0 ? this._knots[index - 1] : this._knots[index];
```
minggo, line 200: What will happen if index == 0?

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,292 @@
...
+
+        if (index >= segments) {
+            return this._knots[this._knots.length - 1];
+        }
+
+        switch (this._mode) {
+            case SplineMode.LINEAR:
+                return Spline.calcLinear(this._knots[index], this._knots[index + 1], t);
+            case SplineMode.BEZIER:
+                return Spline.calcBezier(this._knots[index * 4], this._knots[index * 4 + 1], this._knots[index * 4 + 2], this._knots[index * 4 + 3], t);
```
minggo, line 197: How to make sure `index * 4 + 3` will not exceed the size?

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,292 @@
...
+                const v3 = index + 2 < this._knots.length ? this._knots[index + 2] : this._knots[index + 1];
+                return Spline.calcCatmullRom(v0, this._knots[index], this._knots[index + 1], v3, t);
+            }
+            default:
+                return new Vec3(0.0, 0.0, 0.0);
+        }
+    }
+
+    // get num points from 0 to 1 uniformly with repect to the `index` segment of curve or the whole curve
+    public getPoints(num: number, index: number = SPLINE_WHOLE_INDEX): Vec3[] {
```
minggo, line 209: It seems this method will create a Vec3[] every time. It is time consuming. It is better to avoid it.

cocos/core/pipeline/geometry-renderer.ts
```diff
@@ -897,6 +902,25 @@ export class GeometryRenderer {
         }
     }
 
+    public addSpline (spline: Spline, color: Color, index = 0xffffffff, knotSize = 0.5, segments = 32, depthTest = true) {
```
minggo, line 905: Why default value of  index is 0xffffffff? It is a big number.

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,292 @@
...
+    get type () {
+        return this._type;
+    }
+
+    public getMode () {
+        return this._mode;
+    }
+
+    public setMode (value) {
+        this._mode = value;
```
minggo: Can change mode in runtime? Will it have correct effect?

cocos/core/geometry/spline.ts
stanleyljl: This is consistent with other geometry.

cocos/core/geometry/spline.ts
stanleyljl: ok

cocos/core/geometry/spline.ts
stanleyljl: This can be a bit of a problem, two variables refer to the same data.

cocos/core/geometry/spline.ts
stanleyljl: 1, The result is an interpolated value, it does not have to be read-only. 
2, t is an interpolation coefficient. t goes from 0 to 1 to get points on the entire curve.

cocos/core/geometry/spline.ts
stanleyljl: This method allows the user to modify the return value

cocos/core/geometry/spline.ts
stanleyljl: Bezier mode requires to add multiple of 4 knots to spline.

cocos/core/geometry/spline.ts
minggo: It is a new class, can use correct way to handle it. And it doesn't break anything.

cocos/core/pipeline/geometry-renderer.ts
stanleyljl: It is -1, declare as an uint32_t in native

cocos/core/geometry/spline.ts
stanleyljl: The user need to know what they are doing. There may be a need, the user can reuse the spline.

cocos/core/geometry/spline.ts
minggo: Yep, you are right.

cocos/core/geometry/spline.ts
minggo: The logic is the same.

cocos/core/geometry/spline.ts
minggo: The logic is the same.

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,291 @@
...
+
+    get type () {
+        return this._type;
+    }
+
+    get mode () {
+        return this._mode;
+    }
+
+    get knots () {
```
minggo: Add readonly to return a readonly `Vec3[]`? Or `this._knots` may be modified outside.

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,291 @@
...
+        return this._knots.length; 
+    }
+
+    public addKnot(knot: Vec3) { 
+        this._knots.push(new Vec3(knot)); 
+    }
+
+    public insertKnot(index: number, knot: Vec3) {
+        const item = new Vec3(knot);
+        if (index >= this._knots.length) {
```
minggo, line 147: Is it better to print a message? It seems it is an unexpected behavior.

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,291 @@
...
+            this._knots.push(item);
+            return;
+        }
+    
+        this._knots.splice(index, 0, item);
+    }
+
+    public removeKnot(index: number) {
+        assertIsTrue(index < this._knots.length, "Spline: invalid index");
```
minggo: Should also make sure `index >= 0`?

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,291 @@
...
+
+        this._knots.splice(index, 1);
+    }
+
+    public setKnot(index: number, knot: Vec3) {
+        assertIsTrue(index < this._knots.length, "Spline: invalid index");
```
minggo: The same as above.

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,291 @@
...
+
+    public getKnot(index: number): Readonly<Vec3> {
+        assertIsTrue(index < this._knots.length, "Spline: invalid index");
+
+        return this._knots[index];
+    }
+
+    // get a point at t with repect to the `index` segment of curve or the whole curve.
+    public getPoint(t: number, index: number = SPLINE_WHOLE_INDEX) {
+        t = clamp(t, 0.0, 1.0);
```
minggo, line 175: Should check the value of `t`? What will happen if t > 1 or t < 0?

cocos/core/geometry/spline.ts
```diff
@@ -0,0 +1,291 @@
...
+            return new Vec3(this._knots[this._knots.length - 1]);
+        }
+
+        switch (this._mode) {
+            case SplineMode.LINEAR:
+                return Spline.calcLinear(this._knots[index], this._knots[index + 1], t);
+            case SplineMode.BEZIER:
+                return Spline.calcBezier(this._knots[index * 4], this._knots[index * 4 + 1], this._knots[index * 4 + 2], this._knots[index * 4 + 3], t);
+            case SplineMode.CATMULL_ROM: {
+                const v0 = index > 0 ? this._knots[index - 1] : this._knots[index];
```
minggo, line 199: How about `index < 0`?

cocos/core/pipeline/geometry-renderer.ts
minggo: How about define invalidIndex to replace 0xffffffff?

cocos/core/geometry/spline.ts
stanleyljl: it is an unsigned int

native/cocos/renderer/pipeline/GeometryRenderer.cpp
```diff
@@ -810,6 +811,25 @@ void GeometryRenderer::addBezier(const Vec3 &v0, const Vec3 &v1, const Vec3 &v2,
...
 
+void GeometryRenderer::addSpline(const geometry::Spline &spline, gfx::Color color, uint32_t index, float knotSize, uint32_t segments, bool depthTest) {
+    const auto numPoints = segments + 1;
+    auto       points    = spline.getPoints(numPoints, index);
+
+    for (auto i = 0U; i < segments; i++) {
+        addLine(points[i], points[i + 1], color, depthTest);
+    }
+
+    if (knotSize > 0.0F && index == 0xffffffff) {
```
minggo: It is better to use SPLINE_WHOLE_INDEX instead of 0xffffffff here.

cocos/core/geometry/spline.ts
minggo: but in ts, there is not unsigned int

cocos/core/geometry/spline.ts
stanleyljl: t is an interpolation coefficient. It makes no sense to go beyond this range

native/cocos/renderer/pipeline/GeometryRenderer.cpp
stanleyljl: There is some problem when exporting this value in jsb

minggo: It is in a cpp file, how can it affect jsb?

stanleyljl: 1

minggo (approved):

minggo (approved):

PR #10692, opened 2022-04-08
fix instanced rendering issue by joint index

hana-alice:
Re: cocos-creator/3d-tasks#

Changelog:
 *


hana-alice:
https://github.com/cocos/3d-tasks/issues/11804

minggo (approved):

PR #10709, opened 2022-04-09
Fix native module override

pandamicro:
caused by [#10643](https://github.com/cocos-creator/engine/issues/10643)


pandamicro:
ESLint ci fail is a ci bug, can be ignored

minggo (approved):

Issue #10710, opened 2022-04-09
Design flaw, conflicts between initializing the default platform and creating a custom platform


qiuguohua:
If a platform is customized, the getplatform interface returns a pointer to the default platform, which is an error

qiuguohua:
This is a design defect and a bug. At present, the platform is placed in main, so that developers can customize some interfaces and expand existing interfaces by inheriting and overloading.

However, when users do it through inheritance, it will conflict with the platform created by default. This was not taken into account at that time

PR #10711, opened 2022-04-09
Design flaw, conflicts between initializing the default platform and …

qiuguohua:
https://github.com/cocos/cocos-engine/issues/10710

### DIFF
--- a/native/cocos/platform/BasePlatform.cpp
+++ b/native/cocos/platform/BasePlatform.cpp
@@ -43,10 +43,18 @@
 #endif
 
 namespace cc {
-BasePlatform::BasePlatform()  = default;
-BasePlatform::~BasePlatform() = default;
+BasePlatform* BasePlatform::_currentPlatform = nullptr;
 
-BasePlatform* BasePlatform::getPlatform() {
+BasePlatform::BasePlatform() {
+    CCASSERT(_currentPlatform == nullptr, "The platform has been initialized, only one platform can be initialized");
+    _currentPlatform = this;
+}
+
+BasePlatform::~BasePlatform() {
+    _currentPlatform = nullptr;
+}
+
+BasePlatform* BasePlatform::createDefaultPlatform() {
 #if defined(CC_SERVER_MODE)
     static EmptyPlatform platform;
 #elif (CC_PLATFORM == CC_PLATFORM_WINDOWS)
@@ -66,4 +74,14 @@ BasePlatform* BasePlatform::getPlatform() {
 #endif
     return &platform;
 }
+
+BasePlatform* BasePlatform::getPlatform() {
+    if (_currentPlatform) {
+        return _currentPlatform;
+    }
+    createDefaultPlatform();
+    CCASSERT(_currentPlatform != nullptr, "Need to point to the current platform");
+    return _currentPlatform;
+}
+
 } // namespace cc
--- a/native/cocos/platform/BasePlatform.h
+++ b/native/cocos/platform/BasePlatform.h
@@ -155,6 +155,8 @@ public:
     }
 
 private:
+    static BasePlatform* createDefaultPlatform();
+    static BasePlatform* _currentPlatform;
     std::vector<OSInterface::Ptr> _osInterfaces;
     CC_DISABLE_COPY_AND_MOVE_SEMANTICS(BasePlatform);
 };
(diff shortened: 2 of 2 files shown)


native/cocos/platform/BasePlatform.cpp
```diff
@@ -43,10 +43,18 @@
...
+    _currentPlatform = this;
+}
+
+BasePlatform::~BasePlatform() {
+    _currentPlatform = nullptr;
+}
+
+BasePlatform* BasePlatform::createDefaultPlatform() {
 #if defined(CC_SERVER_MODE)
     static EmptyPlatform platform;
```
timlyeee, line 59: So customize platform is not allowed in Server mode ?

native/cocos/platform/BasePlatform.cpp
```diff
@@ -43,10 +43,18 @@
...
+BasePlatform::BasePlatform() {
+    CCASSERT(_currentPlatform == nullptr, "The platform has been initialized, only one platform can be initialized");
+    _currentPlatform = this;
+}
+
+BasePlatform::~BasePlatform() {
+    _currentPlatform = nullptr;
+}
+
+BasePlatform* BasePlatform::createDefaultPlatform() {
```
timlyeee, line 57: void BasePlatform::createDefaultPlatform;

might be better as no where use it.

timlyeee (changes requested):

native/cocos/platform/BasePlatform.cpp
```diff
@@ -66,4 +74,14 @@ BasePlatform* BasePlatform::getPlatform() {
 #endif
     return &platform;
 }
-} // namespace cc
\ No newline at end of file
+
+BasePlatform* BasePlatform::getPlatform() {
```
minggo, line 78: getPlatform -> getInstance the same rule as other module.

native/cocos/platform/BasePlatform.cpp
```diff
@@ -66,4 +74,14 @@ BasePlatform* BasePlatform::getPlatform() {
...
+    if (_currentPlatform) {
+        return _currentPlatform;
+    }
+    createDefaultPlatform();
```
minggo, line 82: Is `createDefaultPlatform ` needed?

qiuguohua: Yes, this is to work with developers using custom platforms.
At the template layer users can use:
```
class  CustomMacPlatform : public MacPlatform {
public:
     int run() override;
}
```
in the main function：
```
int main( ) {
     CustomMacPlatform platform;
     platform.init();
     return platform.start();
}
```

native/cocos/platform/BasePlatform.cpp
qiuguohua: getPlatform is a static member method, and getInterface is an ordinary member method. Two are not the same. Most of the time, the interface is obtained through the engine, reducing the direct dependence on the platform

native/cocos/platform/BasePlatform.cpp
qiuguohua: Yes here, but better is to assign to currentPlatform. But because of the assignment in the constructor (to avoid letting the user set it). It feels hard to understand here. In fact, is it okay for me to assign a value here, although there is redundancy？

qiuguohua: This will not be modified, it will be strange if the inside of this function is modified. Let's see how this place is handled

native/cocos/platform/BasePlatform.cpp
minggo: In cocos, always use getInstance to get singleton object.

native/cocos/platform/BasePlatform.cpp
minggo: It is a private static function, what's the relationship with developer's codes?

qiuguohua: If the user needs to extend the platform function, there is no need to create it internally

native/cocos/platform/BasePlatform.cpp
qiuguohua: Don't have to use this method? Wouldn't it be better to use more direct semantics? Or change to getCurrentPlatform?

minggo: getInstance has the meaning to get a singleton object.

minggo (approved):

PR #10712, opened 2022-04-09
V3.5.0  Distribute touch events independently

qiuguohua:
https://github.com/cocos/cocos-engine/issues/10677


minggo (approved):

dumganhar:
What does this PR resolve for?

qiuguohua:
> What does this PR resolve for?

The touch events mentioned earlier need to be sent separately. May affect the performance of touch events

dumganhar:
If it's relevant to performance, do you mean that all OSEvent types need to be dynamic_cast to its sub type?
```c++
    template <typename T>
    std::enable_if_t<std::is_base_of<cc::OSEvent, T>::value, const T &> static castEvent(const cc::OSEvent &ev) {
        const T &evDetail = dynamic_cast<const T &>(ev);  // I think here we could use static_cast for the conversion since we have already checked event type before calling castEvent
        return std::move(evDetail);
    }
```

If touch event needs to be handled like this, how about other events, e.g. MouseEvent which will be triggered frequently.

dumganhar:
And please update the PR title, the title `V3.5.0 touch` is confused.

qiuguohua:
> If it's relevant to performance, do you mean that all OSEvent types need to be dynamic_cast to its sub type?

I feel that it should be mixed with other events to cause the touch event to be too late to process
Should be in embedded devices, this situation is more obvious? Shouldn't this be the case on a computer?

qiuguohua:
> And please update the PR title, the title V3.5.0 touch is confused.

Okay, I'm in a hurry. I mentioned it in the issue above

dumganhar (approved):

PR #10713, opened 2022-04-09
fix the performance issue caused by texture mipmap

yiwenxue:
Re: cocos-creator/3d-tasks#

Changelog:
 * 

The default mipmap level range of `Texture2D` and `TextureCube` was [0, 1), which caused a serious performance issue.
This pr changed it so that all mipmaps will be included by default.

### DIFF
--- a/cocos/core/assets/simple-texture.ts
+++ b/cocos/core/assets/simple-texture.ts
@@ -76,7 +76,7 @@ export class SimpleTexture extends TextureBase {
     private _textureHeight = 0;
 
     protected _baseLevel = 0;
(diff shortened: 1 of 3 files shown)


minggo:
Does it fix the performance issue?

yiwenxue:
> Does it fix the performance issue?

I did a simple test on windows (Vulkan and GLES3), but I didn't find any significant improvement. It's weird.

star-e (approved):

yiwenxue:
@cocos-robot run test cases

minggo (approved):

PR #10716, opened 2022-04-10
bypass the bug that vulkan destroys resources that are still in use.

yiwenxue:
Re: cocos-creator/3d-tasks#11829

Changelog:
 * 

This pr bypass the bug that Vulkan destroys resources that are still in use. If the pointer to the texture view equals the pointer to the previous one, the descriptor will not be updated. So creating a new resource before destroying the old one will solve the problem. It is a temporary solution.

### DIFF
--- a/cocos/core/assets/simple-texture.ts
+++ b/cocos/core/assets/simple-texture.ts
@@ -226,8 +226,11 @@ export class SimpleTexture extends TextureBase {
         if (!device) {
             return;
         }
+        // create a new texture view before the destruction of the previous one to bypass the bug that
+        // vulkan destroys textureview in use. This is a temporary solution, should be fixed later.
+        const textureView = this._createTextureView(device);
         this._tryDestroyTextureView();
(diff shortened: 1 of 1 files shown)


yiwenxue:
@cocos-robot run test cases

cocos/core/assets/simple-texture.ts
```diff
@@ -226,8 +226,11 @@ export class SimpleTexture extends TextureBase {
         if (!device) {
             return;
         }
+        // create a new texture view before the destruction of the previous one to bypass the bug that
+        // vulkan destroys textureview in use. This is a temporary solution, should be fixed later.
+        const textureView = this._createTextureView(device);
```
star-e, line 231: Since_createTextureView now can be null, public getGFXTexture () will return null.
Is this correct behabior?

star-e (approved):

minggo (approved):

PR #10722, opened 2022-04-11
Fixed: Return type of `set` is `any` in vec3 and some types

k8w:
Re: cocos-creator/3d-tasks#

Changelog:
 * Fixed: Return type of `set` method is `any` in `Vec2` `Vec3` `Vec4` `Quat` `Mat3` and `Mat4`


pandamicro (approved):

pandamicro:
Thanks, it will be released with v3.5

PR #10728, opened 2022-04-11
The instantiation of animation graph should be done at loading phase

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 *

### DIFF
--- a/cocos/core/animation/marionette/animation-controller.ts
+++ b/cocos/core/animation/marionette/animation-controller.ts
@@ -37,7 +37,7 @@ export class AnimationController extends Component {
 
     private _graphEval: AnimationGraphEval | null = null;
 
-    public start () {
+    public onLoad () {
         if (this.graph) {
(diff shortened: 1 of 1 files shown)


cocos/core/animation/marionette/animation-controller.ts
```diff
@@ -37,7 +37,7 @@ export class AnimationController extends Component {
 
     private _graphEval: AnimationGraphEval | null = null;
 
-    public start () {
+    public onLoad () {
```
SantyWang: What's problem if it's in start phase

shrinktofit: Many applications invoke animation controller API at their start method.

SantyWang (approved):

cocos/core/animation/marionette/animation-controller.ts
SantyWang: User maybe invoke these API  in `onLoad` method, so `__preload` seems better

Issue #10740, opened 2022-04-12
Delete redundant files related to QNX


qiuguohua:
Delete QNX template directory(engine\native\templates\qnx)

The qnx platform has been removed in v3.5, this file is not used, so it can be deleted without affecting any functionality

PR #10741, opened 2022-04-12
Delete qnx related redundant files

qiuguohua:
The qnx platform has been removed in v3.5, this file is not used, so it can be deleted without affecting any functionality


qiuguohua:
https://github.com/cocos/cocos-engine/issues/10740

minggo (approved):

PR #10745, opened 2022-04-13
merge docs and licenses and delete useless code

SantyWang:
Re: cocos-creator/3d-tasks#

Changelog:
 *


docs/NATIVE_LINTER_AUTOFIX_GUIDE.md
```diff
@@ -1,133 +1,133 @@
-# Linter Auto-fix Guide
```
minggo: How about use CPP_LINTER_AUTOFIX_GUIDE instead?

minggo:
- how about use CPP_CODING_STYLE.md instead?
- some licenses are not need
  - LICENSE_CCControlExtension.txt: it is added for cocosbuilder support which is not need now
  - LICENSE_chipmunk.txt: we don't use the chipmunk physics engine now
  - LICENSE_cocos2d-x.txt: it is a license for cocos2d-x, not need now
  - LICENSE_js.txt: it is the license of cocos2d-js, not need now
 
LICENSE_OpenSans.txt: i don't remember we use it. @dumganhar do you know?

minggo (approved):

knoxHuang (approved):

dumganhar:
> LICENSE_OpenSans.txt: i don't remember we use it. @dumganhar do you know?
@minggo I don't know. it seems to be some fonts from https://fonts.google.com/specimen/Open+Sans
Screen Shot 2022-04-13 at 14 33 20

PR #10750, opened 2022-04-13
make sensor usage can be config

bofeng-song:


minggo:
I think we should enable/disable them in runtime.

PR #10753, opened 2022-04-14
Combine hash with validator/agent

timlyeee:
While the texture is destroyed, the old validator is destroyed and new validator inited.

But framebuffer validator maintains the old one as hash is the same, so when it start the validation check, the old texture validator jumped out and assertion failed.

### DIFF
--- a/native/cocos/renderer/gfx-base/GFXDef.cpp
+++ b/native/cocos/renderer/gfx-base/GFXDef.cpp
@@ -113,12 +113,14 @@ bool operator==(const RenderPassInfo& lhs, const RenderPassInfo& rhs) {
 template <>
 size_t Hasher<FramebufferInfo>::operator()(const FramebufferInfo& info) const {
     // render pass is mostly irrelevant
-    size_t seed = (info.colorTextures.size() + 1) * 2;
+    size_t seed = (info.colorTextures.size() + 1) * 3;
     for (auto* colorTexture : info.colorTextures) {
+        boost::hash_combine(seed, colorTexture);
         boost::hash_combine(seed, colorTexture->getRaw());
         boost::hash_combine(seed, colorTexture->getHash());
     }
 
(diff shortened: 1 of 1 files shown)


timlyeee:
## Discussion:

### Is it OK to hash with validator/agent and raw as 2 hash keys?

 I think that won't make error as the combination of Texture Agent and Texture right here is always alive and correct. 

### Is it OK if we disable agent and validator?
 It will hash combine Texture twice.

### Any other solution？

The problem here is that actually FramebufferValidator holds an incorrect Texture Validator while Framebuffer is actually good. But CommandBufferValidator still check the validation here. It's more or less like:

FramebufferValidator(which also might be out of date) ===check====> TextureValidator(which *is* out of date) ==get texture==>NO WHERE
Framebuffer(correct) ====hold====>  Texture(correct)
So Who ====check====>Texture ?

native/cocos/renderer/gfx-base/GFXDef.cpp
```diff
@@ -113,12 +113,14 @@ bool operator==(const RenderPassInfo& lhs, const RenderPassInfo& rhs) {
...
     // render pass is mostly irrelevant
-    size_t seed = (info.colorTextures.size() + 1) * 2;
+    size_t seed = (info.colorTextures.size() + 1) * 3;
     for (auto* colorTexture : info.colorTextures) {
+        boost::hash_combine(seed, colorTexture);
         boost::hash_combine(seed, colorTexture->getRaw());
         boost::hash_combine(seed, colorTexture->getHash());
     }
 
+    boost::hash_combine(seed, info.depthStencilTexture);
```
star-e, line 123: info.depthStencilTexture might be null，we need to add if to check null

star-e (approved):

hana-alice (approved):

minggo (approved):

PR #10770, opened 2022-04-14
disable sensor by default,  game developer can enable it by set 'acce…

bofeng-song:


native/cocos/platform/android/java/src/com/cocos/lib/CocosSensorHandler.java
```diff
@@ -67,13 +69,18 @@ public CocosSensorHandler(final Context context) {
     // Getter & Setter
     // ===========================================================
     public void enable() {
-        mSensorManager.registerListener(this, mAcceleration, mSamplingPeriodUs);
-        mSensorManager.registerListener(this, mAccelerationIncludingGravity, mSamplingPeriodUs);
-        mSensorManager.registerListener(this, mGyroscope, mSamplingPeriodUs);
+        if (mEnableSensor) {
+            mHasRegisterListener = true;
```
minggo: Why need mHasRegisterListener? It seems the same effect as mEnableSensor.

native/cocos/platform/android/java/src/com/cocos/lib/CocosSensorHandler.java
```diff
@@ -67,13 +69,18 @@ public CocosSensorHandler(final Context context) {
...
+            mSensorManager.registerListener(this, mAcceleration, mSamplingPeriodUs);
+            mSensorManager.registerListener(this, mAccelerationIncludingGravity, mSamplingPeriodUs);
+            mSensorManager.registerListener(this, mGyroscope, mSamplingPeriodUs);
+        }
     }
 
     public void disable() {
-        this.mSensorManager.unregisterListener(this);
+        if (mHasRegisterListener) {
```
minggo: enable() is controlled by mEnableSensor, and this function is controlled by mHasRegisterListener. They are not consistent.

native/cocos/platform/android/java/src/com/cocos/lib/CocosSensorHandler.java
bofeng-song: App maybe crash, while disable sensor, but call unregisterListener. Need to be recheck.

native/cocos/platform/android/java/src/com/cocos/lib/CocosSensorHandler.java
bofeng-song: App maybe crash, while disable sensor, but call unregisterListener. Need to be recheck.

native/cocos/platform/android/java/src/com/cocos/lib/CocosSensorHandler.java
minggo: But it is only set to true here. And when it is set to true, no place to set it false.

bofeng-song: The recheck result is unregisterListener can be call multi-times. I have fix it

bofeng-song: unregister implement is:

synchronized (mSensorListeners) {
            SensorEventQueue queue = mSensorListeners.get(listener);
            if (queue != null) {
                boolean result;
                if (sensor == null) {
                    result = queue.removeAllSensors();
                } else {
                    result = queue.removeSensor(sensor, true);
                }
                if (result && !queue.hasSensors()) {
                    mSensorListeners.remove(listener);
                    queue.dispose();
                }
            }
        }
    }

minggo (approved):

PR #10775, opened 2022-04-15
fix vulkan crash while runing in background

bofeng-song:


native/cocos/renderer/gfx-vulkan/VKSwapchain.cpp
```diff
@@ -258,6 +258,9 @@ void CCVKSwapchain::doResize(uint32_t width, uint32_t height, SurfaceTransform /
 }
 
 bool CCVKSwapchain::checkSwapchainStatus(uint32_t width, uint32_t height) {
+    if (_gpuSwapchain->vkSurface == VK_NULL_HANDLE) { // vkSurface will set to VK_NULL_HANDLE after call doDestroySurface
```
minggo: will set -> will be set.

Why this function will be invoked after surface is destroyed.

minggo (approved):

native/cocos/renderer/gfx-vulkan/VKSwapchain.cpp
bofeng-song: called from ts

minggo (approved):

star-e:
Is this problem related to https://github.com/cocos/cocos-engine/pull/10776 ?

bofeng-song:
> Is this problem related to #10776 ?

No, #10776 is trigger by restart Application, but this pr fix running at background.

minggo:
Oh, sorry, it is my mistake. I thought @star-e agree with it.
@star-e if you do not agree with it, i can revert it.

star-e:
I'm not very familiar with Android surface and swapchain. I think this fix is ok. @minggo

minggo:
Ok, thanks.

PR #10780, opened 2022-04-15
Add AnimationMask.joints setter

shrinktofit:
Re: cocos-creator/3d-tasks#

Changelog:
 *

### DIFF
--- a/cocos/core/animation/marionette/animation-mask.ts
+++ b/cocos/core/animation/marionette/animation-mask.ts
@@ -15,6 +15,7 @@ export class AnimationMask extends Asset {
     }
 
     set joints (value) {
+        this.clear();
         for (const joint of value) {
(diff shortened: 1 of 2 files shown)


cocos/core/animation/marionette/animation-mask.ts
```diff
@@ -7,13 +7,20 @@ import { CLASS_NAME_PREFIX_ANIM } from '../define';
...
     private _jointMasks: JointMask[] = [];
 
+    @editable
     get joints (): Iterable<JointMaskInfo> {
         return this._jointMasks;
     }
 
+    set joints (value) {
+        this.clear();
+        for (const joint of value) {
```
shrinktofit, line 19: It's an iterable so we should use for-of

arsen2010 (approved):

SantyWang (approved):

PR #10793, opened 2022-04-18
Use gradle to generate compile-commands.json

PatriceJiang:
Re: cocos-creator/3d-tasks#

Changelog:
 *


PatriceJiang:
included in https://github.com/cocos/cocos-engine/pull/10794

PR #10794, opened 2022-04-18
integrate game-activity

bofeng-song:


.github/workflows/native-linter-android.yml
```diff
@@ -68,12 +68,47 @@ jobs:
...
+          # To speedup CI, compile arm64-v8a only
+          # sed -i "s@^PROP_APP_ABI.*@PROP_APP_ABI=armeabi-v7a:arm64-v8a:x86@g" gradle.properties
+          sed -i "s@^PROP_APP_ABI.*@PROP_APP_ABI=arm64-v8a@g" gradle.properties
 
+          echo "Compile Android - cmake ..."
+          echo "ANDORID_NDK ${ANDROID_NDK} or ${ANDROID_NDK_HOME}"
+          ./gradlew :CocosGame:assembleDebug --quiet
+          echo "Compile Android Debug Done!"
+          echo "Copy Compile database"
+          cp "$(find . -name compile_commands.json)" ${COCOS2DX_ROOT}/
```
minggo: what do these codes do?

bofeng-song: use gradle to download third library

native/templates/android/template/app/cocos_engine.gradle
```diff
@@ -0,0 +1,11 @@
+
+android {
```
minggo: native/templates is moved to cocos-engine/templates. And why add this file?

bofeng-song: game-activity develop earlier than last merged pr. I have fixed it

minggo (approved):

native/cocos/platform/android/jni/JniCocosEntry.cpp
```diff
@@ -23,10 +23,30 @@
...
+extern "C" {
+
+void android_main(struct android_app *app) {
+    cc::FileUtilsAndroid::setassetmanager(app->activity->assetManager);
+
+    auto *platform = cc::BasePlatform::getPlatform();
+    auto *androidPlatform = static_cast<cc::AndroidPlatform *>(platform);
+    androidPlatform->setAndroidApp(app);
+    platform->init();
+    platform->loop();
```
qiuguohua, line 43: Is it possible for us to use 'run' as the main entry, and run the loop inside. 'app' can also be passed in via argc and argv

native/cocos/platform/android/AndroidPlatform.cpp
```diff
@@ -25,79 +25,471 @@
...
         }
-        //std::this_thread::sleep_for(std::chrono::nanoseconds(100));
+
+        void setActiveGameControllerIndex(const int32_t controllerIndex) {
+            _gameControllerIndex = controllerIndex;
+        }
+
+        void handleAppCommand(int32_t cmd) {
+            switch (cmd) {
+                case APP_CMD_SAVE_STATE:
```
qiuguohua, line 260: Where are these commands issued? I don't seem to see
One more question, does HarmonyOS also support GameActivity?
I remember HarmonyOS was very similar to Android before

bofeng-song: APP_CMD_SAVE_STATE is the app lifecycle`s cmd, we just leave this code here.
HarmonyOS doesn't support GameActivity, It's package name doesn't support.

native/cocos/platform/android/jni/JniCocosEntry.cpp
qiuguohua: This entry is the same as other platforms, whether it is also considered to be placed in the template layer. What if the developer wants to extend the interface of the platform?

native/cocos/platform/android/AndroidPlatform.cpp
qiuguohua: Got it

PatriceJiang (approved):

native/cocos/platform/android/jni/JniCocosEntry.cpp
bofeng-song: > This entry is the same as other platforms, whether it is also considered to be placed in the template layer. What if the developer wants to extend the interface of the platform?

Can not move to template layer.  cocos_main will trigger renderer initialize. The old code call cocos_main after ANativeWindow is being set.

bofeng-song: > Is it possible for us to use 'run' as the main entry, and run the loop inside. 'app' can also be passed in via argc and argv

struct android_app ** can not be force cast to const char**

bofeng-song: > Is it possible for us to use 'run' as the main entry, and run the loop inside. 'app' can also be passed in via argc and argv

argc's type is const char **, is it a good way to do this.

qiuguohua: It was char **argv before, but it is const char** argv in IOS. So I changed it to const char** uniformly

bofeng-song: Is it possible to change it to template function ?

qiuguohua: Not very good, because the design of this function is to keep the same parameters passed in as main, and it can be used to parse the command line in the platform.
Can also pass in multiple parameters

qiuguohua (approved):

minggo:
Please add a documentation describe how to upgrade old project.

PR #10803, opened 2022-04-19
update simple fillbuffers for testing 

Zhmz:
Re: cocos-creator/3d-tasks#10539

Changelog:
 * reserve 4 vertices instead of 2 in updateVertexData  of simple, and then UpdateWorldVertex can be unified with other assemblers.


Zhmz:
企业微信截图_16504491314620
The profiling difference on Chrome on PC between v3.6 and this PR.

cocos/2d/assembler/sprite/simple.ts
```diff
@@ -84,60 +86,19 @@ export const simple: IAssembler = {
...
-            vData[27] = ar + cttx;
-            vData[28] = br + dtty;
+
+        const stride = renderData.floatStride;
+
+        for (let i  = 0; i < dataList.length; i++) {
+            const curData = dataList[i];
+            const vec3_temp = new Vec3();
+            Vec3.set(vec3_temp, curData.x, curData.y, 0);
+            Vec3.transformMat4(vec3_temp, vec3_temp, matrix);
```
pandamicro, line 97: It's better to keep a faster approach in ts, but in C++ we can use transformMat4 for all vertex transformation

cocos/2d/assembler/sprite/simple.ts
```diff
@@ -84,60 +86,19 @@ export const simple: IAssembler = {
...
-            vData[19] = bl + dtty;
-            // right top
-            vData[27] = ar + cttx;
-            vData[28] = br + dtty;
+
+        const stride = renderData.floatStride;
+
+        for (let i  = 0; i < dataList.length; i++) {
+            const curData = dataList[i];
+            const vec3_temp = new Vec3();
```
pandamicro: don't create temporary vectors inside logic code, create it in script scope and reuse the same object every time

pandamicro: pay attention to any object creation and new

cocos/2d/assembler/sprite/sliced.ts
```diff
@@ -104,14 +107,31 @@ export const sliced: IAssembler = {
...
-        dataList[1].x = leftWidth * xScale - appX;
-        dataList[1].y = bottomHeight * yScale - appY;
-        dataList[2].x = dataList[1].x + sizableWidth;
-        dataList[2].y = dataList[1].y + sizableHeight;
-        dataList[3].x = width - appX;
-        dataList[3].y = height - appY;
+        const tempList:IRenderData[] = [];
+        // initialize 4 elements
+        for (let i = 0; i < 4; i++) {
+            tempList.push({ x: 0, y: 0, z: 0, u: 0, v: 0, color: new Color() });
```
pandamicro: tempList shouldn't be created inside function logic, try to reuse it for all sliced sprites

cocos/2d/assembler/sprite/simple.ts
```diff
@@ -48,8 +48,10 @@ for (let i = 0; i < 4; i++) {
 export const simple: IAssembler = {
     createData (sprite: Sprite) {
         const renderData = sprite.requestRenderData();
-        renderData.dataLength = 2;
+        renderData.dataLength = 4;
         renderData.resize(4, 6);
+        renderData.vertexRow = 2;
+        renderData.vertexCol = 2;
```
pandamicro, line 54: For the current usage of vertexRow and vertexCol, they are all used inside assemblers, I think it can be an assembler level or type specific constant configuration. They shouldn't be added to RenderData object.

pandamicro (changes requested):

cocos/2d/assembler/sprite/simple.ts
Zhmz: only in Simple, use faster approach. i think we should consider how to unify all fillbuffers firstly, and then we can divide Simple as two sets of codes.

cocos/2d/assembler/sprite/simple.ts
Zhmz: Modified.

cocos/2d/assembler/sprite/sliced.ts
Zhmz: Modified.

cocos/2d/assembler/sprite/simple.ts
pandamicro: OK

cocos/2d/assembler/sprite/sliced.ts
```diff
@@ -28,16 +28,21 @@
...
 import { IBatcher } from '../../renderer/i-batcher';
 import { Sprite } from '../../components';
 import { IAssembler } from '../../renderer/base';
 import { dynamicAtlasManager } from '../../utils/dynamic-atlas/atlas-manager';
 import { StaticVBChunk } from '../../renderer/static-vb-accessor';
+import zlib from '../../../../external/compression/zlib.min';
 
 const vec3_temp = new Vec3();
 const matrix = new Mat4();
+const diagonal_vec3_temps: IRenderData[] = [];
```
pandamicro: name is too long, please modify to a short one, like `tempData`

cocos/2d/assembler/sprite/simple.ts
Zhmz: Such as Tiled, vertexRow and vertexCol cannot be assigned in constructor, since they are assigned in UpdateRenderData. So that we need these value which are from various components instead of some global places.

cocos/2d/assembler/sprite/sliced.ts
Zhmz: Modified to `tempRenderData`

cocos/2d/assembler/sprite/simple.ts
pandamicro: Can we use repeat filter for tiled mode ?

cocos/2d/assembler/sprite/tiled.ts
```diff
@@ -75,8 +87,11 @@ export const tiled: IAssembler = {
 
         const hRepeat = centerWidth === 0 ? sizableWidth : sizableWidth / centerWidth;
         const vRepeat = centerHeight === 0 ? sizableHeight : sizableHeight / centerHeight;
-        const row = Math.ceil(vRepeat + 2);
-        const col = Math.ceil(hRepeat + 2);
+
+        // 有九宫比无九宫多一个点
```
pandamicro, line 91: Don't use Chinese comment

cocos/2d/assembler/sprite/tiled.ts
```diff
@@ -330,42 +352,60 @@ export const tiled: IAssembler = {
...
+        } else {
+            for (let i = 0; i <= col; i++) {
+                if (i === 0) {
+                    data[i].x = -appx;
+                } else if (i > 0 && i < col) {
+                    if (i === 1) {
+                        data[i].x = leftWidth * xScale + Math.min(centerWidth, sizableWidth) - appx;
+                    } else if (centerWidth > 0) {
+                        if (i === (col - 1)) {
+                            data[i].x = leftWidth + offsetWidth + centerWidth * (i - 1) - appx;//无九宫少这一部分
```
pandamicro, line 380: Inline comments always use English

PR #10804, opened 2022-04-19
fix memory leak

stanleyljl:
Re: cocos-creator/3d-tasks#

Changelog:
 * fix some memory leaks

### DIFF
--- a/native/cocos/application/ApplicationManager.cpp
+++ b/native/cocos/application/ApplicationManager.cpp
@@ -33,6 +33,12 @@ ApplicationManager *ApplicationManager::getInstance() {
     return &mgr;
 }
 
+void ApplicationManager::closeAllApplications() {
+    for (auto &app : _apps) {
+        app->close();
+    }
+}
+
 void ApplicationManager::releseAllApplications() {
     _apps.clear();
 }
@@ -50,7 +56,12 @@ ApplicationManager::ApplicationPtr ApplicationManager::getCurrentAppSafe() const
 }
 } // namespace cc
 
-//
+void cocos_close() { // NOLINT(readability-identifier-naming)
+    // Called in the platform layer, because the platform layer is isolated from the application layer
+    // It is the platform layer to drive applications and reclaim resources.
+    cc::ApplicationManager::getInstance()->closeAllApplications();
+}
+
 void cocos_destory() { // NOLINT(readability-identifier-naming)
     // Called in the platform layer, because the platform layer is isolated from the application layer
     // It is the platform layer to drive applications and reclaim resources.
--- a/native/cocos/engine/Engine.cpp
+++ b/native/cocos/engine/Engine.cpp
@@ -51,6 +51,8 @@
 #include "cocos/network/HttpClient.h"
 #include "core/Root.h"
 #include "core/assets/FreeTypeFont.h"
+#include "editor-support/dragonbones-creator-support/ArmatureCacheMgr.h"
+#include "editor-support/spine-creator-support/SkeletonCacheMgr.h"
 #include "platform/interfaces/modules/ISystemWindow.h"
 #include "profiler/DebugRenderer.h"
 #include "profiler/Profiler.h"
@@ -100,22 +102,23 @@ Engine::~Engine() {
     AudioEngine::end();
 #endif
 
-    Root::getInstance()->getPipeline()->destroy();
-
+    dragonBones::ArmatureCacheMgr::destroyInstance();
+    spine::SkeletonCacheMgr::destroyInstance();
+    network::HttpClient::destroyInstance();
     EventDispatcher::destroy();
-    se::ScriptEngine::destroyInstance();
     ProgramLib::destroyInstance();
     BuiltinResMgr::destroyInstance();
-    gfx::DeviceManager::destroy();
-
-    CCObject::deferredDestroy();
-
-    BasePlatform *platform = BasePlatform::getPlatform();
-    platform->setHandleEventCallback(nullptr);
-
+    Root::getInstance()->getPipeline()->destroy();
     CC_PROFILER_DESTROY;
     DebugRenderer::destroyInstance();
     FreeTypeFontFace::destroyFreeType();
+    cc::EventDispatcher::destroy();
+    FileUtils::destroyInstance();
+
+    se::ScriptEngine::getInstance()->cleanup();
+    se::ScriptEngine::destroyInstance();
+    gfx::DeviceManager::destroy();
+    CCObject::deferredDestroy();
 }
 
 int32_t Engine::init() {
@@ -154,12 +157,6 @@ int Engine::restart() {
 }
 
 void Engine::close() { // NOLINT
-    if (cc::EventDispatcher::initialized()) {
-        cc::EventDispatcher::dispatchCloseEvent();
-    }
-
-    auto *scriptEngine = se::ScriptEngine::getInstance();
-
     cc::DeferredReleasePool::clear();
 #if CC_USE_AUDIO
     cc::AudioEngine::stopAll();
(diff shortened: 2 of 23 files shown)


native/cocos/application/ApplicationManager.cpp
```diff
@@ -50,9 +56,14 @@ ApplicationManager::ApplicationPtr ApplicationManager::getCurrentAppSafe() const
 }
 } // namespace cc
 
-//
+void cocos_close() { // NOLINT(readability-identifier-naming)
+    // Called in the platform layer, because the platform layer is isolated from the application layer
+    // It is the platform layer to drive applications and reclaim resources.
+    cc::ApplicationManager::getInstance()->closeAllApplications();
```
minggo, line 62: Can do it in `cocos_destory()`?

native/cocos/engine/Engine.cpp
```diff
@@ -51,6 +51,8 @@
 #include "cocos/network/HttpClient.h"
 #include "core/Root.h"
 #include "core/assets/FreeTypeFont.h"
+#include "editor-support/dragonbones-creator-support/ArmatureCacheMgr.h"
+#include "editor-support/spine-creator-support/SkeletonCacheMgr.h"
```
minggo: spine and dragonbones may not be included into engine. Need to use corresponding macro to include these codes.

native/cocos/renderer/gfx-gles2/GLES2Wrangler.cpp
```diff
@@ -83,3 +114,12 @@ bool gles2wInit() {
 
     return true;
 }
+
+
+bool gles2wExit() {
+    if (!gles2wClose()) {
```
minggo: Why need gles2wExit? Can just use gles2Close?

minggo: I got it, there is a init function.

native/cocos/scene/Camera.cpp
```diff
@@ -110,7 +111,6 @@ void Camera::destroy() {
         _window = nullptr;
     }
     _name.clear();
-    _geometryRenderer->destroy();
```
minggo, line 113: Why not destroy it in destroy function?

native/cocos/application/ApplicationManager.cpp
stanleyljl: The entire exit process needs to be refactored, so here is the simple treatment

native/cocos/engine/Engine.cpp
stanleyljl: Where is the corresponding macro definition？

native/cocos/renderer/gfx-gles2/GLES2Wrangler.cpp
dumganhar: What's the difference between gles2wExit & gles2wClose?

native/cocos/engine/Engine.cpp
```diff
@@ -154,29 +157,18 @@ int Engine::restart() {
 }
 
 void Engine::close() { // NOLINT
-    if (cc::EventDispatcher::initialized()) {
-        cc::EventDispatcher::dispatchCloseEvent();
-    }
```
minggo, line 159: Why not dispatch close event?

native/cocos/scene/Camera.cpp
stanleyljl: The destroy function of Camera is never called.

native/cocos/application/ApplicationManager.cpp
minggo: What's the problem of exit process?

native/cocos/scene/Camera.cpp
dumganhar, line 113: Is it better to keep this line and  set _geometryRenderer to nullptr?
```c++
Camera::~Camera() {
    CC_SAFE_DESTROY_NULL(_geometryRenderer); // if _geometryRenderer is nullptr, then do nothing.
    _frustum->release();
}

void Camera::destroy() {
    if (_window) {
        _window->detachCamera(this);
        _window = nullptr;
    }
    _name.clear();

    CC_SAFE_DESTROY_NULL(_geometryRenderer); // If _geometryRenderer is not nullptr, invoke destroy method and set it to nullptr.
```

native/cocos/engine/Engine.cpp
minggo: You can find them in CMakeLists.txt, such as CC_USE_SPINE.

native/cocos/scene/Camera.cpp
minggo: If so, then the logic is error. We should fix the logic. Currently, destructor invoke destroy function is better.

native/cocos/engine/Engine.cpp
```diff
@@ -154,29 +157,18 @@ int Engine::restart() {
 }
 
 void Engine::close() { // NOLINT
-    if (cc::EventDispatcher::initialized()) {
-        cc::EventDispatcher::dispatchCloseEvent();
```
dumganhar, line 158: Why delete this line ?
It's used for dispatching `onClose` event to JS.

native/cocos/application/ApplicationManager.cpp
stanleyljl: The initialization process and exit process of the entire engine need to be carefully designed, and it is still confusing now.

native/cocos/renderer/gfx-gles2/GLES2Wrangler.cpp
stanleyljl: The exit process mimics the initialization process：gles2wInit->gles2wOpen

native/cocos/application/ApplicationManager.cpp
qiuguohua: Yes, the whole exit needs to be refactored. Calling cocos_close is to call engine.close(). Here, exit(0) will be called; basically, the logic behind will not be called. This was originally done this way, and I've kept the operation.

native/cocos/engine/Engine.cpp
stanleyljl: dispatchCloseEvent is called before somewhere.

native/cocos/scene/Camera.cpp
stanleyljl: 1

stanleyljl: 1

native/cocos/engine/Engine.cpp
stanleyljl: It is called before

native/cocos/scene/Camera.cpp
stanleyljl: 1

native/cocos/application/ApplicationManager.cpp
minggo: > The initialization process and exit process of the entire engine need to be carefully designed, and it is still confusing now.

Ok. Then it is better not export more functions.

native/cocos/engine/Engine.cpp
minggo: @qiuguohua please double check it.

dumganhar (approved):

native/cocos/application/ApplicationManager.cpp
```diff
@@ -33,6 +33,12 @@ ApplicationManager *ApplicationManager::getInstance() {
     return &mgr;
 }
 
+void ApplicationManager::closeAllApplications() {
+    for (auto &app : _apps) {
```
dumganhar, line 37: Use `for (auto* app : _apps)`
`auto*` for pointer rather than `auto&`

stanleyljl: ApplicationPtr is not a raw pointer. The code will compile with errors

native/cocos/engine/Engine.cpp
qiuguohua: > dispatchCloseEvent is called before somewhere.

inside onclose? The timing of these two calls is different. 'close' is an active close, and exit will be called inside, but onclose will not be called?

'onclose' is called when the window is closed

native/cocos/application/ApplicationManager.cpp
dumganhar: Got it.

native/cocos/engine/Engine.cpp
stanleyljl: It is called in Engine::dispatchWindowEvent, which is before Engine::close

qiuguohua (approved):

minggo (approved):

native/cocos/application/ApplicationManager.cpp
minggo: @qiuguohua please create an issue to refactor the init and exit process.

PR #10805, opened 2022-04-19
toggle hdr mode and without diffusemap file, show warnings and fix to…

jk20012001:
… right lighting type

Re: cocos-creator/3d-tasks#

Changelog:
 *

### DIFF
--- a/cocos/core/scene-graph/scene-globals.ts
+++ b/cocos/core/scene-graph/scene-globals.ts
@@ -329,8 +329,9 @@ export class SkyboxInfo {
             }
             this.diffuseMap = this._resource.diffuseMap;
 
-            if (this.diffuseMap == null) {
-                this.applyDiffuseMap = false;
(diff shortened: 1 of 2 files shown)


cocos/core/scene-graph/scene-globals.ts
```diff
@@ -329,8 +329,9 @@ export class SkyboxInfo {
             }
             this.diffuseMap = this._resource.diffuseMap;
 
-            if (this.diffuseMap == null) {
-                this.applyDiffuseMap = false;
+            if (this.diffuseMap == null && this.envLightingType == EnvironmentLightingType.DIFFUSEMAP_WITH_REFLECTION) {
+                this.envLightingType = EnvironmentLightingType.AUTOGEN_HEMISPHERE_DIFFUSE_WITH_REFLECTION;
+                console.error("Can't find corresponding diffuse map, use hemisphere diffuse instead.");
```
pandamicro: Need to use errorID instead of plain text

pandamicro (changes requested):

pandamicro (approved):

PR #10811, opened 2022-04-20
fix PhysX pointer loss percision

SantyWang:
Re: cocos-creator/3d-tasks#

Changelog:
 *

### DIFF
--- a/native/cocos/bindings/jswrapper/v8/Utils.cpp
+++ b/native/cocos/bindings/jswrapper/v8/Utils.cpp
@@ -152,9 +152,7 @@ void setReturnValueTemplate(const Value &data, const T &argv) {
     } else if (data.getType() == Value::Type::Number) {
         argv.GetReturnValue().Set(v8::Number::New(argv.GetIsolate(), data.toDouble()));
     } else if (data.getType() == Value::Type::BigInt) {
-        // Notice: Most return value of type `size_t` should be treated as Number.
-        // argv.GetReturnValue().Set(v8::BigInt::New(argv.GetIsolate(), data.toInt64()));
-        argv.GetReturnValue().Set(v8::Number::New(argv.GetIsolate(), static_cast<double>(data.toInt64())));
+        argv.GetReturnValue().Set(v8::BigInt::New(argv.GetIsolate(), data.toInt64()));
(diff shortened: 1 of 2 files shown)


native/cocos/bindings/jswrapper/v8/Utils.cpp
```diff
@@ -152,9 +152,7 @@ void setReturnValueTemplate(const Value &data, const T &argv) {
     } else if (data.getType() == Value::Type::Number) {
         argv.GetReturnValue().Set(v8::Number::New(argv.GetIsolate(), data.toDouble()));
     } else if (data.getType() == Value::Type::BigInt) {
-        // Notice: Most return value of type `size_t` should be treated as Number.
-        // argv.GetReturnValue().Set(v8::BigInt::New(argv.GetIsolate(), data.toInt64()));
-        argv.GetReturnValue().Set(v8::Number::New(argv.GetIsolate(), static_cast<double>(data.toInt64())));
```
SantyWang, line 157: Due to this line, uintptr_t prointer will lose precision in js

native/cocos/bindings/jswrapper/v8/Utils.cpp
```diff
@@ -152,9 +152,7 @@ void setReturnValueTemplate(const Value &data, const T &argv) {
...
+        argv.GetReturnValue().Set(v8::BigInt::New(argv.GetIsolate(), data.toInt64()));
```
dumganhar, line 155: How about defining a macro for switching to old behavior ?
```c++
#ifdef SE_INT64_VALUE_RETURN_DOUBLE
argv.GetReturnValue().Set(v8::Number::New(argv.GetIsolate(), static_cast<double>(data.toInt64())));
#else
argv.GetReturnValue().Set(v8::BigInt::New(argv.GetIsolate(), data.toInt64()));
#endif
```
And we need to notify developer that the logic has changed.

dumganhar:
After a while, I think the best way is that `se::Value` needs a constructor like
```c++
Value(void* ptr) {
   ... ...
}
```
Only void* type needs to be converted to JS BigInt, other number types will be triggered as `double` (JS number).

@holycanvas @minggo  @PatriceJiang  What do you think?

PR #10820, opened 2022-04-21
Revert removal deprecation for geometry

pandamicro:
Avoid user using static method in deprecated class, ray.create 
need to add a new deprecating function for classes


pandamicro:
Will be resolved in https://github.com/cocos/cocos-engine/issues/10821

PR #10826, opened 2022-04-21
fix memory leak on ts

minggo:


cocos/core/renderer/scene/camera.ts
```diff
@@ -226,10 +226,13 @@ export class Camera {
     }
 
     public destroy () {
+        this.detachFromScene();
```
minggo, line 229: If not detach from scene, camera will hold scene, and it is put to CameraPool, then scene will not be destroyed.

minggo:
@stanleyljl is refactoring exit process. So close this PR first.

Issue #10830, opened 2022-04-21
Saving Illuminance does not work


linruimin:
### Cocos Creator version

3.4.2

### System information

Windows 10 Editor

### Issue description

Edit Illuminance and save it, after change scene and change back, the value back to before edit.
![image](https://user-images.githubusercontent.com/18083506/164433938-f9ac857b-2dda-4d74-9259-54af97b980b9.png)

Try to check HDR, and then it works, the Illuminance value can save after HDR is checked.
![企业微信截图_16505320721382](https://user-images.githubusercontent.com/18083506/164433979-301723d9-1312-4cf3-b02a-9db9ebfe1759.png)

### Steps to reproduce

1.edit Illuminance value
2.save scene
3.change another scene
4.back to scene, then you can see Illuminance value not changed.

PR #10831, opened 2022-04-21
fix ldr intensity overrided by wrong value #10830

jk20012001:
Re: cocos-creator/3d-tasks#

Changelog:
 *


jk20012001:
https://github.com/cocos/cocos-engine/issues/10830

pandamicro (approved):

PR #10832, opened 2022-04-21
Add feature request issue template

jareguo:
You can test the feature request issue template on https://github.com/jareguo/engine/issues/new/choose

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -13,7 +13,7 @@ body:
   attributes:
     label: Use Case
     description: Explain your use case, context, and rationale behind this feature request. More importantly, what is the project you are trying to build that led to the need for this feature? This is important to know the **context** in which the feature is being proposed.
-    placeholder: Example - "I am working on a MOBA game that ... With heavy use of collision detection..."
+    placeholder: Example - "I am working on a MOBA game that ... With heavy use of navigation..."
   validations:
     required: true
 
@@ -27,9 +27,9 @@ body:
 - type: textarea
   attributes:
     label: Proposed Solution
-    description: Describe the solution you'd like in a clear and concise manner.
+    description: Describe the desired workflow or API set to satisfy the need in a clear and concise manner.
   validations:
-    required: true
+    required: false
 
 - type: textarea
   attributes:
@@ -37,7 +37,7 @@ body:
     description: Describe how your proposal will work, with code, pseudo-code, mock-ups, and/or diagrams. If the API needs to be changed, please describe what does the proposed API look like and provide code samples. Note that you can use Markdown to format your code blocks.
     placeholder: Example - "When the user presses ..., select all nodes that have the group..."
   validations:
-    required: true
+    required: false
 
 - type: textarea
   attributes:
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -1,0 +1,60 @@
+name: Feature request
+description: Suggest an idea for Cocos
+labels: ["Needs Triage","Feature Request"]
+body:
+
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,60 @@
...
+  attributes:
+    value: |
+      - Please write a descriptive proposal title above.
+      - Search [open](https://github.com/cocos/cocos-engine/issues) and [closed](https://github.com/cocos/cocos-engine/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been suggested.
+
+- type: textarea
+  attributes:
+    label: Use Case
+    description: Explain your use case, context, and rationale behind this feature request. More importantly, what is the project you are trying to build that led to the need for this feature? This is important to know the **context** in which the feature is being proposed.
+    placeholder: Example - "I am working on a MOBA game that ... With heavy use of collision detection..."
```
pandamicro: navigation might be a better example here

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,60 @@
...
+    description: Please add a clear and concise description of the problem or limitation you are having in your project.
+  validations:
+    required: true
+
+- type: textarea
+  attributes:
+    label: Proposed Solution
+    description: Describe the solution you'd like in a clear and concise manner.
+  validations:
+    required: true
```
pandamicro: Better to be optional

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,60 @@
...
+
+- type: textarea
+  attributes:
+    label: How it works
```
pandamicro, line 36: I think it could be merged with proposed solution

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,60 @@
...
+  attributes:
+    label: Problem Description
+    description: Please add a clear and concise description of the problem or limitation you are having in your project.
+  validations:
+    required: true
+
+- type: textarea
+  attributes:
+    label: Proposed Solution
+    description: Describe the solution you'd like in a clear and concise manner.
```
pandamicro: Propose the desired workflow or API set to satisfy the need in a clear and concise manner.

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,60 @@
...
+  attributes:
+    label: Use Case
+    description: Explain your use case, context, and rationale behind this feature request. More importantly, what is the project you are trying to build that led to the need for this feature? This is important to know the **context** in which the feature is being proposed.
+    placeholder: Example - "I am working on a MOBA game that ... With heavy use of collision detection..."
+  validations:
+    required: true
+
+- type: textarea
+  attributes:
+    label: Problem Description
```
pandamicro, line 22: I think it can be merged in use case

pandamicro (commented):
Generally

.github/ISSUE_TEMPLATE/feature_request.yml
jareguo: I think it's pretty important for an open source project to force people to think about the feasibility of features (or what kind of workflow users want). I would make both of these optional.

.github/ISSUE_TEMPLATE/feature_request.yml
jareguo: I want to force people to describe the source of the need so that we can assess the generality of the need and whether there is a better solution. For similar projects, it is normal to have a little extra cost when raising need. The same rule is true for our internal eco service team.

pandamicro (approved):

PR #10854, opened 2022-04-24
fix debug crash problem

star-e:
fix issue: https://github.com/cocos/3d-tasks/issues/12116
fix debug validation problem by enable input attachment for all platforms.

There are mainly 2 problems:
1.  Currently only one shader descriptor layout can be used across various platforms. As a result, it is impossible to implement certain algorithms.

e.g.
vulkan support color and depth texture as input attachment
gles3 arm_fbf support color and depth as input attachment
gles3 ext_fbf  support color as input attachment
metal support color as input attachment
webgl does not support input attachment.

As a result, at least 4 kinds of descriptor layout are needed to correctly implement deferred lighting. 
Currently implementation might cause gfx validation failure.

2. Currently, the type of descriptor is inferred from both material and resource usage. As a result, the side effect of bindTexture is hard to estimate. Users might be surprised when the shader asset is changed by accident.

Changelog:
 * enable input attachment for all platforms.

### DIFF
--- a/native/cocos/renderer/pipeline/deferred/GbufferStage.cpp
+++ b/native/cocos/renderer/pipeline/deferred/GbufferStage.cpp
@@ -205,9 +205,9 @@ void GbufferStage::render(scene::Camera *camera) {
             static_cast<uint>(pipeline->getWidth() * shadingScale),
             static_cast<uint>(pipeline->getHeight() * shadingScale),
         };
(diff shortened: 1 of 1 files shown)


star-e:
@cocos-robot run test cases

pandamicro (approved):

hana-alice (approved):
Validation passed, but still few issues left on metal, fixing.

hana-alice:
follow-up https://github.com/cocos/cocos-engine/pull/10857

PR #10857, opened 2022-04-24
fix metal ds index

hana-alice:
Re: #

Changelog:
 *

### DIFF
--- a/native/cocos/renderer/gfx-metal/MTLCommandBuffer.mm
+++ b/native/cocos/renderer/gfx-metal/MTLCommandBuffer.mm
@@ -203,8 +203,12 @@ void CCMTLCommandBuffer::beginRenderPass(RenderPass *renderPass, Framebuffer *fb
         for (size_t i = 0; i < subpasses.size(); ++i) {
             for (size_t j = 0; j < subpasses[i].inputs.size(); ++j) {
                 uint32_t input = subpasses[i].inputs[j];
+                if(input >= colorAttachments.size()) {
+                    continue; // depthStencil as input
+                }
                 if (visited[input])
                     continue;
+                
                 auto *ccMtlTexture = static_cast<CCMTLTexture *>(colorTextures[input]);
                 ccMtlRenderPass->setColorAttachment(input, ccMtlTexture, 0);
                 mtlRenderPassDescriptor.colorAttachments[input].clearColor  = mu::toMTLClearColor(colors[input]);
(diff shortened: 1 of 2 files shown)


native/cocos/renderer/gfx-metal/MTLCommandBuffer.mm
```diff
@@ -319,15 +326,25 @@ of this software and associated engine source code (the "Software"), a limited,
 
     const DepthStencilAttachment& dsAttachment  = curRenderPass->getDepthStencilAttachment();
     const SubpassInfo&      subpass             = subpasses[index];
-    if (subpass.depthStencil != INVALID_BINDING) {
+    uint32_t ds = subpass.depthStencil;
+    auto iter = std::find_if(subpass.inputs.begin(), subpass.inputs.end(), [&subpass](uint32_t index){ return index >= subpass.inputs.size(); });
```
star-e: Should it be `index >= subpass.colors.size();` ?

star-e:
the depth texture in subpass.depthStencil is a bit different from the one in subpass.inputs.
subpass.depthStencil is used for depth test/write.
subpass.input is used for framebuffer fetch. 

if  subpass.depthStencil is not specified, there will be no depth test/write, even depth texture is used in inputs

native/cocos/renderer/gfx-metal/MTLCommandBuffer.mm
hana-alice: subpass[0]:
-- input:    []
-- output:  [0, 1, 2]

subpass[1]:
-- input:    [0, 1, 2, 4]
-- output: [3]

to activate Metal Subpass, only 1 renderpassDescriptor is allowed, which means there will be 2 passes once any attachment in slot is changed(because this product another different renderpassDescriptor). So at the very beginning we have to set the slot in renderPassDescriptor: [0, 1, 2, 3, 4], that's why I look for the ds first in subpass.inputs and then in outputs.

hana-alice: append: sth slightly different in Metal: no explicitly input and output attachment, there's just attachments lie in certain slot in renderpass descriptor. 
eg: renderPassDescriptor with slots: [0, 1, 2, 3, 4], in metal shader `[[colors[0]]]` get the attachment at 0, can be used as input OR output

star-e: I see.

star-e (approved):

PR #10873, opened 2022-04-25
V3.6.0 merge v3.5

minggo:
- can work on web and macOS, but doesn't do full test
- should sync codes to c++ for some modules, will do it in another PR


SantyWang (approved):

minggo:
- interface ci error: it is ci script error
- clang tidy error: just merge branch, so do not handle clang tidy error

Issue #10822, opened 2022-04-21
Refactoring the engine exit process


stanleyljl:
### Cocos Creator version

3.6.0

### System information

all platforms

### Issue description

Engine exit process is not standardized, causing many problems.

### Steps to reproduce

Start, run, exit the game normally.

**First iteration result:**

![image](https://user-images.githubusercontent.com/26244076/165235597-d2959692-0e1c-4d16-b6e8-c5c6c72cbe61.png)

PR #10880, opened 2022-04-26
refactor exit process

stanleyljl:
Re: #https://github.com/cocos/cocos-engine/issues/10822

Changelog:
 * refactor exit process

image

### DIFF
--- a/native/cocos/application/CocosApplication.cpp
+++ b/native/cocos/application/CocosApplication.cpp
@@ -121,7 +121,7 @@ void CocosApplication::onResume() {
 }
 
 void CocosApplication::onClose() {
-    // TODO(cc): Handling close events
+    close();
 }
 
 void CocosApplication::setDebugIpAndPort(const ccstd::string &serverAddr, uint32_t port, bool isWaitForConnect) {
--- a/native/cocos/scene/RenderScene.cpp
+++ b/native/cocos/scene/RenderScene.cpp
@@ -109,6 +109,7 @@ void RenderScene::removeCamera(Camera *camera) {
 void RenderScene::removeCameras() {
     for (const auto &camera : _cameras) {
         camera->detachFromScene();
+        CC_SAFE_DESTROY(camera);
     }
     _cameras.clear();
 }
--- a/native/cocos/scene/RenderScene.cpp
+++ b/native/cocos/scene/RenderScene.cpp
@@ -109,7 +109,7 @@ void RenderScene::removeCamera(Camera *camera) {
 void RenderScene::removeCameras() {
     for (const auto &camera : _cameras) {
         camera->detachFromScene();
-        CC_SAFE_DESTROY(camera);
+        camera->destroy();
     }
     _cameras.clear();
 }
--- a/native/cocos/scene/Skybox.cpp
+++ b/native/cocos/scene/Skybox.cpp
@@ -24,7 +24,6 @@
  ****************************************************************************/
 
 #include "scene/Skybox.h"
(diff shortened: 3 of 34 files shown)


stanleyljl:
@cocos-robot run test cases

native/cocos/scene/RenderScene.cpp
```diff
@@ -109,6 +109,7 @@ void RenderScene::removeCamera(Camera *camera) {
 void RenderScene::removeCameras() {
     for (const auto &camera : _cameras) {
         camera->detachFromScene();
+        CC_SAFE_DESTROY(camera);
```
minggo: just invoke `camera->destroy()` as `camera->detachFromScene()` is invoked above.

stanleyljl: Where is camera->detachFromScene() called？

native/cocos/scene/Skybox.cpp
```diff
@@ -42,7 +41,6 @@
 #include "scene/Model.h"
 
 namespace {
-cc::Mesh *    skyboxMesh{nullptr}; // TODO(cjh): How to release ?
 cc::Material *skyboxMaterial{nullptr};
```
dumganhar, line 44: Should `skyboxMaterial` also be a member variable?

stanleyljl: This will also need to be refactored in the future, but it will not leak now.

native/cocos/application/CocosApplication.cpp
```diff
@@ -121,7 +121,7 @@ void CocosApplication::onResume() {
 }
 
 void CocosApplication::onClose() {
-    // TODO(cc): Handling close events
+    close();
```
qiuguohua, line 124: In my opinion, 'close' is the act of actively invoking the shutdown, and 'onclose' is the message processed when closing, and it feels that the concepts of these two operations are confused.

stanleyljl: It depends on how you look at event handling, where onClose is a response to the CLOSE event, not a response to engine::close

native/cocos/scene/RenderScene.cpp
minggo: I mean `camera->detachFromScene()` uses camera without checking camera, so here don't have to check too.

native/cocos/application/CocosApplication.cpp
qiuguohua: 'onClose' is a response to the close event, but another close is called at 'onClose'?

native/cocos/scene/RenderScene.cpp
stanleyljl: 1

minggo (approved):

minggo (approved):

novain1102 (approved):

dumganhar (approved):

qiuguohua (approved):

native/cocos/scene/RenderScene.cpp
```diff
@@ -109,6 +109,7 @@ void RenderScene::removeCamera(Camera *camera) {
 void RenderScene::removeCameras() {
     for (const auto &camera : _cameras) {
         camera->detachFromScene();
+        camera->destroy();
```
dumganhar, line 112: Why adding `destroy` here? It will be different logic with web.

minggo: Yep, should not invoke destroy here. A camera can be attached again or be attached to other scene.

PR #10881, opened 2022-04-26
Sync ts logic to c++

minggo:


minggo:
@jk20012001 please help to review skybox related codes
@yiwenxue please help to review mipmap related codes
@dumganhar please help to review model related codes

dumganhar (approved):

native/cocos/core/assets/SimpleTexture.cpp
```diff
@@ -159,23 +159,24 @@ void SimpleTexture::createTexture(gfx::Device *device) {
...
+    maxLevel - _baseLevel + 1
     );
 
     //TODO(minggo)
 //    if (!textureViewCreateInfo) {
 //        return;
 //    }
     
-    _gfxTextureView = device->createTexture(textureViewCreateInfo);
+    return device->createTexture(textureViewCreateInfo);
```
yiwenxue, line 179: We should also free their memory when destroying `_gfxTexture` and `_gfxTextureView`.

minggo: It is destroyed in `SimpleTexture::tryDestroyTextureView`. And it is the same logic as TS.

jk20012001 (approved):

native/cocos/core/assets/SimpleTexture.cpp
yiwenxue: Yes, exactly, Typescript has a GC system, and the memory is freed automatically. But for C++, things are different, `_gfxTexture->destroy()` will only free the resource on GPU, and we need to call `CC_DELETE(_gfxTexture)` mannually.

minggo: They are IntrusivePtr, destructor will be invoked automatically.

yiwenxue: Ok, fine.

yiwenxue (approved):

PR #10889, opened 2022-04-26
Add the support of `texture2DArray` and `texture3D`.

yiwenxue:
Re: #

Texture2DArray and Texture3D are important features for a modern engine. With `Texture2DArray` we can bind more textures with only one slot. With `Texture3D` we can implement many complex render effects. This PR added support to Texture3D and Texture2DArray.

Changelog:
 * Add support to `texture2DArray`.
 * Add support to `texture3D`.
 * Fix wrong flag.
 * Add new device caps.
 
The modifications were tested using gfx-test-case (now test-gfx) for `Vulkan` and `GLES3`.


yiwenxue:
move to 3.6.1

PR #10896, opened 2022-04-27
fix crash when doing gc

minggo:


PatriceJiang (approved):

native/cocos/bindings/manual/jsb_spine_manual.cpp
```diff
@@ -242,31 +242,7 @@ bool register_all_spine_manual(se::Object *obj) {
...
-
-            se::AutoHandleScope hs;
-            se->clearException();
-
-            // The mapping of native object & se::Object was cleared in above code.
-            // The private data (native object) may be a different object associated with other se::Object.
-            // Therefore, don't clear the mapping again.
-            seObj->clearPrivateData(false);
-            seObj->unroot();
-            seObj->decRef();
```
dumganhar, line 262: Will seObj be leaked?

minggo: @PatriceJiang

PatriceJiang: It's ok here.

dumganhar (approved):

PR #10902, opened 2022-04-27
fix native include problem on windows

LinYunMo:
Fix native include problem on windows with VS 2017


native/cocos/core/assets/Font.cpp
```diff
@@ -34,6 +34,7 @@
 #include "math/Math.h"
 #include "platform/FileUtils.h"
 #include "profiler/Profiler.h"
+#include <cctype>
```
minggo: please add it after `#include <boost/functional/hash.hpp>`

PatriceJiang (approved):

minggo (approved):

PR #10903, opened 2022-04-27
add base-Renderer and separate & rename renderableComp

LinYunMo:
- add base-renderer 
- rename renderable-component to renderer
- separate renderable-component
- rename Renderable2D to UIRenderer

update：
- rename base-renderer to renderer
- rename renderer to model-renderer

### DIFF
--- a/cocos/2d/components/ui-mesh-renderer.ts
+++ b/cocos/2d/components/ui-mesh-renderer.ts
@@ -30,7 +30,7 @@
  */
 
 import { ccclass, help, executionOrder, menu, executeInEditMode } from 'cc.decorator';
-import { RenderableComponent } from '../../core/components/renderable-component';
+import { Renderer } from '../../core/components/renderable-component';
 import { RenderPriority } from '../../core/pipeline/define';
 import { IBatcher } from '../renderer/i-batcher';
 import { Stage } from '../renderer/stencil-manager';
@@ -57,7 +57,7 @@ export class UIMeshRenderer extends Component {
         return this._modelComponent;
     }
 
-    private _modelComponent: RenderableComponent | null = null;
+    private _modelComponent: Renderer | null = null;
 
     public __preload () {
         this.node._uiProps.uiComp = this;
@@ -68,7 +68,7 @@ export class UIMeshRenderer extends Component {
             this.node.addComponent('cc.UITransform');
         }
 
-        this._modelComponent = this.getComponent('cc.RenderableComponent') as RenderableComponent;
+        this._modelComponent = this.getComponent('cc.RenderableComponent') as Renderer;
         if (!this._modelComponent) {
             console.warn(`node '${this.node && this.node.name}' doesn't have any renderable component`);
         }
@@ -78,7 +78,7 @@ export class UIMeshRenderer extends Component {
         if (this.node._uiProps.uiComp === this) {
             this.node._uiProps.uiComp = null;
         }
-        this._modelComponent = this.getComponent('cc.RenderableComponent') as RenderableComponent;
+        this._modelComponent = this.getComponent('cc.RenderableComponent') as Renderer;
         if (!this._modelComponent) {
             return;
         }
--- a/cocos/2d/components/ui-mesh-renderer.ts
+++ b/cocos/2d/components/ui-mesh-renderer.ts
@@ -30,7 +30,7 @@
  */
 
 import { ccclass, help, executionOrder, menu, executeInEditMode } from 'cc.decorator';
-import { Renderer } from '../../core/components/renderable-component';
+import { ModelRenderer } from '../../core/components/model-renderer';
 import { RenderPriority } from '../../core/pipeline/define';
 import { IBatcher } from '../renderer/i-batcher';
 import { Stage } from '../renderer/stencil-manager';
@@ -57,7 +57,7 @@ export class UIMeshRenderer extends Component {
         return this._modelComponent;
     }
 
-    private _modelComponent: Renderer | null = null;
+    private _modelComponent: ModelRenderer | null = null;
 
     public __preload () {
         this.node._uiProps.uiComp = this;
@@ -68,7 +68,7 @@ export class UIMeshRenderer extends Component {
             this.node.addComponent('cc.UITransform');
         }
 
-        this._modelComponent = this.getComponent('cc.RenderableComponent') as Renderer;
+        this._modelComponent = this.getComponent('cc.ModelRenderer') as ModelRenderer;
         if (!this._modelComponent) {
             console.warn(`node '${this.node && this.node.name}' doesn't have any renderable component`);
         }
@@ -78,7 +78,7 @@ export class UIMeshRenderer extends Component {
         if (this.node._uiProps.uiComp === this) {
             this.node._uiProps.uiComp = null;
         }
-        this._modelComponent = this.getComponent('cc.RenderableComponent') as Renderer;
+        this._modelComponent = this.getComponent('cc.ModelRenderer') as ModelRenderer;
         if (!this._modelComponent) {
             return;
         }
--- a/cocos/2d/framework/renderable-2d.ts
+++ b/cocos/2d/framework/renderable-2d.ts
@@ -44,7 +44,7 @@ import { UITransform } from './ui-transform';
 import { Stage } from '../renderer/stencil-manager';
 import { legacyCC } from '../../core/global-exports';
 import { NodeEventType } from '../../core/scene-graph/node-event';
-import { BaseRenderer } from '../../core';
+import { Renderer } from '../../core/components/renderer';
 
 // hack
 ccenum(BlendFactor);
@@ -114,7 +114,7 @@ export enum InstanceMaterialType {
 @requireComponent(UITransform)
 @disallowMultiple
 @executeInEditMode
-export class UIRenderer extends BaseRenderer {
+export class UIRenderer extends Renderer {
     /**
      * @en The blend factor enums
      * @zh 混合模式枚举类型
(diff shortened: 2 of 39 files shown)


cocos/2d/components/ui-mesh-renderer.ts
```diff
@@ -30,7 +30,7 @@
  */
 
 import { ccclass, help, executionOrder, menu, executeInEditMode } from 'cc.decorator';
-import { RenderableComponent } from '../../core/components/renderable-component';
+import { Renderer } from '../../core/components/renderable-component';
```
pandamicro: Shouldn't be using Renderer as class name, keep RenderableComponent is ok

cocos/2d/framework/renderable-2d.ts
```diff
@@ -111,11 +110,11 @@ export enum InstanceMaterialType {
  * @zh 所有支持渲染的 2D 组件的基类。
  * 这个组件会设置 [[Node]] 上的 [[NodeUIProperties.uiComp]]。
  */
-@ccclass('cc.Renderable2D')
+@ccclass('cc.UIRenderer')
 @requireComponent(UITransform)
 @disallowMultiple
 @executeInEditMode
-export class Renderable2D extends RenderableComponent {
+export class UIRenderer extends BaseRenderer {
```
pandamicro: Please keep using `Renderable` instead of `Renderer`, they are objects which can be rendered, so they are renderable, but they are not the renderer, they are not rendering anything, they are only providing data to the renderer

cocos/core/components/base-renderer.ts
```diff
@@ -0,0 +1,220 @@
...
+import { legacyCC } from '../global-exports';
+
+const _matInsInfo: IMaterialInstanceInfo = {
+    parent: null!,
+    owner: null!,
+    subModelIdx: 0,
+};
+
+@ccclass('cc.BaseRenderer')
+export class BaseRenderer extends Component {
```
pandamicro: BaseRenderable

pandamicro: Add API reference for this class

cocos/core/components/renderable-component.ts
```diff
@@ -28,26 +28,17 @@
...
-const _matInsInfo: IMaterialInstanceInfo = {
-    parent: null!,
-    owner: null!,
-    subModelIdx: 0,
-};
-
-@ccclass('cc.RenderableComponent')
-export class RenderableComponent extends Component {
+@ccclass('cc.Renderer')
+export class Renderer extends BaseRenderer {
```
pandamicro: Renderable or keep RenderableComponent

cocos/core/components/renderable-component.ts
```diff
@@ -236,17 +66,8 @@ export class RenderableComponent extends Component {
...
-
-    protected _clearMaterials () {
-    }
-
     protected _onVisibilityChange (val) {
     }
 }
 
-legacyCC.RenderableComponent = RenderableComponent;
+legacyCC.Renderer = Renderer;
```
pandamicro: If modify name, no need to add to legacyCC again

pandamicro (changes requested):

SantyWang:
If it's an interface, I think it's more appropriate to call it IRenderable, but if it's a type, I think it's better to call it Renderer, even if it's an abstract type. `Renderer` is used to express that his job is rendering, not to describe that he is a type that supports rendering

cocos/core/components/base-renderer.ts
minggo: BaseRenderer -> Renderer?

cocos/core/components/base-renderer.ts
```diff
@@ -0,0 +1,220 @@
...
+    @type(Material)
+    @displayOrder(0)
+    @displayName('Materials')
+    get sharedMaterials () {
+        // if we don't create an array copy, the editor will modify the original array directly.
+        return (EDITOR && this._materials.slice()) || this._materials;
+    }
+
+    set sharedMaterials (val) {
+        for (let i = 0; i < val.length; i++) {
```
minggo: I found much TS codes caches length like this
```javascript
for (let i = 0, len = val.length; i < len; ++)
```
Does caching length will improve performance?

cocos/core/components/index.ts
```diff
@@ -30,6 +30,7 @@ export { default as MissingScript } from './missing-script';
 export { EventHandler } from './component-event-handler';
 export { Component } from './component';
 export { Camera } from './camera-component';
-export { RenderableComponent } from './renderable-component';
+export { Renderer } from './renderable-component';
+export { BaseRenderer } from './base-renderer';
```
minggo: What's difference between `BaseRenderer` and `Renderer`?

cocos/core/components/base-renderer.ts
```diff
@@ -0,0 +1,220 @@
+/*
+ Copyright (c) 2020 Xiamen Yaji Software Co., Ltd.
```
minggo: 2020 -> 2022

LinYunMo:
I added a comparison chart of the structure before and after the change

RendererExtends (Before)

RendererExtends (After)

minggo:
> I added a comparison chart of the structure before and after the change

Thanks. I think change like this is better
- `BaseRenderer` -> `Renderer`
- `Renderer` -> `ObjectRenderer`? or other more meaningful name,  i am not sure

SantyWang:
> Renderer -> ObjectRenderer? or other more meaningful name, i am not sure

how about this:
* BaseRenderer -> Renderer
* Renderer -> ModelRenderer ? 

The main difference between UIRenderer and ModelRenderer is that ModelRenderer is based on model

minggo:
> > Renderer -> ObjectRenderer? or other more meaningful name, i am not sure
> 
> how about this:
> 
> * BaseRenderer -> Renderer
> * Renderer -> ModelRenderer ?
> 
> The main difference between UIRenderer and ModelRenderer is that ModelRenderer is based on model

It looks good to me.

pandamicro:
> If it's an interface, I think it's more appropriate to call it IRenderable, but if it's a type, I think it's better to call it Renderer, even if it's an abstract type. `Renderer` is used to express that his job is rendering, not to describe that he is a type that supports rendering

I see your point, but please keep in mind, we already have a `renderer` module, so it not acceptable to have a Class `Renderer`, it will create more confusions.

Thanks for @LinYunMo diagram to make it more clear, we have `MeshRenderer`, `SkinnedMeshRenderer`, so I will not insist on using Renderer in the name, but I do insist we shouldn't create a `Renderer` class. I suggest using 

- `BaseRenderer`
  - `ModelRenderer`
  - `UIRenderer`

minggo:
Every engine contains a module named renderer explicitly or implicitly, adding a class named `Renderer ` seems ok. I think Renderer is more concise than BaseRenderer.

minggo (approved):

SantyWang (approved):

SantyWang (approved):

cocos/2d/framework/renderable-2d.ts
```diff
@@ -111,11 +110,11 @@ export enum InstanceMaterialType {
  * @zh 所有支持渲染的 2D 组件的基类。
  * 这个组件会设置 [[Node]] 上的 [[NodeUIProperties.uiComp]]。
  */
-@ccclass('cc.Renderable2D')
+@ccclass('cc.UIRenderer')
 @requireComponent(UITransform)
 @disallowMultiple
 @executeInEditMode
-export class Renderable2D extends RenderableComponent {
```
pandamicro, line 118: Please also change the file name

cocos/core/components/model-renderer.ts
```diff
@@ -0,0 +1,70 @@
...
+import {
+    ccclass, serializable,
+} from 'cc.decorator';
+import { scene } from '../renderer';
+import { Layers } from '../scene-graph/layers';
+import { legacyCC } from '../global-exports';
+import { Renderer } from './renderer';
+
+@ccclass('cc.ModelRenderer')
+export class ModelRenderer extends Renderer {
```
pandamicro, line 38: API doc for the class and all public APIs

cocos/core/components/model-renderer.ts
```diff
@@ -0,0 +1,70 @@
...
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ THE SOFTWARE.
+ */
+
+/**
+ * @packageDocumentation
+ * @module model
+ */
```
pandamicro: No need to put this anymore

cocos/core/components/renderer.ts
```diff
@@ -35,31 +35,16 @@ import {
...
-    get visibility () {
-        return this._visFlags;
-    }
-
-    set visibility (val) {
-        this._visFlags = val;
-        this._onVisibilityChange(val);
-    }
-
+@ccclass('cc.Renderer')
```
pandamicro, line 45: Also add some docs on it

cocos/core/components/renderer.ts
```diff
@@ -35,31 +35,16 @@ import {
...
+export class Renderer extends Component {
```
pandamicro, line 46: Renderer is still too general for me, maybe AbstractRenderer, would that sound better @minggo

pandamicro (commented):
Please check whether any native class need a name refactor

pandamicro (approved):

PR #10918, opened 2022-04-29
V3.6.0 add instanced particle

zxx43:
Re: #

Changelog:
 *


cocos/particle/renderer/particle-system-renderer-base.ts
```diff
@@ -60,16 +60,28 @@ export interface IParticleSystemRenderer {
...
+    setUseInstance (value: boolean): void;
+    getUseInstance (): boolean;
 }
 
 export abstract class ParticleSystemRendererBase implements IParticleSystemRenderer {
     protected _particleSystem: any = null;
     protected _model: ParticleBatchModel | null = null;
     protected _renderInfo: ParticleSystemRenderer | null = null;
     protected _vertAttrs: Attribute[] = [];
+    protected _useInstance;
```
star-e: lint problem

star-e, line 72: lint problem

star-e:
Currently we have used _useInstance to differ plain/instanced rendering strategies. It is ok for current situation.
However, if we need to add new features in the future, ParticleBatchModel will need some refactoring.

zxx43:
> Currently we have used _useInstance to differ plain/instanced rendering strategies. It is ok for current situation. However, if we need to add new features in the future, ParticleBatchModel will need some refactoring.

Split current ParticleBatchModel into BatchModel & InstanceModel

star-e (approved):

Issue #10840, opened 2022-04-22
v3.5 api-task


UniqueDisplayNameIsInUse:
- [x]  中文版显示为英文注释
- [x]  有索引列表，注释过长时，会被遮挡  如 Macro（Window10 + Chrome 测试未发现该问题，需在Mac+Safari下测试）
- [x]  英文版有中文翻译，例如： degreesToRadians，每个页面仅展示 1 个图片。
1. https://test-docs.cocos.com/creator/3.5/api/en/class/Graphics
![image](https://user-images.githubusercontent.com/97574620/164616221-985c13da-5eeb-40db-a07b-433790839a28.png)
2.https://test-docs.cocos.com/creator/3.5/api/en/class/UI  
![image](https://user-images.githubusercontent.com/97574620/164616658-cd7e5e16-bed4-4a39-b48c-a0cc844f896e.png)
3.https://test-docs.cocos.com/creator/3.5/api/en/class/UITransform
![image](https://user-images.githubusercontent.com/97574620/164616999-18a85db5-ff4c-435c-9e8f-cae766c10fba.png)
4.https://test-docs.cocos.com/creator/3.5/api/en/class/Mesh
![image](https://user-images.githubusercontent.com/97574620/164617379-67bf2c3b-b515-422e-a350-cbc330b45d0f.png)
5.https://test-docs.cocos.com/creator/3.5/api/en/class/SkelAnimDataHub
![image](https://user-images.githubusercontent.com/97574620/164617554-6e99c3c0-5ab4-41f5-8da8-fff8dcd39564.png)
6.https://test-docs.cocos.com/creator/3.5/api/en/class/SkeletalAnimation
![image](https://user-images.githubusercontent.com/97574620/164617605-c27ead69-00ed-4a1f-bfb4-1798eed05dd1.png)
7.https://test-docs.cocos.com/creator/3.5/api/en/class/SkinnedMeshBatchRenderer
![image](https://user-images.githubusercontent.com/97574620/164617753-fe474709-1f7f-4bec-8572-3c649358950c.png)
8.https://test-docs.cocos.com/creator/3.5/api/en/class/SkinnedMeshRenderer
![image](https://user-images.githubusercontent.com/97574620/164617893-a1eeb65a-72d3-427e-957f-1db6b7344923.png)
9.https://test-docs.cocos.com/creator/3.5/api/en/namespace/geometry.distance
[...]

PR #10928, opened 2022-05-05
try fix doc problem

star-e:
Try fixing gfx doc problem.

Changelog:
 * added en/zh comment to gfx/base/define.ts


minggo (approved):

pandamicro:
https://github.com/cocos/cocos-engine/issues/10840

PR #10930, opened 2022-05-05
listen WINDOW_RESIZE event by application

timlyeee:
Re: #https://github.com/cocos/3d-tasks/issues/12150
internal issue： Once resized, Windows restart will show as 800x600 surface view as original size.
Changelog:
 * No public API changed.
 * save changed size to _windowInfo

### DIFF
--- a/native/cocos/application/BaseGame.h
+++ b/native/cocos/application/BaseGame.h
@@ -52,6 +52,6 @@ protected:
     std::string  _xxteaKey;
     DebuggerInfo _debuggerInfo;
     WindowInfo   _windowInfo;
-    std::once_flag _windowIsCreated;
+    std::once_flag _windowCreateFlag;
 };
 } // namespace cc
--- a/native/cocos/application/BaseGame.h
+++ b/native/cocos/application/BaseGame.h
@@ -54,41 +54,4 @@ protected:
     WindowInfo   _windowInfo;
     std::once_flag _windowIsCreated;
 };
-
-int BaseGame::init() {
-#if CC_PLATFORM == CC_PLATFORM_WINDOWS || CC_PLATFORM == CC_PLATFORM_LINUX || CC_PLATFORM == CC_PLATFORM_QNX || CC_PLATFORM == CC_PLATFORM_MAC_OSX
-        // override default value
-        //_windowInfo.x      = _windowInfo.x == -1 ? 0 : _windowInfo.x;
-        //_windowInfo.y      = _windowInfo.y == -1 ? 0 : _windowInfo.y;
-        _windowInfo.width  = _windowInfo.width == -1 ? 800 : _windowInfo.width;
-        _windowInfo.height = _windowInfo.height == -1 ? 600 : _windowInfo.height;
-        _windowInfo.flags  = _windowInfo.flags == -1 ? cc::ISystemWindow::CC_WINDOW_SHOWN |
-                                                          cc::ISystemWindow::CC_WINDOW_RESIZABLE |
-                                                          cc::ISystemWindow::CC_WINDOW_INPUT_FOCUS
-                                                     : _windowInfo.flags;
-        std::call_once(_windowIsCreated, [&]() {
-            if (_windowInfo.x == -1 || _windowInfo.y == -1) {
-                createWindow(_windowInfo.title.c_str(), _windowInfo.width, _windowInfo.height, _windowInfo.flags);
-            } else {
-                createWindow(_windowInfo.title.c_str(), 
-                    _windowInfo.x, _windowInfo.y, _windowInfo.width, _windowInfo.height, _windowInfo.flags);
-            }
-        });
-
-#endif
-
-        if (_debuggerInfo.enabled) {
-            setDebugIpAndPort(_debuggerInfo.address, _debuggerInfo.port, _debuggerInfo.pauseOnStart);
-        }
-
-        int ret = cc::CocosApplication::init();
-        if (ret != 0) {
-            return ret;
-        }
-
-        setXXTeaKey(_xxteaKey);
-        runScript("jsb-adapter/jsb-builtin.js");
-        runScript("main.js");
-        return 0;
-}
 } // namespace cc
--- a/native/cocos/application/BaseGame.h
+++ b/native/cocos/application/BaseGame.h
@@ -46,7 +46,16 @@ public:
     };
 
     BaseGame() = default;
-    int init() override {
+    int init() override;
+
+protected:
+    std::string  _xxteaKey;
+    DebuggerInfo _debuggerInfo;
+    WindowInfo   _windowInfo;
+    std::once_flag _windowIsCreated;
+};
+
+int BaseGame::init() {
 #if CC_PLATFORM == CC_PLATFORM_WINDOWS || CC_PLATFORM == CC_PLATFORM_LINUX || CC_PLATFORM == CC_PLATFORM_QNX || CC_PLATFORM == CC_PLATFORM_MAC_OSX
         // override default value
         //_windowInfo.x      = _windowInfo.x == -1 ? 0 : _windowInfo.x;
@@ -56,11 +65,11 @@ public:
         _windowInfo.flags  = _windowInfo.flags == -1 ? cc::ISystemWindow::CC_WINDOW_SHOWN |
                                                           cc::ISystemWindow::CC_WINDOW_RESIZABLE |
                                                           cc::ISystemWindow::CC_WINDOW_INPUT_FOCUS
-                                                    : _windowInfo.flags;
-        if (!isDesktopWindowCreated) {
+                                                     : _windowInfo.flags;
+        std::call_once(_windowIsCreated, [&]() {
             createWindow(_windowInfo.title.c_str(), _windowInfo.width, _windowInfo.height, _windowInfo.flags);
-            isDesktopWindowCreated = true;
-        }
+        });
+
 #endif
 
         if (_debuggerInfo.enabled) {
@@ -76,14 +85,5 @@ public:
         runScript("jsb-adapter/jsb-builtin.js");
         runScript("main.js");
         return 0;
-    }
-
-protected:
-    std::string  _xxteaKey;
-    DebuggerInfo _debuggerInfo;
-    WindowInfo   _windowInfo;
-
-private:
-    bool         isDesktopWindowCreated { false };
-};
+}
 } // namespace cc
--- a/native/cocos/application/BaseGame.h
+++ b/native/cocos/application/BaseGame.h
@@ -57,12 +57,11 @@ public:
                                                           cc::ISystemWindow::CC_WINDOW_RESIZABLE |
                                                           cc::ISystemWindow::CC_WINDOW_INPUT_FOCUS
                                                     : _windowInfo.flags;
-        if (_windowInfo.x == -1 || _windowInfo.y == -1) {
-            createWindow(_windowInfo.title.c_str(), _windowInfo.width, _windowInfo.height, _windowInfo.flags);
-        } else {
-            createWindow(_windowInfo.title.c_str(),
-                         _windowInfo.x, _windowInfo.y, _windowInfo.width, _windowInfo.height, _windowInfo.flags);
+        if (!isDesktopWindowCreated) {
+                createWindow(_windowInfo.title.c_str(), _windowInfo.width, _windowInfo.height, _windowInfo.flags);
+                isDesktopWindowCreated = true;
         }
+        
 #endif
 
         if (_debuggerInfo.enabled) {
@@ -75,18 +74,9 @@ public:
         }
 
         setXXTeaKey(_xxteaKey);
-        registerListeners();
         runScript("jsb-adapter/jsb-builtin.js");
         runScript("main.js");
         return 0;
-    }
-    void registerListeners() {
-        EventDispatcher::addCustomEventListener("event_resize", [&](const CustomEvent& event) {
-            //TODO: get width and height
-            //_windowInfo.width = event.args
-            _windowInfo.width  = event.args[1].longVal;
-            _windowInfo.height = event.args[2].longVal;
-        });
     }
 
 protected:
@@ -94,5 +84,7 @@ protected:
     DebuggerInfo _debuggerInfo;
     WindowInfo   _windowInfo;
 
+private:
+    bool         isDesktopWindowCreated { false };
 };
 } // namespace cc
--- a/native/cocos/application/BaseGame.h
+++ b/native/cocos/application/BaseGame.h
@@ -67,7 +67,12 @@ int BaseGame::init() {
                                                           cc::ISystemWindow::CC_WINDOW_INPUT_FOCUS
                                                      : _windowInfo.flags;
         std::call_once(_windowIsCreated, [&]() {
-            createWindow(_windowInfo.title.c_str(), _windowInfo.width, _windowInfo.height, _windowInfo.flags);
+            if (_windowInfo.x == -1 && _windowInfo.y == -1) {
+                createWindow(_windowInfo.title.c_str(), _windowInfo.width, _windowInfo.height, _windowInfo.flags);
+            } else {
+                createWindow(_windowInfo.title.c_str(), 
+                    _windowInfo.x, _windowInfo.y, _windowInfo.width, _windowInfo.height, _windowInfo.flags);
+            }
         });
 
 #endif
--- a/native/cocos/application/BaseGame.h
+++ b/native/cocos/application/BaseGame.h
@@ -58,10 +58,9 @@ public:
                                                           cc::ISystemWindow::CC_WINDOW_INPUT_FOCUS
                                                     : _windowInfo.flags;
         if (!isDesktopWindowCreated) {
-                createWindow(_windowInfo.title.c_str(), _windowInfo.width, _windowInfo.height, _windowInfo.flags);
-                isDesktopWindowCreated = true;
+            createWindow(_windowInfo.title.c_str(), _windowInfo.width, _windowInfo.height, _windowInfo.flags);
+            isDesktopWindowCreated = true;
         }
-        
 #endif
 
         if (_debuggerInfo.enabled) {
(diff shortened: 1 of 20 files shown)


timlyeee:
Read me plz.

So I'm not sure whether it's ok to listen resize event here, but for native codes, it's impossible to get width and height as no information was given. here's the original code
```cpp
void EventDispatcher::doDispatchEvent(const char *eventName, const char *jsFunctionName, const std::vector<se::Value> &args) {
    if (!se::ScriptEngine::getInstance()->isValid()) {
        return;
    }

    if (eventName) {
        CustomEvent event;
        event.name = eventName;
#if CC_PLATFORM == CC_PLATFORM_WINDOWS
        CCASSERT(CC_GET_PLATFORM_INTERFACE(ISystemWindow) != nullptr, "System window interface does not exist");
        event.args->ptrVal = reinterpret_cast<void *>(CC_GET_PLATFORM_INTERFACE(ISystemWindow)->getWindowHandler());
#endif
        EventDispatcher::dispatchCustomEvent(event);
    }

    // dispatch to Javascript
    if (!se::ScriptEngine::getInstance()->isValid()) {
        return;
    }

    se::AutoHandleScope scope;
    assert(inited);

    se::Value func;
    __jsbObj->getProperty(jsFunctionName, &func);
    if (func.isObject() && func.toObject()->isFunction()) {
        func.toObject()->call(args, nullptr);
    }
}
``` 
when EventDispatcher::dispatchCustomEvent(event) is called, there's no information stored in this event, so for all listener in c++, they know nothing except "EVENT NAME".

My idea is that we have to load essential params into the event, so some parts of eventListener could be refactored.

native/cocos/bindings/event/EventDispatcher.cpp
```diff
@@ -256,12 +256,43 @@ void EventDispatcher::dispatchResizeEvent(int width, int height) {
...
+    
     se::ValueArray args;
     args.emplace_back(se::Value(jsResizeEventObj));
-    EventDispatcher::doDispatchEvent(EVENT_RESIZE, "onResize", args);
+    const char *jsFunctionName = "onResize";
+    //EventDispatcher::doDispatchEvent(EVENT_RESIZE, "onResize", args);
+
+    if (!se::ScriptEngine::getInstance()->isValid()) {
+        return;
+    }
```
minggo: no need here, as
- these codes will be invoked again in following codes
- can dispatch events to c++ event script engine is not ready

native/cocos/bindings/event/EventDispatcher.cpp
```diff
@@ -256,12 +256,43 @@ void EventDispatcher::dispatchResizeEvent(int width, int height) {
...
+
+    if (true) {
+        CustomEvent event;
+        event.name = EVENT_RESIZE;
+#if CC_PLATFORM == CC_PLATFORM_WINDOWS
```
minggo: why only pass width and height on windows?

native/cocos/bindings/event/EventDispatcher.cpp
```diff
@@ -256,12 +256,43 @@ void EventDispatcher::dispatchResizeEvent(int width, int height) {
...
+    // dispatch to Javascript
+    if (!se::ScriptEngine::getInstance()->isValid()) {
+        return;
+    }
+
+    se::Value func;
+    __jsbObj->getProperty(jsFunctionName, &func);
+    if (func.isObject() && func.toObject()->isFunction()) {
+        func.toObject()->call(args, nullptr);
+    }
```
minggo: Why not invoke `EventDispatcher::doDispatchEvent(EVENT_RESIZE, "onResize", args);`?

native/cocos/application/BaseGame.h
```diff
@@ -57,12 +57,11 @@ class BaseGame : public CocosApplication {
                                                           cc::ISystemWindow::CC_WINDOW_RESIZABLE |
                                                           cc::ISystemWindow::CC_WINDOW_INPUT_FOCUS
                                                     : _windowInfo.flags;
-        if (_windowInfo.x == -1 || _windowInfo.y == -1) {
-            createWindow(_windowInfo.title.c_str(), _windowInfo.width, _windowInfo.height, _windowInfo.flags);
-        } else {
-            createWindow(_windowInfo.title.c_str(),
-                         _windowInfo.x, _windowInfo.y, _windowInfo.width, _windowInfo.height, _windowInfo.flags);
+        if (!isDesktopWindowCreated) {
```
minggo: Use [std::call_once](https://en.cppreference.com/w/cpp/thread/call_once) ?

native/cocos/bindings/event/EventDispatcher.cpp
timlyeee: I've modified the logic of all, no need to consider it now

native/cocos/application/BaseGame.h
timlyeee: truly better!

native/cocos/bindings/event/EventDispatcher.cpp
timlyeee: updated, it's outdated now

native/cocos/platform/win32/modules/SystemWindow.cpp
```diff
@@ -86,5 +86,9 @@ void SystemWindow::copyTextToClipboard(const std::string &text) {
 SystemWindow::Size SystemWindow::getViewSize() const {
     return Size{static_cast<float>(_width), static_cast<float>(_height)};
 }
+void SystemWindow::setViewSize(uint32_t width, uint32_t height) {
```
qiuguohua: Does the mac platform also need to add this interface?

native/cocos/application/BaseGame.h
```diff
@@ -56,13 +65,11 @@ class BaseGame : public CocosApplication {
...
-                                                    : _windowInfo.flags;
-        if (_windowInfo.x == -1 || _windowInfo.y == -1) {
+                                                     : _windowInfo.flags;
+        std::call_once(_windowIsCreated, [&]() {
             createWindow(_windowInfo.title.c_str(), _windowInfo.width, _windowInfo.height, _windowInfo.flags);
-        } else {
-            createWindow(_windowInfo.title.c_str(),
-                         _windowInfo.x, _windowInfo.y, _windowInfo.width, _windowInfo.height, _windowInfo.flags);
-        }
+        });
```
qiuguohua: Need to call the interface to change the x, y coordinates of the window？

timlyeee: no need, as x, y coordinates means actually nothing here..

native/cocos/platform/win32/modules/SystemWindow.h
```diff
@@ -47,6 +47,7 @@ class CC_DLL SystemWindow : public ISystemWindow {
                            int h, int flags) override;
     uintptr_t getWindowHandler() const override;
     Size getViewSize() const override;
+    void      setViewSize(uint32_t width, uint32_t height) override;
```
minggo: unneeded spaces

native/cocos/application/BaseGame.h
qiuguohua: This is an interface that provides additional functionality. Although it is not currently called, developers may use it.
For example, we are currently displayed in the lower left corner. If the developer wants to change this operation, just set the coordinates of x and y directly.

timlyeee: understand, still need to check x and y for the first init

native/cocos/platform/interfaces/modules/ISystemWindow.h
```diff
@@ -89,6 +89,7 @@ class CC_DLL ISystemWindow : public OSInterface {
     }
     virtual uintptr_t getWindowHandler() const = 0;
     virtual Size      getViewSize() const      = 0;
+    virtual void      setViewSize(uint32_t width, uint32_t height) = 0;
```
qiuguohua: Is it better to write a non-pure virtual function here?

native/cocos/application/BaseGame.h
```diff
@@ -46,7 +46,16 @@ class BaseGame : public CocosApplication {
...
+    int init() override;
+
+protected:
+    std::string  _xxteaKey;
+    DebuggerInfo _debuggerInfo;
+    WindowInfo   _windowInfo;
+    std::once_flag _windowIsCreated;
+};
+
+int BaseGame::init() {
```
dumganhar: Need to put BaseGame::init to BaseGame.cpp file.

native/cocos/platform/interfaces/modules/ISystemWindow.h
timlyeee: Possibly, I can change it if you want

native/cocos/bindings/event/EventDispatcher.cpp
```diff
@@ -144,7 +143,8 @@ void EventDispatcher::dispatchTouchEvent(const TouchEvent &touchEvent) {
 
     se::ValueArray args;
     args.emplace_back(se::Value(jsTouchObjArray));
-    EventDispatcher::doDispatchEvent(nullptr, eventName, args);
+    EventDispatcher::doDispatchJsEvent(eventName, args);
+    EventDispatcher::doDispatchCustomEvent(nullptr, 0);
```
dumganhar: Why separate doDispatchEvent into two functions?
If will be easier to forget one of the invocation.

And what's the purpose of passing `nullptr, 0` to doDispatchCustomEvent?

native/cocos/application/BaseGame.h
```diff
@@ -46,44 +46,12 @@ class BaseGame : public CocosApplication {
...
-        runScript("main.js");
-        return 0;
-    }
+    int init() override;
 
 protected:
     std::string  _xxteaKey;
     DebuggerInfo _debuggerInfo;
     WindowInfo   _windowInfo;
+    std::once_flag _windowIsCreated;
```
dumganhar: It's a `once flag` rather than a `boolean` value, so the variable name `_windowIsCreated` is not suitable here.

native/cocos/bindings/event/EventDispatcher.h
```diff
@@ -212,19 +212,19 @@ class KeyboardEvent : public OSEvent {
     bool   shiftKeyActive = false;
     // TODO(mingo): support caps lock?
 };
-
+union ccUnionArg {
```
minggo: - please do not start with lowercase characters
- please use more meaningful type name, such as `UnionEventParameterType`

native/cocos/bindings/event/EventDispatcher.cpp
timlyeee: The mechanism of the event dispatcher does not support the situation if we want to dispatch the native event with several arguments. for example:
```cpp
//Somewhere customize the render event
EventDispatcher::addCustomEventListener(EVENT_RESIZE, [&](CustomEvent e){
 texture.size.x = e.args[0].intVal;
 texture.size.y = e.args[1].intVal;
 windowHandle = e.args[2].ptrVal;
})
```

but doDispatchEvent will trigger Js events and Native events at same time, without adding arguments into the CustomEvent, so most of time, listeners can only be triggered with an event name.
```cpp
void EventDispatcher::doDispatchEvent(const char *eventName, const char *jsFunctionName, const std::vector<se::Value> &args) {
...
if (eventName) {
        CustomEvent event;
        event.name = eventName;
        EventDispatcher::dispatchCustomEvent(event);
    }
...
}
```
also, the listener will get errors if there are no args in the event.
That's why I want to separate it to 2 methods

timlyeee: > And what's the purpose of passing nullptr, 0 to doDispatchCustomEvent?

I use va_list to accept multiple arguments, the disadvantage is that a descriptor should be given to specified the va_end

native/cocos/bindings/event/EventDispatcher.h
```diff
@@ -276,7 +269,8 @@ class EventDispatcher {
     static void     dispatchCustomEvent(const CustomEvent &event);
 
 private:
-    static void doDispatchEvent(const char *eventName, const char *jsFunctionName, const std::vector<se::Value> &args);
+    static void doDispatchJsEvent(const char *jsFunctionName, const std::vector<se::Value> &args);
+    static void doDispatchCustomEvent(const char *eventName, int argNum, ...);
```
minggo: Does this function only sent event to cpp codes, not sent to JS? If so, please use a more meaningful name.

timlyeee: change to doDispatchNativeEvent, "CustomEvent" means actually nothing here

native/cocos/bindings/event/EventDispatcher.h
timlyeee: use `EventParameterType` instead

qiuguohua (approved):

dumganhar (approved):

native/cocos/bindings/event/EventDispatcher.cpp
```diff
@@ -259,9 +260,17 @@ void EventDispatcher::dispatchResizeEvent(int width, int height) {
 
     jsResizeEventObj->setProperty("width", se::Value(width));
     jsResizeEventObj->setProperty("height", se::Value(height));
+    
     se::ValueArray args;
     args.emplace_back(se::Value(jsResizeEventObj));
-    EventDispatcher::doDispatchEvent(EVENT_RESIZE, "onResize", args);
+    const char *jsFunctionName = "onResize";
+    EventDispatcher::doDispatchJsEvent("onResize", args);
```
minggo, line 266: Why not use jsFunctionName?

native/cocos/bindings/event/EventDispatcher.cpp
```diff
@@ -259,9 +260,17 @@ void EventDispatcher::dispatchResizeEvent(int width, int height) {
...
+#if CC_PLATFORM == CC_PLATFORM_WINDOWS
+    //EventDispatcher::doDispatchNativeEvent(EVENT_RESIZE, 3, reinterpret_cast<void *>(CC_GET_PLATFORM_INTERFACE(ISystemWindow)->getWindowHandler()), width, height);
```
minggo: Please remove unneeded codes.

native/cocos/bindings/event/EventDispatcher.cpp
```diff
@@ -289,49 +293,52 @@ void EventDispatcher::dispatchOrientationChangeEvent(int orientation) {
...
 
 void EventDispatcher::dispatchCloseEvent() {
-    EventDispatcher::doDispatchEvent(EVENT_CLOSE, "onClose", se::EmptyValueArray);
+    EventDispatcher::doDispatchJsEvent("onClose", se::EmptyValueArray);
+    EventDispatcher::dispatchCustomEvent(EVENT_CLOSE, 0);
 }
 
 void EventDispatcher::dispatchDestroyWindowEvent() {
-    EventDispatcher::doDispatchEvent(EVENT_DESTROY_WINDOW, "", se::EmptyValueArray);
+    EventDispatcher::doDispatchJsEvent("", se::EmptyValueArray);
```
minggo: "" -> nullptr

native/cocos/bindings/event/EventDispatcher.cpp
```diff
@@ -259,9 +260,12 @@ void EventDispatcher::dispatchResizeEvent(int width, int height) {
 
     jsResizeEventObj->setProperty("width", se::Value(width));
     jsResizeEventObj->setProperty("height", se::Value(height));
+    
     se::ValueArray args;
     args.emplace_back(se::Value(jsResizeEventObj));
-    EventDispatcher::doDispatchEvent(EVENT_RESIZE, "onResize", args);
+    const char *jsFunctionName = "onResize";
+    EventDispatcher::doDispatchJsEvent(jsFunctionName, args);
```
minggo: Is jsFunctionName needed?

native/cocos/bindings/event/EventDispatcher.cpp
```diff
@@ -423,7 +430,23 @@ void EventDispatcher::removeAllEventListeners() {
...
+void EventDispatcher::dispatchCustomEvent(const char *eventName, int argNum, ...) {
+    if (!eventName || !strcmp(eventName, "")) {
+        return;
+    }
+    CustomEvent event;
+    event.name = eventName;    
+    va_list vl;
+    va_start(vl, argNum);
+    // Step through the list.
+    for (int i = 0; i < argNum; ++i) {
```
minggo, line 437: argNum is not needed, please refer the [doc](https://www.cplusplus.com/reference/cstdarg/va_end/#:~:text=Performs%20the%20appropriate%20actions%20to,been%20invoked%20from%20that%20function.)

timlyeee: Yes, I can make it with while-loop

timlyeee: Hard to end the loop, as all parameters can be translate to EventParameterType, so I finally keep the argNum..

native/cocos/bindings/event/EventDispatcher.cpp
timlyeee: As the `nullptr` means no jsFunction will be called, just delete it.

native/cocos/bindings/event/EventDispatcher.cpp
```diff
@@ -423,7 +427,23 @@ void EventDispatcher::removeAllEventListeners() {
     hashListenerId = 1;
 }
 
-void EventDispatcher::dispatchCustomEvent(const CustomEvent &event) {
+void EventDispatcher::dispatchCustomEvent(const char *eventName, int argNum, ...) {
+    if (!eventName || !strcmp(eventName, "")) {
```
minggo: May be it is better to use assert. And is it need to check `""`?

timlyeee: it's better to just return if eventName is nullptr or empty, it's not an error but a code with forgivable usage.

minggo: As this function will be invoked many times a frame. So i suggest to use asset instead. I think it is an error.

timlyeee: replaced with CCASSERT

minggo: please use CC_ASSERT instead, CCASSERT is removed in v3.6.

minggo (approved):

PR #10932, opened 2022-05-05
No longer use animation state to implement clip motion

shrinktofit:
Re: #

Formerly we internally use `AnimationState` to implement clip motion in animation graphs for fast prototyping. Now I decide to drop such implementation instead take up the "raw" animation evaluation way to support animation event and comming soon event-like feature.

### DIFF
--- a/cocos/core/animation/wrap.ts
+++ b/cocos/core/animation/wrap.ts
@@ -1,0 +1,74 @@
+import { WrapMode, WrapModeMask, WrappedInfo } from './types';
+
+export function wrap (
+    elapsedTime: number,
+    duration: number,
+    wrapMode: WrapMode,
+    repeatCount: number,
+    negativeSpeed: boolean,
+    info: WrappedInfo,
+): WrappedInfo {
+    let stopped = false;
+
+    let currentIterations = elapsedTime > 0 ? (elapsedTime / duration) : -(elapsedTime / duration);
+    if (currentIterations >= repeatCount) {
+        currentIterations = repeatCount;
+
+        stopped = true;
+        let tempRatio = repeatCount - (repeatCount | 0);
+        if (tempRatio === 0) {
+            tempRatio = 1;  // 如果播放过，动画不复位
+        }
+        elapsedTime = tempRatio * duration * (elapsedTime > 0 ? 1 : -1);
+    }
+
+    if (elapsedTime > duration) {
+        const tempTime = elapsedTime % duration;
+        elapsedTime = tempTime === 0 ? duration : tempTime;
+    } else if (elapsedTime < 0) {
+        elapsedTime %= duration;
+        if (elapsedTime !== 0) { elapsedTime += duration; }
+    }
+
+    let needReverse = false;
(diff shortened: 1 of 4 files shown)


SantyWang:
Why dont use animation state any more

shrinktofit:
> Why dont use animation state any more

We never said it to users that our motion states are driven by animation state internally.

Motion states have no concept of "play", "pause", "resume" or "stop", neither no invocation should be made on them nor conrresponding event should be triggered. It just does sample. Where, when and whether to sample is controlled by animation graph it within. For this reason, I think the animation evaluation should be the underlying basic building block.

In the other hand, the speed, wrap mode, repeat count etc. of animation clips can not be controlled directly by users though they can be in the case of animation states.

However what actually enforces this change is an in-development feature we're planing to introduce is latest: subregion play. This feature by design requires certain [interoperations(for demonstration only)](https://github.com/shrinktofit/engine/blob/29923b088939d394f631676c6300bcef6a9ee764/cocos/core/animation/animation-clip.ts#L944-L1035) with animation clip evaluator and notifies the evaluator about whether the state of the animation is "play", "pause", "stop". If we insist on using animation states, we have to call playback methods(`play()`, `stop()`, `pause()` etc) on animation states. Playback methods has theirown semantic on animation states. However as I mentioned above, the playback of animation clip should be controlled by animation graph instead.

SantyWang:
OK，I got your point

cocos/core/animation/wrap.ts
```diff
@@ -0,0 +1,74 @@
...
+    let stopped = false;
+
+    let currentIterations = elapsedTime > 0 ? (elapsedTime / duration) : -(elapsedTime / duration);
+    if (currentIterations >= repeatCount) {
+        currentIterations = repeatCount;
+
+        stopped = true;
+        let tempRatio = repeatCount - (repeatCount | 0);
+        if (tempRatio === 0) {
+            tempRatio = 1;  // 如果播放过，动画不复位
```
SantyWang, line 20: dont use chinese in comment

cocos/core/animation/wrap.ts
```diff
@@ -0,0 +1,74 @@
+import { WrapMode, WrapModeMask, WrappedInfo } from './types';
+
+export function wrap (
```
SantyWang, line 3: I feel that this name may not be clear enough, `wrap` is too general, this function is used to calculate the time and period information in the corresponding wrap mode, so `calcWrappedInfo` , `calcWrappedCycle` or other similar names maybe better?

holycanvas, line 3: I feel that this name may not be clear enough, `wrap` is too general, this function is used to calculate the time and period information in the corresponding wrap mode, so `calcWrappedInfo` , `calcWrappedCycle` or other similar names maybe better?

shrinktofit:
I'll temporarily draft this PR, to wait for another.

shrinktofit:
Edit: this PR is now marked as "under consideration" by myself.

shrinktofit:
Closed in favor of https://github.com/cocos/cocos-engine/pull/13503

PR #10941, opened 2022-05-05
load libGLESv3.so by default while user choose glesv3

bofeng-song:


native/CMakeLists.txt
```diff
@@ -2383,13 +2383,10 @@ if(ANDROID)
     )
 
     find_library(LIB_EGL NAMES EGL)
-    find_library(LIB_GLESv2 NAMES GLESv2)
     target_link_libraries(cocos2d PUBLIC
         android
         OpenSLES
         ${LIB_EGL}
-        ${LIB_GLESv2}
-        GLESv2
```
minggo, line 2392: What will happen if want to use gles2?

native/cocos/renderer/gfx-gles3/GLES3Wrangler.cpp
```diff
@@ -60,7 +60,7 @@ bool gles3wOpen() {
     #if __OHOS__
     libgles = dlopen("libGLESv3.so", RTLD_LAZY | RTLD_GLOBAL);
     #else
-    libgles = dlopen("libGLESv2.so", RTLD_LAZY | RTLD_GLOBAL);
+    libgles = dlopen("libGLESv3.so", RTLD_LAZY | RTLD_GLOBAL);
```
minggo: Don't need special codes for __OHOS__

native/CMakeLists.txt
bofeng-song: Nothing will happen, we load libGLESv2 dynamically

native/cocos/renderer/gfx-gles3/GLES3Wrangler.cpp
bofeng-song: > Don't need special codes for **OHOS**

has been fixed

minggo (approved):

PR #10945, opened 2022-05-05
fix update octree twice

stanleyljl:
Re: #https://github.com/cocos/3d-tasks/issues/9210

Changelog:
 * fix update octree twice

### DIFF
--- a/native/cocos/scene/Model.h
+++ b/native/cocos/scene/Model.h
@@ -122,7 +122,7 @@ public:
     inline void setBounds(geometry::AABB *world) {
         _worldBounds = world;
         _modelBounds->set(_worldBounds->getCenter(), _worldBounds->getHalfExtents());
-        _needUpdateOctree = true;
+        _worldBoundsDirty = true;
     }
     inline void setInstancedAttributeBlock(const InstancedAttributeBlock &val) {
         _instanceAttributeBlock = val;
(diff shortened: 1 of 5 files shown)


native/cocos/scene/Model.h
```diff
@@ -188,6 +193,7 @@ class Model : public RefCounted {
 
     Type _type{Type::DEFAULT};
     bool _localDataUpdated{false};
+    bool _needUpdateOctree{true};
```
minggo: When set `_needUpdateOctree ` to true?

native/cocos/scene/Model.h
```diff
@@ -95,6 +95,7 @@ class Model : public RefCounted {
 
     virtual void updateTransform(uint32_t stamp);
     virtual void updateUBOs(uint32_t stamp);
+    void updateOctree();
```
minggo, line 98: Why Model has a function to update octree?

native/cocos/scene/RenderScene.cpp
```diff
@@ -79,6 +79,7 @@ void RenderScene::update(uint32_t stamp) {
         if (model->isEnabled()) {
             model->updateTransform(stamp);
             model->updateUBOs(stamp);
+            model->updateOctree();
```
minggo, line 82: octree will be updated many times

native/cocos/scene/Model.h
stanleyljl: while world bound is updated.

native/cocos/scene/Model.h
stanleyljl: model update itself in octree

native/cocos/scene/RenderScene.cpp
stanleyljl: If world bound is updated,  the model in octree need update, not update the whole octree

native/cocos/scene/Model.h
minggo: Got it. I suggest to change `_needUpdateOctree ` to `_isWorldBoundsDirty`.

stanleyljl: 1

minggo (approved):

PR #10948, opened 2022-05-05
fix skeleton animation scene destroy crash caused by inactive Skinned…

zhakesi:
* https://github.com/cocos/3d-tasks/issues/12154


shrinktofit (changes requested):
Please supply this change with an unit test

PR #10956, opened 2022-05-06
fix math comments

stanleyljl:
Re: #

Changelog:
 * fix math comments

### DIFF
--- a/cocos/core/math/utils.ts
+++ b/cocos/core/math/utils.ts
@@ -23,8 +23,6 @@
  THE SOFTWARE.
  */
 
-
-
 import { ValueType } from '../value-types';
 import { IVec3Like } from './type-define';
 
--- a/cocos/core/math/utils.ts
+++ b/cocos/core/math/utils.ts
@@ -245,8 +245,8 @@ export function inverseLerp (from: number, to: number, value: number) {
 /**
  * @en Compare the absolute values of all components and the component with the largest absolute value will be returned.
  * @zh 对所有分量的绝对值进行比较大小，返回绝对值最大的分量。
- * @param v 类 Vec3 结构
- * @returns 绝对值最大的分量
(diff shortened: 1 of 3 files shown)


cocos/core/math/utils.ts
```diff
@@ -245,8 +245,8 @@ export function inverseLerp (from: number, to: number, value: number) {
 /**
  * @en Compare the absolute values of all components and the component with the largest absolute value will be returned.
  * @zh 对所有分量的绝对值进行比较大小，返回绝对值最大的分量。
- * @param v 类 Vec3 结构
- * @returns 绝对值最大的分量
+ * @param v vec3 like value
+ * @returns max absolute component
```
minggo, line 247: It is better to keep both English and Chinese.

cocos/core/value-types/value-type.ts
```diff
@@ -59,8 +59,8 @@ export class ValueType {
      * @zh
      * 判断当前值是否与指定值相等。此判断应当具有交换性，即满足 `this.equals(other) === other.equals(this)`。
      * 本方法的基类版本简单地返回 `false`。
-     * @param other 相比较的值。
-     * @returns 相等则返回 `true`，否则返回 `false`。
+     * @param other The other object
+     * @returns `true` if equal, otherwise returns `false`
```
minggo, line 62: The same as above.

stanleyljl:
skip ci

cocos/core/math/utils.ts
stanleyljl: A large number of parameters are lack of Chinese notes, so it is suggested to change them uniformly later.

minggo (approved):

pandamicro:
https://github.com/cocos/cocos-engine/issues/10840

PR #10964, opened 2022-05-06
Spotlight interface description change for v3.5.1

troublemaker52025:
Re: #

Changelog:
 *

### DIFF
--- a/cocos/core/pipeline/pipeline-ubo.ts
+++ b/cocos/core/pipeline/pipeline-ubo.ts
@@ -277,7 +277,7 @@ export class PipelineUBO {
             Mat4.invert(_matShadowView, (light as any).node.getWorldMatrix());
             Mat4.toArray(sv, _matShadowView, UBOShadow.MAT_LIGHT_VIEW_OFFSET);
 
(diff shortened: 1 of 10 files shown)


troublemaker52025:
* https://github.com/cocos/3d-tasks/issues/12184

minggo (approved):

pandamicro:
It does't seem appropriate, if aspect for the shadow projection matrix is useful and user need to control it, then we should expose it at component level, not removing it

jk20012001:
Aspect is an attribute of camera, not light source. It use for fit render target x/y. Shadowmap width always equals to height, so dirlight / spotlight / pointlight should use 1.0 in shadow project matrix.
If you modify shadowmap aspect, lighting model also needs to be modified, but we don't have any aspect-related calculations in lighting.
UE4 and Unity don't have this attribute at all. So user don't need to obtain or modify this value, I suggest that remove it from light component.

pandamicro:
Ok, got it, @troublemaker52025 please resolve conflicts

troublemaker52025:
@cocos-robot run test cases

PR #10975, opened 2022-05-06
Remove unused code: defineOwnProperty("__native_ptr__",...)

dumganhar:
Re: #

Changelog:
 *

### DIFF
--- a/native/cocos/bindings/jswrapper/sm/Object.cpp
+++ b/native/cocos/bindings/jswrapper/sm/Object.cpp
@@ -580,7 +580,6 @@ void Object::setPrivateObject(PrivateObjectBase *data) {
     internal::setPrivate(__cx, obj, data, this, &_internalData, _finalizeCb); //TODO(cjh): how to use _internalData?
     NativePtrToObjectMap::emplace(data->getRaw(), this);
     _privateObject = data;
-    defineOwnProperty("__native_ptr__", se::Value(static_cast<uint64_t>(reinterpret_cast<uintptr_t>(data->getRaw()))), false, false, false);
 }
 
 PrivateObjectBase *Object::getPrivateObject() const {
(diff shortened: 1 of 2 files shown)


minggo:
Is there a way to distinguish native binding object or pure javascript object?

PatriceJiang (approved):

dumganhar:
> Is there a way to distinguish native binding object or pure javascript object?
What kinds of situations that need to distinguish? 
If we really need to distinguish, we should do it in another way in class registration. For example, do it in register class method. Define property in runtime dynamically will make code run slower.
```c++
bool js_register_ccc(se::Object* obj) // NOLINT(readability-identifier-naming)
{
    auto* cls = se::Class::create("ccc", obj, nullptr, _SE(js_ccc_constructor));
    ......
    cls->defineProperty("nativePtr", _SE(js_ccc_get_nativePtr), nullptr); // Add this line 
    cls->defineStaticProperty("isJSBClass", _SE(js_ccc_static_get_isJSBClass), nullptr);// Add this line 
    cls->install();
    JSBClassType::registerClass<cc::IPassInfoFull>(cls);

    __jsb_cc_ccc_proto = cls->getProto();
    __jsb_cc_ccc_class = cls;

    se::ScriptEngine::getInstance()->clearException();
    return true;
}
```

minggo:
> What kinds of situations that need to distinguish?

As some classes are implemented both in TS and c++, sometimes we want to make sure if TS related codes are deleted correctly, then need it.

minggo:
This PR looks good to me, just need another way to distinguish if an object is a pure TS object or a bound object in debug mode.

minggo (approved):

dumganhar:
> This PR looks good to me, just need another way to distinguish if an object is a pure TS object or a bound object in debug mode.

Fixed at https://github.com/cocos/cocos-engine/pull/10993

PR #10984, opened 2022-05-07
fix android crash due to surface not ready

bofeng-song:


bofeng-song:
gles will cache the cmd, but vulkan will call the api immediately

minggo (approved):

star-e (approved):

PR #10990, opened 2022-05-07
[Fix bug] Play and pause audio in the same frame, the pause operation won't work on native

timlyeee:
Re: 
 * https://github.com/cocos/3d-tasks/issues/10825
 * https://github.com/cocos/3d-tasks/issues/12036

Changelog:
 * Now once the AudioEngine runs play2d, it will directly set AudioInfo.state correspond to AudioState::PLAYING, which was changed by a scheduler. 
 * Since the AudioEngine has to actually "trust" the AudioEngine-impl, then all errors should be asserted by AudioEngine-impl rather than working with delay schedule.

### DIFF
--- a/native/cocos/audio/apple/AudioPlayer.mm
+++ b/native/cocos/audio/apple/AudioPlayer.mm
@@ -204,9 +204,6 @@ bool AudioPlayer::play2d() {
             ALOGE("%s:alSourcePlay error code:%x", __PRETTY_FUNCTION__, alError);
             break;
         }
-
-        ALint state;
-        alGetSourcei(_alSource, AL_SOURCE_STATE, &state);
         /** Due to the bug of OpenAL, when the second time OpenAL trying to mix audio into bus, the mRampState become kRampingComplete, and for those oalSource whose mRampState == kRampingComplete, nothing happens.
          * OALSource::Play{
          *      switch(mState){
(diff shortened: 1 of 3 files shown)


native/cocos/audio/apple/AudioPlayer.mm
```diff
@@ -207,7 +207,19 @@ of this software and associated engine source code (the "Software"), a limited,
 
         ALint state;
         alGetSourcei(_alSource, AL_SOURCE_STATE, &state);
```
dumganhar: Is `alGetSourcei` still needed?

native/cocos/audio/apple/AudioPlayer.mm
```diff
@@ -207,7 +207,19 @@ of this software and associated engine source code (the "Software"), a limited,
...
+         * OALSource::Play{
+         *      switch(mState){
+         *       case kTransitionToStop:
+         *       case kTransitionToStop:
+         *         if(mRampState != kRampingComplete){..}
+         *         break;
+         *      }
+         * }
+         * So the assert here will trigger this bug as aolSource is reused.
+         * Replace OpenAL with AVAudioEngine on V3.6 mightbe helpful
```
minggo, line 217: Good comment.

minggo (approved):

native/cocos/audio/apple/AudioPlayer.mm
timlyeee: No, I'll delete it

dumganhar (approved):

PR #11009, opened 2022-05-10
move font builtin resource from 3d to base

stanleyljl:
Re: #

Changelog:
 * move font builtin resource from 3d to base


stanleyljl:
skip ci

SantyWang (approved):

PR #11012, opened 2022-05-10
[JSB] Check whether node._children && node._children.length are valid in updateChildrenForDeserialize

dumganhar:
[JSB] Optimize nodeProto._onBatchCreated, reduce the count of visiting properties.

### DIFF
--- a/cocos/core/utils/jsb-utils.ts
+++ b/cocos/core/utils/jsb-utils.ts
@@ -110,26 +110,25 @@ export function updateChildrenForDeserialize (node: Node) {
     if (!node) {
         return;
     }
+
     // @ts-expect-error: jsb related codes.
-    node._setChildren(node._children);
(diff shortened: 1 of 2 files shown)


cocos/core/utils/jsb-utils.ts
```diff
@@ -110,26 +110,25 @@ export function updateChildrenForDeserialize (node: Node) {
     if (!node) {
         return;
     }
+
     // @ts-expect-error: jsb related codes.
-    node._setChildren(node._children);
+    const children = node._children;
+    if (!children) {
+        return;
```
dumganhar, line 117: Fix incorrect animation state of https://github.com/cocos-creator/example-marionette demo .
node._children is probably an `undefined` value.

minggo (approved):

PR #11015, opened 2022-05-10
reset gradle & jdk11 dependency

bofeng-song:


native/cocos/platform/android/libgame-sdk/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,4 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.cocos.gamesdk">
+
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
```
minggo: Why need network permission?

minggo:
Why need both source codes and jar?

native/cocos/platform/android/libgame-sdk/src/main/AndroidManifest.xml
bofeng-song: Copy from libcocos, it can be remove

bofeng-song:
> Why need both source codes and jar?

We've change some code, source code only store. App uses jar only

PatriceJiang (approved):

minggo (approved):

PR #11023, opened 2022-05-10
fix etc1 related compress texture crash

novain1102:


minggo (approved):

yiwenxue (approved):

cocos/core/assets/asset-enum.ts
```diff
@@ -131,7 +131,7 @@ export enum PixelFormat {
      * @zh 包含 RGBA 通道的 PVR 2BPP 压缩纹理格式
      * 这种压缩纹理格式贴图的高度是普通 RGB_PVRTC_2BPPV1 贴图高度的两倍，使用上半部分作为原始 RGB 通道数据，下半部分用来存储透明通道数据。
      */
-    RGB_A_PVRTC_2BPPV1 = CUSTOM_PIXEL_FORMAT++,
+    RGB_A_PVRTC_2BPPV1 = CUSTOM_PIXEL_FORMAT,
```
minggo, line 134: It is better to set CUSTOM_PIXEL_FORMAT const as ESLint said.
And it is more clear now.

novain1102: got it.

LinYunMo (approved):

PR #11036, opened 2022-05-11
fix tween-action related change scene crash

novain1102:
Temporary fix: need to refactory after **https://github.com/cocos/3d-tasks/issues/11110**

### DIFF
--- a/cocos/tween/actions/action-manager.ts
+++ b/cocos/tween/actions/action-manager.ts
@@ -30,6 +30,7 @@ import { errorID, logID, assertID } from '../../core/platform/debug';
 import { Action } from './action';
 import { Node, CCObject } from '../../core';
 import { legacyCC } from '../../core/global-exports';
+import { JSB } from '../../core/default-constants';
 
 let ID_COUNTER = 0;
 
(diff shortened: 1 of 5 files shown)


dumganhar (approved):

cocos/tween/actions/action-manager.ts
```diff
@@ -457,7 +458,14 @@ export class ActionManager {
             locCurrTarget = this._currentTarget;
 
             const target = locCurrTarget.target;
-            if (target instanceof CCObject && !target.isValid) {
+            let isCCObject = target instanceof CCObject;
```
SantyWang: It might be better to encapsulate a method to do this

minggo: Yep, it is better. But to implement it in which file?

novain1102: Encapsulated it into Object.ts as a helper funcniton

SantyWang (approved):

dumganhar (approved):

minggo (approved):

PR #11052, opened 2022-05-12
Instantiate all FastDefined objects as value type

jareguo:
Re: [#12248](https://github.com/cocos/3d-tasks/issues/12248)

This is an unproven fix because it will deep instantiate all fast defined object.
**Please look for all classes registered with fastDefine, such as RenderPipeline?**
And make sure they are not referenced by CCClass property, so that multiple copies are not accidentally instantiated when copying scene nodes.

* Instantiate all FastDefined object as value type

Compatibility Impact:
* Change instantiate behaviour for user custom classes defined with `fastDefine`.

### DIFF
--- a/cocos/core/data/class.ts
+++ b/cocos/core/data/class.ts
@@ -406,14 +406,15 @@ export function CCClass<TFunction> (options: {
  * 检查构造函数是否由 `Class` 创建。
  * @method _isCCClass
  * @param {Function} constructor
+ * @param {Boolean} [allowFastDefined=false]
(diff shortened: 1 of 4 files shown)


shrinktofit:
I‘ve summitted a [PR](https://github.com/jareguo/engine/pull/6) which, instead of adding new parameter,  introduce a new method to decide if a class is cc-class or is fast defined. I also provided an unit test.

shrinktofit (approved):

SantyWang (approved):

PR #11066, opened 2022-05-13
V3.5.1 fix clear particle bug

zxx43:
Re: #

Changelog:
 *


cocos/particle/renderer/trail.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-len */
```
pandamicro: Please fix the lengthy lines, it's not that complicated, just break the lines

cocos/particle/renderer/trail.ts
```diff
@@ -755,6 +756,10 @@ export default class TrailModule {
         }
     }
 
+    public getModel () {
```
pandamicro: Please add public method at correct place, along with other public methods

cocos/particle/particle-system.ts
```diff
@@ -1130,6 +1130,19 @@ export class ParticleSystem extends RenderableComponent {
...
         }
+
+        if (this.getParticleCount() > 0) {
+            if (!this._isCulled) {
+                if (!this.processor.getModel()?.scene) {
+                    this.processor.attachToScene();
+                }
+                if (this._trailModule && this._trailModule.enable) {
+                    if (!this._trailModule.getModel()?.scene) {
+                        this._trailModule._attachToScene();
```
pandamicro: It's not optimal to add such logic in update phase, we should be precise about when we need to reattach processor and trail models

pandamicro (commented):
From the issue, I think it's better just to clear previous position for trail module. I'm not sure the purpose of managing scene attachment in `update` and `beforeRender`, need a little bit explanation on the approach

pandamicro (approved):

PR #11078, opened 2022-05-14
when user's font is deleted, the _userDefinedFont should be assigned as null

Zhmz:
Re: [#12270](https://github.com/cocos/3d-tasks/issues/12270)

### Changelog

* when user's font is deleted, the _userDefinedFont should be assigned as null

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/2d/components/label.ts
+++ b/cocos/2d/components/label.ts
@@ -439,8 +439,12 @@ export class Label extends Renderable2D {
         // if delete the font, we should change isSystemFontUsed to true
         this._isSystemFontUsed = !value;
 
-        if (EDITOR && value) {
-            this._userDefinedFont = value;
+        if (EDITOR) {
+            if (value) {
+                this._userDefinedFont = value;
+            } else {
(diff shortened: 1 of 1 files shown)


LinYunMo (approved):

cocos/2d/components/label.ts
```diff
@@ -439,8 +439,12 @@ export class Label extends Renderable2D {
         // if delete the font, we should change isSystemFontUsed to true
         this._isSystemFontUsed = !value;
```
VisualSJ, line 440: Why do these properties need to change together
Each property should take care of itself

Zhmz: Syntactically, _userDefinedFont is just a field instead of an attribute, it doesn't have getter or setter.
Functionally, _userDefinedFont is a auxiliary variable for usesystemfont and font attributes, its assignments should be synchronized with the assignments of font.

PR #11085, opened 2022-05-14
v3.5.1 Support NX platform and NX windows platform

qiuguohua:
Support NX platform and NX windows platform

https://github.com/cocos/3d-tasks/issues/10893


PPpro (approved):

star-e (approved):

native/cmake/predefine.cmake
```diff
@@ -215,6 +218,8 @@ elseif(ANDROID)
 elseif(OHOS)
     include(${CMAKE_CURRENT_LIST_DIR}/../templates/cmake/ohos.cmake)
 elseif(QNX)
+elseif(NX OR NX_WINDOWS)
```
dumganhar, line 221: Leave empty for elseif(QNX) ??  Delete the empty block?

oahcy (approved):

native/cmake/predefine.cmake
qiuguohua: Because of the addition of error handling, when it is not specified, an error will be reported and the execution will be terminated.
WXWorkCapture_16525152617142
Another way is to remove error handling

native/cocos/bindings/auto/jsb_spine_auto.cpp
```diff
@@ -15095,7 +15098,9 @@ bool js_register_spine_SkeletonAnimation(se::Object* obj) // NOLINT(readability-
     cls->defineFunction("setTrackEventListener", _SE(js_spine_SkeletonAnimation_setTrackEventListener));
     cls->defineFunction("setTrackInterruptListener", _SE(js_spine_SkeletonAnimation_setTrackInterruptListener));
     cls->defineFunction("setTrackStartListener", _SE(js_spine_SkeletonAnimation_setTrackStartListener));
+#if SCRIPT_ENGINE_TYPE != SCRIPT_ENGINE_QUICKJS
     cls->defineFunction("ctor", _SE(js_spine_SkeletonAnimation_ctor));
+#endif
```
dumganhar, line 15103: Modify tools/spine.ini, 
Make `classes_need_extend` empty.
It's used in Creator 2.x.
Since Creator 3.x use typescript for writing engine code, there is no need to have `ctor` stuff now.

minggo: I think it is better to create another issue for it.

minggo:
effects may change after releasing v3.5.1

native/cocos/bindings/auto/jsb_spine_auto.cpp
dumganhar: spine doesn't use the ctor functionality. so the toos/spine.ini doesn't need to set `SkeletonAnimation` to classes_need_extend section.
I think it's old code and not deleted in 3.x.

I create an issue here: https://github.com/cocos/cocos-engine/issues/11091

dumganhar (approved):

native/cmake/predefine.cmake
dumganhar: OK, just keep it as it's now.

star-e (approved):

minggo (approved):

Issue #11027, opened 2022-05-11
Subsystem start up and shut down


minggo:
Currently, engine uses the codes to initialize a subsystems
```c++
Engine::Engine() {
    _scheduler = std::make_shared<Scheduler>();
    FileUtils::getInstance()->addSearchPath("Resources", true);
    FileUtils::getInstance()->addSearchPath("data", true);
    EventDispatcher::init();
    se::ScriptEngine::getInstance();

#if CC_USE_PROFILER
    Profiler::getInstance();
#endif
}
```

It has some problems:
- use different ways to initialize a subsystem
   - use `XXX::getInstance()`
   - use `XXX::init()`
   - just construct a object instance
- use the same method to initialize a subsystem and get the singleton object
  - it is not a clear
  - it is uncontrollable, each invoking of the static method may initialize it again

# How to improve

It is better to have a unify way to initialize subsystems. There are two ways to do it as "Game Engine Architecture" suggests:
- Method1: use global static object with default constructor, and have startup and shutdown function
```c++
// System1.cpp
class SubSystem1 {
public:
    SubSystem1() { // do nothing }
    ~SubSystem1() { // do nothing }
    void startUp() { // start up the subsystem }
    void shutDown() { // shut down the subsystem }
};

// Engine.cpp
Subsystem1 gSubSystem1;
Engine() {
    gSubSystem1.startUp();
    ...
}

~Engine() {
    gSubSystem1.shutDown();
}
```
- Method2: only creates subsystem instance in engine constructor
```c++
// System1.cpp
class SubSystem1 {
public:
    SubSystem1() { // do nothing }
    ~SubSystem1() { // do nothing }
    void startUp() { // start up the subsystem }
    void shutDown() { // shut down the subsystem }
};

// Engine.h
class Engine {
private:
    SubSystem1 *_subSystem1{nullptr}
};

// Engine.cpp
Engine() {
     _subSystem1 = new SubSystem1;
     subsystem1->startUp();
}

~Engine() {
    _subSystem1->shutDown();
    delete _subSystem1;
}
```

# How to get the subsystem instance

There are two ways to get the subsystem:
- use engine object to get subsystem
```c++
auto *subSystem1 = engine->getSubSystem1();
subSystem1->doSomething();
```
- use subsystem static method
```
SubSystem1::getInstance()->doSomething();
```

I think use subsystem static method is more convenient

stanleyljl:
[...]

PR #11089, opened 2022-05-14
refactor startup/shutdown of file system

minggo:
Re: #https://github.com/cocos/cocos-engine/issues/11027

### Changelog

* do not use `FileUtils::getInstance()` to initialize FileUtils, `FileUtils::getInstance()` only returns the singleton, after we implements RTTI, will deprecated `FileUtils::getInstance()` and use `engine->getSubSystem()` instead
* deprecated `FileUtils::destroyInstance()`, just delete the singleton in engine.

### DIFF
--- a/native/cocos/platform/qnx/FileUtils-qnx.h
+++ b/native/cocos/platform/qnx/FileUtils-qnx.h
@@ -30,6 +30,9 @@ namespace cc {
 class CC_DLL FileUtilsQNX : public FileUtils {
 public:
     friend class FileUtils;
+    FileUtilsQNX();
+    ~FileUtils() override = default;
+ 
     bool isFileExistInternal(const ccstd::string &filename) const override;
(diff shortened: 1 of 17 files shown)


native/cocos/platform/qnx/FileUtils-qnx.h
```diff
@@ -30,6 +30,9 @@ namespace cc {
 class CC_DLL FileUtilsQNX : public FileUtils {
 public:
     friend class FileUtils;
+    FileUtilsQNX();
+    ~FileUtils() override = default;
```
dumganhar: ~FileUtils() -> ~FileUtilsQNX()

minggo: Good catch.

dumganhar (approved):

qiuguohua (approved):

PR #11094, opened 2022-05-14
Sorting Layer & Sorting order add

LinYunMo:
Re: https://github.com/cocos/3d-tasks/issues/12273
Re: https://github.com/cocos/3d-tasks/issues/12287

### Changelog

* add Sorting Layer & Sorting order
* wait the editor support

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/core/components/model-renderer.ts
+++ b/cocos/core/components/model-renderer.ts
@@ -24,11 +24,16 @@
  */
 
 import {
-    ccclass, serializable,
+    ccclass, range, serializable,
 } from 'cc.decorator';
 import { scene } from '../renderer';
 import { Layers } from '../scene-graph/layers';
 import { Renderer } from './renderer';
+import { clamp } from '../math/utils';
+import { SortingManager } from '../scene-graph/sorting-manager';
+
+const MAX_UINT_NUM = (1 << 15) - 1;
+const MIN_UINT_NUM = -1 << 15;
 
 /**
  * @en Base class for all rendering components containing model.
@@ -38,6 +43,11 @@ import { Renderer } from './renderer';
 export class ModelRenderer extends Renderer {
     @serializable
     protected _visFlags = Layers.Enum.NONE;
+
+    @serializable
+    protected _sortingLayerID = 0;
+    @serializable
+    protected _sortingOrder = 0;
 
     /**
      * @zh 组件所属层，影响该组件下的所有 model 的 visFlags
@@ -50,6 +60,44 @@ export class ModelRenderer extends Renderer {
     set visibility (val) {
         this._visFlags = val;
         this._onVisibilityChange(val);
+    }
+
+    /**
+     * @zh 组件所属排序层 id，影响组件的渲染排序。
+     * @en The sorting layer id of the component, which affects the rendering order of the component.
+     */
+    // Todo,how to show on inspector
+    get sortingLayerID () {
+        return this._sortingLayerID;
+    }
+    set sortingLayerID (val) {
+        if (val === this._sortingLayerID || !SortingManager.idIsValid(val)) return;
+        this._sortingLayerID = val;
+        this._updateSortingPriority();
+    }
+
+    /**
+     * @zh 组件在当前排序层中的顺序。
+     * @en Model Renderer's order within a sorting layer.
+     */
+    @range([MIN_UINT_NUM, MAX_UINT_NUM, 1])
+    get sortingOrder () {
+        return this._sortingOrder;
+    }
+    set sortingOrder (val) {
+        if (val === this._sortingOrder) return;
+        this._sortingOrder = clamp(val, MIN_UINT_NUM, MAX_UINT_NUM);
+        this._updateSortingPriority();
+    }
+
+    /**
+     * @zh 通过 name 设置组件的 sortingLayer
+     * @en Set the sortingLayer of the component by name
+     * @param name
+     */
+    public setSortingLayerByName (name: string) {
+        const id = SortingManager.getSortingIDFromName(name);
+        this.sortingLayerID = id;
     }
 
     protected _models: scene.Model[] = [];
@@ -70,5 +118,15 @@ export class ModelRenderer extends Renderer {
     }
 
     protected _onVisibilityChange (val) {
+    }
+
+    protected _updateSortingPriority () {
+        const sortingLayerValue = SortingManager.getSortingLayerIndex(this._sortingLayerID);
+        const sortingPriority = SortingManager.getSortingPriority(sortingLayerValue, this._sortingOrder);
+        if (this._models.length > 0) {
+            for (let i = 0; i < this._models.length; i++) {
+                this._models[i].sortingPriority = sortingPriority;
+            }
+        }
     }
 }
--- a/cocos/core/components/model-renderer.ts
+++ b/cocos/core/components/model-renderer.ts
@@ -41,14 +41,6 @@ const MIN_UINT_NUM = -1 << 15;
  */
 @ccclass('cc.ModelRenderer')
 export class ModelRenderer extends Renderer {
-    @serializable
-    protected _visFlags = Layers.Enum.NONE;
-
-    @serializable
-    protected _sortingLayerID = 0;
-    @serializable
-    protected _sortingOrder = 0;
-
     /**
      * @zh 组件所属层，影响该组件下的所有 model 的 visFlags
      * @en The layer of the current component, which affects all the visFlags of the models belonging to this component.
@@ -67,12 +59,12 @@ export class ModelRenderer extends Renderer {
      * @en The sorting layer id of the component, which affects the rendering order of the component.
      */
     // Todo,how to show on inspector
-    get sortingLayerID () {
-        return this._sortingLayerID;
+    get sortingLayer () {
+        return this._sortingLayer;
     }
-    set sortingLayerID (val) {
-        if (val === this._sortingLayerID || !SortingManager.idIsValid(val)) return;
-        this._sortingLayerID = val;
+    set sortingLayer (val) {
+        if (val === this._sortingLayer || !SortingManager.isLayerValid(val)) return;
+        this._sortingLayer = val;
         this._updateSortingPriority();
     }
 
@@ -96,11 +88,9 @@ export class ModelRenderer extends Renderer {
      * @param name
      */
     public setSortingLayerByName (name: string) {
-        const id = SortingManager.getSortingIDFromName(name);
-        this.sortingLayerID = id;
+        const id = SortingManager.getLayerByName(name);
+        this.sortingLayer = id;
     }
-
-    protected _models: scene.Model[] = [];
 
     /**
      * @zh 收集组件中的 models
@@ -109,6 +99,20 @@ export class ModelRenderer extends Renderer {
      */
     public _collectModels (): scene.Model[] {
         return this._models;
+    }
+
+    @serializable
+    protected _visFlags = Layers.Enum.NONE;
+
+    @serializable
+    protected _sortingLayer = 0;
+    @serializable
+    protected _sortingOrder = 0;
+
+    protected _models: scene.Model[] = [];
+
+    protected onEnable () {
+        this._updateSortingPriority();
     }
 
     protected _attachToScene () {
@@ -121,7 +125,7 @@ export class ModelRenderer extends Renderer {
     }
 
     protected _updateSortingPriority () {
-        const sortingLayerValue = SortingManager.getSortingLayerIndex(this._sortingLayerID);
+        const sortingLayerValue = SortingManager.getLayerIndex(this._sortingLayer);
         const sortingPriority = SortingManager.getSortingPriority(sortingLayerValue, this._sortingOrder);
         if (this._models.length > 0) {
             for (let i = 0; i < this._models.length; i++) {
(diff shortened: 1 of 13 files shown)


cocos/core/renderer/scene/model.ts
```diff
@@ -218,6 +227,8 @@ export class Model {
     protected _enabled = true;
     protected _visFlags = Layers.Enum.NONE;
 
+    protected _sortingPriority: number = SortingManager.getSortingPriority();
```
SantyWang: getDefaultSortingPriority seems better

templates/launcher/application.ejs
```diff
@@ -137,6 +137,14 @@ function initializeGame (cc, settings, findCanvas) {
                 cc.Layers.addLayer(layer.name, layer.bit);
             });
         }
+        if (settings.sortingLayersUsed) {
+            cc.sortingManager.initLayerIDFromSettings(settings.sortingLayersUsed);
+        }
+        if (settings.sortingLayers) {
+            settings.sortingLayers.forEach((sortingLayer) => {
```
SantyWang, line 144: hack for now, I'm working on restructure the launch process，and after that `game.init` will be like this 
image
should adjust it after refactoring

cocos/core/scene-graph/sorting-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+            return false;
+        }
+    }
+
+    // Editor Function
+
+    public static addSortingLayer () {
+        this.ID++;
+        if (this.ID > 65535) {
+            console.error('too many layers');
```
SantyWang: use assert instead

SantyWang: use assert instead

cocos/core/scene-graph/sorting-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    public static initLayerFromSettings (layerID, layerName, layerIndex) {
+        this.nameMap.set(layerID, layerName);
+        this.indexMap.set(layerID, layerIndex);
+    }
+
+    public static initLayerIDFromSettings (ID) {
+        // Error check
+        if (ID > 65535) {
+            console.error('Invalid layer id.');
```
SantyWang: assert

SantyWang: assert

cocos/core/components/model-renderer.ts
```diff
@@ -24,11 +24,16 @@
...
+    ccclass, range, serializable,
 } from 'cc.decorator';
 import { scene } from '../renderer';
 import { Layers } from '../scene-graph/layers';
 import { Renderer } from './renderer';
+import { clamp } from '../math/utils';
+import { SortingManager } from '../scene-graph/sorting-manager';
+
+const MAX_UINT_NUM = (1 << 15) - 1;
+const MIN_UINT_NUM = -1 << 15;
```
pandamicro: UINT shouldn't be using -1, it's just not unsigned

cocos/core/components/model-renderer.ts
```diff
@@ -52,6 +62,44 @@ export class ModelRenderer extends Renderer {
         this._onVisibilityChange(val);
     }
 
+    /**
+     * @zh 组件所属排序层 id，影响组件的渲染排序。
+     * @en The sorting layer id of the component, which affects the rendering order of the component.
+     */
+    // Todo,how to show on inspector
+    get sortingLayerID () {
```
pandamicro: Use `sortingLayer` directly in model renderer, and use `sortingLayerIndex` in manager

pandamicro: Put public properties on top of all protected properties

cocos/core/scene-graph/sorting-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+        let index = 0;
+        if (this.indexMap.has(layerID)) {
+            index = this.indexMap.get(layerID)!;
+        } else {
+            console.error('invalid ID');
+        }
+        return index;
+    }
+
+    public static getSortingIndexFromName (name: string): number {
```
pandamicro: I think we can remove `sorting` from all API names in `SortingManager`, like `getLayerIndexByName` here (using by to be consistent with other APIs)

cocos/core/scene-graph/sorting-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+*/
+
+import { legacyCC } from '../global-exports';
+
+export class SortingManager {
+    public static getSortingPriority (layer = 0, order = 0): number {
+        return (((layer + (1 << 15)) << 16) | (order + (1 << 15))) >>> 0;
+    }
+
+    public static getSortingLayerIndex (layerID = 0): number {
```
pandamicro: simplify all `layerID` to `layer`, when we use `layer`, it's always the ID

cocos/core/scene-graph/sorting-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+        let key = 0;
+        for (let i = 0; i < count; i++) {
+            key = keyIterator.next().value;
+            if (this.nameMap.get(key) === name) return key;
+        }
+        console.warn('invalid name');
+        return 0;
+    }
+
+    public static idIsValid (id: number): boolean {
```
pandamicro: isLayerValid

cocos/core/scene-graph/sorting-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+        if (this.nameMap.has(layerID)) {
+            name = this.nameMap.get(layerID)!;
+        } else {
+            console.error('invalid ID');
+        }
+        return name;
+    }
+
+    // Name To ID
+    public static getSortingIDFromName (name: string): number {
```
pandamicro: getLayerByName

cocos/core/scene-graph/sorting-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+        } else {
+            console.error('invalid ID');
+            return false;
+        }
+    }
+
+    // Editor Function
+
+    public static addSortingLayer () {
+        this.ID++;
```
pandamicro: If we separate layer and layer index, why not give the ability to user to add a specific layer id

cocos/core/scene-graph/sorting-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+        if (this.ID > 65535) {
+            console.error('too many layers');
+        }
+        this.nameMap.set(this.ID, `New Layer${this.ID}`);
+        this.indexMap.set(this.ID, this.ID);
```
pandamicro: The logic here will lead to misunderstanding that layer and layer index is bound and equals

cocos/core/scene-graph/sorting-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+        // Todo: need update all component sorting priority in active scene
+    }
+
+    private static defaultIndex = 0;
+    private static defaultName = 'default';
+    private static nameMap = new Map<number, string>();
+    private static indexMap = new Map<number, number>();
+    private static ID = 0;
+
+    public static initLayerFromSettings (layerID, layerName, layerIndex) {
```
pandamicro: There is no settings in the input parameter, I think it's the same ability as `addLayer`, we can have 

```
public static addLayer(layer, layerName, layerIndex)
```

Do some assert and check in the function is enough, don't over complicate the usage here

cocos/core/pipeline/render-queue.ts
```diff
@@ -74,6 +75,7 @@ export class RenderQueue {
     constructor (desc: IRenderQueueDesc) {
         this._passDesc = desc;
         this._passPool = new RecyclePool<IRenderPass>(() => ({
+            sortingPriority: SortingManager.getSortingPriority(),
```
pandamicro: It shouldn't get sorting priority here, just use a default value and make sure it will updated

cocos/core/renderer/scene/model.ts
pandamicro: No need to initialize with SortingManager, just make sure it will be updated

pandamicro (changes requested):

cocos/core/components/model-renderer.ts
LinYunMo: should I change a name? It`s just specify the range of values

cocos/core/components/model-renderer.ts
LinYunMo: Modified

cocos/core/pipeline/render-queue.ts
LinYunMo: OK，it will updated when insertRenderPass

cocos/core/renderer/scene/model.ts
LinYunMo: It will be update when need

cocos/core/scene-graph/sorting-manager.ts
LinYunMo: Why custom need this ability？

cocos/core/scene-graph/sorting-manager.ts
LinYunMo: It will result in ID needing to be managed externally, rather than assigned internally.

cocos/core/scene-graph/sorting-manager.ts
LinYunMo: Modified

cocos/core/scene-graph/sorting-manager.ts
LinYunMo: Modified

cocos/core/scene-graph/sorting-manager.ts
LinYunMo: Modified

cocos/core/scene-graph/sorting-manager.ts
LinYunMo: Modified

templates/launcher/application.ejs
LinYunMo: I‘ll update when it change

cocos/core/components/model-renderer.ts
pandamicro: MAX_INT16
MIN_INT16
would be more accurate

cocos/core/components/model-renderer.ts
```diff
@@ -52,7 +54,43 @@ export class ModelRenderer extends Renderer {
...
+        this._sortingOrder = clamp(val, MIN_UINT_NUM, MAX_UINT_NUM);
+        this._updateSortingPriority();
+    }
+
+    /**
+     * @zh 通过 name 设置组件的 sortingLayer
+     * @en Set the sortingLayer of the component by name
+     * @param name
+     */
+    public setSortingLayerByName (name: string) {
```
pandamicro, line 100: Function should be placed with functions

cocos/core/scene-graph/sorting-manager.ts
pandamicro: If it's only editor function, we should give editor the ability to set any layer index with any layer name, the editor is responsible to ensure the layer index is compact.

We should also enclose all code in Editor functions inside `if (EDITOR)`

pandamicro (changes requested):

cocos/core/components/model-renderer.ts
LinYunMo: Modified

LinYunMo:
This PR is waiting for Editor support.

cocos/core/scene-graph/sorting-manager.ts
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+
+    private static defaultIndex = 0;
+    private static defaultName = 'default';
+    private static nameMap = new Map<number, string>();
+    private static indexMap = new Map<number, number>();
+    private static ID = 0;
+
+    public static initLayerFromSettings (layer, layerName, layerIndex) {
```
pandamicro, line 133: It's meaningless to mention settings here, the behavior is just to set a layer with the layer name and layer index, I suggest to modify to 

```
public static setLayer (layer, layerName?, layerIndex?)
```

And this function is enough, no need for addLayer

cocos/core/scene-graph/sorting-manager.ts
```diff
@@ -0,0 +1,147 @@
...
+        this.nameMap.set(layer, layerName);
+        this.indexMap.set(layer, layerIndex);
+    }
+
+    public static initLayerIDFromSettings (ID) {
```
pandamicro, line 138: This is none sense

cocos/core/scene-graph/sorting-manager.ts
```diff
@@ -0,0 +1,147 @@
...
+            // Todo: need update index map
+            // Todo: need update all component sorting priority in active scene
+        }
+    }
+
+    private static defaultIndex = 0;
+    private static defaultName = 'default';
+    private static nameMap = new Map<number, string>();
+    private static indexMap = new Map<number, number>();
+    private static ID = 0;
```
pandamicro, line 131: Put all properties declaration before functions

pandamicro (commented):
The whole layer management in SortingManager is not well designed, need to reconsider what we need to expose and how they are used by the editor

PR #11111, opened 2022-05-16
Add FixedDeltaTime/FixedUpdate mode

finscn:
新增 FixedDeltaTime ( FixedUpdate )模式.
具体用法见注释
实现原理见源码

### DIFF
--- a/cocos2d/core/CCDirector.js
+++ b/cocos2d/core/CCDirector.js
@@ -1117,6 +1117,7 @@ cc.Director.EVENT_AFTER_PHYSICS = 'director_after_physics';
 /**
  * !#en Enable FixedDeltaTime mode.
  * !#zh 开启 FixedDeltaTime 模式.
+ * @method enableFixedDeltaTime
  * @param {number} options - the value of fixedDeltaTime
  * @example
  *   cc.director.enableFixedDeltaTime()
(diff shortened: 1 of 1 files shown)


pandamicro:
Hi, @finscn First, I agree with you for the need of this PR.
But I'm afraid the implementation of this PR doesn't fit the engine architecture, it's quite a hack way to add this feature. The main problem is that this shouldn't be Director's responsibility, there should be a global timer which provide this feature.

So we are adding [Pacer in game](https://github.com/cocos/cocos-engine/blob/v3.6.0/cocos/core/game.ts#L891-L897) in v3.6, it will be its responsibility to update the game with fixed delta time, it's also its responsibility for global time scale.

idchlife:
> Hi, @finscn First, I agree with you for the need of this PR. But I'm afraid the implementation of this PR doesn't fit the engine architecture, it's quite a hack way to add this feature. The main problem is that this shouldn't be Director's responsibility, there should be a global timer which provide this feature.
> 
> So we are adding [Pacer in game](https://github.com/cocos/cocos-engine/blob/v3.6.0/cocos/core/game.ts#L891-L897) in v3.6, it will be its responsibility to update the game with fixed delta time, it's also its responsibility for global time scale.

Is there working link to "pacer in game"?

Wondering what this is and if actual fixedUpdate can be added into cocos engine.

Maybe this mysterious pacer can help with that?

PR #11114, opened 2022-05-16
refactor startup/shutdown process of DebugRenderer

minggo:

### Changelog

* as the title said

### DIFF
--- a/native/cocos/engine/Engine.h
+++ b/native/cocos/engine/Engine.h
@@ -41,6 +41,7 @@ class ScriptEngine;
 namespace cc {
 
 class FileUtils;
+class DebugRenderer;
 class Profiler;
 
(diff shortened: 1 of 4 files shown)


native/cocos/engine/Engine.h
```diff
@@ -139,6 +140,7 @@ class CC_DLL Engine : public BaseEngine {
     // Subsystems
     FileUtils *_fs{nullptr};
 	Profiler *_profiler{nullptr};
+    DebugRenderer *_debugRenderer{nullptr};
```
stanleyljl, line 145: Add #if CC_USE_PROFILER/#endif for _profiler declaration.

minggo: It is fixed.

stanleyljl (approved):

PR #11123, opened 2022-05-17
Subsystem startup shutdown debug renderer

minggo:

### Changelog
* remove DeviceManager::destroy(), it is only used in engine, so not problem
* create gfx::Device in cpp, make the life cycle more clear,

### DIFF
--- a/native/cocos/engine/Engine.cpp
+++ b/native/cocos/engine/Engine.cpp
@@ -29,11 +29,11 @@
 #include <sstream>
 #include "base/DeferredReleasePool.h"
 #include "base/Macros.h"
-#include "cocos/bindings/jswrapper/SeApi.h"
-#include "cocos/core/builtin/BuiltinResMgr.h"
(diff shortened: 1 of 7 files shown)


native/cocos/engine/Engine.cpp
```diff
@@ -97,6 +97,8 @@ namespace cc {
 Engine::Engine() {
     _scheduler = std::make_shared<Scheduler>();
     _fs = createFileUtils();
+    // May create gfx device in render subsystem in future.
+    _gfxDevice = gfx::DeviceManager::create();
```
star-e, line 101: If we create gfx device before script engine, user might lost device configuration ability in js. 
However, as in comment, we may create gfx device in render subsystem in the future.

star-e (approved):

native/cocos/engine/Engine.cpp
minggo: Yep, i think we need a render subsystem in future, and gfx device will be created there.

PR #11132, opened 2022-05-18
Update SkeletonBinary.cpp

xianyinchen:
### Changelog

* Fix mesh attachment decode crash, borrowing from js version implementation.


fish9167:
@cocos-robot run test cases

PR #11140, opened 2022-05-18
do not handle Root in engine

minggo:

### Changelog
* Root is created by TS, and ScriptEngine::cleanup() will handle it, so remove redundant codes.

### DIFF
--- a/native/cocos/core/Root.cpp
+++ b/native/cocos/core/Root.cpp
@@ -62,8 +62,9 @@ Root::Root(gfx::Device *device)
 }
 
 Root::~Root() {
-    instance = nullptr;
+    destroy();
     CC_SAFE_DELETE(_eventProcessor);
+    instance = nullptr;
 }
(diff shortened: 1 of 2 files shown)


native/cocos/core/Root.cpp
```diff
@@ -62,8 +62,9 @@ Root::Root(gfx::Device *device)
 }
 
 Root::~Root() {
-    instance = nullptr;
+    destroy();
     CC_SAFE_DELETE(_eventProcessor);
+    instance = nullptr;
```
dumganhar, line 67: Will `~Root` be invoked?

dumganhar (approved):

native/cocos/core/Root.cpp
minggo: Yep, it will be invoked when invoking `ScriptEngine::cleanup()`.

PR #11146, opened 2022-05-19
V3.6.0 ios input box

timlyeee:
Re: internal
* https://github.com/cocos/3d-tasks/issues/11051
* https://github.com/cocos/3d-tasks/issues/9517
* https://github.com/cocos/3d-tasks/issues/10605

### Changelog

* A better view for input box on IOS. Now we're using InputAccessoryView to draw the textView and textField instead, it's fully supported by Apple and we don't need to configure the position of it.
image
image
* The inputbox(textView and textField) will automatically hide once orientation changed.
* The inputbox(textView and textField) will automatically hide once the user click outside of the keyboard, on Android and IOS
* NULL detection for Android Editbox
* No public api or file structure is changed.
-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/bindings/event/EventDispatcher.cpp
+++ b/native/cocos/bindings/event/EventDispatcher.cpp
@@ -144,6 +144,7 @@ void EventDispatcher::dispatchTouchEvent(const TouchEvent &touchEvent) {
     se::ValueArray args;
     args.emplace_back(se::Value(jsTouchObjArray));
     EventDispatcher::doDispatchJsEvent(eventName, args);
+    EventDispatcher::dispatchCustomEvent(eventName,0);
 }
 
 void EventDispatcher::dispatchMouseEvent(const MouseEvent &mouseEvent) {
@@ -289,6 +290,7 @@ void EventDispatcher::dispatchOrientationChangeEvent(int orientation) {
         args.emplace_back(se::Value(jsOrientationEventObj));
         func.toObject()->call(args, nullptr);
     }
+    EventDispatcher::dispatchCustomEvent("orientation_changed", 0);
 }
 
 void EventDispatcher::dispatchEnterBackgroundEvent() {
--- a/native/cocos/bindings/event/EventDispatcher.cpp
+++ b/native/cocos/bindings/event/EventDispatcher.cpp
@@ -290,7 +290,6 @@ void EventDispatcher::dispatchOrientationChangeEvent(int orientation) {
         args.emplace_back(se::Value(jsOrientationEventObj));
         func.toObject()->call(args, nullptr);
     }
-    EventDispatcher::dispatchCustomEvent("orientation_changed", 0);
 }
 
 void EventDispatcher::dispatchEnterBackgroundEvent() {
--- a/native/cocos/platform/android/java/src/com/cocos/lib/CocosEditBoxActivity.java
+++ b/native/cocos/platform/android/java/src/com/cocos/lib/CocosEditBoxActivity.java
@@ -277,14 +277,21 @@ public class CocosEditBoxActivity extends Activity {
             extras = intent.getExtras();
         }
         if(extras == null){
-            extras = new Bundle();
-        }
-        show(extras.getString("defaultValue"),
+            show("",
+                10,
+                false,
+                false,
+                "done",
+                "text"
+                );
+        } else {
+            show(extras.getString("defaultValue"),
                 extras.getInt("maxLength"),
                 extras.getBoolean("isMultiline"),
                 extras.getBoolean("confirmHold"),
                 extras.getString("confirmType"),
                 extras.getString("inputType"));
+        }
     }
 
     /***************************************************************************************
--- a/native/cocos/platform/android/java/src/com/cocos/lib/CocosEditBoxActivity.java
+++ b/native/cocos/platform/android/java/src/com/cocos/lib/CocosEditBoxActivity.java
@@ -264,7 +264,14 @@ public class CocosEditBoxActivity extends Activity {
 
         this.addItems(frameLayout);
 
-        Bundle extras = getIntent().getExtras();
+        Intent intent = getIntent();
+        Bundle extras = null;
+        if (null != intent) {
+            extras = intent.getExtras();
+        }
+        if(extras == null){
+            extras = new Bundle();
+        }
         show(extras.getString("defaultValue"),
                 extras.getInt("maxLength"),
                 extras.getBoolean("isMultiline"),
--- a/native/cocos/platform/android/java/src/com/cocos/lib/CocosEditBoxActivity.java
+++ b/native/cocos/platform/android/java/src/com/cocos/lib/CocosEditBoxActivity.java
@@ -260,6 +260,13 @@ public class CocosEditBoxActivity extends Activity {
                         ViewGroup.LayoutParams.MATCH_PARENT);
         RelativeLayout frameLayout = new RelativeLayout(this);
         frameLayout.setLayoutParams(frameLayoutParams);
+        frameLayout.setClickable(true);
+        frameLayout.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View view) {
+                CocosEditBoxActivity.this.hide();
+            }
+        });
         setContentView(frameLayout);
 
         this.addItems(frameLayout);
--- a/native/cocos/ui/edit-box/EditBox-ios.mm
+++ b/native/cocos/ui/edit-box/EditBox-ios.mm
@@ -53,7 +53,7 @@
 #include "EditBox.h"
 #include "cocos/bindings/jswrapper/SeApi.h"
 #include "cocos/bindings/manual/jsb_global.h"
-
+#include "cocos/bindings/event/EventDispatcher.h"
 
 #import <UIKit/UIKit.h>
 
@@ -71,6 +71,7 @@ const bool INPUTBOX_HIDDEN = true; // Toggle if Inputbox is visible
 @interface Editbox_impl : NSObject
 + (instancetype)sharedInstance;
 - (void)        show:(const cc::EditBox::ShowInfo*)showInfo;
+- (void)        hide;
 - (UIView*)     getCurrentViewInUse;
 - (NSString*)   getCurrentText;
 @end
@@ -197,7 +198,9 @@ NSString *getHash(const cc::EditBox::ShowInfo* showInfo) {
     //TODO: get hash with different type of showInfo, for example different inputAccessoryView
     return showInfo->isMultiline?@"textView":@"textField";
 }
-
+void onParentViewTouched(const cc::CustomEvent &touchEvent){
+    [[Editbox_impl sharedInstance] hide];
+}
 } // namespace
 
 /*************************************************************************
@@ -287,6 +290,8 @@ static Editbox_impl *instance = nil;
     dispatch_once(&pred, ^{
         instance = [[super allocWithZone:NULL] init];
     });
+    
+
     return instance;
 }
 + (id)allocWithZone:(struct _NSZone*)zone {
@@ -395,8 +400,6 @@ static Editbox_impl *instance = nil;
     TextFieldDelegate* delegate = [[TextFieldDelegate alloc] initWithPairs:[inputbox inputOnView] and:textField];
     textField.delegate = delegate;
     [textField addTarget:delegate action:@selector(textFieldDidChange:) forControlEvents:UIControlEventEditingChanged];
-    setTextFieldReturnType(textField, showInfo->confirmType);
-    setTextFieldKeyboardType(textField, showInfo->inputType);
     UIBarButtonItem *textFieldItem = [[UIBarButtonItem alloc] initWithCustomView:textField];
     
     if (!btnHandler){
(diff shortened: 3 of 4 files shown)


native/cocos/ui/edit-box/EditBox-ios.mm
```diff
@@ -126,7 +115,7 @@ void getTextInputCallback() {
     }
 }
 
-void callJSFunc(const ccstd::string &type, const ccstd::string &text) {
+void callJSFunc(const std::string &type, const std::string &text) {
```
minggo: Please use ccstd::string instead.

minggo:
And please add some information about this PR, such as
- what's changed
- the new effect

native/cocos/ui/edit-box/EditBox-ios.mm
timlyeee: 1

native/cocos/bindings/event/EventDispatcher.cpp
```diff
@@ -289,6 +290,7 @@ void EventDispatcher::dispatchOrientationChangeEvent(int orientation) {
         args.emplace_back(se::Value(jsOrientationEventObj));
         func.toObject()->call(args, nullptr);
     }
+    EventDispatcher::dispatchCustomEvent("orientation_changed", 0);
```
minggo: Please use macro for event name as other events.

timlyeee:
## What's changed?

1. **InputAccessoryView mechanism**: We used to draw TextView and TextField by calculating pixels and offsets with the height and length of the keyboard, which InputAccessoryView can do. So in this PR, InputAccessoryView is used to replace the layout of TextView+UIButton. and optimize the effect by using borderRadius and so on.

2. **EditboxManager**:Editbox_impl is renamed as EditboxManager as singleton mode, to reduce the global variables usages. Some variables are still left but could be optimized in the future, such as g_isMultiline and g_maxLength to represent the max length or character that can be typed in, and whether to show a multiline edit box.

3. **Inputbox Pair**: Get hash is a simple method that can be expended, the developer might be possible to add their own items on the InputAccessoryView, such as "open iphone camera", so for each customized InputBox, we can make a hash and save in the dictionnary. For now, it return the hash by checking multiline only.

native/cocos/ui/edit-box/EditBox-ios.mm
```diff
@@ -259,86 +191,55 @@ CGRect getSafeAreaRect() {
...
-    else
-        initTextField(rect, showInfo);
-
-    UIView *textInput = getCurrentView();
-    UIView *view = UIApplication.sharedApplication.delegate.window.rootViewController.view;
-    [view addSubview:textInput];
-    [textInput becomeFirstResponder];
+//TODO: Get hash with different type of showInfo, for example different inputAccessoryView
+NSString *getHash(const cc::EditBox::ShowInfo* showInfo) {
+    return showInfo->isMultiline?@"textView":@"textField";
```
dumganhar, line 196: Currently,  getHash is not a real hash computation function. It's better to rename it to some thing like `getTextInputType`.

native/cocos/ui/edit-box/EditBox-ios.mm
```diff
@@ -259,86 +191,55 @@ CGRect getSafeAreaRect() {
...
-    textView.frame = CGRectMake(screenRect.origin.x,
-                                screenRect.size.height - textHeight - kbSize.height,
-                                screenRect.size.width,
-                                textHeight);
+@implementation TextViewDelegate {
+    UITextView* tViewOnView;
+    UITextView* tViewOnToolbar;
 }
+- (id) initWithPairs:(UITextView*) inputOnView and:(UITextView*) inputOnToolbar {
+    self = [super init];
```
dumganhar: it's better to check `self` like:
```oc
if (self = [super init]) {
    tViewOnView = inputOnView;
    tViewOnToolbar = inputOnToolbar;
}
return self;
```

native/cocos/ui/edit-box/EditBox-ios.mm
```diff
@@ -259,86 +191,55 @@ CGRect getSafeAreaRect() {
...
 @end
+@implementation TextFieldDelegate {
+    UITextField* textFieldOnView;
+    UITextField* textFieldOntoolbar;
+}
 
-@implementation TextFieldDelegate
+- (id) initWithPairs:(UITextField*) inputOnView and:(UITextField*) inputOnToolbar {
+    self = [super init];
+    textFieldOnView = inputOnView;
```
dumganhar: Some here. Need check self.

native/cocos/ui/edit-box/EditBox-ios.mm
```diff
@@ -363,76 +265,315 @@ - (BOOL)textFieldShouldReturn:(UITextField *)textField {
...
+    EditBox::_isShown = true;
 }
 
 bool EditBox::complete() {
-    if (!_isShown)
-        return false;
-
-    NSString *text = getCurrentText();
+    if(!EditBox::_isShown)
+        return;
```
dumganhar: return void ?  Should return false instead.

native/cocos/ui/edit-box/EditBox-ios.mm
```diff
@@ -363,76 +265,315 @@ - (BOOL)textFieldShouldReturn:(UITextField *)textField {
...
+
+- (id)copyWithZone:(struct _NSZone*)zone {
+    return [EditboxManager sharedInstance];
+}
+
+- (void)onOrientationChanged{
+    cc::EditBox::complete();
+}
+- (id)init {
+    self          = [super init];
```
dumganhar: Need to check `self`.

native/cocos/ui/edit-box/EditBox-ios.mm
timlyeee: well, it's actually better to make function void because the behavior does not need to check the stats.

But right there I will return true here.

native/cocos/ui/edit-box/EditBox-ios.mm
timlyeee: Sure.

native/cocos/ui/edit-box/EditBox-ios.mm
timlyeee: 1

native/cocos/ui/edit-box/EditBox-ios.mm
timlyeee: 1

native/cocos/ui/edit-box/EditBox-ios.mm
timlyeee: 1

native/cocos/bindings/event/EventDispatcher.cpp
timlyeee: Code here is no more needed, I will delete

native/cocos/ui/edit-box/EditBox-ios.mm
```diff
@@ -363,76 +267,315 @@ - (BOOL)textFieldShouldReturn:(UITextField *)textField {
...
+    [confirmBtn addTarget:btnHandler
+                   action:@selector(buttonTapped:)
+         forControlEvents:UIControlEventTouchUpInside];
+    [confirmBtn setTitle:[NSString stringWithUTF8String:showInfo->confirmType.c_str()]
+             forState:UIControlStateNormal];
+    [confirmBtn setTitleColor:[UIColor systemBlueColor]
+                  forState:UIControlStateNormal];
+    [confirmBtn setTitleColor:[UIColor darkTextColor]
+                  forState:UIControlStateHighlighted]; // Hight light state triggered when the button is tapped.
+    UIBarButtonItem *confirm = [[UIBarButtonItem alloc]initWithCustomView:confirmBtn];
```
minggo, line 457: Is `confirmBtn` a strong reference or not? If it is a strong reference, then should release it here, or will cause memory leak. Should check other places too.

timlyeee: Add release logic when leave the blocks.

minggo (approved):

native/cocos/platform/android/java/src/com/cocos/lib/CocosEditBoxActivity.java
```diff
@@ -260,11 +260,25 @@ protected void onCreate(Bundle savedInstanceState) {
...
         this.addItems(frameLayout);
 
-        Bundle extras = getIntent().getExtras();
+        Intent intent = getIntent();
+        Bundle extras = null;
+        if (null != intent) {
+            extras = intent.getExtras();
+        }
+        if(extras == null){
+            extras = new Bundle();
```
dumganhar: Empty Bundle will return null for extras.getString("defaultValue"). Could `show` handle null ?

native/cocos/ui/edit-box/EditBox-ios.mm
```diff
@@ -363,76 +267,331 @@ - (BOOL)textFieldShouldReturn:(UITextField *)textField {
...
+                                                   safeView.size.width,
+                                                   TEXT_LINE_HEIGHT + ITEM_MARGIN_HEIGHT)];
+    [toolbar setBackgroundColor:[UIColor darkGrayColor]];
+    
+    UITextField* textField = [[UITextField alloc] init];
+    textField.borderStyle = UITextBorderStyleRoundedRect;
+    textField.textColor = [UIColor blackColor];
+    textField.backgroundColor = [UIColor whiteColor];
+    textField.text = [NSString stringWithUTF8String:showInfo->defaultValue.c_str()];
+    TextFieldDelegate* delegate = [[TextFieldDelegate alloc] initWithPairs:[inputbox inputOnView] and:textField];
```
dumganhar, line 434: where is `delegate` released ?

native/cocos/ui/edit-box/EditBox-ios.mm
```diff
@@ -363,76 +267,331 @@ - (BOOL)textFieldShouldReturn:(UITextField *)textField {
...
+        [self setInputWidthOf:[[ret inputOnView] inputAccessoryView] ];
+    } else {
+        ret = [[InputBoxPair alloc] init];
+        [ret setInputOnView:[[UITextView alloc]
+               initWithFrame:CGRectMake(showInfo->x,
+                                        viewRect.size.height - showInfo->y - showInfo->height,
+                                        showInfo->width,
+                                        showInfo->height)]];
+        [textInputDictionnary setValue:ret forKey:inputType];
+        [self addInputAccessoryViewForTextView:ret with:showInfo];
```
dumganhar, line 486: Should invoke `[ret release]` ?

native/cocos/ui/edit-box/EditBox-ios.mm
```diff
@@ -363,76 +267,331 @@ - (BOOL)textFieldShouldReturn:(UITextField *)textField {
...
+    [view addSubview:[curView inputOnView]];
+    [[curView inputOnView] becomeFirstResponder];
+    [[curView inputOnToolbar] becomeFirstResponder];
+}
+// Change the focus point to the TextField or TextView on the toolbar.
+- (void) hide {
+    [[curView inputOnView] becomeFirstResponder];
+    [[curView inputOnView] removeFromSuperview];
+    [[curView inputOnToolbar] resignFirstResponder];
+    [[curView inputOnView] resignFirstResponder];
```
dumganhar, line 564: Should release curView here?

native/cocos/platform/android/java/src/com/cocos/lib/CocosEditBoxActivity.java
timlyeee: No, Editbox::show can handle a default value, but not null.

native/cocos/ui/edit-box/EditBox-ios.mm
timlyeee: Once the game ended. It's the same behavior as the static member of global namespace as v3.5.0. As it might be recreated again and again, no need to release it in a certain point.

native/cocos/ui/edit-box/EditBox-ios.mm
timlyeee: No! It's week ref here, use it but not release it.

native/cocos/ui/edit-box/EditBox-ios.mm
timlyeee: > As it might be recreated again and again

So to prevent this behavior, just keep the week reference in InputBox.

But yes, the delegate should be release to decrease the retainCount once Inputbox is deleted

native/cocos/platform/android/java/src/com/cocos/lib/CocosEditBoxActivity.java
timlyeee: Added a default show when no information was given

native/cocos/ui/edit-box/EditBox-ios.mm
timlyeee: added a release operation for delegate

native/cocos/ui/edit-box/EditBox-ios.mm
```diff
@@ -211,8 +211,11 @@ - (id) initWithPairs:(UITextView*) inputOnView and:(UITextView*) inputOnToolbar
     if (self = [super init]) {
         tViewOnView = inputOnView;
         tViewOnToolbar = inputOnToolbar;
+        return self;
+    } else {
+        [self release];
```
dumganhar: [nil release] ?

timlyeee: removed

dumganhar (approved):

PR #11160, opened 2022-05-19
Animation graph transition destination start

shrinktofit:
Re: #11157 

Changelog:
 *


shrinktofit:
Edit: I feel "destination start" is more comprehensive than "offset" :)

cocos/core/animation/marionette/graph-eval.ts
```diff
@@ -1248,8 +1259,8 @@ export class MotionStateEval extends StateEval {
         return stateStatus;
     }
 
-    public resetToPort () {
-        this._toPort.progress = 0.0;
+    public resetToPort (at: number) {
+        this._toPort.progress = at;
```
arsen2010, line 1296: should take some actions when 'at' is  unlegal data

shrinktofit: The valid input is non-negative finite numbers for now. I think it's safe to use it without assertion.

arsen2010 (approved):

SantyWang (approved):

PR #11179, opened 2022-05-23
fixed #10929: Hash computation in native engine causes loss of precision, use 32 bits hash for all platforms

### LINKED ISSUE #10929: Hash computation in native engine causes loss of precision
## Background

Javascript VM (e.g. V8) uses `Double` to store number for representing both Integer & Double. There is a SMI (Small Integer) optimization that if an integer is less than 31 bits on 32  bits architecture device or less than 32 bits  on 64 bits architecture device. SMI is not what we need to discuss here. We're discussing the situation of how JS VM bind an integer that larger than 32 bits.

Javascript defines `Safe Integer` to remind developer that integer represented with Number has a range between `-2^53` to `2^53`, "Safe" means integers that can be represented one-by-one, that is, in the range (-2^53, 2^53), doubles and integers are one-to-one, and conversely, in Within this range, all integers have a unique floating-point representation, which is called safe integers. Beyond this range, there will be two or more integers whose double-precision representations are the same; on the other hand, beyond this range, some integers cannot be represented exactly, and can only be rounded to their nearest floating-point representations. Such cases are called unsafe integers. This is not a JS-specific property, but that is how IEEE754 double-precision floating-point numbers are defined.

### What happen if we bind a 64 bit integer to JS VM

In Cocos Creator native engine, we have many Hash computation logic to compute a hash value for some resources and export 64 bits integer (`size_t`) to JS. For instance, there is an interface called `Sampler::getHash` in gfx.

https://github.com/cocos/cocos-engine/blob/271cc5cf232e88238a8e5c35458c2e77bfd1bee5/native/cocos/renderer/gfx-base/states/GFXSampler.h#L42

```c++
class CC_DLL Sampler : public GFXObject {
public:
    explicit Sampler(const SamplerInfo &info);

    static size_t computeHash(const SamplerInfo &info);
    static SamplerInfo unpackFromHash(size_t hash);

    inline const SamplerInfo &getInfo() const { return _info; }
    inline const size_t &getHash() const { return _hash; } // Return size_t that is 64bits when targeting 64bits device.

protected:
    SamplerInfo _info;
    size_t _hash{0U};
};
```

https://github.com/cocos/cocos-engine/blob/271cc5cf232e88238a8e5c35458c2e77bfd1bee5/native/cocos/bindings/auto/jsb_gfx_auto.cpp

```c++
static bool js_gfx_Sampler_getHash(se::State& s) 
{
    auto* cobj = SE_THIS_OBJECT<cc::gfx::Sampler>(s);
    SE_PRECONDITION2(cobj, false, "js_gfx_Sampler_getHash : Invalid Native Object");
    const auto& args = s.args();
    size_t argc = args.size();
    CC_UNUSED bool ok = true;
    if (argc == 0) {
        const size_t& result = cobj->getHash();
// s.rval() will store int64_t with BigInt type, that's ok by now.
        ok &= nativevalue_to_se(result, s.rval(), nullptr /*ctx*/); 
        SE_PRECONDITION2(ok, false, "js_gfx_Sampler_getHash : Error processing arguments");
        SE_HOLD_RETURN_VALUE(result, s.thisObject(), s.rval());
        return true;
    }
    SE_REPORT_ERROR("wrong number of arguments: %d, was expecting %d", (int)argc, 0);
    return false;
}
SE_BIND_FUNC(js_gfx_Sampler_getHash)
```

```c++
template <typename T>
void setReturnValueTemplate(const Value &data, const T &argv) {
    if (data.getType() == Value::Type::Undefined) {
        argv.GetReturnValue().Set(v8::Undefined(argv.GetIsolate()));
    } else if (data.getType() == Value::Type::Null) {
        argv.GetReturnValue().Set(v8::Null(argv.GetIsolate()));
    } else if (data.getType() == Value::Type::Number) {
        argv.GetReturnValue().Set(v8::Number::New(argv.GetIsolate(), data.toDouble()));
    } else if (data.getType() == Value::Type::BigInt) {
        // We cast the int64 to a double value and construct a v8::Number here. Precision will lost here.
        // If a 64 bit hash value is beyond the range of `-2^53 <-> 2^53`, JS will get the wrong hash value and hash conflict may happen.
        argv.GetReturnValue().Set(v8::Number::New(argv.GetIsolate(), static_cast<double>(data.toInt64())));
    }
    ......
}

void setReturnValue(const Value &data, const v8::FunctionCallbackInfo<v8::Value> &argv) {
    setReturnValueTemplate(data, argv);
}

void setReturnValue(const Value &data, const v8::PropertyCallbackInfo<v8::Value> &argv) {
    setReturnValueTemplate(data, argv);
}

#define SE_BIND_FUNC(funcName)                                                                                                              \
    void funcName##Registry(const v8::FunctionCallbackInfo<v8::Value> &_v8args) {                                                           \
        JsbInvokeScope(#funcName);                                                                                                          \
        bool                   ret      = false;                                                                                            \
        v8::Isolate *          _isolate = _v8args.GetIsolate();                                                                             \
        v8::HandleScope        _hs(_isolate);                                                                                               \
        se::ValueArray &       args = se::gValueArrayPool.get(_v8args.Length());                                                            \
        se::CallbackDepthGuard depthGuard{args, se::gValueArrayPool._depth};                                                                \
        se::internal::jsToSeArgs(_v8args, args);                                                                                            \
        se::PrivateObjectBase *privateObject = static_cast<se::PrivateObjectBase *>(se::internal::getPrivate(_isolate, _v8args.This(), 0)); \
        se::Object *           thisObject    = reinterpret_cast<se::Object *>(se::internal::getPrivate(_isolate, _v8args.This(), 1));       \
        se::State              state(thisObject, privateObject, args);                                                                      \
        ret = funcName(state);                                                                                                              \
        if (!ret) {                                                                                                                         \
            SE_LOGE("[ERROR] Failed to invoke %s, location: %s:%d\n", #funcName, __FILE__, __LINE__);                                       \
        }                                                                                                                                   \
        // We convert state.rval() which is represented with int64_t to v8::Number
        se::internal::setReturnValue(state.rval(), _v8args);                                                                                \
    }
```

https://github.com/cocos/cocos-engine/blob/271cc5cf232e88238a8e5c35458c2e77bfd1bee5/cocos/2d/renderer/batcher-2d.ts#L357

```ts
export class Batcher2D implements IBatcher {
......
    public commitComp (comp: Renderable2D, renderData: BaseRenderData|null, frame: TextureBase|SpriteFrame|null, assembler, transform: Node|null) {
        ......

        if (this._currHash !== dataHash || dataHash === 0 || this._currMaterial !== mat
            || this._currDepthStencilStateStage !== depthStencilStateStage) {
            ......
            if (frame) {
                this._currTexture = frame.getGFXTexture();
                this._currSampler = frame.getGFXSampler();
                this._currTextureHash = frame.getHash();
                this._currSamplerHash = this._currSampler.hash; // Call JSB method: js_gfx_Sampler_getHash
            } else {
                ......
            }
        }

        assembler.fillBuffers(comp, this);
    }
......
}
```

### Possible Solutions

- Solution (1)

Cocos Creator Web Engine uses 32 bits hash  computation by the third-party library `murmurhash2_32_gc` now. If native engine keeps the same strategy of using 32 bits hash (Use `uint32_t` for hash value) , hash value will not be clipped in the static_cast<double> conversion. Because double could represent the whole range of 32bits integer, we could make sure the return value is safe (not clipped) for JS.

But @star-e suggest that 32 bit hash value will cause Hash conflict easily in games.

- Solution (2)

Both Cocos Creator Web engine & native engine use BigInt for representing hash value.  In the hash implementation on web platform, `murmurhash2_32_gc` needs to be extend to something like `murmurhash2_64_gc` and the result value of this function needs to be changed to BigInt too.

And if bit operation is needed, the left and the right side of bit operation should both be a BigInt value, otherwise JS VM will trigger an error. For example:

Screen Shot 2022-05-05 at 11 15 28

### My Advice

We need to keep the same logic for both Web & Native platform. Solution (1) and (2) are both ok for me. But I'm not familiar with Javascript language, perhaps use 64 bit hash computation will make the performance worse on web platform. 

Suggestions are appreciated.

dumganhar:
Re: #10929

### Changelog

* fixed #10929: Hash computation in native engine causes loss of precision, use 32 bits hash for all platforms

-------

### Continuous Integration

This pull request:

* [X] needs automatic test cases check // Manual trigger with `@cocos-robot run test cases` afterward

-------

### Compatibility Check

This pull request:


native/cocos/base/std/hash/hash_fwd.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include <boost/config/workaround.hpp>
+#include <cstddef>
+
+#if defined(BOOST_HAS_PRAGMA_ONCE)
+#pragma once
+#endif
+
+namespace ccstd
+{
+    using hash_t = std::uint32_t;
```
dumganhar, line 22: Define hash_t as std::uint32_t.

dumganhar: If we need use 64bits someday, we could just modify this line to
```c++
using hash_t = std::uint64_t;
```

native/cocos/core/assets/Font.cpp
```diff
@@ -25,24 +25,24 @@
...
 #include "profiler/Profiler.h"
 
 namespace cc {
 
-std::size_t KerningHash::operator()(const KerningPair &k) const {
-    size_t seed = 2;
-    boost::hash_combine(seed, k.prevCode);
-    boost::hash_combine(seed, k.nextCode);
-    return seed;
+size_t KerningHash::operator()(const KerningPair &k) const {
```
star-e: Should return value be hash_t？

native/cocos/core/assets/Font.h
```diff
@@ -71,7 +71,7 @@ struct KerningPair {
 };
 
 struct KerningHash {
-    std::size_t operator()(const KerningPair &k) const;
+    size_t operator()(const KerningPair &k) const;
```
star-e: Should return value be hash_t？

native/cocos/scene/Pass.cpp
```diff
@@ -116,8 +116,8 @@ uint64_t Pass::getPassHash(Pass *pass) {
     res << serializeRasterizerState(pass->_rs);
 
     ccstd::string str{res.str()};
-    std::size_t seed = 666;
-    boost::hash_range(seed, str.begin(), str.end());
+    ccstd::hash_t seed = 666;
+    ccstd::hash_range(seed, str.begin(), str.end());
     return static_cast<uint64_t>(seed);
```
dumganhar: Don't need cast now. I will remove the cast.

native/cocos/math/MathBase.h
```diff
@@ -72,11 +73,14 @@
 NS_CC_MATH_BEGIN
 
 template <typename T, typename Enable = std::enable_if_t<std::is_class<T>::value>>
-struct Hasher final { size_t operator()(const T &info) const; };
+struct Hasher final {
+    size_t operator()(const T &info) const;
```
star-e: Should return value be hash_t？

native/cocos/renderer/gfx-base/GFXDef.cpp
```diff
@@ -35,52 +35,72 @@ namespace gfx {
...
+    ccstd::hash_combine(seed, info.preserves);
+    ccstd::hash_combine(seed, info.depthStencil);
+    ccstd::hash_combine(seed, info.depthStencilResolve);
+    ccstd::hash_combine(seed, info.depthResolveMode);
+    ccstd::hash_combine(seed, info.stencilResolveMode);
     return seed;
 }
 
+template <>
+size_t Hasher<SubpassInfo>::operator()(const SubpassInfo &info) const {
```
star-e, line 100: Should return value be hash_t？

native/cocos/renderer/gfx-base/GFXDef.cpp
```diff
@@ -110,25 +135,30 @@ bool operator==(const RenderPassInfo &lhs, const RenderPassInfo &rhs) {
...
-        boost::hash_combine(seed, colorTexture->getHash());
+        ccstd::hash_combine(seed, colorTexture);
+        ccstd::hash_combine(seed, colorTexture->getRaw());
+        ccstd::hash_combine(seed, colorTexture->getHash());
     }
     return seed;
 }
 
+template <>
+size_t Hasher<FramebufferInfo>::operator()(const FramebufferInfo &info) const {
```
star-e, line 158: Should return value be hash_t？

native/cocos/renderer/gfx-base/GFXDef.h
```diff
@@ -27,16 +27,20 @@
...
 #include <functional>
 #include "GFXDef-common.h"
+#include "base/std/hash/hash.h"
 
 namespace cc {
 namespace gfx {
 
 template <typename T, typename Enable = std::enable_if_t<std::is_class<T>::value>>
-struct Hasher final { size_t operator()(const T &info) const; };
+struct Hasher final {
```
star-e, line 36: Should return value be hash_t？

native/cocos/math/MathBase.h
dumganhar: I think it's probably not, the operator() is used for interacting with `stl containers`.

```c++
ccstd::unordered_map<Vec4, gfx::InputAssembler *, Hasher<Vec4>> _quadIA;
```

So I defined two functions, one is functor for interacting with `stl containers` which returns size_t.
Another is a static function for used in our internal code.
```c++
template <typename T, typename Enable = std::enable_if_t<std::is_class<T>::value>>
struct Hasher final {
    size_t operator()(const T &info) const;
    static ccstd::hash_t hashValue(const T &info);
};
```

Using static `hashValue` function could also avoid constructing a temporary `Hasher` object.

native/cocos/renderer/gfx-base/GFXDef.cpp
```diff
@@ -35,52 +35,72 @@ namespace gfx {
...
+    return static_cast<size_t>(hashValue(info));
+}
+
 bool operator==(const ColorAttachment &lhs, const ColorAttachment &rhs) {
     return !memcmp(&lhs, &rhs, sizeof(ColorAttachment));
 }
 
 template <>
-size_t Hasher<DepthStencilAttachment>::operator()(const DepthStencilAttachment &info) const {
+ccstd::hash_t Hasher<DepthStencilAttachment>::hashValue(const DepthStencilAttachment &info) {
```
minggo: Why need to add this function?

dumganhar: The reason is the same as what I have replied.
> I think it's probably not, the operator() is used for interacting with `stl containers`.
> 
> ```c++
> ccstd::unordered_map<Vec4, gfx::InputAssembler *, Hasher<Vec4>> _quadIA;
> ```
>
> So I defined two functions, one is functor for interacting with `stl containers` which returns size_t.
> Another is a static function for used in our internal code.
> ```c++
> template <typename T, typename Enable = std::enable_if_t<std::is_class<T>::value>>
> struct Hasher final {
>     size_t operator()(const T &info) const;
>     static ccstd::hash_t hashValue(const T &info);
> };
> ```
> 
> Using static `hashValue` function could also avoid constructing a temporary `Hasher` object.


hashValue returns ccstd::hash_t and is used in our engine code.
`size_t operator()` is used for custom hash of std containers since std containers use size_t for hash. If using ccstd::hash_t, it may cause warnings.

dumganhar: Furthermore, keeping `size_t operator()` will make it easier to convert 64 bits hash to 32 bits.
Suppose we wanna use 64 bits hash on 32 bits devices some day. Then, std containers will still use `size_t` as hash value which is 32 bits on 32 bits architecture. We could do the conversion in operator() method like this:

```c++
template <>
ccstd::hash_t Hasher<SubpassDependency>::hashValue(const SubpassDependency &info) {
    return quickHashTrivialStruct(&info);
}

template <>
size_t Hasher<SubpassDependency>::operator()(const SubpassDependency &info) const {
    ccstd::hash_t value = hashValue(info); // ccstd::hash_t  is 64bits, size_t is 32bits
    return static_cast<size_t>((value >> 32) ^ (value & 0xFFFFFFFF) ); // Convert 64 bits hash to 32 bits
}
```

minggo: I think just return hash_t for operator makes things simple. hast_t is a special type used in engine, then the corresponding Hasher is a special type too. 

All codes use the same result is better. If using two functions, then some codes will use `operator()`, some use `hashValue`, which makes things deprecated. We let JS and C++ use the same type, then use hash_t, we should also use hash_t for all codes too.

dumganhar: It's the best situation we hope. But remember that, stl container ( std::unordered_map ) is external code that will use `operator()`, it's out of our control. 

If we use hash_t for the return value of `operator()`, 
- 1. it will generate compiler warning of type conversion
- 2. If we use 64 bits hash someday, hash value may be cut from 64 bits to 32 bits in std container code without notifying us, it will be bug prone.

dumganhar:
@cocos-robot run test cases

native/cocos/renderer/gfx-base/GFXDef.cpp
minggo: > it will generate compiler warning of type conversion

I think it is ok, because it doesn't loose any information. Does it block the CI?

> If we use 64 bits hash someday, hash value may be cut from 64 bits to 32 bits in std container code without notifying us, it will be bug prone.

We can add some comments. And in current implementation, we also need to remove the `hashValue` function too.

fish9167:
@cocos-robot run test cases

star-e (approved):

native/cocos/renderer/gfx-base/GFXDef.cpp
dumganhar: DONE in the latest commit

dumganhar: DONE

minggo (approved):

PR #11194, opened 2022-05-24
add nx webp lib

oahcy:


native/cocos/platform/Image.cpp
```diff
@@ -917,8 +917,8 @@ bool Image::initWithWebpData(const unsigned char *data, ssize_t dataLen) {
 
         ret = true;
     } while (false);
-    return ret;
 #endif // CC_USE_WEBP
+    return ret;
```
minggo, line 921: So still doesn't support webp?

oahcy: support， 
single step debugging have strangely behave before support。

minggo (approved):

PR #11200, opened 2022-05-24
Cancel sharing index buffer for per block

MSoft1115:
Re: #

### Changelog
fix https://github.com/cocos/3d-tasks/issues/12281
* 

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/terrain/terrain.ts
+++ b/cocos/terrain/terrain.ts
@@ -397,7 +397,7 @@ export class TerrainBlock {
         ];
 
         this._renderable._meshData = new RenderingSubMesh([vertexBuffer], gfxAttributes,
-            PrimitiveMode.TRIANGLE_LIST, this._terrain._getSharedIndexBuffer());
+            PrimitiveMode.TRIANGLE_LIST, this._terrain._createSharedIndexBuffer());
         this._renderable._model = (legacyCC.director.root as Root).createModel(scene.Model);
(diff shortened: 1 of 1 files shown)


hana-alice (approved):

SantyWang (approved):

cocos/terrain/terrain.ts
```diff
@@ -397,7 +397,7 @@ export class TerrainBlock {
         ];
 
         this._renderable._meshData = new RenderingSubMesh([vertexBuffer], gfxAttributes,
-            PrimitiveMode.TRIANGLE_LIST, this._terrain._getSharedIndexBuffer());
+            PrimitiveMode.TRIANGLE_LIST, this._terrain._createSharedIndexBuffer());
```
star-e, line 400: use newly created index buffer might increase memory usage.
I think we need to modify rendering sub mesh and add a flag to rendering submesh.
if the rendering submesh owning the index buffer, then it needs to destroy the buffer.
if the rendering submesh not owning the index buffer, then it should not destroy the buffer

star-e: when creating the RenderingSubMesh, we might add a new parameter indicating whether the RenderingSubMesh owning the index buffer.
This is a temporary solution.

PR #11244, opened 2022-05-26
use Plane object instead

minggo:


native/cocos/core/geometry/AABB.cpp
```diff
@@ -127,7 +127,7 @@ bool AABB::aabbFrustum(const Frustum &frustum) const {
     return std::all_of(planes.begin(),
                        planes.end(),
                        // frustum plane normal points to the inside
-                       [self](const Plane *plane) { return self->aabbPlane(*plane) != -1; });
+                       [self](const Plane plane) { return self->aabbPlane(plane) != -1; });
```
PatriceJiang, line 130: Use reference type is better

troublemaker52025 (approved):

native/cocos/core/geometry/Sphere.cpp
```diff
@@ -203,7 +203,7 @@ bool Sphere::sphereFrustum(const Frustum &frustum) const {
     return std::all_of(planes.begin(),
                        planes.end(),
                        // frustum plane normal points to the inside
-                       [self](const Plane *plane) { return self->interset(*plane) != -1; });
+                       [self](const Plane plane) { return self->interset(plane) != -1; });
```
PatriceJiang, line 206: Reference type?

minggo:
Need to use pointer to share with JS, so close it.

Issue #11297, opened 2022-05-31
[3.6.0] Loading problem of ttf font


xianyinchen:
### Cocos Creator version

3.4.2,3.5

### System information

Android 12

### Issue description

**different rendering result in editor and runtime.**

editor picture:
![企业微信截图_16539688169869](https://user-images.githubusercontent.com/8321365/171088863-531ac60c-80b5-42f5-82c1-dd3f5c4d549e.png)

runtime picture:
![Screenshot_20220531-114501_test](https://user-images.githubusercontent.com/8321365/171088717-7222fac5-9c92-483d-98a8-fcd4d69de752.png)

In andorid platform, I find  that ttf font with splice string "-Bold" or  "-Italic" can not found in sTypefaceCache.
![1653968904(1)](https://user-images.githubusercontent.com/8321365/171088990-cf024167-58d3-4c03-b54a-c7697045e7fb.png)

### Steps to reproduce
try fix.
![Screenshot_20220531-113714_test](https://user-images.githubusercontent.com/8321365/171087987-dcb351b6-4b16-4a4b-b7c5-a28b0e7b90c9.png)

test code:
[test (2).zip](https://github.com/cocos/cocos-engine/files/8801880/test.2.zip)

Try to fix.
https://github.com/cocos/cocos-engine/pull/11298

PR #11298, opened 2022-05-31
fix ttf font render on android

xianyinchen:
Re: https://github.com/cocos/cocos-engine/issues/11297

### Changelog

*  fix ttf font  load issue.


bofeng-song (approved):

zhefengzhang:
The same engine code is available on 2.x versions. So the same problem occurs. @xianyinchen We need to sync the changes to the 2.x version.

minggo (approved):

Issue #11124, opened 2022-05-17
Does it need to delete gfx device when doing restart javascript vm?


minggo:
Now `Engine::restartVM()` will destroy and delete gfx device. Is it needed? If not delete the gfx device, will it cause memory leak? Especially GPU memory leak.

dumganhar:
When script engine restarts, all GFX binding objects must get collected by GC, the relevant c++ GFX objects must be destroyed. In this case, whether to delete GFXDevice will make no much differences, it is just the case of whether to reuse GFXDevice instance.

> If not delete the gfx device, will it cause memory leak?

So if there are memory leaks, they need to be fixed.

star-e:
If some part of native engine is not destroyed, and these parts hold some gfx objects. Then GFXDevice shall not be destroyed.

minggo:
Can not make sure if any object holds gfx object until invoke `DeviceValidator::doDestroy()`. So may be move the logic to another function to check GPU memory leak?

minggo:
Can not make sure if any object holds gfx object until invoke `DeviceValidator::doDestroy()`. So may be move the logic to another function to check GPU memory leak?

star-e:
> Can not make sure if any object holds gfx object until invoke DeviceValidator::doDestroy(). So may be move the logic to another function to check GPU memory leak?

Yes, gfx need to provide these statistics.

minggo:
@holycanvas does TS engine destroy all GPU resources before restarting?

SantyWang:
> @holycanvas does TS engine destroy all GPU resources before restarting?

For now, it's not. In the long term, I think it depends on how we define ***restarting***，If we want to achieve the cleanest reboot, then I think we need to go through the destruction of all the modules and then go through the boot process again, not just restart the virtual machine

minggo:
> > @holycanvas does TS engine destroy all GPU resources before restarting?
> 
> For now, it's not. In the long term, I think it depends on how we define _**restarting**_，If we want to achieve the cleanest reboot, then I think we need to go through the destruction of all the modules and then go through the boot process again, not just restart the virtual machine

Then what's the difference between restart VM and re-launch application?

SantyWang:
> Then what's the difference between restart VM and re-launch application?

Oh，I'm talking about restarting, not just restart VM. For restart vm, I don't think it's necessary to clean up the gfx device

minggo:
[...]

PR #11301, opened 2022-05-31
delete/recreate all resources when restart app

minggo:
https://github.com/cocos/cocos-engine/issues/11124

### DIFF
--- a/native/cocos/application/BaseGame.cpp
+++ b/native/cocos/application/BaseGame.cpp
@@ -50,14 +50,14 @@ int BaseGame::init() {
     });
 
 #endif
-
-    if (_debuggerInfo.enabled) {
-        setDebugIpAndPort(_debuggerInfo.address, _debuggerInfo.port, _debuggerInfo.pauseOnStart);
-    }
-
+    
     int ret = cc::CocosApplication::init();
     if (ret != 0) {
         return ret;
+    }
+
+    if (_debuggerInfo.enabled) {
+        setDebugIpAndPort(_debuggerInfo.address, _debuggerInfo.port, _debuggerInfo.pauseOnStart);
     }
 
     setXXTeaKey(_xxteaKey);
--- a/native/cocos/engine/Engine.cpp
+++ b/native/cocos/engine/Engine.cpp
@@ -249,7 +249,7 @@ void Engine::tick() {
         CC_PROFILE(EngineTick);
 
         if (_needRestart) {
-            restartVM();
+            doRestart();
             _needRestart = false;
         }
 
--- a/native/cocos/engine/Engine.cpp
+++ b/native/cocos/engine/Engine.cpp
@@ -92,8 +92,7 @@ bool setCanvasCallback(se::Object * /*global*/) {
 
 namespace cc {
 
-Engine::Engine() {
-}
+Engine::Engine() = default;
 
 Engine::~Engine() {
     destroy();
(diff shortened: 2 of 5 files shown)


native/cocos/renderer/GFXDeviceManager.h
```diff
@@ -139,6 +128,17 @@ class CC_DLL DeviceManager final {
     }
 
 private:
+    static void addSurfaceEventListener() {
```
dumganhar, line 131: So `addSurfaceEventListener` will never be invoked now?
If so, could it be removed?

native/cocos/engine/Engine.cpp
```diff
@@ -297,12 +297,9 @@ int32_t Engine::restartVM() {
 
     delete _programLib;
     delete _builtinResMgr;
-    CC_SAFE_DESTROY_AND_DELETE(_gfxDevice);
```
timlyeee, line 300: The reason why we should delete the GFXDevice, is that once the app start, it will create a brandly new GFXSwapchain, yet no where to delete it. 

It should be deleted on ts engine, do we have that step?

native/cocos/renderer/GFXDeviceManager.h
minggo: It is used in DeviceManager internally.

suqionglian:
@cocos-robot run test cases

dumganhar (approved):

native/cocos/engine/Engine.cpp
```diff
@@ -311,7 +308,6 @@ int32_t Engine::restartVM() {
     cc::EventDispatcher::init();
     CC_CURRENT_APPLICATION()->init();
 
-    cc::gfx::DeviceManager::addSurfaceEventListener();
```
novain1102, line 314: This line seems necessary since new EventDispatcher init above.

minggo: Yep, you are right.

minggo: But it means all events should be re-added, not only DeviceManager.

suqionglian:
@cocos-robot run test cases

qiuguohua (approved):

native/cocos/engine/Engine.cpp
minggo: I modify the restart logic.

native/cocos/engine/Engine.cpp
minggo: I modify the logic, please review again.

novain1102 (approved):

timlyeee (approved):

suqionglian:
@cocos-robot run test cases

suqionglian:
@cocos-robot run test cases

sushanhong:
@cocos-robot run test cases

suqionglian:
@cocos-robot run test cases

native/cocos/application/BaseGame.cpp
```diff
@@ -50,16 +50,16 @@ int BaseGame::init() {
...
-    }
-
+    
     int ret = cc::CocosApplication::init();
     if (ret != 0) {
         return ret;
     }
 
+    if (_debuggerInfo.enabled) {
+        setDebugIpAndPort(_debuggerInfo.address, _debuggerInfo.port, _debuggerInfo.pauseOnStart);
```
dumganhar, line 60: Why change the order of setDebugIpAndPort after CocosApplication::init? 
This change will make the JS debugger disabled all the time even in debug mode.
Screen Shot 2022-06-03 at 23 16 52

minggo: As it depends on ScriptEngine, so should execute the codes after ScriptEngine is created.

Issue #11306, opened 2022-05-31
dispatchOrientationChangeEvent is not called


qiuguohua:
### Cocos Creator version

3.5.0 3.5.1

### System information

Andorid/IOS

### Issue description

In 3.5 and 3.5.1, the dispatchOrientationChangeEvent function is not called, so the js layer cannot call back the onOrientationChanged function

### Steps to reproduce

The mobile terminal triggers the screen orientation switch

Issue #10955, opened 2022-05-06
DeviceEvent prefix is redundant, parameters are not clear


qiuguohua:
DeviceEvent prefix is redundant, parameters are not clear

PR #11308, opened 2022-05-31
deviceevent related issues

qiuguohua:
1、Fix the issue that dispatchOrientationChangeEvent is not called（https://github.com/cocos/cocos-engine/issues/11306 )
2、Modify the problem of DeviceEvent naming（https://github.com/cocos/cocos-engine/issues/10955  )
3、The touch event is called back to the application

### DIFF
--- a/native/cocos/bindings/event/EventDispatcher.h
+++ b/native/cocos/bindings/event/EventDispatcher.h
@@ -242,12 +242,12 @@ class DeviceEvent : public OSEvent {
 public:
     CONSTRUCT_EVENT(DeviceEvent, OSEventType::DEVICE_OSEVENT)
     enum class Type {
-        DEVICE_MEMORY,
-        DEVICE_ORIENTATION,
+        MEMORY,
+        ORIENTATION,
         UNKNOWN
     };
     EventParameterType args[3];
-    Type type{Type::DEVICE_MEMORY}; // NOLINT(modernize-use-nullptr)
+    Type type{Type::UNKNOWN}; // NOLINT(modernize-use-nullptr)
 };
 
 class EventDispatcher {
--- a/native/cocos/engine/Engine.cpp
+++ b/native/cocos/engine/Engine.cpp
@@ -349,14 +349,15 @@ Engine::SchedulerPtr Engine::getScheduler() const {
 }
 
 bool Engine::dispatchDeviceEvent(const DeviceEvent &ev) { // NOLINT(readability-convert-member-functions-to-static)
+    bool isHandled = false;
     if (ev.type == DeviceEvent::Type::MEMORY) {
         cc::EventDispatcher::dispatchMemoryWarningEvent();
-        return true;
+        isHandled = true;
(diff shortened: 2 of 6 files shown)


native/cocos/engine/Engine.cpp
```diff
@@ -341,17 +341,20 @@ bool Engine::handleEvent(const OSEvent &ev) {
 
 bool Engine::handleTouchEvent(const TouchEvent &ev) { // NOLINT(readability-convert-member-functions-to-static)
     cc::EventDispatcher::dispatchTouchEvent(ev);
-    return true;
+    return dispatchEventToApp(OSEventType::TOUCH_OSEVENT, ev);
```
qiuguohua, line 344: The touch event is called back to the application

native/cocos/bindings/event/EventDispatcher.h
```diff
@@ -242,12 +242,12 @@ class DeviceEvent : public OSEvent {
 public:
     CONSTRUCT_EVENT(DeviceEvent, OSEventType::DEVICE_OSEVENT)
     enum class Type {
-        DEVICE_MEMORY,
-        DEVICE_ORIENTATION,
+        MEMORY,
```
minggo, line 245: What's memory event?

native/cocos/bindings/event/EventDispatcher.h
```diff
@@ -242,12 +242,12 @@ class DeviceEvent : public OSEvent {
...
+        ORIENTATION,
```
minggo, line 246: orientation changed?

qiuguohua: Yep

native/cocos/bindings/event/EventDispatcher.h
qiuguohua: DeviceEvent is two-dimensional, and memory is related to memory, such as insufficient memory, memory failure, etc. Currently, there are only insufficient memory events.Subtypes may be added in the future

native/cocos/engine/Engine.cpp
dumganhar: Why need to dispatch touch event to application? 
I think only JS code needs touch event.

qiuguohua: Developers have the ability to listen to all events

native/cocos/bindings/event/EventDispatcher.h
minggo: If so, then should it be more specific? Such as OUT_OF_MEMORY or so on.

dumganhar (approved):

native/cocos/bindings/event/EventDispatcher.h
qiuguohua: Yes, but currently there is only one kind of event and only one interface to the upper layer(**dispatchMemoryWarningEvent**). We can expand in the future

minggo (approved):

PR #11318, opened 2022-06-01
add copy constructor, move constructor, operator= functions

minggo:


native/cocos/core/geometry/Frustum.h
```diff
@@ -140,12 +140,40 @@ class Frustum final : public ShapeBase {
...
         for (auto *plane : planes) {
             plane->release();
         }
     }
 
+    // Can remove these operator override functions if not using Plane* in planes array.
+    Frustum &operator=(const Frustum &rhs) {
+        setType(rhs.getType());
+        for (size_t i = 0; i < planes.size(); ++i) { // NOLINT(modernize-loop-convert)
+            planes[i] = ccnew Plane();
```
dumganhar, line 161: Need to check old planes if they exist?
```
if (planes[i] == nullptr) {
    planes[i] = ccnew Plane();
    planes[i]->addRef();
}

```

native/cocos/core/geometry/Frustum.h
```diff
@@ -140,12 +140,40 @@ class Frustum final : public ShapeBase {
...
+            planes[i]->addRef();
+            Plane::copy(planes[i], *(rhs.planes[i]));
+        }
+        return *this;
+    }
+
+    Frustum &operator=(Frustum &&rhs) {
+        setType(rhs.getType());
+        for (size_t i = 0; i < planes.size(); ++i) { // NOLINT(modernize-loop-convert)
+            planes[i] = rhs.planes[i];
```
dumganhar, line 171: Need to release old planes
```c++
CC_SAFE_RELEASE(planes[i]);
planes[i] = rhs.planes[i];
```

dumganhar: After looking at the source code of Frustum, I think planes should not be released.  Instead Plane needs a `reset` function to reset values to its original state.

dumganhar:
Plz read the comments of step in the following demo code.

```c++
Frustum &operator=(const Frustum &rhs) {
    if (this == &rhs) { // 1. Check whether rhs is self, if so, just return self.
        return *this;
    }

    setType(rhs.getType());
    for (size_t i = 0; i < planes.size(); ++i) {
        if (planes[i] == nullptr) { // 2. Check whether self's planes[i] is nullptr, it may be nullptr since copy constructor invoke copy operator directly
            planes[i] = ccnew Plane();
            planes[i]->addRef();
        }

        Plane::copy(planes[i], *(rhs.planes[i])); // 3. Copy the values in rhs.planes[i] to this->planes[i]
    }

    // 4. Copy vertices
    this->vertices = rhs.vertices;

    return *this;
}

Frustum &operator=(Frustum &&rhs) {
    if (this == &rhs) { // 1. Check whether rhs is self, if so, just return self.
        return *this;
    }

    setType(rhs.getType());
    for (size_t i = 0; i < planes.size(); ++i) {
        if (planes[i] == nullptr) { // 2. Check whether self's planes[i] is nullptr, it may be nullptr since move constructor invoke move operator directly
            planes[i] = ccnew Plane();
            planes[i]->addRef();
        }

        Plane::copy(planes[i], *(rhs.planes[i])); // 3. Copy the values in rhs.planes[i] to this->planes[i] even we're in move operator since planes in Frustum should not be shared
        rhs.planes[i].reset(); // 4. Reset rhs.planes[i] to its orignal value since we are in move operator.
    }

    // 5. Copy vertices, no need to move since it's ccstd::array.
    this->vertices = rhs.vertices;

    return *this;
}


```

Therefore, the difference between copy operator & move operator is that whether to reset the rhs.planes.

minggo:
@dumganhar yep, i fixed the logic.

dumganhar (approved):

PR #11348, opened 2022-06-02
fixed #11339 : [v3.6] Crash after hot update by invoking cc.game.restart().

### LINKED ISSUE #11339: [v3.6] Crash after hot update
It's dangerous to use a thread that is detached.

WeChatd045b816d0efec0f19686a16d4572108
WeChatb2483f91871ca867ebd70bf33ae4e764

WeChatc9d23b370d6c5540a407838ee7b08dd2

WeChat456acf2d194dda4563e9be21df6715e2

## Step of code execute flow, read the `// -->` comments

```c++
void MessageQueue::runConsumerThread() noexcept {
    if (_immediateMode || _workerAttached) return;

    _reader.terminateConsumerThread = false;
    _reader.flushingFinished = false;

    std::thread consumerThread(&MessageQueue::consumerThreadLoop, this);
    consumerThread.detach(); // --> 1. consumerThreadLoop is a detached thread
    _workerAttached = true;
}

void DeviceAgent::doDestroy() {
    ......

    // --> 2. terminateConsumerThread in main thread
    _mainMessageQueue->terminateConsumerThread();

    _mainMessageQueue->~MessageQueue();
    CC_FREE_ALIGN(_mainMessageQueue);// --> 9. Delete MessageQueue instance, release its memory
    _mainMessageQueue = nullptr;
}

void MessageQueue::terminateConsumerThread() noexcept {
    if (_immediateMode || !_workerAttached) return;

    EventSem event;
    EventSem *const pEvent = &event;

    ReaderContext *const pR = &_reader;

    ccnew_placement(allocate<TerminateConsumerThreadMessage>(1)) TerminateConsumerThreadMessage(pEvent, pR); // --> 3. Send a terminate message to queue

    kick(); // --> 4. Tell the consumer thread to work
    event.wait(); // --> 5. Wait the last TerminateConsumerThreadMessage to be executed
    // --> 8. terminateConsumerThread done
}

void MessageQueue::consumerThreadLoop() noexcept {
    while (!_reader.terminateConsumerThread) {// --> 10. Bang~~~, MessageQueue was freed in step 9. Here checks the freed memory. Crash is found by Clang Address Sanitizer
        AutoReleasePool autoReleasePool;
        flushMessages(); // --> 6. flushMessages all messages in queue, handle the last terminate message
        // --> 8. continue to while loop
    }

    _workerAttached = false;// --> 11. Bang~~~, Here checks the freed memory second time. Crash is found by Clang Address Sanitizer
}

void TerminateConsumerThreadMessage::execute() noexcept {
    _reader->terminateConsumerThread = true;
    _reader->flushingFinished = true;
    _event->signal(); // --> 7. Notify main thread the last message was done, it just set two flags in reader.
}

```

dumganhar:
MessageQueue::terminateConsumerThread needs to wait consumer thread to exit. So don't detach consumer thread and we need to join the thread at the end of terminateConsumerThread.

Re: #11339

### Changelog

* Fix a crash after hot update by invoking cc.game.restart().

-------

### Continuous Integration

This pull request:

* [X] needs automatic test cases check.

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/base/threading/MessageQueue.cpp
+++ b/native/cocos/base/threading/MessageQueue.cpp
@@ -124,8 +124,7 @@ void MessageQueue::runConsumerThread() noexcept {
     _reader.terminateConsumerThread = false;
     _reader.flushingFinished = false;
 
-    std::thread consumerThread(&MessageQueue::consumerThreadLoop, this);
-    consumerThread.detach();
+    _consumerThread = ccnew std::thread(&MessageQueue::consumerThreadLoop, this);
     _workerAttached = true;
 }
 
@@ -141,6 +140,10 @@ void MessageQueue::terminateConsumerThread() noexcept {
 
     kick();
     event.wait();
+
+    if (_consumerThread->joinable()) {
+        _consumerThread->join();
+    }
 }
 
 void MessageQueue::finishWriting() noexcept {
--- a/native/cocos/base/threading/MessageQueue.cpp
+++ b/native/cocos/base/threading/MessageQueue.cpp
@@ -141,9 +141,12 @@ void MessageQueue::terminateConsumerThread() noexcept {
     kick();
     event.wait();
 
-    if (_consumerThread->joinable()) {
-        _consumerThread->join();
+    if (_consumerThread != nullptr) {
+        if (_consumerThread->joinable()) {
+            _consumerThread->join();
+        }
     }
+    CC_SAFE_DELETE(_consumerThread);
 }
 
 void MessageQueue::finishWriting() noexcept {
--- a/native/cocos/base/threading/MessageQueue.h
+++ b/native/cocos/base/threading/MessageQueue.h
@@ -179,6 +179,7 @@ private:
     bool _immediateMode{true};
     bool _workerAttached{false};
     bool _freeChunksByUser{true}; // recycled chunks will be stashed until explicit free instruction
+    std::thread *_consumerThread{nullptr};
 
     friend class MemoryChunkSwitchMessage;
 };


native/cocos/base/threading/MessageQueue.cpp
```diff
@@ -141,6 +140,10 @@ void MessageQueue::terminateConsumerThread() noexcept {
 
     kick();
     event.wait();
+
+    if (_consumerThread->joinable()) {
+        _consumerThread->join();
```
star-e: after join, does it need to delete _consumerThread？

dumganhar: Yes, need delete it. I will fix it. Thank you.

star-e (approved):

hana-alice:
what if we are running on triple-buffer? eg. MainmessageQueue can be running ahead of RenderQueue 3 frames before, now we have to wait until RenderQueue finished, which means it's almost running synchronously, in the aspect of frame.

dumganhar:
> what if we are running on triple-buffer? eg. MainmessageQueue can be running ahead of RenderQueue 3 frames before, now we have to wait until RenderQueue finished, which means it's almost running synchronously, in the aspect of frame.

The key point of this issue is that the MessageQueue is deleted before RenderThread exits. So whether to use triple-buffer may not solve this issue. :(

hana-alice:
yeah.
I mean does this pr has any effect on further implementation on triple-buffer? because we have to wait at end of the logic thread every frame, this may stops us from doing things like logic thread goes ahead of render thread 2 or 3 frames. And maybe this is why it's originally use 'detach'.

dumganhar:
> I mean does this pr has any effect on further implementation on triple-buffer? because we have to wait at end of the logic thread every frame, this may stops us from doing things like logic thread goes ahead of render thread 2 or 3 frames. And maybe this is why it's originally use 'detach'.

@hana-alice  My change only affects the logic of destroying MessageQueue, not affects the logic of every frame.

hana-alice (approved):

native/cocos/base/threading/MessageQueue.cpp
```diff
@@ -141,6 +140,13 @@ void MessageQueue::terminateConsumerThread() noexcept {
 
     kick();
     event.wait();
+
+    if (_consumerThread != nullptr) {
```
minggo, line 144: After `event.wait()`, the thread is not exited?

minggo (approved):

xjtulk:
I met a similar crash about MessageQueue and i use v3.5.2 engine, After apply the same code change in my branch, it seems the crash disappear, but i am confused about the change, i think `event.Wait()` already ensure the thread exit cause it only be signaled when the TerminateConsumerThreadMessage is executed and the quit flags be set, so can you guys explain why the code works @dumganhar @minggo @hana-alice

dumganhar:
@xjtulk The explanation is in the issue: https://github.com/cocos/cocos-engine/issues/11339

xjtulk:
Thanks very much for the detailed explanation @dumganhar

PR #11370, opened 2022-06-06
fixed #11362: [3.6] No debugger could be attached.

### LINKED ISSUE #11362: [3.6] No debugger could be attached.
Once the application is already init, the code to set debugger's info makes no meaning here.
```cpp
    int ret = cc::CocosApplication::init();
    if (ret != 0) {
        return ret;
    }

    if (_debuggerInfo.enabled) {
        setDebugIpAndPort(_debuggerInfo.address, _debuggerInfo.port, _debuggerInfo.pauseOnStart);
    }

```
Thus we have no debugger tool to use or attached.

dumganhar:
Re: #11362

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/bindings/jswrapper/v8/ScriptEngine.h
+++ b/native/cocos/bindings/jswrapper/v8/ScriptEngine.h
@@ -353,9 +353,13 @@ public:
         uint32_t port{0};
         bool isWait{false};
         inline bool isValid() const { return !serverAddr.empty() && port != 0; }
-        inline void reset() { serverAddr.clear(); port = 0; isWait = false; }
+        inline void reset() {
+            serverAddr.clear();
+            port = 0;
+            isWait = false;
+        }
     };
-    static void _setDebuggerInfo(const DebuggerInfo& info); // NOLINT(readability-identifier-naming)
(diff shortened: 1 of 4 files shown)


native/cocos/bindings/jswrapper/v8/ScriptEngine.h
```diff
@@ -348,6 +348,18 @@ class ScriptEngine final {
     v8::Local<v8::Context> _getContext() const;                   // NOLINT(readability-identifier-naming)
     void _setGarbageCollecting(bool isGarbageCollecting);         // NOLINT(readability-identifier-naming)
 
+    struct DebuggerInfo {
+        ccstd::string serverAddr;
```
timlyeee, line 352: there's a default value for serverAddr? like 127.0.0.1?

dumganhar: I think it doesn't need to be 127.0.0.1 as its default value since we check whether it's valid by the length of serverAddr string.

minggo (approved):

PR #11406, opened 2022-06-08
Fix non-foreground state processing time-consuming tasks, resulting in anr

bofeng-song:


native/cocos/platform/android/AndroidPlatform.cpp
```diff
@@ -398,6 +400,10 @@ class GameInputProxy {
         return _isVisible && _hasWindow;
     }
 
+    bool isForeground() const {
```
dumganhar: Add `inline`

minggo: It is better to rename to `isActive`.

minggo (approved):

PR #11461, opened 2022-06-10
fix crash if reset texture property of a Material

minggo:
This issue fix the crash if reset Material's texture property.

Steps to reproduce the issue:
- run test case 3d
- click `mipMapCheck`
- it will crash after a while on macOS

The reasons are:
- native implementation of `Material.setProperty()` doesn't handle null value
- `Pass::resetTexture()` may pass negative index to `DescriptorSet::bindTexture()` which doesn't reset texture correctly, then the descriptor set will hold a released texture and cause crash.

The web implementation of `Pass.resetTexture()` also has similar issue, it doesn't handle undefined value.

### DIFF
--- a/native/cocos/scene/Pass.cpp
+++ b/native/cocos/scene/Pass.cpp
@@ -316,7 +316,11 @@ void Pass::resetUniform(const ccstd::string &name) {
     _rootBufferDirty = true;
 }
 
-void Pass::resetTexture(const ccstd::string &name, index_t index /* = CC_INVALID_INDEX */) {
+void Pass::resetTexture(const ccstd::string &name) {
+    resetTexture(name, 0);
+}
+
+void Pass::resetTexture(const ccstd::string &name, index_t index) {
     const uint32_t handle = getHandle(name);
     if (0 == handle) {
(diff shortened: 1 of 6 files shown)


native/cocos/scene/Pass.cpp
```diff
@@ -316,7 +316,11 @@ void Pass::resetUniform(const ccstd::string &name) {
...
 }
 
-void Pass::resetTexture(const ccstd::string &name, index_t index /* = CC_INVALID_INDEX */) {
+void Pass::resetTexture(const ccstd::string &name) {
+    resetTexture(name, 0);
+}
+
+void Pass::resetTexture(const ccstd::string &name, index_t index) {
     const uint32_t handle = getHandle(name);
     if (0 == handle) {
```
dumganhar, line 325: Should check whether index is CC_INVALID_INDEX here?

minggo: I will change the index type in other PR.

dumganhar (approved):

yiwenxue (approved):

PR #11480, opened 2022-06-13
refactoring pass.ts

star-e:
Major change to Pass.

Added setTexture, setSampler, setTextureAndSampler and their Array version to Pass.
These functions uses `name` to set texture/sampler, instead of `binding`, because binding might change when effectAsset is loaded / unloaded when using custom pipeline.

In fact, almost all 'pass.bindTexture' usage uses `name` to access descriptor already. This pr does not change many usages.

User should not uses binding to access descriptors, `pass.bindTexture` should be marked as deprecated.

In the future, we might pre-compute descriptor binding id during build time. At that time binding id will be more stable.

### Changelog

* changed pass, material.
* changed related usages.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


star-e:
@cocos-robot run test cases

cocos/core/assets/material.ts
```diff
@@ -441,32 +441,51 @@ export class Material extends Asset {
...
             }
-            pass.bindTexture(binding, texture, index);
-            pass.bindSampler(binding, val.getGFXSampler(), index);
+            pass.setTextureAndSampler(name, texture, val.getGFXSampler());
         }
     }
 
+    protected _bindTextureArray (pass: Pass, name: string, val: MaterialPropertyFull[]) {
+        const textures = new Array<Texture>(val.length);
+        const samplers = new Array<Sampler>(val.length);
```
dumganhar, line 470: Will _bindTextureArray be invoked frequently? 
`new Array`  will return GC objects, if it could be invoked frequently, it will make lots of garbage.

star-e: It will be triggered when properties are loaded or changed, like changing textures.
This should not happen frequently.
If a material need changing textures, user should consider using other rendering mechanism.

minggo (approved):

pandamicro (approved):

suqionglian:
@cocos-robot run test cases

star-e:
Currently, gfx::Shader is exposed through Pass, Submodel, Model to the outside.
As a result, 
1. Without major refactoring, change shader at runtime is impossible. 
2. Shader initialization time is uncontrollable. 

This pr is not helpful under current situation, and shader binding must be rearranged during build time.
This pr will be postponed until Pass, Submodel, Model are refactored.

PR #11487, opened 2022-06-13
Remove native camera scale matrix

GengineJS:


native/cocos/scene/Camera.cpp
```diff
@@ -190,6 +190,9 @@ void Camera::update(bool forceUpdate /*false*/) {
 
     // view-projection
     if (viewProjDirty) {
+        Mat4 scaleMat{};
+        scaleMat.scale(_node->getWorldScale());
+        Mat4::multiply(scaleMat, _matView, &_matView);
```
minggo: Have two questions:
- why need to remove scale matrix
- the logic seems to multiply scale matrix

GengineJS: for this issue：https://github.com/cocos/3d-tasks/issues/10377

GengineJS: Because _matView is the inverse world matrix of the node, it needs to be corrected by the world scale of the node

GengineJS: as shown below：

image

minggo: It will do reverse every frame. I think can cache the result. It is updated only `_matView` is changed.

GengineJS: like this?

Mat4::multiply(scaleMat, _matView, &_tempMatView); ?

GengineJS: The position has been changed, it will not update every frame now

native/cocos/scene/Camera.cpp
```diff
@@ -160,7 +160,10 @@ void Camera::update(bool forceUpdate /*false*/) {
     if (_node->getChangedFlags() || forceUpdate) {
         _matView = _node->getWorldMatrix().getInversed();
         _forward.set(-_matView.m[2], -_matView.m[6], -_matView.m[10]);
-
+        Mat4 scaleMat{};
+        scaleMat.scale(_node->getWorldScale());
+        // remove scale
+        Mat4::multiply(scaleMat, _matView, &_matView);
```
minggo, line 166: Why native implementation is different from ts version?

GengineJS: native's dst is on the right, typescript is on the left. so it is consistent.

static void cc::Mat4::multiply(const cc::Mat4 &m1, const cc::Mat4 &m2, cc::Mat4 *dst)

minggo (approved):

star-e (approved):

PR #11498, opened 2022-06-14
fix crash when delete transform

minggo:
Fix crash if running `SwitchAnimation` test case.


novain1102 (approved):

novain1102:
related issue: https://github.com/cocos/3d-tasks/issues/12473

dumganhar (approved):

PR #11501, opened 2022-06-14
[v3.5.2] Solved the problem of switching scene shadows to disappear

troublemaker52025:
Re: #

### Changelog

* 

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/core/root.ts
+++ b/cocos/core/root.ts
@@ -601,22 +601,18 @@ export class Root {
                 switch (l.type) {
                 case LightType.SPHERE:
                     l.scene.removeSphereLight(l as SphereLight);
-                    p.free(l);
-                    l.destroy();
                     break;
                 case LightType.SPOT:
                     l.scene.removeSpotLight(l as SpotLight);
-                    p.free(l);
-                    l.destroy();
                     break;
                 case LightType.DIRECTIONAL:
                     l.scene.removeDirectionalLight(l as DirectionalLight);
-                    p.free(l);
-                    l.destroy();
                     break;
                 default:
                     break;
                 }
+                p.free(l);
+                l.destroy();
             }
         }
     }
--- a/cocos/core/root.ts
+++ b/cocos/core/root.ts
@@ -596,21 +596,23 @@ export class Root {
     public destroyLight (l: Light) {
         const p = this._lightPools.get(l.constructor as Constructor<Light>);
         if (p) {
-            p.free(l);
             if (l.scene) {
                 switch (l.type) {
                 case LightType.SPHERE:
                     l.scene.removeSphereLight(l as SphereLight);
+                    p.free(l);
+                    l.destroy();
                     break;
                 case LightType.SPOT:
                     l.scene.removeSpotLight(l as SpotLight);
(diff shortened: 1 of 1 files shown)


troublemaker52025:
@cocos-robot run test cases

cocos/core/root.ts
```diff
@@ -596,21 +596,23 @@ export class Root {
     public destroyLight (l: Light) {
         const p = this._lightPools.get(l.constructor as Constructor<Light>);
         if (p) {
-            p.free(l);
```
troublemaker52025, line 599: 会把本来不需要处理的 direction-light 意外的释放掉

SantyWang (approved):

troublemaker52025:
* https://github.com/cocos/3d-tasks/issues/12525

jk20012001 (approved):

pandamicro (commented):
1. Please rebase to v3.6, v3.5.2 doesn't accept new PRs
2. This pr doesn't seem complete, if we don't release DirectionalLight in render scene, there need to be a synchronize process during scene transition.
3. Currently render scene is managing directional light just like other light source, using an array and push any attached light to it, it's just setting one main light.

So the real important questions are

1. How to manage directional lights in render scene ? 
    - In an array and manage just like other light sources, only one of the directional lights is the main light and activated
    - Only use one main light reference, and replace it each time a new directional light is attached
    - Retain one main DirectionalLight source object, each time update the settings with the main DirectionalLightComponent setup when new scene loaded. Be noticed, the DirectionalLightComponent will no longer have a DirectionalLight instance any more
3. Who is in charge of destroying the directional light ? The render scene or the DirectionalLightComponent

cocos/core/root.ts
```diff
@@ -605,12 +605,16 @@ export class Root {
...
                     l.scene.removeSpotLight(l as SpotLight);
                     break;
+                case LightType.DIRECTIONAL:
+                    l.scene.removeDirectionalLight(l as DirectionalLight);
+                    break;
                 default:
                     break;
                 }
+                p.free(l);
+                l.destroy();
```
pandamicro, line 615: If light is destroyed, it shouldn't be free to the light pool, otherwise unexpected behavior will occur when it's reused

pandamicro: 1. `destroyLight` should only destroy it, shouldn't free it
2. We can add a `recycleLight` function to free it but not destroy it
3. All light component's `onDestroy` should only invoke `recycleLight` so that the light object will be reused

pandamicro (changes requested):

cocos/core/root.ts
troublemaker52025: sure，I will change that on the v3.6.0 version

pandamicro: v3.5.2 won't accept this PR, please directly rebase the most suitable modification to v3.6

troublemaker52025:
rebase the most suitable modification to v3.6
* https://github.com/cocos/cocos-engine/pull/11506

PR #11516, opened 2022-06-15
use uint32_t for index

minggo:


native/cocos/core/assets/EffectAsset.h
```diff
@@ -404,7 +404,7 @@ struct IBlockInfo {
 };
 
 struct ISamplerTextureInfo {
-    int32_t binding{-1};
+    uint32_t binding{0};
```
star-e: binding has to be 0xFFFFFFFF, which is the default (null) value.

native/cocos/core/assets/Material.h
```diff
@@ -311,7 +311,7 @@ class Material : public Asset {
 protected:
     void update(bool keepProps = true);
     bool uploadProperty(scene::Pass *pass, const ccstd::string &name, const MaterialPropertyVariant &val);
-    void bindTexture(scene::Pass *pass, uint32_t handle, const MaterialProperty &val, index_t index = CC_INVALID_INDEX);
+    void bindTexture(scene::Pass *pass, uint32_t handle, const MaterialProperty &val, uint32_t index = 0);
```
star-e, line 314: ~~`uint32_t index` should be CC_INVALID_INDEX, which is 0xFFFFFFFF~~
We can use 0 as default value

native/cocos/core/assets/EffectAsset.h
minggo: Yep, seems reasonable.

minggo: fixed

dumganhar (approved):

star-e (approved):

Issue #11363, opened 2022-06-06
Apple platform macros need to be adjusted


dumganhar:
Currently, 
```
CC_PLATFORM_MAC_IOS  // It's iOS, but it has MAC keyword, it's confused.
CC_PLATFORM_MAC_OSX // It's macOS
```

Suggestion:
```
CC_PLATFORM_IOS
CC_PLATFORM_MACOS
CC_PLATFORM_WATCHOS
CC_PLATFORM_TVOS
```

PR #11528, opened 2022-06-15
Change macro name

minggo:
https://github.com/cocos/cocos-engine/issues/11363


dumganhar:
Should deprecate these macros?

minggo:
Is there any better way to deprecate the macro?

minggo:
> Should deprecate these macros?

Fixed, keep old macros too.

dumganhar (approved):

PR #11564, opened 2022-06-16
Cleanup code in Node.h/.cpp

dumganhar:
-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/core/scene-graph/Node.h
+++ b/native/cocos/core/scene-graph/Node.h
@@ -266,9 +266,9 @@ public:
     inline Node *getParent() const { return _parent; }
     inline NodeEventProcessor *getEventProcessor() const { return _eventProcessor; }
 
-    Node *getChildByUuid(const ccstd::string &) const;
-    Node *getChildByName(const ccstd::string &) const;
-    Node *getChildByPath(const ccstd::string &) const;
+    Node *getChildByUuid(const ccstd::string &uuid) const;
+    Node *getChildByName(const ccstd::string &name) const;
+    Node *getChildByPath(const ccstd::string &path) const;
     inline index_t getSiblingIndex() const { return _siblingIndex; }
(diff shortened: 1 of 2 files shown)


native/cocos/core/scene-graph/Node.h
```diff
@@ -357,10 +357,10 @@ class Node : public BaseNode {
     /**
      * @en Inversely transform a point from world coordinate system to local coordinate system.
      * @zh 逆向变换一个空间点，一般用于将世界坐标转换到本地坐标系中。
-     * @param out The result point in local coordinate system will be stored in this vector
      * @param p A position in world coordinate system
+     * @return The result point in local coordinate system will be stored in this vector
      */
-    void inverseTransformPoint(Vec3 &out, const Vec3 &p);
+    Vec3 inverseTransformPoint(const Vec3 &p);
```
minggo, line 363: Why change it. It is different from TS implementation.

dumganhar: There are two main reasons:

1. `Vec3& out` is not user friendly for c++ developers, c++ has stack for returning a temporary variable, but JS objects are allocated on heap, so JS needs a `out` parameter for returning value without generating GC objects.

2. JS binding could not bind a function with a `out` parameter, the original `inverseTransformPoint` auto binding code will not work properly.

Actually, we need to bind `inverseTransformPoint` manually. I will do it later.

minggo: Ok. Will modify in this PR, right?

dumganhar: No, I will create a new issue for that.

dumganhar: https://github.com/cocos/cocos-engine/issues/11570

minggo (approved):

Issue #11158, opened 2022-05-19
Implement transition interruption in animation graph


shrinktofit:

shrinktofit:
https://github.com/cocos/cocos-engine/pull/11566

PR #11566, opened 2022-06-17
Animation graph transition interruption

shrinktofit:
Re: #11158

### Changelog

* Supports transition interruption.

Design notes: https://github.com/cocos/3d-tasks/discussions/12274

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/core/animation/marionette/animation-graph.ts
+++ b/cocos/core/animation/marionette/animation-graph.ts
@@ -65,6 +65,14 @@ export type { TransitionView as Transition };
 
 export type TransitionInternal = Transition;
 
+export enum TransitionInterruption {
+    NONE,
+    CURRENT_STATE,
+    NEXT_STATE,
+    CURRENT_STATE_THEN_NEXT_STATE,
+    NEXT_STATE_THEN_CURRENT_STATE,
+}
+
 @ccclass(`${CLASS_NAME_PREFIX_ANIM}AnimationTransition`)
 class AnimationTransition extends Transition {
     /**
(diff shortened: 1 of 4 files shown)


cocos/core/animation/marionette/animation-graph.ts
```diff
@@ -65,6 +65,14 @@ export type { TransitionView as Transition };
 
 export type TransitionInternal = Transition;
 
+export enum TransitionInterruption {
```
SantyWang: TransitionInterruptionMode or TransitionInterruptionType?

cocos/core/animation/marionette/graph-eval.ts
```diff
@@ -1099,27 +1360,59 @@ interface TransitionMatch {
...
+    from: MotionStateEval;
+}
+
 class TransitionMatchCache {
     public transition: TransitionMatch['transition'] | null = null;
 
-    public requires = 0.0;
+    public requires = Infinity;
+
+    public hasMinimalCost (): this is TransitionMatch {
```
SantyWang: I feel that this name may cause problems of understanding, "minimal" should mean the smallest of a bunch of data. But there seems to be no comparison with other data here, so it would be more accurate to use hasZeroCost？

cocos/core/animation/marionette/animation-graph.ts
shrinktofit: I'll rename it into `TransitionInterruptionSource`.

cocos/core/animation/marionette/graph-eval.ts
shrinktofit: Good catch. Changed.

shrinktofit:
Edit: in current edition. I want the usage of transition interruption to be positioned as "whether the transition is interruptible or not"(the newly added `.interruptible: boolean`) instead of "how can 
 the transition be interrupted"(.interruptionSource: TransitionInterruptionSource). Once we collect feedbacks we may open the interruption source property.

SantyWang (approved):

PR #11576, opened 2022-06-17
android platform use system's original size, do not need to rotate

bofeng-song:


star-e (approved):

minggo:
Do not need pre transform any more.

GengineJS (approved):

bofeng-song:
> Do not need pre transform any more.

has been refined

minggo (approved):

PR #11588, opened 2022-06-18
remove duplicate codes

minggo:


native/cocos/core/memop/CachedArray.h
```diff
@@ -1,192 +1,167 @@
...
-     * 得到数组中指定位置的元素
-     * @param idx The index of the requested element
-     * @return The element at given index
-     */
-    T *get(uint32_t idx) const {
-        if (idx < _length) {
-            return _array[idx];
+    CachedArray &operator=(CachedArray &&other) noexcept {
+        if (this != &other) {
+            delete[] _array;
```
dumganhar, line 74: Use CC_SAFE_DELETE_ARRAY instead of delete[] ?

dumganhar (approved):

native/cocos/core/memop/CachedArray.h
minggo: Can do in other PR.

PR #11598, opened 2022-06-18
use uint32_t instead of uint

star-e:
use uint32_t instead of uint

### Changelog

* renderer uses uint32_t 

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


native/cocos/renderer/core/ProgramLib.cpp
```diff
@@ -390,7 +390,7 @@ IProgramInfo *ProgramLib::define(IShaderInfo &shader) {
             tmplInfo.blockSizes.emplace_back(getSize(block));
             tmplInfo.bindings.emplace_back();
             auto &bindingsInfo = tmplInfo.bindings.back();
-            bindingsInfo.binding = static_cast<uint>(block.binding);
+            bindingsInfo.binding = static_cast<uint32_t>(block.binding);
```
minggo: Don't need `static_cast` as `block.binding` and `bindingsInfo.binding` are the same type. I modified it in https://github.com/cocos/cocos-engine/pull/11596.

star-e: ok，I will revert the change and wait for #11596 first.

minggo: It is merged. Can just fix conflict.

minggo (approved):

PR #11599, opened 2022-06-18
The base version of sorting priority add

LinYunMo:
Re: https://github.com/cocos/cocos-engine/pull/11094

The base version of sorting，included some features in https://github.com/cocos/cocos-engine/pull/11094

Cause：https://github.com/cocos/3d-tasks/issues/12273#issuecomment-1158724102

Re: https://github.com/cocos/3d-tasks/issues/12287


SantyWang (approved):

SantyWang (approved):

cocos/core/components/model-renderer.ts
```diff
@@ -49,9 +49,23 @@ export class ModelRenderer extends Renderer {
...
 
+    /**
+     * @en The transparent sorting priority which will be applied to the committed models.
+     * @zh 应用于所有提交渲染的 Model 在透明队列中的排序优先级
+     */
+    get transparentPriority () {
+        return this._transparentPriority;
+    }
+
+    set transparentPriority (val) {
```
minggo: What does this function used for? When need to invoke this function?

native/cocos/bindings/auto/jsb_scene_auto.cpp
```diff
@@ -6573,6 +6573,25 @@ static bool js_scene_Model_getTransform(se::State& s) // NOLINT(readability-iden
 }
 SE_BIND_FUNC_AS_PROP_GET(js_scene_Model_getTransform)
 
+static bool js_scene_Model_getTransparentPriority(se::State& s) // NOLINT(readability-identifier-naming)
```
minggo: Don't need to upload these changes, it will be generated by CI automatically.

cocos/core/components/model-renderer.ts
```diff
@@ -49,9 +49,23 @@ export class ModelRenderer extends Renderer {
         this._onVisibilityChange(val);
     }
 
+    /**
+     * @en The transparent sorting priority which will be applied to the committed models.
+     * @zh 应用于所有提交渲染的 Model 在透明队列中的排序优先级
+     */
+    get transparentPriority () {
```
pandamicro: priority is priority, it shouldn't be coupled with transparent or opaque queue

pandamicro: I remember there was sorting priority added or is it the same PR ?

pandamicro (changes requested):

cocos/core/components/model-renderer.ts
LinYunMo: When the user wants to control the priority directly

cocos/core/components/model-renderer.ts
SantyWang: This value is not meaningful for opaque queues. the opaque queue has depth writing turned on, so no matter what order is used to render, the ones closer to the screen are rendered to the front, so this value only works for the translucent queue

pandamicro (approved):

minggo (approved):

PR #11617, opened 2022-06-20
Animation graph preview/variable manipulation

shrinktofit:
Re: #

### Changelog

* Add some editor decicated APIs to support preview animation blend/animation transition in animation graphs, as well as more flexibly manipulate variables in animation graphs. 

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


editor/src/marionette/preview.ts
```diff
@@ -0,0 +1,357 @@
...
+
+    public destroy() {
+    }
+
+    public evaluate() {
+        this._blendBuffer.apply();
+    }
+
+    public addVariable(id: string, type: VariableType, value: Value) {
+        this._varInstances[id] = new VarInstance(type, value);
```
SantyWang: What if add variable with same name twice?

editor/src/marionette/preview.ts
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    public removeVariable(id: string) {
+        delete this._varInstances[id];
+    }
+
+    public updateVariable(id: string, value: Value) {
+        this._varInstances[id].value = value;
```
SantyWang: Need to check if variable exists?

editor/src/marionette/preview.ts
```diff
@@ -0,0 +1,357 @@
...
+            this._timelineStatsDirty = false;
+        }
+        return this._timelineStats;
+    }
+
+    /**
+     * Gets an iterable to the weights of each motion(that has runtime ID).
+     */
+    public queryWeights(): Iterable<[RuntimeID, number]> {
+        const getWeightsRecursive = function* (motionEval: MotionEval, weight: number): Iterable<[RuntimeID, number]> {
```
SantyWang, line 94: What does function* mean?

editor/src/marionette/preview.ts
```diff
@@ -0,0 +1,357 @@
...
+        return [];
+    }
+
+    public setMotion(motion: Motion) {
+        this._motionEval = super.createMotionEval(motion);
+        this._timelineStatsDirty = true;
+    }
+
+    public setTime(time: number) {
+        this._time = time;
```
SantyWang, line 128: should set timeline stats dirty?

editor/src/marionette/preview.ts
shrinktofit: It shall not take effect now in 7a7bb82049ffe5b5eab6695a2f1a7ad6bda1879d

editor/src/marionette/preview.ts
shrinktofit: Solved in 7a7bb82049ffe5b5eab6695a2f1a7ad6bda1879d

editor/src/marionette/preview.ts
shrinktofit: Time does not belong to timeline stats as suggested in https://github.com/shrinktofit/engine/blob/AnimationGraphEditorAPI/editor/src/marionette/preview.ts#L161

editor/src/marionette/preview.ts
shrinktofit: It's a generator whose usage for example is:

```ts
const [[id, weight]] = queryWeights(); // Get the first motion's id and weight
for (const [id, weight] of queryWeights()) { // Iterate motion weights
  if (id === 'a') {
     break;
  }
}
```

Generators save us from array construction/concatenation e.g.

SantyWang (approved):

PR #11623, opened 2022-06-20
Initialize libcurl's environment interface on the NX platform

qiuguohua:
Add curl_global_init to initialize libcurl's environment interface

### DIFF
--- a/native/cocos/network/Downloader-curl.cpp
+++ b/native/cocos/network/Downloader-curl.cpp
@@ -234,6 +234,18 @@ public:
     //        : _thread(nullptr)
     {
         DLLOG("Construct DownloaderCURL::Impl %p", this);
+        #if (CC_PLATFORM == CC_PLATFORM_NX)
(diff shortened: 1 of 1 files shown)


native/cocos/network/Downloader-curl.cpp
```diff
@@ -234,6 +234,15 @@ class DownloaderCURL::Impl : public enable_shared_from_this<DownloaderCURL::Impl
     //        : _thread(nullptr)
     {
         DLLOG("Construct DownloaderCURL::Impl %p", this);
+        #if (CC_PLATFORM == CC_PLATFORM_NX)
```
minggo, line 237: Please add comments about why need to do special operation on NX.

qiuguohua: In addition to the NX platform, there will be no crashes on other platforms. Not sure if it's a version issue or something else

minggo (approved):

PR #11642, opened 2022-06-22
use reference count to share objects

minggo:
`render pass/depth stencil texture/color textures` will be shared outside, so use reference count

### DIFF
--- a/native/cocos/base/RefVector.h
+++ b/native/cocos/base/RefVector.h
@@ -224,7 +224,7 @@ public:
         return _data[idx];
     }
 
-    T operator[](uint32_t idx) const {
+    const T&operator[](uint32_t idx) const {
         return _data[idx];
     }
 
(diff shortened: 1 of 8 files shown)


freeboy (approved):

native/cocos/base/RefVector.h
```diff
@@ -220,6 +220,14 @@ class RefVector {
         return *this;
     }
 
+    T &operator[](uint32_t idx) {
+        return _data[idx];
+    }
+
+    T operator[](uint32_t idx) const {
```
star-e: should return type be `const T&` ?

minggo: Yep, `const T&` is better. Fixed.

star-e (approved):

yiwenxue (approved):

PR #11650, opened 2022-06-22
Modify the path for nx windows to read sqlite

qiuguohua:
Modify the path for nx windows to read sqlite
Use the log log of the nx platform


minggo (approved):

minggo:
Please fix clang tidy error.

PR #11706, opened 2022-06-25
fix forceUpdate

star-e:
Fixes forceUpdate. 
Currently gfx has multiple instance of DescriptorSet(DescriptorSetAgent, DescriptorSetValidator, Derived class of DescriptorSet).  
We must implement forceUpdate on all backend and delegators. This is required by the gfx architecture.

### Changelog

* added forceUpdate to all backends

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/renderer/gfx-agent/DescriptorSetAgent.cpp
+++ b/native/cocos/renderer/gfx-agent/DescriptorSetAgent.cpp
@@ -88,6 +88,17 @@ void DescriptorSetAgent::update() {
         });
 }
 
+void DescriptorSetAgent::forceUpdate() {
+    _isDirty = false;
+    ENQUEUE_MESSAGE_1(
+        DeviceAgent::getInstance()->getMessageQueue(),
+        DescriptorSetForceUpdate,
+        actor, getActor(),
+        {
+            actor->forceUpdate();
+        });
+}
+
 void DescriptorSetAgent::bindBuffer(uint32_t binding, Buffer *buffer, uint32_t index) {
     DescriptorSet::bindBuffer(binding, buffer, index);
 
--- a/native/cocos/renderer/gfx-base/GFXDescriptorSet.cpp
+++ b/native/cocos/renderer/gfx-base/GFXDescriptorSet.cpp
@@ -48,11 +48,6 @@ void DescriptorSet::initialize(const DescriptorSetInfo &info) {
     doInit(info);
 }
 
-void DescriptorSet::forceUpdate() {
-    _isDirty = true;
-    update();
-}
-
 void DescriptorSet::destroy() {
     doDestroy();
 
--- a/native/cocos/renderer/gfx-validator/DescriptorSetValidator.cpp
+++ b/native/cocos/renderer/gfx-validator/DescriptorSetValidator.cpp
@@ -105,6 +105,7 @@ void DescriptorSetValidator::update() {
 }
 
 void DescriptorSetValidator::forceUpdate() {
+    _isDirty = true;
     _actor->forceUpdate();
     _isDirty = false;
 }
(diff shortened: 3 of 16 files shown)


star-e:
@cocos-robot run test cases

native/cocos/renderer/gfx-validator/DescriptorSetValidator.cpp
```diff
@@ -104,6 +104,11 @@ void DescriptorSetValidator::update() {
     _isDirty = false;
 }
 
+void DescriptorSetValidator::forceUpdate() {
+    _actor->forceUpdate();
+    _isDirty = false;
```
minggo, line 110: Other backends set `_isDirty` first, should use the same logic?

star-e: There are two types of logic.
1. DescriptorSetValidator and DescriptorSetAgent are like delegators. They just forward `forceUpdate` to its child.
2. For graphics backend like VKDescriptorSet, they will implement `forceUpdate` and call the real `update`

If we reuse base class's implementation, it will become very confusing. Since there are two possibilities.

native/cocos/renderer/gfx-agent/DescriptorSetAgent.cpp
```diff
@@ -88,6 +88,17 @@ void DescriptorSetAgent::update() {
...
 
+void DescriptorSetAgent::forceUpdate() {
+    _isDirty = false;
+    ENQUEUE_MESSAGE_1(
+        DeviceAgent::getInstance()->getMessageQueue(),
+        DescriptorSetForceUpdate,
+        actor, getActor(),
+        {
+            actor->forceUpdate();
+        });
```
star-e, line 99: forward `forceUpdate` to graphics backend

native/cocos/renderer/gfx-base/GFXDescriptorSet.cpp
```diff
@@ -48,11 +48,6 @@ void DescriptorSet::initialize(const DescriptorSetInfo &info) {
     doInit(info);
 }
 
-void DescriptorSet::forceUpdate() {
-    _isDirty = true;
-    update();
-}
-
```
star-e, line 55: Remove default implementation. Let derived class provide implementation.

native/cocos/renderer/gfx-validator/DescriptorSetValidator.cpp
```diff
@@ -104,6 +104,11 @@ void DescriptorSetValidator::update() {
...
+}
```
star-e: forward `forceUpdate` to the actual actor.

native/cocos/renderer/gfx-validator/DescriptorSetValidator.cpp
minggo: I agree to implement it in each backend. I just mean, should it set dirty first in `DescriptorSetValidator::forceUpdate()`.

star-e: Since `_actor` does not know Validator, set `_isDirty = true` theoretically has no impact on `_actor`'s behaviour.
I can add some comment, but a bit reluctant to add `_isDirty = true`.

minggo: Yep, i think just to let them have the same logic is better.

minggo (approved):

troublemaker52025 (approved):

native/cocos/renderer/gfx-validator/DescriptorSetValidator.cpp
star-e: I decided to set `_isDirty = true`.

PR #11715, opened 2022-06-27
V3.6.0 particle material switch error fix

zxx43:
Re: #

### Changelog

* 

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


zxx43:
@cocos-robot run test cases

cocos/particle/renderer/particle-system-renderer-data.ts
```diff
@@ -151,12 +152,67 @@ export default class ParticleSystemRenderer {
...
+    public get cpuMaterial () {
+        return this._cpuMaterial;
+    }
+
+    public set cpuMaterial (val: Material | null) {
+        if (val === null) {
+            return;
+        } else {
+            const effectName = val.effectName;
+            if (effectName.indexOf('particle') === -1 || effectName.indexOf('particle-gpu') !== -1) {
```
SantyWang, line 172: Why check "particle-gpu" ?

AndyYuTian:
At the current stage, this change will lead to all users previous made particles system lost their material, which is unacceptable for users. it should according to the user existing setup, put users' material into the right property. Plus, the user interface doesn't update immediately after user update the useGPU settings, which does not supposed to happen as well.

cocos/particle/renderer/particle-system-renderer-data.ts
zxx43: do not use effect include gpu for cpu material

zxx43: > Why check "particle-gpu" ?

do not use effect include gpu for cpu material

zxx43:
> At the current stage, this change will lead to all users previous made particles system lost their material, which is unacceptable for users. it should according to the user existing setup, put users' material into the right property. Plus, the user interface doesn't update immediately after user update the useGPU settings, which does not supposed to happen as well.

fixed

LIUYIa:
@cocos-robot run test cases

SantyWang (approved):

SantyWang (approved):

PR #11716, opened 2022-06-27
memory usage check

minggo:


troublemaker52025 (approved):

native/cocos/renderer/pipeline/PipelineSceneData.cpp
```diff
@@ -66,23 +66,16 @@ void PipelineSceneData::activate(gfx::Device *device) {
 }
 
 void PipelineSceneData::destroy() {
-    for (const auto &pair : _shadowFrameBufferMap) {
-        pair.second->destroy();
```
star-e, line 70: Framebuffer::destroy() should be called explicitly . It will not be called in the destructor.

native/cocos/renderer/pipeline/PipelineSceneData.cpp
```diff
@@ -66,23 +66,16 @@ void PipelineSceneData::activate(gfx::Device *device) {
...
+    _occlusionQueryInputAssembler = nullptr;
+    _occlusionQueryVertexBuffer = nullptr;
+    _occlusionQueryIndicesBuffer = nullptr;
 }
 
 void PipelineSceneData::initOcclusionQuery() {
-    if (!_occlusionQueryInputAssembler) {
-        _occlusionQueryInputAssembler = createOcclusionQueryIA();
-    }
+    _occlusionQueryInputAssembler = createOcclusionQueryIA();
```
star-e, line 79: We might `CC_ASSERT(!_occlusionQueryInputAssembler)` first.

native/cocos/renderer/pipeline/PipelineSceneData.cpp
minggo: The sub class's destructor(for example, MTLFrameBuffer) will invoke destroy().

native/cocos/renderer/pipeline/PipelineSceneData.cpp
minggo: It is better to add assert.

native/cocos/renderer/pipeline/PipelineSceneData.cpp
star-e: Ah, yes. I see.

star-e (approved):

PR #11729, opened 2022-06-28
use IntrusivePtr to share object

minggo:
- _actor uses IntrusivePtr
- _cmdBuff, _queryPool and _queue defined in gfx::Device use IntrusivePtr
- make some `doDestroy()` to be reentrant

### DIFF
--- a/native/cocos/renderer/gfx-agent/BufferAgent.cpp
+++ b/native/cocos/renderer/gfx-agent/BufferAgent.cpp
@@ -43,7 +43,7 @@ BufferAgent::~BufferAgent() {
         BufferDestruct,
         actor, _actor,
         {
-            CC_SAFE_DELETE(actor);
+            actor = nullptr;
         });
 }
 
--- a/native/cocos/base/Agent.h
+++ b/native/cocos/base/Agent.h
@@ -27,6 +27,7 @@
 
 #include <memory>
 #include "Macros.h"
+#include "base/Ptr.h"
 
 namespace cc {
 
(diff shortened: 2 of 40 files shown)


native/cocos/renderer/gfx-agent/BufferAgent.cpp
```diff
@@ -43,7 +43,7 @@ BufferAgent::~BufferAgent() {
         BufferDestruct,
         actor, _actor,
         {
-            CC_SAFE_DELETE(actor);
+            actor = nullptr;
```
minggo, line 46: Indeed, can do nothing here. As `_actor` is an IntrusivePtr, it will be assigned to `actor` which will add reference count automatically.

minggo: Add to the message queue just extend the life cycle of `_actor`.

freeboy (approved):

star-e (approved):

native/cocos/renderer/gfx-agent/BufferAgent.cpp
star-e: Shall we add .reset() to IntrusivePtr? It will be the same as all other smart pointers, such as shared_ptr, unique_ptr

minggo: Yep, i think can add .reset(). I will create an issue for it. It has the same effect here.

hana-alice (approved):

yiwenxue (approved):

PR #11737, opened 2022-06-28
fix crash

minggo:


native/cocos/renderer/pipeline/shadow/ShadowFlow.cpp
```diff
@@ -201,7 +201,7 @@ void ShadowFlow::resizeShadowMap(const scene::Light *light, gfx::DescriptorSet *
 
     auto renderTargets = framebuffer->getColorTextures();
 
-    IntrusivePtr<gfx::Texture> texture = gfx::Device::getInstance()->createTexture({
+    auto *texture = gfx::Device::getInstance()->createTexture({
```
star-e, line 204: Here, I think it might be better to use smart ptr.

minggo: IntrusivePtr can convert from a raw pointer automatically. So i think just use raw pointer here is enough, then:
- don't create/destroy an Intrusive instance
- don't need to use .get() to get raw pointer later

star-e: If other maintainer did some change and the `texture` is not put into `_usedTextures`, it might cause memory leak.

minggo: The only situation that will cause memory leak is that;
- use raw pointer here and don't put into `_usedTextures`
- other place also use raw pointer and do not delete it

In this situation, if use IntrusivePtr, will cause crash as IntrusivePtr will delete the raw pointer.

minggo: Don't use IntrusivePtr here is because it is temple variable. The only place use IntrusivePtr here is:
- create a texture
- other places use the texture and do not keep it, such as assign to a class member variable

I don't think there is a situation that needs a temple texture.

star-e (approved):

native/cocos/renderer/pipeline/shadow/ShadowFlow.cpp
star-e: I see.

troublemaker52025 (approved):

PR #11762, opened 2022-06-29
[fix bug]1. crash when create mesh with _struct undefined, 2. when get property from material

timlyeee:
to fix: https://github.com/cocos/3d-tasks/issues/12716 https://github.com/cocos/3d-tasks/issues/12732 

* check if _struct is undefined or empty, both valid.
* fix get property error when no passIdx is given, use default method.
* fix std::variant to private object
* fix get property failed when try to get from Vec2 or Color etc.

no public api changed

### DIFF
--- a/cocos/3d/assets/mesh.jsb.ts
+++ b/cocos/3d/assets/mesh.jsb.ts
@@ -63,7 +63,7 @@ Object.defineProperty(meshAssetProto, 'struct', {
 });
 
 meshAssetProto.onLoaded = function () {
-    // might be undefine
+    // might be undefined
     if (this._struct != undefined) {
         this.setStruct(this._struct);
     }
--- a/cocos/3d/assets/mesh.jsb.ts
+++ b/cocos/3d/assets/mesh.jsb.ts
@@ -64,7 +64,7 @@ Object.defineProperty(meshAssetProto, 'struct', {
 
 meshAssetProto.onLoaded = function () {
     // might be undefine
-    if (this._struct) {
+    if (this._struct != undefined) {
         this.setStruct(this._struct);
     }
     // Set to null to release memory in JS
--- a/cocos/3d/assets/mesh.jsb.ts
+++ b/cocos/3d/assets/mesh.jsb.ts
@@ -63,7 +63,10 @@ Object.defineProperty(meshAssetProto, 'struct', {
 });
 
 meshAssetProto.onLoaded = function () {
-    this.setStruct(this._struct);
+    // might be undefine
+    if (this._struct) {
+        this.setStruct(this._struct);
+    }
     // Set to null to release memory in JS
     this._struct = null;
 };
--- a/cocos/core/assets/material.jsb.ts
+++ b/cocos/core/assets/material.jsb.ts
@@ -88,7 +88,17 @@ function wrapSetProperty (cb: setProperyCB, target: Material, name: string, val:
         cb.call(target, name, val);
     }
 }
-
+enum  MathType {
+    VEC2 = 0,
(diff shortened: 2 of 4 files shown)


dumganhar (approved):

minggo (approved):

timlyeee:
Add bug fix for issue https://github.com/cocos/3d-tasks/issues/12732 when entering to scene Physics

cocos/core/assets/material.jsb.ts
```diff
@@ -156,31 +166,36 @@ matProto.setProperty = function (name: string, val: MaterialPropertyFull | Mater
 };
 
 matProto.getProperty = function (name: string, passIdx?: number) {
-    const val = this._getProperty(name, passIdx);
+    let val: any;
+    if (passIdx) {
```
PatriceJiang: `passIdx !== undefined` ?

timlyeee: undefine or null both make it no sense, so I suppose that `if(passIdx)` is ok here

PatriceJiang: Should zero be treated as undefined?

cocos/core/assets/material.jsb.ts
```diff
@@ -88,7 +88,17 @@ function wrapSetProperty (cb: setProperyCB, target: Material, name: string, val:
         cb.call(target, name, val);
     }
 }
-
+enum  MathType {
```
dumganhar: Remove a space between enum and MathType.

minggo (approved):

native/cocos/bindings/manual/jsb_conversions_spec.cpp
```diff
@@ -133,6 +134,17 @@ enum class DataType {
     INT,
     FLOAT
 };
+enum class MathType {
```
dumganhar, line 136: Add a `NOTE` comment that MathType here should synchronize which in material.jsb.ts.

timlyeee: I guess the note should be added in material.jsb.ts? that's more logical

cocos/core/assets/material.jsb.ts
timlyeee: 1

cocos/core/assets/material.jsb.ts
timlyeee: passIdx might be 0, right about that

cocos/3d/assets/mesh.jsb.ts
```diff
@@ -63,7 +63,10 @@ Object.defineProperty(meshAssetProto, 'struct', {
 });
 
 meshAssetProto.onLoaded = function () {
-    this.setStruct(this._struct);
+    // might be undefine
```
dumganhar: undefine -> undefined

timlyeee: 1

cocos/core/assets/material.jsb.ts
```diff
@@ -156,31 +167,36 @@ matProto.setProperty = function (name: string, val: MaterialPropertyFull | Mater
 };
 
 matProto.getProperty = function (name: string, passIdx?: number) {
-    const val = this._getProperty(name, passIdx);
+    let val: any;
+    if (passIdx != undefined) {
```
PatriceJiang: Should use `!==` here

timlyeee: 1

PatriceJiang (approved):

Issue #11288, opened 2022-05-30
[ Feature ] Reduce the memory usage of GFXBuffers and mesh arrays.


timlyeee:
### Use Case

Right now the graphics memory usage is much higher than Unity, as we would draw precisely the same object but with a greater number of triangles, for example, a car with 500,000 triangles. 

The principal cause of it is that we use FLOAT and UNSIGNED_INT to store all data, which is not necessary for normal and UV, as well as a mesh with less than 65535 triangles.

### Problem Description

If we're concentrating on the code, for gles render backend, we shall see the codes to bind buffer is using the same type from the beginning to the end.
```cpp
for (auto &gpuInput : gpuPipelineState->gpuShader->glInputs) {
                    for (size_t a = 0; a < gpuInputAssembler->attributes.size(); ++a) {
                        const GLES3GPUAttribute &gpuAttribute = gpuInputAssembler->glAttribs[a];
                        if (gpuAttribute.name == gpuInput.name) {
                            GL_CHECK(glBindBuffer(GL_ARRAY_BUFFER, gpuAttribute.glBuffer));

                            for (uint32_t c = 0; c < gpuAttribute.componentCount; ++c) {
                                GLuint   glLoc        = gpuInput.glLoc + c;
                                uint32_t attribOffset = gpuAttribute.offset + gpuAttribute.size * c;
                                GL_CHECK(glEnableVertexAttribArray(glLoc));

                                cache->glEnabledAttribLocs[glLoc] = true;
                                GL_CHECK(glVertexAttribPointer(glLoc, gpuAttribute.count, gpuAttribute.glType, gpuAttribute.isNormalized, gpuAttribute.stride, BUFFER_OFFSET(attribOffset)));
                                GL_CHECK(glVertexAttribDivisor(glLoc, gpuAttribute.isInstanced ? 1 : 0));
                            }
                            break;
                        }
                    }
                }
```
The `gpuAttribute.glType` is hesitated.
And according to the editor's behavior, all meshes are using the same type of UINT even if we support drawing elements with USHORT.

### Alternatives Considered

Considering the whole workflow from the editor to the engine, if we can set mesh data type in the editor or we should save it intelligently?

timlyeee:
As it's mentioned before, right now I realize that it all depends on how JS shows the data in mesh.ts, which should be checked first

pandamicro:
@timlyeee It's because the mesh asset is using the specific format for those attributes, the engine itself won't change format settings.

[...]

PR #11775, opened 2022-06-30
[feature] 1. Add support for converting float to half from some vertex attributes at runtime; 2. Add Mesh.allowDataAccess property for releasing data after it is uploaded to GPU

dumganhar:
Re: #11288

### Changelog

* [feature] Add support for converting float to half from some vertex attributes at runtime, it is disabled by default. Developer could enable it by changing CC_OPTIMIZE_MESH_DATA to 1 in Mesh.cpp
* [feature] Add Mesh.allowDataAccess property for releasing data after it is uploaded to GPU

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


native/cocos/3d/assets/Mesh.cpp
```diff
@@ -1095,6 +1253,17 @@ bool Mesh::validate() const {
...
 
+void Mesh::setAllowReadWriteData(bool allowReadWriteData) {
+    _allowReadWriteData = allowReadWriteData;
+    if (_isMeshDataUploaded && _allowReadWriteData) {
+        releaseData();
+    }
+}
+
+void Mesh::releaseData() {
+    _data.clear();
```
stanleyljl, line 1273: Need to call shrink_to_fit？

native/cocos/3d/assets/Mesh.cpp
```diff
@@ -292,15 +294,26 @@ void Mesh::initialize() {
 
                 uint32_t dstStride = idxView.stride;
                 uint32_t dstSize = idxView.length;
-                if (dstStride == 4 && !gfxDevice->hasFeature(gfx::Feature::ELEMENT_INDEX_UINT)) {
+                if (dstStride == 4) {
                     uint32_t vertexCount = _struct.vertexBundles[prim.vertexBundelIndices[0]].view.count;
-                    if (vertexCount >= 65536) {
-                        debug::warnID(10001, vertexCount, 65536);
+#if CC_OPTIMIZE_MESH_DATA
+                    if (vertexCount < 65536) {
```
minggo, line 357: What happen if `vertexCount >= 65536`?

native/cocos/renderer/gfx-metal/MTLPipelineState.mm
```diff
@@ -230,7 +230,15 @@ of this software and associated engine source code (the "Software"), a limited,
                 break;
             }
         }
-        streamOffsets[inputAttrib.stream] += GFX_FORMAT_INFOS[(int)inputAttrib.format].size;
+
+        uint32_t attributeSize = GFX_FORMAT_INFOS[(int)inputAttrib.format].size;
+        // Metal requires 4 bytes alignment for attribute
+        uint32_t remain = attributeSize % 4;
+        if (remain != 0) {
+            attributeSize = attributeSize + remain;
```
hana-alice: is it `attributeSize = attributeSize + 4 - remain;` ?  or we have utils::alignTo(attributeSize, 4);

native/cocos/3d/assets/Mesh.cpp
dumganhar: No.
_data is cc::Uint8Array which is a reference of JS Uint8Array Object. Could refer to the new method `clear` in TypedArrayTemp template class.
```c++
    void clear() {
        if (_jsTypedArray != nullptr) {
            _jsTypedArray->unroot(); // Unrooted the JS object reference, so v8 gc could see the object and collect it 
            _jsTypedArray->decRef();
            _jsTypedArray = nullptr;
        }
        _buffer = nullptr;
        _byteLength = 0;
        _byteOffset = 0;
        _byteEndPos = 0;
    }
```

native/cocos/3d/assets/Mesh.cpp
dumganhar: When CC_OPTIOMIZE_MESH_DATA is enabled, if vertexCount >= 65536, there are two situations:
1. GPU support UINT element index type, then do nothing
2. GPU doesn't support UNIT element type, just ignore the primitive (mesh) to be uploaded to GPU.


if vertexCount is < 65536, convert index type to ushort.

When CC_OPTIOMIZE_MESH_DATA is disabled, if vertexCount is < 65536 and GPU doesn't support UINT element index type, converting index type to ushort, otherwise do nothing. This is the original logic.

native/cocos/3d/assets/Mesh.cpp
stanleyljl: Oh, It is not a std::vector

native/cocos/3d/assets/Mesh.cpp
```diff
@@ -1071,9 +1091,148 @@ void Mesh::accessAttribute(index_t primitiveIndex, const char *attributeName, co
...
+                if (attribute.format == gfx::Format::RGB32F) {
+                    attributeIndicsNeedConvert.emplace_back(attributeIndex);
+                    attribute.format = gfx::Format::RGB16F;
+    #if (CC_PLATFORM == CC_PLATFORM_IOS) || (CC_PLATFORM == CC_PLATFORM_MACOS)
+                    dstStride -= 4; // NOTE: Metal needs 4 bytes alignment
+    #else
+                    dstStride -= 6;
+    #endif
+                }
+            } else if (attribute.name == gfx::ATTR_NAME_TEX_COORD || attribute.name == gfx::ATTR_NAME_TEX_COORD1 || attribute.name == gfx::ATTR_NAME_TEX_COORD2 || attribute.name == gfx::ATTR_NAME_TEX_COORD3 || attribute.name == gfx::ATTR_NAME_TEX_COORD4 || attribute.name == gfx::ATTR_NAME_TEX_COORD5 || attribute.name == gfx::ATTR_NAME_TEX_COORD6 || attribute.name == gfx::ATTR_NAME_TEX_COORD7 || attribute.name == gfx::ATTR_NAME_TEX_COORD8) {
```
minggo: Use `switch case` instead? It is too long here.

native/cocos/3d/assets/Mesh.cpp
```diff
@@ -1071,9 +1091,148 @@ void Mesh::accessAttribute(index_t primitiveIndex, const char *attributeName, co
     }
 }
 
-/* static */
+void Mesh::updateVertexFormat() {
```
minggo: It is better to rename `convertVertexData`  and divide it into some small functions, such as:
- convertRGB32F_2_RGB16F
- ...

stanleyljl (approved):

native/cocos/3d/assets/Mesh.cpp
dumganhar: The name is `std::string` type, could not be use in switch case.

native/cocos/3d/assets/Mesh.cpp
```diff
@@ -356,6 +427,11 @@ void Mesh::initialize() {
         if (_struct.morph.has_value()) {
             morphRendering = createMorphRendering(this, gfxDevice);
         }
+
+        _isMeshDataUploaded = true;
+        if (!_allowReadWriteData) {
+            releaseData();
+        }
```
minggo, line 434: How about put these logic into `releaseData()`, like
```c++
releaseData() {
    if (!_isMeshDataUploaded || _allowReadWriteData) {
        return;
    }
    ...
}
```

minggo (approved):

native/cocos/3d/assets/Mesh.cpp
dumganhar: It could be.
But I like the current code more than that. Because while reading the code, I could simply know the logic (condition) before doing the real `releaseData` operation. E.g.

```c++
_isMeshDataUploaded = true;
  if (!_allowReadWriteData) { // This `if` code say that: I'm a mesh that doesn't allow read/write operation, so please releaseData now
      releaseData();
  }
```

```c++
_isMeshDataUploaded = true;
releaseData(); // Could data be released? may not! Because it depends on the logic inside releaseData. While reading code here without knowing the implementation of releaseData function, code reader may assume that it just does the actual `release` operation.
```

hana-alice (commented):
sorry for this late comment, reviewed but forgot to submit this review

native/cocos/3d/assets/Mesh.cpp
minggo: You can rename to `releaseDataIfPossible`

star-e (approved):

native/cocos/renderer/gfx-metal/MTLPipelineState.mm
dumganhar: It's `+ remain`  rather than `- remain`.

> or we have utils::alignTo(attributeSize, 4);

Yes, could use that utils function.

hana-alice (approved):

SantyWang (approved):

minggo:
Please fix conflict.

PR #11779, opened 2022-06-30
Support vec2 to float4 property set.

timlyeee:
* Fix the crash when trying to set property "intensitySize" with Vec2 yet conversion failed. The behavior is actually supported in Web by setting vec4.z. vec4.w to NaN

### DIFF
--- a/native/cocos/renderer/core/PassUtils.cpp
+++ b/native/cocos/renderer/core/PassUtils.cpp
@@ -191,6 +191,12 @@ const ccstd::unordered_map<gfx::Type, GFXTypeWriterCallback> type2writer = {
              a[idx + 1] = quat.y;
              a[idx + 2] = quat.z;
              a[idx + 3] = quat.w;
+         } else if (ccstd::holds_alternative<Vec2>(v)) {
+             const auto &vec2 = ccstd::get<Vec2>(v);
+             a[idx] = vec2.x;
+             a[idx + 1] = vec2.y;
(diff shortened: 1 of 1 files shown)


minggo (approved):

hana-alice:
do not change here, this case is under gfx::Type::FLOAT4, which means it shouldn't be VEC2, and vec2 has been handled somewhere else. 

this is a problem in example case:
https://github.com/cocos/cocos-example-projects/blob/34634d52d312c6e480a2c86844647764d3f16475/show-cases/assets/prefabs/light-hint/light-hint.ts#L18
make it pass vec4.

dumganhar (approved):

native/cocos/renderer/core/PassUtils.cpp
```diff
@@ -191,6 +191,12 @@ const ccstd::unordered_map<gfx::Type, GFXTypeWriterCallback> type2writer = {
              a[idx + 1] = quat.y;
              a[idx + 2] = quat.z;
              a[idx + 3] = quat.w;
+         } else if (ccstd::holds_alternative<Vec2>(v)) {
+             const auto &vec2 = ccstd::get<Vec2>(v);
+             a[idx] = vec2.x;
+             a[idx + 1] = vec2.y;
+             a[idx + 2] = 0.f;
+             a[idx + 3] = 0.f;
```
dumganhar: Coding style: 0.f -> 0.F

PR #11844, opened 2022-07-05
do not use align malloc

minggo:
c++17 supports create an aligned struct

### DIFF
--- a/native/cocos/renderer/gfx-agent/CommandBufferAgent.cpp
+++ b/native/cocos/renderer/gfx-agent/CommandBufferAgent.cpp
@@ -91,19 +91,14 @@ void CommandBufferAgent::initMessageQueue() {
     DeviceAgent *device = DeviceAgent::getInstance();
     device->_cmdBuffRefs.insert(this);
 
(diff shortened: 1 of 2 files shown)


star-e (approved):

PatriceJiang (approved):

yiwenxue:
https://github.com/cocos/cocos-engine/blob/eaddf1dab7cdef8a24eca7cc85d01267abd01a82/native/cocos/renderer/gfx-agent/CommandBufferAgent.cpp#L95
https://github.com/cocos/cocos-engine/blob/d1e141990bd28bab86597589c1c2672e057e265c/native/cocos/renderer/gfx-agent/DeviceAgent.cpp#L306-L307
they should be adapted as well?

yiwenxue (approved):

star-e:
Is c++17 aligned new supported on old iOS and Android？

PatriceJiang:
These two lines 
image
should be revert according to origin commit
image

minggo:
> Is c++17 aligned new supported on old iOS and Android？

Yep, i tested with iPhone 6s with iOS 11.2. And c++17 always works on Android.

PR #11925, opened 2022-07-11
Property handle zero duration animation clips(2nd fix)

shrinktofit:
Re: #11879

### Changelog

* In #11879 I proposed a PR to handle zero duration animation, but it doesn't seems correct since single frame clips(created by editor, having zero duration) can not be sampled. In this PR I'll try to strike this.

After this PR, given zero duration clip `C`, an animation state `S` of `C`, the following should always hold:

- As long as the `S`'s playing or `S.sample()` is called, the clip `C` will be sampled at time 0.

  > In #11879 this is not true!

- In anytime, `S.current === S.ratio === 0`.

- In anytime, if "wrap info" is inspected on `S`(from within other engine code), the wrap info's following properties keeps constant:

  - `ratio` - 0.0
  - `time` - 0.0
  - `iteration` - 1.0(which means the starting point of first iteration)
  - `direction` - 1.0(which means forwarding iteration)
  - `stopped` - `true` if `S.repeatCount` is finite, `false` otherwise. In other words, any single update invocation on `S` stops `S` no matter how long the update delta time, if `S` has finite repeat count(for example, non-loop wrap mode, manually set finite repeat count). In other case, `S` keeps loop forever.
    >  In #11879 `stopped` would be always `true`.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/core/animation/animation-state.ts
+++ b/cocos/core/animation/animation-state.ts
@@ -494,9 +494,6 @@ export class AnimationState extends Playable {
 
     public sample () {
         const info = this.getWrappedInfo(this.time, this._wrappedInfo);
-        if (this._playbackDuration === 0.0) {
-            return info;
-        }
         this._sampleCurves(info.time);
         if (!EDITOR || legacyCC.GAME_VIEW) {
             this._sampleEvents(info);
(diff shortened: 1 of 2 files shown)


shrinktofit:
@cocos-robot run test cases

cocos/core/animation/animation-state.ts
```diff
@@ -494,9 +494,6 @@ export class AnimationState extends Playable {
 
     public sample () {
         const info = this.getWrappedInfo(this.time, this._wrappedInfo);
-        if (this._playbackDuration === 0.0) {
```
shrinktofit, line 497: Added in #11879

cocos/core/animation/animation-state.ts
```diff
@@ -582,14 +579,17 @@ export class AnimationState extends Playable {
     private simpleProcess () {
         const playbackStart = this._playbackRange.min;
         const playbackDuration = this._playbackDuration;
-        if (playbackDuration === 0.0) {
```
shrinktofit, line 585: Added in #11879

tests/animation/animation-state.test.ts
```diff
@@ -28,18 +28,22 @@ describe('Animation state', () => {
         // - `time`  - 0.0
         // - `iteration` - 1.0(starting point of first iteration)
         // - `direction` - 1.0(forwarding iteration)
+        // - `stopped` - true if `.repeatCount` is finite false otherwise
         // Zero duration states won't do sample even there are curves in zero time.
         expectState(state);
         expectWrappedInfo(state.sample());
         blendStateBuffer.apply();
-        expect(node.position.x).toBeCloseTo(0.0);
+        expect(node.position.x).toBeCloseTo(2.0);
```
shrinktofit, line 36: It should be 2.0 as this PR mainly to solve with

SantyWang (approved):

PR #11997, opened 2022-07-14
simplify logic of Node::walk()

minggo:


pandamicro:
Related: https://github.com/cocos/cocos-engine/issues/11998

pandamicro (approved):

PR #12024, opened 2022-07-15
use ccstd::hash_combine to improve performance

minggo:
### Changelog

* add hash related files in CMakeLists.txt
* use ccstd::hash_combine to improve performance


minggo:
@cocos-robot run test cases

dumganhar (approved):

PR #12032, opened 2022-07-16
Implement a new file system

qiuguohua:
Design Features:
1. 'filesystem' contains a collection of all file systems;
2. The operation of the file system is performed through the 'filesystem' class, and all path operations are performed through FilePath;
3. Windows only has localFileSystem, while Apple and Android include resource file systems;

Unfinished part:
1. Unit test all interfaces;
2. The interface of many file systems is not perfect
3. The FileUtils class is not removed
4. Linux, nx, qnx, HarmonyOS systems are not adapted


native/cocos/base/ZipUtils.cpp
```diff
@@ -581,6 +581,39 @@ unsigned char *ZipFile::getFileData(const ccstd::string &fileName, uint32_t *siz
     return buffer;
 }
 
+bool ZipFile::getFileData(const ccstd::string &fileName, const char *buffer, size_t bufferSize) {
```
qiuguohua: The buffer here cannot be customized in size.If the buffer is too small, there should be problems with decompression

native/cocos/base/ZipUtils.cpp
```diff
@@ -581,6 +581,39 @@ unsigned char *ZipFile::getFileData(const ccstd::string &fileName, uint32_t *siz
...
+    bool res = false;
+    do {
+        auto zipFile = _data->zipFile.lock();
```
qiuguohua, line 601: Check if _data is empty

native/cocos/engine/Engine.cpp
```diff
@@ -104,7 +105,7 @@ Engine::~Engine() {
 
 int32_t Engine::init() {
     _scheduler = std::make_shared<Scheduler>();
-    _fs = createFileUtils();
+    _fs = new FileSystem();
```
qiuguohua: Use createFileUtils

native/cocos/platform/FileUtils.cpp
```diff
@@ -556,45 +567,40 @@ Data FileUtils::getDataFromFile(const ccstd::string &filename) {
...
     }
+    return FileUtils::Status::OK;
+}
 
-    return Status::OK;
+ccstd::string FileUtils::getWritablePath() const {
+    return FileSystem::getInstance()->getUserAppDataPath().value();
 }
 
 unsigned char *FileUtils::getFileDataFromZip(const ccstd::string &zipFilePath, const ccstd::string &filename, uint32_t *size) {
```
qiuguohua, line 620: The archive file system of other platforms is not implemented

native/cocos/platform/FileUtils.cpp
```diff
@@ -748,21 +752,12 @@ void FileUtils::setSearchPaths(const ccstd::vector<ccstd::string> &searchPaths)
...
-    if (!isAbsolutePath(searchpath)) {
-        prefix = _defaultResRootPath;
-    }
-
-    ccstd::string path = prefix + searchpath;
-    if (!path.empty() && path[path.length() - 1] != '/') {
-        path += "/";
-    }
     if (front) {
         _originalSearchPaths.insert(_originalSearchPaths.begin(), searchpath);
```
qiuguohua, line 783: These two paths are the same, is it possible to remove one?

native/cocos/platform/FileUtils.cpp
```diff
@@ -783,175 +778,57 @@ ccstd::string FileUtils::getFullPathForDirectoryAndFilename(const ccstd::string
...
-
-                if (tinydir_next(&dir) == -1) {
-                    // Error getting next file
-                    break;
-                }
-            }
-        }
-        tinydir_close(&dir);
-    }
+    FileSystem::getInstance()->listFilesRecursively(dirPath, &ret);
```
qiuguohua: Use files instead of temporary variables

native/cocos/platform/FileUtils.cpp
```diff
@@ -783,175 +778,57 @@ ccstd::string FileUtils::getFullPathForDirectoryAndFilename(const ccstd::string
...
-    return false;
+
+    return FileSystem::getInstance()->exist(dirPath);
 }
 
 bool FileUtils::createDirectory(const ccstd::string &path) {
-    // FileUtils not support createDirectory.
-    CC_ASSERT(false);
-    return false;
+    return FileSystem::getInstance()->exist(path);
```
qiuguohua: Error calling function

native/cocos/platform/android/AndroidFileSystem.cpp
```diff
@@ -0,0 +1,121 @@
...
+    return false;
+}
+
+
+bool AndroidFileSystem::renameFile(const FilePath& oldFilepath, const FilePath& newFilepath) {
+    return false;
+}
+
+std::unique_ptr<IFileHandle> AndroidFileSystem::open(const FilePath& filepath, AccessFlag flag) {
+    if (filepath.empty()) {
```
qiuguohua: Remove unused functions

native/cocos/platform/android/AndroidPlatform.cpp
```diff
@@ -456,7 +455,7 @@ void gameControllerStatusCallback(const int32_t controllerIndex,
 AndroidPlatform::~AndroidPlatform() = default;
 
 int AndroidPlatform::init() {
-    cc::FileUtilsAndroid::setassetmanager(_app->activity->assetManager);
+    cc::ResourceFileSystem::setassetmanager(_app->activity->assetManager);
```
qiuguohua, line 458: Is it possible to use filesystem operations?

native/cocos/platform/android/ResourceFileHandle.cpp
```diff
@@ -0,0 +1,79 @@
...
+ THE SOFTWARE.
+ ****************************************************************************/
+
+#include "cocos/platform/android/ResourceFileHandle.h"
+#include "base/Log.h"
+
+namespace cc {
+
+ResourceFileHandle::ResourceFileHandle(const FilePath& path, AAsset* asset)
+: _asset(asset) {
```
qiuguohua, line 33: path not used

native/cocos/platform/apple/ResourceFileSystem.mm
```diff
@@ -0,0 +1,262 @@
...
+    for (auto iter = dict.begin(); iter != dict.end(); ++iter) {
+        addCCValueToNSDictionary(iter->first, iter->second, nsDict);
+    }
+
+    NSString *file = [NSString stringWithUTF8String:fullPath.c_str()];
+    // do it atomically
+    return [nsDict writeToFile:file atomically:YES];
+}
+
+bool FileUtils::writeValueVectorToFile(const ValueVector &vecData, const ccstd::string &fullPath) {
```
qiuguohua, line 190: This module has not yet been adapted

native/cocos/platform/FileUtils.cpp
```diff
@@ -556,45 +567,40 @@ Data FileUtils::getDataFromFile(const ccstd::string &filename) {
...
-    FILE *fp = fopen(fs->getSuitableFOpen(fullPath).c_str(), "rb");
-    if (!fp) {
-        return Status::OPEN_FAILED;
+    FileSystem *filesystem = FileSystem::getInstance();
+    // read the file from hardware
+    FilePath fullPath = fullPathForFilename(filename);
+    std::unique_ptr<IFileHandle> fileHandle(
+        filesystem->open(fullPath, IFileSystem::AccessFlag::READ_ONLY));
+    if (!fileHandle) {
+        return FileUtils::Status::NOT_EXISTS;
```
qiuguohua: return FileUtils::Status::OPEN_FAILED

native/cocos/platform/filesystem/archive-filesystem/ZipFileSystem.h
```diff
@@ -0,0 +1,65 @@
...
+        return false;
+    }
+    bool removeDirectory(const FilePath& dirPath) override {
+        return false;
+    }
+    FilePath getUserAppDataPath() const override {
+        return FilePath("");
+    }
+private:
+    ZipFile* _zipFile;
```
qiuguohua: initialization value？

native/cocos/platform/filesystem/archive-filesystem/ZipFileSystem.h
```diff
@@ -0,0 +1,65 @@
...
+public:
+    ZipFileSystem(const FilePath& assetsPath);
+    ~ZipFileSystem() override;
+    bool exist(const FilePath& filepath) const override;
+    std::unique_ptr<IFileHandle> open(const FilePath& filepath, AccessFlag flag) override;
+
+    bool createDirectory(const FilePath& path) override {
+        return false;
+    }
+    int64_t getFileSize(const FilePath& filepath) override {
```
qiuguohua: Can use const？

native/cocos/base/ZipUtils.cpp
```diff
@@ -581,6 +594,39 @@ unsigned char *ZipFile::getFileData(const ccstd::string &fileName, uint32_t *siz
     return buffer;
 }
 
+bool ZipFile::getFileData(const ccstd::string &fileName, uint8_t *buffer, size_t bufferSize) {
+    bool res = false;
```
dumganhar, line 598: Why need to add this function?
`bool ZipFile::getFileData(const ccstd::string &fileName, ResizableBuffer *buffer)`  could not achieve what you want?
The implementations of these two functions are too similar.

qiuguohua: This is the most primitive interface. This interface is added because of the abstract class of IFileHandle.

Why do we need a basic interface for reading files
Because sometimes it is not necessary to read all the contents of the file at one time, because if the file is large, it will consume a lot of resources

native/cocos/platform/filesystem/FilePath.h
```diff
@@ -0,0 +1,67 @@
...
+
+    const ccstd::string& value() const { return _path; }
+    bool empty() const { return _path.empty(); };
+    void clear() { _path.clear(); }
+
+    FilePath baseName() const;
+    FilePath dirName() const;
+    void removeLastSeparator();
+
+    ccstd::string finalExtension(bool tolower = true) const;
```
minggo, line 55: What's `final extension`?

qiuguohua: The extension may contain multiple, linux-like compressed files
Example: 'package.tar.gz', 'final extension' refers to 'gz', not 'tar.gz'

native/cocos/platform/filesystem/IFileSystem.h
```diff
@@ -0,0 +1,89 @@
...
+
+namespace cc {
+class CC_DLL IFileSystem {
+public:
+    enum class AccessFlag {
+        READ_ONLY  = 0x0000,
+        WRITE_ONLY = 0x0001 << 0,
+        READ_WRITE = 0x0001 << 1,
+        APPEND     = 0x0001 << 2,
+    };
```
minggo, line 43: The flag takes effect for file, but define here is weird.

minggo: Define it in IFileHandle is more suitable.

native/cocos/platform/filesystem/IFileHandle.h
```diff
@@ -0,0 +1,53 @@
...
+namespace cc {
+
+class CC_DLL IFileHandle {
+public:
+    enum class MoveMethod{
+        FILE_SEEK_CUR = 0,
+        FILE_SEEK_END = 1,
+        FILE_SEEK_SET = 2,
+    };
+    IFileHandle() = default;
```
minggo, line 41: I think the constructor can be private, as it seems meaningless to new a IFileHandle in codes.

native/cocos/platform/filesystem/IFileSystem.h
qiuguohua: The behavior of opening a file is in the FileSystem, and it seems that it is not a good idea to put it in the IFileHandle. IFileHandle is only the operation of the file content, the behavior of opening the file does not occur in it

Maybe put it globally?

native/cocos/platform/filesystem/IFileHandle.h
qiuguohua: This is a pure virtual function and cannot be constructed.

native/cocos/platform/filesystem/IFileSystem.h
minggo: It is a state of a FileHandle object. So i think it is reasonable to put it in IFileHandle. Put it in global seems ok. And how about put all file system related codes into `fs` namespace?

native/cocos/platform/filesystem/IFileHandle.h
minggo: Yep, we should forbid its children too.

native/cocos/platform/filesystem/FilePath.h
minggo: Is it a common sense?

native/cocos/platform/filesystem/IFileHandle.h
minggo: Sorry, i mean its sub class.

native/cocos/platform/filesystem/FileSystem.h
```diff
@@ -0,0 +1,65 @@
...
+
+#pragma once
+
+#include <vector>
+#include "base/Macros.h"
+#include "cocos/platform/filesystem/IFileSystem.h"
+
+namespace cc {
+class LocalFileSystem;
+class CC_DLL FileSystem : public IFileSystem {
```
minggo, line 35: It is weird to have both iFileSystem and FileSystem.

native/cocos/platform/filesystem/IFileHandle.h
qiuguohua: Yes, derived classes are appropriate, but friend classes may be used, created in FileSystem

native/cocos/platform/filesystem/IFileSystem.h
qiuguohua: If it is opened in FileHandle, I think it is ok, but since the open operation will depend on FileSystem, I put it in FileSystem.

On the other hand, file opening, file moving, file deletion, and file naming are all operations on files. File read, file write, this is the operation of the file content

native/cocos/platform/filesystem/FilePath.h
qiuguohua: Most files have only one extension. May need to be obtained separately in the future

native/cocos/platform/filesystem/FileSystem.h
qiuguohua: The current design is that all file operations are operated through filesystem, so the initial assumption is that filesystem is also inherited from 'IFilesystem'.

If the virtual file system is designed, it will be removed here.

PR #12039, opened 2022-07-16
[v3.6.0] Optimize se::Object::getPrivateData, math::IsEqualF, Vec3::transformMat4

dumganhar:
-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


native/cocos/bindings/jswrapper/v8/Object.h
```diff
@@ -300,7 +300,7 @@ class Object final : public RefCounter {
      *  @return A void* that is the object's private data, if the object has private data, otherwise nullptr.
      */
     inline void *getPrivateData() const {
-        return _privateObject ? _privateObject->getRaw() : nullptr;
```
dumganhar, line 303: getRaw is a virtual function invocation, remove it.

native/cocos/math/Math.h
```diff
@@ -78,11 +74,11 @@ inline bool IsPowerOfTwo(T n) {
 }
 
 inline bool IsEqualF(float lhs, float rhs, float precision = 0.000001f) {
-    return (Abs<float>(lhs - rhs) < precision);
+    return (std::abs(lhs - rhs) < precision);
```
dumganhar, line 68: std::abs has better performance than what we implemented.

dumganhar:
@cocos-robot run test cases

SantyWang (approved):

native/cocos/math/Vec3.h
```diff
@@ -615,6 +615,10 @@ class CC_DLL Vec3 {
     static const Vec3 UNIT_Z;
     /** equals to Vec3(0,0,-1) */
     static const Vec3 FORWARD;
+
+private:
+    void transformMat4_c(const Vec3 &v, const Mat4 &m);
+    void transformMat4_neon(const Vec3 &v, const Mat4 &m);
```
PatriceJiang: These methods can be anonymous in Vec3.cpp?

PatriceJiang (approved):

native/cocos/math/Vec3.h
dumganhar: No, it's because `MathUtil::transformVec4` is private method and Vec3 is the friend class of MathUtil.

native/cocos/math/Vec3.cpp
```diff
@@ -27,6 +27,36 @@
...
+        #define USE_NEON64
+        #define INCLUDE_NEON64
+    #elif defined(__ARM_NEON__)
+        #define INCLUDE_NEON32
+    #endif
+#endif
+
+#if defined(USE_NEON64) || defined(USE_NEON32) || defined(INCLUDE_NEON32)
+    #include "enoki/array.h"
+    #ifndef ENOKI_ARM_NEON
```
minggo, line 50: Where defines this macro?

native/cocos/math/Vec3.cpp
```diff
@@ -27,6 +27,36 @@
...
+#elif (CC_PLATFORM == CC_PLATFORM_ANDROID)
+    #if defined(__arm64__) || defined(__aarch64__)
+        #define USE_NEON64
+        #define INCLUDE_NEON64
+    #elif defined(__ARM_NEON__)
+        #define INCLUDE_NEON32
+    #endif
+#endif
+
+#if defined(USE_NEON64) || defined(USE_NEON32) || defined(INCLUDE_NEON32)
```
minggo, line 48: Where defines INCLUDE_NEON32?

minggo, line 48: What's the difference between `USE_NEON32 ` and `INCLUDE_NEON32 `?

native/cocos/math/Vec3.cpp
```diff
@@ -192,15 +222,63 @@ void Vec3::transformMat3(const Vec3 &v, const Mat3 &m) {
...
+
+    row0 *= v.x;
+    row1 *= v.y;
+    row2 *= v.z;
+
+    row0 = row0 + row1 + row2 + row3;
+    enoki::store(tmpV0, row0);
+    float rhw;
+    // FIXME: use std::numeric_limits<float>::epsilon() instead ?
+    static constexpr float FLT_PRECISION = 0.000001F;
```
minggo: Can define it in Math.h? As other codes may need it too.

native/cocos/math/Vec3.cpp
```diff
@@ -192,15 +222,63 @@ void Vec3::transformMat3(const Vec3 &v, const Mat3 &m) {
...
+    if (CC_PREDICT_TRUE(std::abs(tmpV0[3]) > FLT_PRECISION)) {
```
minggo: Can add a `isGreaterThanF()`? As there is `isEqualF()`.

native/cocos/math/Vec3.cpp
dumganhar: ENOKI_ARM_NEON is defined in `enoki/fwd.h`, I checked it here for ensuring neon is enabled in enoki.

native/cocos/math/Vec3.cpp
dumganhar: It's the old logic in MathUtil.

When iOS 32 bits support is not removed, USE_NEON32 and INCLUDE_NEON32 is both enabled
On Android, only INCLUDE_NEON32 is enabled, USE_NEON32 is disabled.

native/cocos/math/Vec3.cpp
dumganhar: the magic number 0.000001F is from 

```
inline bool isEqualF(float lhs, float rhs, float precision = 0.000001F) {
    return (std::abs(lhs - rhs) < precision);
}

inline bool isNotEqualF(float lhs, float rhs, float precision = 0.000001F) {
    return (std::abs(lhs - rhs) > precision);
}
```

It's a little tricky, I don't wanna use this for checking float number, so don't define it in Math.h
In fact, it's better to use `std::numeric_limits<float>::epsilon()` for checking float value.

Here, I just keep the old logic and add a FIXME.

native/cocos/math/Vec3.cpp
dumganhar: For optimization reason.
It originally invokes `math::isNotEqualF`.
```c++
float rhw = math::isNotEqualF(tmp[3], 0.0F) ? 1 / tmp[3] : 1; // Pass extra number 0.0F to isNotEqualF, it's not needed.
```
I extend it here just because I don't wanna subtract `0` here.
```c++
inline bool isNotEqualF(float lhs, float rhs, float precision = 0.000001F) { // rhs is 0
    return (std::abs(lhs - rhs) > precision);
}
```

dumganhar: Yes, I think we need to add a new function called `isNotZeroF`

dumganhar: Add math::isNotZeroF, please review again.

minggo (approved):

PR #12069, opened 2022-07-19
Adjust node code struct

minggo:
- remove unused const member variables
  - Node::DESTROYING
  - Node::DEACTIVATING
  - Node::DONT_DESTROY
 - remove duplicated `public`
 - merge `protected` to `private`
 - declare static variables and functions first
 - declare virtual functions before other functions


native/cocos/core/scene-graph/Node.h
```diff
@@ -63,19 +63,14 @@ class Node : public BaseNode {
     public:
         ~UserData() override = default;
 
-    protected:
```
dumganhar: Why remove protected ?
We keep UserData's constructor protected since we don't wanna developer uses `new UserData()`. Instead, developer should inherit UserData. `new UserData()` is meaningless

minggo: I will revert it. Remove it to do debug.

dumganhar (approved):

PR #12118, opened 2022-07-21
[V3.6.0 gfx] auto barrier added

star-e:
Added gfx auto barrier.

Although this pr changed gfx implementation, it has no impact on current pipeline.

Passed all tests.

### Changelog

* enhanced gfx barrier.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


star-e:
@cocos-robot run test cases

native/cocos/renderer/gfx-base/GFXDef-common.h
```diff
@@ -1289,7 +1309,13 @@ using SubpassInfoList = ccstd::vector<SubpassInfo>;
...
     uint32_t srcSubpass{0};
     uint32_t dstSubpass{0};
-    GeneralBarrier *barrier{nullptr};
+    GeneralBarrier *generalBarrier{nullptr};
+    BufferBarrier **bufferBarriers{nullptr};
+    Buffer **buffers{nullptr};
+    uint32_t bufferBarrierCount{0};
+    TextureBarrier **textureBarriers{nullptr};
+    Texture **textures{nullptr};
+    uint32_t textureBarrierCount{0};
```
minggo, line 1318: Can declare pointers first, then uint32_t types? It will reduce padding.

native/cocos/renderer/gfx-base/GFXDevice.h
```diff
@@ -118,6 +118,9 @@ class CC_DLL Device : public RefCounted {
     template <typename ExecuteMethod>
     void registerOnAcquireCallback(ExecuteMethod &&execute);
 
+    inline void setOptions(const DeviceOptions &opts) { _options = opts; }
+    inline DeviceOptions getOptions() const { return _options; }
```
minggo: Can return a const reference?

native/cocos/renderer/gfx-base/GFXDef-common.h
star-e: In this particular case, it doesn't matter.
SubpassDependency is 64byte, fits in 64byte cache line.  So it is better setting `alignas` to 64 in the future.
If we reorder the uint32_t, the size is reduced to 56 byte, it might cause more cache misses (across two 64-byte cache line).
Another reason is in typescript, it is more error prone for the two consecutive number in constructor. (which is generated from GFXDef-common.h). Some linter might report this warning.

native/cocos/renderer/gfx-base/GFXDevice.h
star-e: Changed to const reference.

star-e:
@cocos-robot

native/cocos/renderer/gfx-base/GFXDef-common.h
minggo: > If we reorder the uint32_t, the size is reduced to 56 byte, it might cause more cache misses (across two 64-byte cache line).

Good to know. Is there any document about it?

star-e: What I have said previously is not very accurate.
SubpassDependency is typically store in a ccstd::vector, whose memory is continuous.
If SubpassDependency is 56 byte, there will be some alignment mismatch, and most elements will be across two 64 cache-line.
If SubpassDependency is `alignas(64)`, it is then aligned with 64 cache-line. It will reduce cache usage.

https://stackoverflow.com/questions/14707803/line-size-of-l1-and-l2-caches

pandamicro (approved):

native/cocos/renderer/gfx-base/GFXDef-common.h
minggo: Yep, if align with 64 bytes, then can reduce cache miss. But both align with 64 bytes or not, use 56 bytes will not have more cache miss i think.

star-e: The problem is called false sharing. https://en.wikipedia.org/wiki/False_sharing

minggo: Yep. If it is not aligned with cache line size, then a SubpassDependency instance may be loaded into two cache lines.

PR #12126, opened 2022-07-21
Fix bug when JsbBridge test case close window

timlyeee:
* engine-close will not dispatchCloseEvent as it's already dispatch otherwhere
* jsbBridgeWrapper will now release JsbBridgeWrapper sharedInstance rather than jsbBridge


native/cocos/engine/Engine.cpp
```diff
@@ -222,7 +222,6 @@ void Engine::close() { // NOLINT
     cc::DeferredReleasePool::clear();
     _scheduler->removeAllFunctionsToBePerformedInCocosThread();
     _scheduler->unscheduleAll();
-    cc::EventDispatcher::dispatchCloseEvent();
```
minggo, line 225: Why remove it?

qiuguohua (approved):

minggo (approved):

PR #12130, opened 2022-07-21
add protect condition of kerningDict and fontDefDictionary are undefined

Zhmz:
Re: [#13071](https://github.com/cocos/3d-tasks/issues/13071)

### Changelog

* kerningDict is undefined caused crash in _computeHorizontalKerningForText
* fontDefDictionary is undefined caused crash in _getFirstWordLen
* create a new fontAtlas when the parameter is null 
* newly created label-atlas which doesn't have its spriteFrame is dragged to scene, and the data in fontConfig are not initialized. This is the reason of these 2 point above.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/2d/assembler/label/bmfontUtils.ts
+++ b/cocos/2d/assembler/label/bmfontUtils.ts
@@ -163,7 +163,7 @@ export const bmfontUtils = {
         shareLabelInfo.fontAtlas = fontAsset.fontDefDictionary;
         if (!shareLabelInfo.fontAtlas) {
             if (comp.cacheMode === CacheMode.CHAR) {
-                shareLabelInfo.fontAtlas = new LetterAtlas(1024, 1024);
+                shareLabelInfo.fontAtlas = new LetterAtlas(64, 64);
             } else {
                 shareLabelInfo.fontAtlas = new FontAtlas(null);
             }
--- a/cocos/2d/assembler/label/bmfontUtils.ts
+++ b/cocos/2d/assembler/label/bmfontUtils.ts
@@ -24,15 +24,14 @@
 */
 
 import { JSB } from 'internal:constants';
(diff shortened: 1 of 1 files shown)


pandamicro (approved):

cocos/2d/assembler/label/bmfontUtils.ts
```diff
@@ -161,6 +161,13 @@ export const bmfontUtils = {
         _spriteFrame = fontAsset.spriteFrame;
         _fntConfig = fontAsset.fntConfig;
         shareLabelInfo.fontAtlas = fontAsset.fontDefDictionary;
+        if (!shareLabelInfo.fontAtlas) {
+            if (comp.cacheMode === CacheMode.CHAR) {
+                shareLabelInfo.fontAtlas = new LetterAtlas(1024, 1024);
```
pandamicro: It feels strange, if fontAtlas is missing and if it's an unexpected scenario, then maybe we should just skip rendering.
We should be always vigilant about memory creation, especially on GPU side, please justify the modification with fine cause

pandamicro (changes requested):

cocos/2d/assembler/label/bmfontUtils.ts
Zhmz: image
There are many usage of fontAtlas, we should create a default fontAtlas unless we make null judgment protection in each of the above places

Zhmz: I create two global variables for assignment. It can avoid memory allocations.

SantyWang (approved):

pandamicro (approved):

cocos/2d/assembler/label/bmfontUtils.ts
pandamicro: Ok, 64*64 seems much smaller and acceptable

PR #12131, opened 2022-07-21
add callbacksInvoker missing api

timlyeee:
re: #https://github.com/cocos/3d-tasks/issues/12875


pandamicro (changes requested):
This is not how it should be fixed, please find out which end class are missing  the APIs and try to fix it in end class level. CallbacksInvoker doesn't have once and targetOff, the implementations are also meaningless.
Be more rigorous

pandamicro (approved):

dumganhar (approved):

minggo (approved):

PR #12159, opened 2022-07-22
[3.6.0] Support device vs design aspect ratio in Splash Screen

mikecoker:
Currently the Splash Screen in the engine uses a hard coded displayRatio setting in the build configuration and I guess that's trying to fill the role of what I actually need. I need this value to be dynamic based on the device that's running the game.

Mobile web devices come in various aspect ratios that may or may not match the design ratio. We need to compare the design aspect ratio to the device aspect ratio and use that as a multiplier on the splash screen to ensure it fits. Previously iPhone had 1.77 aspect ratio but newer devices seem to be 2.16. 

Typically designers would be designing for a 1.77 aspect ratio, so we would need a ratio modifier based on the device it runs on. It would be 1.0 for older devices and 1.22 for newer. But rather than hard code these values, I want to use the aspect ratio of the device and the aspect ratio of the design resolution to calculate it dynamically.
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/core/splash-screen.ts
+++ b/cocos/core/splash-screen.ts
@@ -173,13 +173,17 @@ export class SplashScreen {
             device.capabilities.clipSpaceSignY, swapchain.surfaceTransform);
         const dw = swapchain.width; const dh = swapchain.height;
         const refW = dw < dh ? dw : dh;
+        const deviceAspect = dw / dh;
+        const designRes = window._CCSettings.designResolution;
+        const designtAspect = designRes.width / designRes.height;
+        const ratio = deviceAspect / designtAspect;
         // update logo uniform
         this._curTime += deltaTime * 1000;
         const percent = clamp01(this._curTime / settings.totalTime);
         let u_p = easing.cubicOut(percent);
         if (settings.effect === 'NONE') u_p = 1.0;
         const logoTW = this.logoTexture.width; const logoTH = this.logoTexture.height;
-        const logoW = refW * settings.displayRatio;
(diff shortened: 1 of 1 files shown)


cocos/core/splash-screen.ts
```diff
@@ -173,13 +173,17 @@ export class SplashScreen {
             device.capabilities.clipSpaceSignY, swapchain.surfaceTransform);
         const dw = swapchain.width; const dh = swapchain.height;
         const refW = dw < dh ? dw : dh;
+        const deviceAspect = dw / dh;
+        const designRes = window._CCSettings.designResolution;
```
SantyWang: It not working now, please see `settings.querySettings`

PPpro (approved):

cocos/core/splash-screen.ts
Zhmz: consider using view.getDesignResolutionSize() to get design resolution?
or screen.windowSize.

mikecoker: ok, I ended up adding the design resolution to the settings using the `settings.querySettings` call

pandamicro:
Hi, @mikecoker we are about to submit last round QA for v3.6, and new features won't be accepted in this version, but it will be taken into v3.6.1

pandamicro (approved):

SantyWang:
Thanks for your contribution,  we have refactor the splash screen in v3.7. And I think this PR is unnecessary. Feel free if you rebased this to v3.7

PR #12165, opened 2022-07-23
fix wrong transfrom calc when scale value is negative

novain1102:
Re: #https://github.com/cocos/3d-tasks/issues/12675

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


dumganhar:
Please rebase your changes.

novain1102:
wecom-temp-d4eeb806d57b805ba9138155aedd2598
This change will cause the problem that when setting Position & eulerAngles or Position & Scale run well, but when setting eulerAngles and Scale at one time will calculate wrong transform data. 
The reason maybe we cannot calculate rotation and scale separately from matrix transform.

Still need to figure out the right way to calculate.

PR #12184, opened 2022-07-25
fix ui Model proxy memory leak 

LinYunMo:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/2d/components/graphics.ts
+++ b/cocos/2d/components/graphics.ts
@@ -751,7 +751,7 @@ export class Graphics extends UIRenderer {
                     for (let i = 0; i < renderDataList.length; i++) {
                         renderDataList[i].setRenderDrawInfoAttributes();
                     }
-                    this._graphicsNativeProxy.activeSubModel();
+                    this._graphicsNativeProxy.activeSubModels();
                 }
                 this._graphicsNativeProxy.uploadData();
                 this._isNeedUploadData = false;
--- a/cocos/2d/components/graphics.ts
+++ b/cocos/2d/components/graphics.ts
@@ -746,12 +746,7 @@ export class Graphics extends UIRenderer {
                     for (let i = 0; i < renderDataList.length; i++) {
                         renderDataList[i].setRenderDrawInfoAttributes();
                     }
-                    const len = this._graphicsNativeProxy.getModel().subModels.length;
-                    if (renderDataList.length > len) {
-                        for (let i = len; i < renderDataList.length; i++) {
-                            this._graphicsNativeProxy.activeSubModel(i);
-                        }
-                    }
+                    this._graphicsNativeProxy.activeSubModel();
                 }
                 this._graphicsNativeProxy.uploadData();
                 this._isNeedUploadData = false;
--- a/cocos/2d/components/graphics.ts
+++ b/cocos/2d/components/graphics.ts
@@ -220,6 +220,9 @@ export class Graphics extends UIRenderer {
     public static LineJoin = LineJoin;
     public static LineCap = LineCap;
     public impl: Impl | null = null;
+    /**
+     * @deprecated since v3.6.0, this is an engine private interface that will be removed in the future.
+     */
     public model: scene.Model | null = null;
     @serializable
     protected _lineWidth = 1;
--- a/cocos/2d/components/graphics.ts
+++ b/cocos/2d/components/graphics.ts
@@ -267,6 +267,7 @@ export class Graphics extends UIRenderer {
         super.onLoad();
         if (JSB) {
             this._graphicsNativeProxy.initModel(this.node);
+            this.model = this._graphicsNativeProxy.getModel();
         } else {
             this.model = director.root!.createModel(scene.Model);
             this.model.node = this.model.transform = this.node;
@@ -292,6 +293,7 @@ export class Graphics extends UIRenderer {
         this._sceneGetter = null;
         if (JSB) {
             this.graphicsNativeProxy.destroy();
+            this.model = null;
         } else {
             if (this.model) {
                 director.root!.destroyModel(this.model);
--- a/cocos/2d/components/graphics.ts
+++ b/cocos/2d/components/graphics.ts
@@ -262,12 +262,13 @@ export class Graphics extends UIRenderer {
 
     public onLoad () {
         super.onLoad();
-        this.model = director.root!.createModel(scene.Model);
-        this.model.node = this.model.transform = this.node;
-        this._flushAssembler();
         if (JSB) {
             this._graphicsNativeProxy.initModel(this.node);
+        } else {
+            this.model = director.root!.createModel(scene.Model);
+            this.model.node = this.model.transform = this.node;
         }
+        this._flushAssembler();
     }
 
     // hack for mask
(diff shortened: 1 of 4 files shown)


native/cocos/2d/renderer/UIModelProxy.cpp
```diff
@@ -111,8 +110,14 @@ void UIModelProxy::uploadData() {
 void UIModelProxy::destroy() {
     if (_model != nullptr) {
         Root::getInstance()->destroyModel(_model);
-        _model = nullptr;
+        CC_SAFE_DELETE(_model);
```
dumganhar: `_model = nullptr` is enough.
Don't call `delete` for `IntrusivePtr`.

cocos/2d/components/graphics.ts
```diff
@@ -738,7 +746,7 @@ export class Graphics extends UIRenderer {
                     for (let i = 0; i < renderDataList.length; i++) {
                         renderDataList[i].setRenderDrawInfoAttributes();
                     }
-                    const len = this.model!.subModels.length;
+                    const len = this._graphicsNativeProxy.getModel().subModels.length;
```
dumganhar: Visiting `subModels` property of `Model` object really cost a lot.
In Model::getSubModels, it returns `const std::vector<cc::IntrusivePtr<cc::scene::SubModel>>&` , it will be converted to a new js array (which is a GC object ) when visiting the getter.
```c++
static bool js_scene_Model_getSubModels(se::State& s) // NOLINT(readability-identifier-naming)
{
    auto* cobj = SE_THIS_OBJECT<cc::scene::Model>(s);
    // SE_PRECONDITION2(cobj, false, "Invalid Native Object");
    if (nullptr == cobj) return true;
    const auto& args = s.args();
    size_t argc = args.size();
    CC_UNUSED bool ok = true;
    if (argc == 0) {
        const std::vector<cc::IntrusivePtr<cc::scene::SubModel>>& result = cobj->getSubModels();
        ok &= nativevalue_to_se(result, s.rval(), nullptr /*ctx*/);
        SE_PRECONDITION2(ok, false, "Error processing arguments");
        SE_HOLD_RETURN_VALUE(result, s.thisObject(), s.rval());
        return true;
    }
    SE_REPORT_ERROR("wrong number of arguments: %d, was expecting %d", (int)argc, 0);
    return false;
}
SE_BIND_FUNC_AS_PROP_GET(js_scene_Model_getSubModels)

template <typename T>
inline bool nativevalue_to_se(const ccstd::vector<T> &from, se::Value &to, se::Object *ctx) { // NOLINT(readability-identifier-naming)
    se::HandleObject array(se::Object::createArrayObject(from.size()));
    se::Value tmp;
    for (size_t i = 0; i < from.size(); i++) {
        // If from[i] is on stack, then should create a new object, or
        // JS will hold a freed object.
        if CC_CONSTEXPR (!std::is_pointer<T>::value && is_jsb_object_v<T>) {
            auto *pFrom = ccnew T(from[i]);
            nativevalue_to_se(pFrom, tmp, ctx);
            tmp.toObject()->clearPrivateData();
            tmp.toObject()->setPrivateData(pFrom);
        } else {
            nativevalue_to_se(from[i], tmp, ctx);
        }

        array->setArrayElement(static_cast<uint32_t>(i), tmp);
    }
    to.setObject(array, true);
    return true;
}
```
So, if .subModels is visited frequently, I suggest don't invoke it in JS.

dumganhar: Could it be done in CPP with the following `activeSubModel` & `updateData` ?

native/cocos/2d/renderer/UIModelProxy.cpp
```diff
@@ -33,7 +33,6 @@ UIModelProxy::UIModelProxy() {
 }
 
 UIModelProxy::~UIModelProxy() {
-    destroy();
```
dumganhar: Why delete this?

native/cocos/2d/renderer/UIModelProxy.cpp
```diff
@@ -111,8 +110,14 @@ void UIModelProxy::uploadData() {
 void UIModelProxy::destroy() {
     if (_model != nullptr) {
         Root::getInstance()->destroyModel(_model);
-        _model = nullptr;
+        CC_SAFE_DELETE(_model);
     }
+
+    for (auto subMesh: _graphicsUseSubMeshes) {
```
dumganhar: `auto` -> `auto&` for IntrusivePtr

native/cocos/2d/renderer/UIModelProxy.cpp
```diff
@@ -111,8 +110,14 @@ void UIModelProxy::uploadData() {
...
+        subMesh->destroy();
+        CC_SAFE_DELETE(subMesh);
```
dumganhar: Don't need to invoke delete for IntrusivePtr

cocos/2d/components/graphics.ts
```diff
@@ -262,12 +262,13 @@ export class Graphics extends UIRenderer {
 
     public onLoad () {
         super.onLoad();
-        this.model = director.root!.createModel(scene.Model);
```
dumganhar, line 265: `this.model` is a public property, will it be accessed in other modules?
If it does, it will be `undefined` on JSB which will break the compatibility.

fish9167:
@cocos-robot run test cases

fish9167:
@cocos-robot run test cases

cocos/2d/components/graphics.ts
```diff
@@ -738,12 +746,7 @@ export class Graphics extends UIRenderer {
                     for (let i = 0; i < renderDataList.length; i++) {
                         renderDataList[i].setRenderDrawInfoAttributes();
                     }
-                    const len = this.model!.subModels.length;
-                    if (renderDataList.length > len) {
-                        for (let i = len; i < renderDataList.length; i++) {
-                            this._graphicsNativeProxy.activeSubModel(i);
-                        }
-                    }
+                    this._graphicsNativeProxy.activeSubModel();
```
dumganhar: activeSubModel -> activeSubModels ?

native/cocos/2d/renderer/UIModelProxy.cpp
```diff
@@ -43,35 +42,41 @@ void UIModelProxy::initModel(Node* node) {
...
-    if (_model->getSubModels().size() <= val) {
-        auto* entity = static_cast<RenderEntity*>(_node->getUserData());
-        RenderDrawInfo* drawInfo = entity->getDynamicRenderDrawInfo(val);
-        if (drawInfo == nullptr) {
-            return;
+    auto* entity = static_cast<RenderEntity*>(_node->getUserData());
+    auto drawInfoSize = entity->getDynamicRenderDrawInfos().size();
+    auto subModelSize = _model->getSubModels().size();
+    if (drawInfoSize > subModelSize) {
+        for (uint8_t i = subModelSize; i < drawInfoSize; i++) {
```
dumganhar: `uint8_t i` -> `size_t i`

dumganhar (approved):

SantyWang (approved):

PR #12190, opened 2022-07-26
Fixed the problem that the script callback function could not be registered in game.cpp

qiuguohua:
1. Add engine start event
2. When the engine is created, create a script engine at the same time

### DIFF
--- a/native/cocos/engine/Engine.cpp
+++ b/native/cocos/engine/Engine.cpp
@@ -96,7 +96,9 @@ bool setCanvasCallback(se::Object * /*global*/) {
 
 namespace cc {
 
-Engine::Engine() = default;
+Engine::Engine() {
+    _scriptEngine = ccnew se::ScriptEngine();
+}
 
 Engine::~Engine() {
     destroy();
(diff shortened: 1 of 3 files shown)


native/cocos/engine/Engine.cpp
```diff
@@ -117,8 +119,9 @@ int32_t Engine::init() {
 #if CC_USE_PROFILER
     _profiler = ccnew Profiler();
 #endif
-
-    _scriptEngine = ccnew se::ScriptEngine();
+    if (!_scriptEngine) {
+        _scriptEngine = ccnew se::ScriptEngine();
+    }
```
minggo: Please add comments why need to do it like this.

qiuguohua: OK

minggo: > The script engine is created in the engine constructor for the first time.

Please add a reason for it.

dumganhar: Could users's custom JSB registeration in Game.cpp be invoked again while restarting?

dumganhar (approved):

minggo (approved):

PR #12290, opened 2022-08-01
Fix the problem that the mac cannot set the frame rate

qiuguohua:


minggo:
It is not an urgent issue, please send to develop.

PR #12318, opened 2022-08-03
bind missing functions

minggo:


native/tools/tojs/geometry.ini
```diff
@@ -43,7 +43,7 @@ classes = AABB Obb Line Plane Frustum Capsule Sphere Triangle Ray Spline ShapeBa
 # functions from all classes.
 skip = AABB::[getBoundary aabbAabb aabbFrustum aabbPlan merge transform transformExtentM4 isValid setValid set fromPoints (g|s)etCenter (g|s)etHalfExtents],
        Frustum::[update type planes split],
-       Plane::[clone copy normalize getSpotAngle fromNormalAndPoint fromPoints set],
+       Plane::[getSpotAngle],
```
dumganhar, line 46: The generated code of `copy` by bindings-generator is not correct since there is a `out` argument which  bindings-generator could not handle. So bind `copy` is useless. 
Probably, we need to implement it in TS.

dumganhar: The same as `normalize`, `fromNormalAndPoint `, `fromPoints `, `set`.

dumganhar (approved):

PR #12319, opened 2022-08-03
add Node::getPathInHierarchy

novain1102:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/core/scene-graph/Node.cpp
+++ b/native/cocos/core/scene-graph/Node.cpp
@@ -889,7 +889,7 @@ void Node::clearNodeArray() {
 ccstd::string Node::getPathInHierarchy() const {
     ccstd::string result = getName();
     Node *curNode = getParent();
-    while (curNode) {
+    while (curNode && curNode->getParent()) { 
         result.insert(0, "/").insert(0, curNode->getName());
         curNode = curNode->getParent();
     }
--- a/native/cocos/core/scene-graph/Node.cpp
+++ b/native/cocos/core/scene-graph/Node.cpp
@@ -889,7 +889,7 @@ void Node::clearNodeArray() {
 ccstd::string Node::getPathInHierarchy() const {
     ccstd::string result = getName();
     Node *curNode = getParent();
-    while (curNode && curNode->getParent()) {
+    while (curNode) {
         result.insert(0, "/").insert(0, curNode->getName());
         curNode = curNode->getParent();
     }
--- a/native/cocos/core/scene-graph/Node.cpp
+++ b/native/cocos/core/scene-graph/Node.cpp
@@ -886,6 +886,16 @@ void Node::clearNodeArray() {
     }
 }
 
+ccstd::string Node::getPathInHierarchy() const {
+    ccstd::string result = getName();
+    Node *curNode = getParent();
+    while (curNode && dynamic_cast<Scene *>(curNode) == nullptr) {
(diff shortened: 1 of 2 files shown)


native/cocos/core/scene-graph/Node.cpp
```diff
@@ -886,6 +886,16 @@ void Node::clearNodeArray() {
     }
 }
 
+ccstd::string Node::getPathInHierarchy() const {
+    ccstd::string result = getName();
+    Node *curNode = getParent();
+    while (curNode && dynamic_cast<Scene *>(curNode) == nullptr) {
```
novain1102: In TS, curNode  is `BaseNode` Type ;`Scene` and `Node` is derived class of `BaseNode`

```ts
        let curNode: BaseNode | null = this.parent;
        while (curNode && curNode instanceof Node) {
           ...
        }
```

minggo: dynamic_cast<Scene *>(curNode) == nullptr -> dynamic_cast<Scene *>(curNode) != nullptr

novain1102: As I can see, the original idea in TS is to judge that curNode is not a Scene @Zhmz

native/cocos/core/scene-graph/Node.cpp
```diff
@@ -886,6 +886,16 @@ void Node::clearNodeArray() {
...
+        result = ccstd::string(curNode->getName()).append("/").append(result);
```
minggo: can reuse result, just insert contents ahead of it, such as
```c++
result.insert(0, "/").insert(0, curNode->getName());
```

novain1102: It's a good idea!

native/cocos/core/scene-graph/Node.cpp
minggo: If so, then its parent is null, don't have to check type. Please double check the meaning of it.

Zhmz: using "parent == nullptr" can exactly determine that the curNode is a scene type.
All comman nodes have their own parent but scenes don't have.

Zhmz (changes requested):

native/cocos/core/scene-graph/Node.cpp
novain1102: > If so, then its parent is null, don't have to check type. Please double check the meaning of it.

Yes. You're right. Wait for @Zhmz to make sure.

native/cocos/core/scene-graph/Node.cpp
```diff
@@ -886,6 +886,16 @@ void Node::clearNodeArray() {
     }
 }
 
+ccstd::string Node::getPathInHierarchy() const {
+    ccstd::string result = getName();
+    Node *curNode = getParent();
+    while (curNode && curNode->getParent()) {
```
minggo: Don't have to check its parent. Current logic will exclude Scene.

minggo: Don't have to check its parent. Current logic will exclude Scene.

minggo (approved):

native/cocos/core/scene-graph/Node.cpp
minggo: Sorry, it is different from TS. So should keep it.

PR #12326, opened 2022-08-04
Wgpu 3.6.0

hana-alice:
Re: # free to try https://github.com/cocos/3d-tasks/issues/13520

everything up-to-date: emsdk 3.1.17, chromium 105.

- [ ] sth wrong in this test case:`testlist`, try to figure out how to deal with copyWebPngToNativeTexture.
- [ ] mem leak either in wgpu native or wasm, working on it.


hana-alice:
those changes like in `texture-2d.ts` should be removed, they  were used to work with c++ native object export by emscripten, it's out-dated.

PR #12369, opened 2022-08-09
Fix pipeline dangling ptr

yiwenxue:
Re: #https://github.com/cocos/3d-tasks/issues/13069, https://github.com/cocos/3d-tasks/issues/13340#issue-1316141034

### Changelog

* use intrusive ptr to fix this problem completely

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/renderer/pipeline/RenderFlow.cpp
+++ b/native/cocos/renderer/pipeline/RenderFlow.cpp
@@ -51,24 +51,24 @@ void RenderFlow::activate(RenderPipeline *pipeline) {
         return s1->getPriority() < s2->getPriority();
     });
 
-    for (auto *const stage : _stages) {
+    for (auto const &stage : _stages) {
         stage->activate(pipeline, this);
     }
 }
 
 void RenderFlow::render(scene::Camera *camera) {
-    for (auto *const stage : _stages) {
+    for (auto const &stage : _stages) {
         stage->render(camera);
     }
 }
 
 void RenderFlow::destroy() {
     if (_isResourceOwner) {
-        for (auto *stage : _stages) {
+        for (auto &stage : _stages) {
             CC_SAFE_DESTROY_AND_DELETE(stage);
         }
     } else {
-        for (auto *stage : _stages) {
+        for (auto &stage : _stages) {
             CC_SAFE_DESTROY(stage);
         }
     }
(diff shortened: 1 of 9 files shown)


star-e (approved):

bluesky013 (approved):

native/cocos/renderer/pipeline/RenderFlow.cpp
```diff
@@ -51,33 +51,27 @@ void RenderFlow::activate(RenderPipeline *pipeline) {
...
-        for (auto *stage : _stages) {
-            CC_SAFE_DESTROY_AND_DELETE(stage);
-        }
-    } else {
-        for (auto *stage : _stages) {
-            CC_SAFE_DESTROY(stage);
-        }
+    for (auto &stage : _stages) {
+        CC_SAFE_DESTROY(stage);
     }
```
minggo, line 68: In C++, i think it is better to invoke destroy in destructor, then don't need to destroy here. Even better is that, don't need destroy function.

native/cocos/renderer/pipeline/RenderPipeline.cpp
```diff
@@ -150,14 +149,8 @@ void RenderPipeline::updateGeometryRenderer(const ccstd::vector<scene::Camera *>
...
-        for (auto *flow : _flows) {
-            CC_SAFE_DESTROY_AND_DELETE(flow);
-        }
-    } else {
-        for (auto *flow : _flows) {
-            CC_SAFE_DESTROY(flow);
-        }
+    for (auto const &flow : _flows) {
+        CC_SAFE_DESTROY(flow);
     }
```
minggo, line 154: The same as above.

native/cocos/renderer/pipeline/RenderFlow.cpp
star-e: The reason is that pipeline contains gfx resources which need to be destroyed immediately instead of GC. 
`destroy` is dispatched from `root.destroy->Root::destroy->RenderPipeline.cpp->RenderFlow.cpp`. So the timing of destroy might be several seconds before `pipeline` is garbage collected.
We will try to provide a better solution to the resource life-cycle problem, in the new pipeline.

minggo: Ok.

minggo (approved):

PR #12386, opened 2022-08-10
[v3.6.1-shader] add deferred shader compiling

star-e:
Deferred shader compiling. Only compile shader when actually used.

### Changelog

* refactored all shader backends

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/core/gfx/webgl/webgl-shader.ts
+++ b/cocos/core/gfx/webgl/webgl-shader.ts
@@ -31,7 +31,10 @@ import { IWebGLGPUShader, IWebGLGPUShaderStage } from './webgl-gpu-objects';
 
 export class WebGLShader extends Shader {
     get gpuShader (): IWebGLGPUShader {
-        return  this._gpuShader!;
+        if (this._gpuShader!.glProgram === null) {
+            WebGLCmdFuncCreateShader(WebGLDeviceManager.instance, this._gpuShader!);
+        }
+        return this._gpuShader!;
     }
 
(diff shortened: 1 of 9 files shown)


star-e:
@cocos-robot run test cases

suqionglian:
@cocos-robot run test cases

star-e:
@cocos-robot run test cases

minggo (approved):

hana-alice:
In my understanding this delay the shader initialization from init to first time get the gpuObject, it decreases peak compilation time. @bluesky013 once he told me maybe all gfx object initialization can be done as a task(by TBB/taskflow). Do we have any plan on send these initializations to a task? if we do, does this pr has any harm on how we will implement the `task`?

star-e:
I can add a macro to disable deferred shader compiling.

The purpose of this pr is to reduce the number of shaders that need compiling.
I don't think this is a good way to solve this problem, but it is the only practical way I can think of.
Actually, it's user's responsibility to specify the subset of shader need compiling and we should provide this feature in the future.
Deferred shader compiling should be turned off, when shader subset-compiling is supported.

We should also provide batch/multi-threaded/async shader compiling in the future.

hana-alice (approved):

star-e:
相关issue  https://github.com/cocos/3d-tasks/issues/13642

star-e:
Related issue in cocos-engine  https://github.com/cocos/cocos-engine/issues/12396

PR #12483, opened 2022-08-19
Wgpu on 3.6.1

hana-alice:
### Changelog
webgpu now available.
- [x] still some issues: https://github.com/cocos/3d-tasks/issues/13711
- [x] start flow integrated like physic-wasm: @PPpro 
- [x] docs preparing: https://github.com/cocos/3d-tasks/blob/c204cf7eb5359036397f828dc7028cb41281cefe/design-docs/gfx/backend/WebGPU/readme.md
- [ ] ci fixing


hana-alice:
emscripten/wasm related, how to debug with wgpu, issues of wgpu now --- docs are preparing

cc.config.json
```diff
@@ -1,32 +1,57 @@
...
-            "bcd64cc6-2dd9-43f6-abbe-66318d332032",
-            "d930590d-bb92-4cc8-8bd1-23cd027f9edf",
-            "a3cd009f-0ab0-420d-9278-b9fdab939bbc",
-            "9361fd90-ba52-4f84-aa93-6e878fd576ca"
-        ]},
-        "gfx-webgl": { "modules": [ "gfx-webgl" ] },
-        "gfx-webgl2": { "modules": [ "gfx-webgl2" ] },
-        "3d": { 
-            "modules": [ "3d" ], 
+        "base": {
```
minggo: So many changed in this file. Use different format setting?

star-e: consider reverting code formatting.

cocos/core/gfx/device-manager.ts
```diff
@@ -27,7 +28,7 @@ import { JSB } from 'internal:constants';
 import { legacyCC } from '../global-exports';
 import { error, getError } from '../platform/debug';
 import { sys } from '../platform/sys';
-import { BindingMappingInfo, DeviceInfo, SwapchainInfo } from './base/define';
+import { BindingMappingInfo, DeviceInfo, SwapchainInfo } from '.';
```
dumganhar, line 31: Why change this line?

cocos/core/renderer/scene/submodel.ts
```diff
@@ -371,7 +371,13 @@ export class SubModel {
...
         const drawInfo = this._subMesh.drawInfo;
         if (this._inputAssembler && drawInfo) {
-            this._inputAssembler.drawInfo.copy(drawInfo);
+            this._inputAssembler.drawInfo.vertexCount = drawInfo.vertexCount;
+            this._inputAssembler.drawInfo.firstVertex = drawInfo.firstVertex;
+            this._inputAssembler.drawInfo.indexCount = drawInfo.indexCount;
+            this._inputAssembler.drawInfo.firstIndex = drawInfo.firstIndex;
+            this._inputAssembler.drawInfo.vertexOffset = drawInfo.vertexOffset;
+            this._inputAssembler.drawInfo.instanceCount = drawInfo.instanceCount;
+            this._inputAssembler.drawInfo.firstInstance = drawInfo.firstInstance;
```
dumganhar: drawInfo.copy does the same logic. Why did you remove `copy` invocation and wrote the logic again?

dumganhar:
Screen Shot 2022-08-22 at 10 31 05

The wasm binary file is too large, would it be better to place it in external repo?
Or could we compile it when needed rather than commit to source code repo directly.

native/cocos/renderer/gfx-wgpu/WGPUDevice.h
```diff
@@ -28,37 +28,35 @@
 #if (CC_PLATFORM == CC_PLATFORM_WINDOWS)
     #include <windows.h>
 #endif
-
-#include <emscripten/bind.h>
-#include <emscripten/val.h>
-#include "WGPUDef.h"
+#ifdef CC_WGPU_WASM
+    #include "WGPUDef.h"
```
dumganhar, line 32: Why need to add the `ifdef CC_WGPU_WASM` logic in WebGPU gfx backend?

native/cocos/renderer/gfx-wgpu/WGPUObject.h
```diff
@@ -245,6 +254,158 @@ struct CCWGPUQueryPoolObject {
...
+constexpr uint32_t INIT_BUFFER_SIZE = 1024 * 1024 * 1; // 1MB
+
+class CCWGPUStagingBuffer {
+    CCWGPUStagingBuffer(const CCWGPUStagingBuffer &) = delete;
+    CCWGPUStagingBuffer &operator=(const CCWGPUStagingBuffer &) = delete;
+    CCWGPUStagingBuffer() = default;
+    ~CCWGPUStagingBuffer() = default;
+
+public:
+    explicit CCWGPUStagingBuffer(WGPUDevice device, std::function<void(WGPUBuffer)> recycleFunc)
```
dumganhar: to `const std::function<void(WGPUBuffer)> &recycleFunc`

native/cocos/renderer/gfx-wgpu/WGPUPipelineLayout.h
```diff
@@ -24,34 +24,48 @@
...
+    // bindgroup not ready yet so delay creation
     void prepare(const ccstd::set<uint8_t> &setInUse);
 
-    // const ccstd::vector<void*> & layouts()const{return _bgLayouts;}
+    const ccstd::vector<void *> &layouts() const { return _bgLayouts; }
+    const ccstd::vector<void *> &cclayouts() const { return _ccbgLayouts; }
+
+    static ccstd::map<ccstd::hash_t, void *> layoutMap;
+
+    inline ccstd::hash_t getHash() { return _hash; }
```
dumganhar: Add const suffix.
`ccstd::hash_t getHash() const`

native/cocos/renderer/gfx-wgpu/WGPUPipelineState.h
```diff
@@ -33,19 +36,25 @@ namespace gfx {
...
-
+    void check(RenderPass *renderPass, bool forceUpdate = false);
     void prepare(const ccstd::set<uint8_t> &setInUse);
+    PipelineLayout *layout() const { return _pipelineLayout; }
 
     void *ppl() const { return _ppl; }
 
+    static ccstd::map<ccstd::hash_t, void *> pipelineMap;
+
+    ccstd::hash_t getHash() {
```
dumganhar: inline ccstd::hash_t getHash() const {...}

cocos/webgpu/webgpu_wasm.js
```diff
@@ -0,0 +1,5151 @@
+
+var wasmDevice = (() => {
+  var _scriptDir = typeof document !== 'undefined' && document.currentScript ? document.currentScript.src : undefined;
```
minggo: use `-` to connect words.

cocos/core/renderer/scene/submodel.ts
```diff
@@ -371,7 +371,13 @@ export class SubModel {
         // update draw info
         const drawInfo = this._subMesh.drawInfo;
         if (this._inputAssembler && drawInfo) {
-            this._inputAssembler.drawInfo.copy(drawInfo);
+            this._inputAssembler.drawInfo.vertexCount = drawInfo.vertexCount;
```
minggo: Any problem with copy?

native/cocos/renderer/gfx-wgpu/WGPUTexture.cpp
```diff
@@ -99,6 +105,7 @@ void CCWGPUTexture::doInit(const TextureViewInfo &info) {
 
 void CCWGPUTexture::doInit(const SwapchainTextureInfo &info) {
     if (_swapchain) {
+        printf("swapchain %d, %d\n", info.width, info.height);
```
dumganhar, line 108: Why not to use `CC_LOG_DEBUG` ?

cocos/core/pipeline/pipeline-ubo.ts
```diff
@@ -511,7 +511,7 @@ export class PipelineUBO {
         globalDS.bindTexture(UNIFORM_SHADOWMAP_BINDING, builtinResMgr.get<Texture2D>('default-texture').getGFXTexture()!);
         globalDS.bindTexture(UNIFORM_SPOT_SHADOW_MAP_TEXTURE_BINDING, builtinResMgr.get<Texture2D>('default-texture').getGFXTexture()!);
         globalDS.update();
-        globalDS.getBuffer(UBOShadow.BINDING).update(this._shadowUBO);
+        this._pipeline.commandBuffers[0].updateBuffer(globalDS.getBuffer(UBOShadow.BINDING), this._shadowUBO);
```
minggo, line 517: Why change it like this?

cocos/core/gfx/device-manager.ts
hana-alice: still some circular dependencies in ts, thanks for figuring it out

cocos/core/renderer/scene/submodel.ts
hana-alice: `DrawInfo` has been exported by emscripten, ems value cannot be copied from js drawInfo, so change copy here

native/cocos/base/StringUtil.cpp
```diff
@@ -28,7 +28,9 @@
 #include <algorithm>
 #include <cctype>
 #include <cstdarg>
-#include "base/ZipUtils.h"
+#ifndef CC_WGPU_WASM
+    #include "base/ZipUtils.h"
+#endif
```
minggo, line 33: Why change it?

hana-alice:
> Screen Shot 2022-08-22 at 10 31 05
> 
> The wasm binary file is too large, would it be better to place it in external repo? Or could we compile it when needed rather than commit to source code repo directly.

`-g` flag added in cmakelist for debug, so symbols has been preserved in wasm, this is not the final version, anyway i'll remove this flag later, by the way, size of release version for wasm is 280KB
move to external is fine with me.

cocos/webgpu/webgpu_wasm.js
hana-alice: this file is generated by emscripten, maybe ems's rules of format is different from us

native/cocos/renderer/gfx-wgpu/WGPUDevice.h
hana-alice: for emscripten wasm only, as you mentioned before we can run wgpu on native by dawn, so there is also CC_WGPU_DAWN, and CC_WGPU_RS

cocos/core/renderer/scene/submodel.ts
hana-alice: `DrawInfo ` has been exported by emscripten, ems value cannot be copied from js drawInfo, so change copy here

native/cocos/renderer/gfx-wgpu/WGPUTexture.cpp
hana-alice: this is a "issue" using with emscripten. When using CC_LOG_XXXX/std::cout..., not sure why some function like `get_local_time` has been called, and those functions doesn't work happy with emscripten, maybe more flags are needed to export system funcs. While printf needs nothing.

cocos/core/pipeline/pipeline-ubo.ts
hana-alice: It's a bug, it should be updated by commandbuffer

native/cocos/base/StringUtil.cpp
hana-alice: this denpends on libz, while wgpu does not.

cocos/webgpu/webgpu_wasm.js
minggo: Ok

cocos/core/renderer/scene/submodel.ts
minggo: The drawinfo exported by emscripten is not a standard js object?

native/cocos/renderer/gfx-wgpu/WGPUDevice.h
dumganhar: That's cool. I got it now.

cocos/core/pipeline/pipeline-ubo.ts
minggo: Can any test case trigger the error?

cocos/core/renderer/scene/submodel.ts
dumganhar: So `this._inputAssembler.drawInfo` and `drawInfo` are different types on WebGPU?
Could we export copy method to JS and implement copy in C++ by emscripten::val ?

hana-alice: >The drawinfo exported by emscripten is not a standard js object

No, it's more like a jsb object.
>So this._inputAssembler.drawInfo and drawInfo are different types on WebGPU?

No, they are different. We can meke them the same type, but performance may get affected. Considering `commandBuffer.draw` may be called frequently per frame, to reduce js->c++ data conversion so I decide to export drawInfo. This is just a choice of `performance` or `usage`.

* If `DrawInfo` exported by emscripten, we can fill data and pass it directly to c++ interface in js. If not, we have to pass js object `DrawInfo` wrapped by `emscripten::val`, and then read property one by one and fill c++ `DrawInfo` in native. This may be a performance harzard.

cocos/core/pipeline/pipeline-ubo.ts
hana-alice: Nope, this works well with webGL, but webGL seems delayed every update until it has to. Same implementation in c++ doing this way.

native/cocos/renderer/gfx-wgpu/WGPUTexture.cpp
minggo: Can modify CC_LOG_DEBUG to adapt wgpu.

cocos/core/pipeline/pipeline-ubo.ts
minggo: Ok.

cocos/core/renderer/scene/submodel.ts
dumganhar: Okay, is there any other places which also needs to do the `copy` logic? Shall we add a helper function for this copy purpose instead of writing it in many places?

hana-alice: this is the only place I found. `copy` will be added later.

cocos/core/gfx/override.ts
```diff
@@ -0,0 +1,45 @@
...
+    RenderPass = gfx.CCWGPURenderPass;
+    Framebuffer = gfx.CCWGPUFramebuffer;
+    DescriptorSet = gfx.CCWGPUDescriptorSet;
+    DescriptorSetLayout = gfx.CCWGPUDescriptorSetLayout;
+    PipelineLayout = gfx.CCWGPUPipelineLayout;
+    PipelineState = gfx.CCWGPUPipelineState;
+    CommandBuffer = gfx.CCWGPUCommandBuffer;
+    GeneralBarrier = gfx.WGPUGeneralBarrier;
+    TextureBarrier = gfx.WGPUTextureBarrier;
+    BufferBarrier = gfx.WGPUBufferBarrier;
```
star-e, line 45: What does override do?

hana-alice: At first when gfx has not been instantiated from wasm file, we exported Device/Queue... as any just to make it can be successfully imported. After gfx was extracted from wasm file, use native object gfx.CCWGPUDevice to replace what we defined before(any), and that's where C++ symbols are loaded in TS.

Any code refer to gfx type needs to be executed after `
WEBGPU && promiseForWebGPUInstantiation.then`.

scripts/build-engine/src/index.ts
```diff
@@ -537,6 +540,11 @@ async function doBuild ({
         defaultHandler(warning);
     };
 
+    rollupPlugins.unshift(assetUrl({
+        engineRoot,
+        useWebGPU,
+    }));
+
```
star-e, line 547: What does rollupPlugins do?

PPpro: copy wasm from `engine/native/external/emscripten` to the output engine lib `cococs-js/assets/`

hana-alice:
move to https://github.com/cocos/cocos-engine/pull/12731

PR #12619, opened 2022-08-30
spine/dragonbones renderdata use rgba8 color format

zhakesi:
spine and dragonbones use simple color 4byte

### DIFF
--- a/cocos/2d/renderer/mesh-buffer.ts
+++ b/cocos/2d/renderer/mesh-buffer.ts
@@ -25,7 +25,7 @@
 
 import { JSB } from 'internal:constants';
 import { Device, BufferUsageBit, MemoryUsageBit, Attribute, Buffer, BufferInfo, InputAssembler, InputAssemblerInfo } from '../../core/gfx';
-import { getComponentPerVertex } from './vertex-format';
+import { getAttributeStride, getComponentPerVertex } from './vertex-format';
 import { getError, warnID } from '../../core/platform/debug';
 import { sys } from '../../core';
 import { assertIsTrue } from '../../core/data/utils/asserts';
@@ -187,7 +187,7 @@ export class MeshBuffer {
         this._attributes = attrs;
 
         //sync to native
-        this.floatsPerVertex = getComponentPerVertex(attrs);
+        this.floatsPerVertex = getAttributeStride(attrs) >> 2;
 
         assertIsTrue(this._initVDataCount / this._floatsPerVertex < 65536, getError(9005));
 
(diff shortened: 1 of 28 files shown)


cocos/2d/renderer/render-data.ts
```diff
@@ -227,7 +227,7 @@ export class BaseRenderData {
  */
 export class RenderData extends BaseRenderData {
     public static add (vertexFormat = vfmtPosUvColor, accessor?: StaticVBAccessor) {
-        const rd = new RenderData(vertexFormat);
```
SantyWang, line 230: RenderData's contructor do not accept accessor parameter

cocos/2d/renderer/mesh-buffer.ts
```diff
@@ -198,7 +198,14 @@ export class MeshBuffer {
...
         if (JSB) {
-            this._nativeObj.initialize(device, attrs, vFloatCount, iCount);
+            // default vfmtPosUvColor
+            let vfm = Vertex2DFormat.V3F_T2F_C4F;
+            if (attrs === vfmtPosUvColor4B) {
+                vfm = Vertex2DFormat.V3F_T2F_C4B;
+            } else if (attrs === vfmtPosUvTwoColor4B) {
+                vfm = Vertex2DFormat.V3F_T2F_C4B_C4B;
+            }
+            this._nativeObj.initialize(device, vfm, vFloatCount, iCount);
```
SantyWang: Why don't pass attributes array

native/cocos/2d/renderer/UIMeshBuffer.cpp
```diff
@@ -40,9 +59,24 @@ void UIMeshBuffer::setIData(uint16_t* iData) {
     _iData = iData;
 }
 
-void UIMeshBuffer::initialize(gfx::Device* /*device*/, ccstd::vector<gfx::Attribute*>&& attrs, uint32_t /*vFloatCount*/, uint32_t /*iCount*/) {
-    if (attrs.size() == 4) {
-        _attributes.push_back(gfx::Attribute{gfx::ATTR_NAME_COLOR2, gfx::Format::RGBA32F});
+void UIMeshBuffer::initialize(gfx::Device* /*device*/,Vertex2DFormat vfm, uint32_t /*vFloatCount*/, uint32_t /*iCount*/) {
```
SantyWang: Need to figure out why can not pass array

zhakesi: in c++ get a item value with null vector. i think jsb not convert class. 
i have no time to find root cause, so use a simple way to avoid this problem

SantyWang (approved):

PR #12623, opened 2022-08-31
fix native spine beginSchedule

zhakesi:


platforms/native/engine/jsb-spine-skeleton.js
```diff
@@ -307,6 +307,9 @@ const cacheManager = require('./jsb-cache-manager');
             nativeSkeleton = new spine.SkeletonAnimation();
             try {
                 spine.initSkeletonRenderer(nativeSkeleton, uuid);
+                if (this._objFlags !== 0) {
+                    nativeSkeleton.beginSchedule();
+                }
```
SantyWang, line 312: Why need this check??

zhakesi: no active state, the onenable function not be called, should not call beginSchedule.
or it's onDestory not work, nativeSkeleton.stopSchedule also can not be called

PR #12678, opened 2022-09-05
add XR macro for gfx

hana-alice:
A brief thought:
In my opinion GFX is an abstraction layer of Graphics API, which does not include XR. 
If there is something we have to put it into halfway of a GFX function, it means GFX does not have the ability to make it at that time, which we may work on in the future.
A middleware can be added between XR and GFX, or GFX support custom `initialize`, `acquire`, `present`, etc.

append:
at the very first beginning I just try to add macro in GFXDevice, can't only do that because it won't compile on windows with GLES/VK.

code below
```
if(_xr) {
    // A
} else {
    // B
}
```
translated into
```
#if CC_USE_XR
if(_xr) {
    // A
}
#else
// B
#endif
```
considering `_xr` is the only instance of `XRInstance`.

besides:
accidently found `GFX:VK` relies on module `Application` by getting swapchain width/height from `CC_CURRENT_ENGINE()->getInterface<cc::ISystemWindow>();`.
### not checked on any XR environment.
-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


hana-alice:
@cocos-robot run test cases

minggo (approved):

PR #12727, opened 2022-09-08
improve: update engine to support xr development

fqamic:
1.Fixed slider moving in wrong direction
Use local coordinates for determination
2.fix: XR error opening thread detach
The build supports Vulkan and OpenGLES, but when the device does not support Vulkan, the XR platform cannot support thread detach
3.update xr common

Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/renderer/GFXDeviceManager.h
+++ b/native/cocos/renderer/GFXDeviceManager.h
@@ -85,7 +85,7 @@ public:
 
 #ifdef CC_USE_VULKAN
     #if XR_OEM_PICO
-        Device::SUPPORT_DETACH_DEVICE_THREAD = false;
+        Device::isSupportDetachDeviceThread = false;
     #endif
         if (tryCreate<CCVKDevice>(info, &device)) return device;
 #endif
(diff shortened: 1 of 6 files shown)


minggo (approved):

native/cocos/renderer/GFXDeviceManager.h
```diff
@@ -110,8 +110,8 @@ class CC_DLL DeviceManager final {
         return nullptr;
     }
 
-    static constexpr bool isDetachDeviceThread() {
-        return DETACH_DEVICE_THREAD;
+    static bool isDetachDeviceThread() {
+        return DETACH_DEVICE_THREAD && Device::isSupportDetachDeviceThread;
```
star-e, line 114: I suggest changing to
```c++
static constexpr bool isDetachDeviceThread() {
#if XR_OEM_PICO
    constexpr bool isSupportDetachDeviceThread{false};
#else
    constexpr bool isSupportDetachDeviceThread{true};
#endif
    return DETACH_DEVICE_THREAD && isSupportDetachDeviceThread;
}
```
And remove `isSupportDetachDeviceThread` from `Device`.
The reason is that the behavior of `Device::isSupportDetachDeviceThread` depends on the side-effects of user's setting. It is hard to maintain.

star-e: isDetachDeviceThread() will be moved to Device in 3.6.2

star-e (approved):

SantyWang (approved):

PR #12732, opened 2022-09-09
gpu objects use smart pointer

bluesky013:
Re: #

### Changelog

* gpu objects use smart pointer
* update bufferView binding in descriptor set after buffer resize

-------

### Continuous Integration

This pull request:

* [x] needs automatic test cases check.
  > Manual trigger with `@cocos-robot run test cases` afterward.
* [ ] does not change any runtime related code or build configuration
  > If any reviewer thinks the CI checks are needed, please uncheck this option, then close and reopen the issue.

-------

### Compatibility Check

This pull request:

* [ ] changes public API, and have ensured backward compatibility with [deprecated features](https://github.com/cocos/cocos-engine/blob/v3.5.0/docs/contribution/deprecated-features.md).
* [ ] affects platform compatibility, e.g. system version, browser version, platform sdk version, platform toolchain, language version, hardware compatibility etc.
* [ ] affects file structure of the build package or build configuration which requires user project upgrade.
* [ ] **introduces breaking changes**, please list all changes, affected features and the scope of violation.

PR #12789, opened 2022-09-16
fix:engine crash sometime

oahc09:
_queues's element may became dangling pointer, before use cached buffer we call clear. 
(Pass destroy the buffer, but RenderQueue doesn't know)

Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/renderer/pipeline/RenderBatchedQueue.cpp
+++ b/native/cocos/renderer/pipeline/RenderBatchedQueue.cpp
@@ -35,9 +35,6 @@ namespace cc {
 namespace pipeline {
 
 void RenderBatchedQueue::clear() {
-    for (auto *it : _queues) {
-        it->clear();
-    }
     _queues.clear();
 }
 
--- a/native/cocos/renderer/pipeline/RenderInstancedQueue.cpp
+++ b/native/cocos/renderer/pipeline/RenderInstancedQueue.cpp
@@ -35,9 +35,6 @@ namespace cc {
 namespace pipeline {
 
 void RenderInstancedQueue::clear() {
-    for (auto *it : _queues) {
-        it->clear();
-    }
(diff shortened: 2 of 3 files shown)


oahc09:
@cocos-robot run test cases

minggo:
How about change to use IntrusivePtr instead of raw pointer.

star-e:
lizhen is currently working on replacing raw pointer with intrusive ptr.

oahc09:
> How about change to use IntrusivePtr instead of raw pointer.

Not suitable, weak ptr is better

minggo:
> lizhen is currently working on replacing raw pointer with intrusive ptr.

Could you please add the link of the PR? Then we may close this one if it is fixed in better way.

star-e:
Here is the pr.
I'm not sure whether lizhen has already fixed this problem.
We need to check
https://github.com/cocos/cocos-engine/pull/12732

oahc09:
> > lizhen is currently working on replacing raw pointer with intrusive ptr.
> 
> Could you please add the link of the PR? Then we may close this one if it is fixed in better way.

I think this pr can be closed.

bluesky013:
> > > lizhen is currently working on replacing raw pointer with intrusive ptr.
> > 
> > 
> > Could you please add the link of the PR? Then we may close this one if it is fixed in better way.
> 
> I think this pr can be closed.

bluesky013:
> Here is the pr. I'm not sure whether lizhen has already fixed this problem. We need to check #12732

#12732 focus on gfx-vulkan and dit not fix this issue.

bluesky013:
I think IntrusivePtr can solve this problem. ```InstancedBuffer*``` here can be a strong reference.

PR #12804, opened 2022-09-19
Fix: use objectId instead of pointer to check for dirty flag in descr…

bluesky013:
…iptorSet.

Re: #

### Changelog

* use objectId instead of pointer to check for dirty flag in descriptorSet.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/renderer/gfx-base/GFXDescriptorSet.cpp
+++ b/native/cocos/renderer/gfx-base/GFXDescriptorSet.cpp
@@ -25,6 +25,9 @@
 
 #include "GFXDescriptorSet.h"
 #include "GFXDescriptorSetLayout.h"
+#include "GFXBuffer.h"
+#include "GFXTexture.h"
+#include "states/GFXSampler.h"
 #include "GFXObject.h"
 
 namespace cc {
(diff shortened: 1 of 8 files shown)


bluesky013:
@cocos-robot run test cases

minggo (approved):

minggo:
Can i know what issue does it fix?

star-e (approved):

star-e:
When deleting a Texture and then creating a Texture, the returned pointer address might be the same.
If we compare Texture using pointer address, they will be the same, although their contents are different.
This will crash the program.
So we need objectID to distinguish different gfx resources.

bluesky013:
If DescriptorSet holds strong reference of bound resources, the newly created resource pointer would never be the same and pointer can be used to campare.
If it holds weak reference as it does now, DescriptorSet doesn't know bound resource is a valid pointer.

```IntrusivePtr``` can not be used unless that all bound resources are guaranteed to be managed with smart pointer.

minggo:
Got it, thanks.

minggo:
Got it, thanks.

bluesky013:
split from https://github.com/cocos/cocos-engine/pull/12732

PR #12814, opened 2022-09-19
[Optimize] Cache current time for audio player in minigame.

timlyeee:
Re: #https://github.com/cocos/3d-tasks/issues/12423

### Changelog

* Cache current time for minigames, in case the developer cannot set current time before play

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/pal/audio/minigame/player-minigame.ts
+++ b/pal/audio/minigame/player-minigame.ts
@@ -2,7 +2,7 @@ import { minigame } from 'pal/minigame';
 import { systemInfo } from 'pal/system-info';
 import { EventTarget } from '../../../cocos/core/event';
 import { AudioEvent, AudioPCMDataView, AudioState, AudioType } from '../type';
-import { clamp, clamp01, sys } from '../../../cocos/core';
+import { clamp, clamp01 } from '../../../cocos/core';
 import { enqueueOperation, OperationInfo, OperationQueueable } from '../operation-queue';
 
 export class OneShotAudioMinigame {
@@ -48,7 +48,8 @@ export class AudioPlayerMinigame implements OperationQueueable {
     private _innerAudioContext: InnerAudioContext;
     private _state: AudioState = AudioState.INIT;
     private _cacheTime = 0;
-    private _needSeekAfterPlay = false;
+    private _needSeek = false;
+    private _seeking = false;
 
     private _onPlay: () => void;
     private _onPause: () => void;
@@ -57,6 +58,11 @@ export class AudioPlayerMinigame implements OperationQueueable {
     private _onEnded: () => void;
     private _readyToHandleOnShow = false;
 
+    private resetSeekCache () {
+        this._cacheTime = 0;
+        this._needSeek = false;
+        this._seeking = false;
+    }
     /**
      * @deprecated since v3.5.0, this is an engine private interface that will be removed in the future.
      */
@@ -77,10 +83,9 @@ export class AudioPlayerMinigame implements OperationQueueable {
         this._onPlay = () => {
             this._state = AudioState.PLAYING;
             eventTarget.emit(AudioEvent.PLAYED);
-            if (this._needSeekAfterPlay) {
-                this._needSeekAfterPlay = false;
-                this._innerAudioContext.seek(this._cacheTime);
-                this._cacheTime = 0;
+            if (this._needSeek) {
+                this._needSeek = false;
+                this.seek(this._cacheTime).catch((e) => {});
             }
         };
         innerAudioContext.onPlay(this._onPlay);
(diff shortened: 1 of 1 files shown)


pal/audio/minigame/player-minigame.ts
```diff
@@ -225,13 +238,18 @@ export class AudioPlayerMinigame implements OperationQueueable {
...
     seek (time: number): Promise<void> {
         return new Promise((resolve) => {
-            time = clamp(time, 0, this.duration);
-            if (time === 0 && this.currentTime === 0) {
-                // skip invalid seek
-                resolve();
-            } else {
+            // time = clamp(time, 0, this.duration);
+            if (this._state === AudioState.PLAYING) {
+                time = clamp(time, 0, this.duration);
```
PPpro, line 252: need clamp time when state is not PLAYING ?

pal/audio/minigame/player-minigame.ts
```diff
@@ -211,6 +221,9 @@ export class AudioPlayerMinigame implements OperationQueueable {
         // KNOWN ISSUES: currentTime doesn't work well
         // on Baidu: currentTime returns without numbers on decimal places
         // on WeChat iOS: we can't reset currentTime to 0 when stop audio
+        if (this._state !== AudioState.PLAYING) {
```
minggo, line 233: Can remove the comments now?

pal/audio/minigame/player-minigame.ts
```diff
@@ -2,7 +2,7 @@ import { minigame } from 'pal/minigame';
 import { systemInfo } from 'pal/system-info';
 import { EventTarget } from '../../../cocos/core/event';
 import { AudioEvent, AudioPCMDataView, AudioState, AudioType } from '../type';
-import { clamp, clamp01 } from '../../../cocos/core';
+import { clamp, clamp01, sys } from '../../../cocos/core';
```
PPpro: it seems that we don't need to import sys

pal/audio/minigame/player-minigame.ts
timlyeee: Yes, some comments can be removed

PPpro: please ensure the platforms have fixed these known issues
then we can remove this KNOWN ISSUES comments

pal/audio/minigame/player-minigame.ts
timlyeee: We will clamp the time when we're truly ready to seek.

In the callback OnPlay();

pal/audio/minigame/player-minigame.ts
timlyeee: 1

pal/audio/minigame/player-minigame.ts
PPpro: I see that you call `this._innerAudioContext.seek()` instead of `this.seek()` In the callback OnPlay()

timlyeee: Yes, this.seek() return a promise and a little bit slower

pal/audio/minigame/player-minigame.ts
timlyeee: On Baidu Android, the problem stll exist, On Wechat ios, we can reset it manually

minggo: So this PR can not fix issue on all platforms?

pal/audio/minigame/player-minigame.ts
```diff
@@ -55,6 +58,11 @@ export class AudioPlayerMinigame implements OperationQueueable {
     private _onEnded: () => void;
     private _readyToHandleOnShow = false;
 
+    private resetSeekCache () {
```
PPpro:
```suggestion
    private _resetSeekCache () {
```

PPpro: we name private method / property as `_xxxx`

pal/audio/minigame/player-minigame.ts
```diff
@@ -225,14 +247,16 @@ export class AudioPlayerMinigame implements OperationQueueable {
...
+            // KNOWN ISSUES: on Baidu: currentTime returns without numbers on decimal places
+            if (this._state === AudioState.PLAYING && !this._seeking) {
+                time = clamp(time, 0, this.duration);
+                this._seeking = true;
                 this._innerAudioContext.seek(time);
+            } else if (this._cacheTime !== time) { // Skip the invalid seek
+                this._cacheTime = time;
+                this._needSeek = true;
             }
+            resolve();
```
PPpro, line 259: if we called innerAudioContext.seek(), please resolve in the callback of onSeek()

pal/audio/minigame/player-minigame.ts
```diff
@@ -225,14 +247,16 @@ export class AudioPlayerMinigame implements OperationQueueable {
...
     seek (time: number): Promise<void> {
         return new Promise((resolve) => {
-            time = clamp(time, 0, this.duration);
-            if (time === 0 && this.currentTime === 0) {
-                // skip invalid seek
-                resolve();
-            } else {
-                this._eventTarget.once(AudioEvent.SEEKED, resolve);
+            // KNOWN ISSUES: on Baidu: currentTime returns without numbers on decimal places
+            if (this._state === AudioState.PLAYING && !this._seeking) {
```
PPpro, line 251: what if the call stack like this
```js
audio.play()
audio.seek(10)
audio.seek(20)  // now we are seeking, will this seek call be invalid ?
```

pal/audio/minigame/player-minigame.ts
timlyeee: > So this PR can not fix issue on all platforms?

No, specifically, there's no bug on Baidu minigame, we can still seek and play. It's just a wrong behavior

pal/audio/minigame/player-minigame.ts
timlyeee: Actually, if we use cache time to save middle state, we can directly resolve here.

pal/audio/minigame/player-minigame.ts
timlyeee: Will be seek later.

pal/audio/minigame/player-minigame.ts
minggo: But can also use cached value event it doesn't have problem. Right?

timlyeee: Yes, the reason to cache the seek time, is because on some platform, we cannot seek before play. Even if Baidu return a number without decimal places, it doesn't influence the behavior of the cached time

minggo (approved):

PPpro (approved):

PR #12826, opened 2022-09-20
fix download url request not encode space

novain1102:
Re: #https://github.com/cocos/3d-tasks/issues/7874

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/native-binding/index.ts
+++ b/cocos/native-binding/index.ts
@@ -408,7 +408,9 @@ export declare namespace native {
         /**
          * @en create a download task. The maximum size for a single download file is 4GB.
          * @zh 创建一个下载任务. 单个下载文件最大为4GB.
-         * @param requestURL @en Request download resource URL  @zh 请求下载资源的URL
+         * @param requestURL
+         * @en Request download resource URL. Node: Users need to encode the URL containing special characters except spaces (such as Chinese, etc.)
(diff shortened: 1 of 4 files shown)


PatriceJiang:
Should we implement `url_encode` ?

minggo:
As discussed, may let developers encode the url if needed.

novain1102:
> Should we implement `url_encode` ?

In native, the encoding methods in different platforms are not consistent. Eg, Windows needs locate some last of char like `/`to substr the string and then call `curl_easy_escape()`, which will affect the performance when deal with amount of requests, And another potential problem is that if it is a `POST` request, which may be very long and cotains special char after `/`, the position of this substr is not accurate. so I think developer do the encoding if needed is a better approach.

minggo:
It is better to provide a function to encode the string.

novain1102:
> It is better to provide a function to encode the string.

We can't predict the url request pass in from the user. Eg:it may be an http resource request(in this case we can find last of '/' to encode) or a very long POST request with special characters (in this case, user must locate and encode the POST parameters separately, otherwise an error will occur). So different urls need to encode in different way, and the engine cannot encode uniformly. The better approach is to let the user do encoding before passing in.

minggo:
What i mean is to encode the space, as did in the PR.

minggo (approved):

PR #12835, opened 2022-09-21
Reduce jsb glue code size

PatriceJiang:
Re: #

### Changelog

* Un-inline jsb wrapper functions, reduce the size of libcocos.so by around 1MB

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/bindings/jswrapper/State.cpp
+++ b/native/cocos/bindings/jswrapper/State.cpp
@@ -27,24 +27,4 @@
 #include "State.h"
 
 namespace se {
-
-State::State(Object *thisObject)
-: _thisObject(thisObject) {
-    if (_thisObject != nullptr) {
-        _thisObject->incRef();
-    }
-}
-
-State::State(Object *thisObject, const ValueArray &args)
-: _thisObject(thisObject),
-  _args(&args) {
-    if (_thisObject != nullptr) {
-        _thisObject->incRef();
-    }
-}
-
-State::~State() {
-    SAFE_DEC_REF(_thisObject);
-}
-
 } // namespace se
--- a/native/cocos/bindings/manual/jsb_cocos_manual.cpp
+++ b/native/cocos/bindings/manual/jsb_cocos_manual.cpp
@@ -36,7 +36,7 @@
 
 extern se::Object *__jsb_cc_FileUtils_proto; // NOLINT(readability-redundant-declaration, readability-identifier-naming)
 
-static bool jsb_ccx_empty_func(const se::State & /*s*/) { // NOLINT(readability-identifier-naming)
+static bool jsb_ccx_empty_func(se::State & /*s*/) { // NOLINT(readability-identifier-naming)
     return true;
 }
 SE_BIND_FUNC(jsb_ccx_empty_func) // NOLINT(readability-identifier-naming)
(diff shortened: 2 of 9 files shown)


native/cocos/bindings/jswrapper/State.cpp
```diff
@@ -27,24 +27,4 @@
...
-State::State(Object *thisObject, const ValueArray &args)
-: _thisObject(thisObject),
-  _args(&args) {
-    if (_thisObject != nullptr) {
-        _thisObject->incRef();
-    }
-}
-
-State::~State() {
-    SAFE_DEC_REF(_thisObject);
```
dumganhar, line 47: Moving the code to header could make compiled code smaller?

PatriceJiang: To inline the destructor and code size is bearly not aaffected.

dumganhar (approved):

PatriceJiang:
@cocos-robot run test cases

PatriceJiang:
@cocos-robot run test cases

native/cocos/bindings/manual/jsb_cocos_manual.cpp
```diff
@@ -36,7 +36,7 @@
 
 extern se::Object *__jsb_cc_FileUtils_proto; // NOLINT(readability-redundant-declaration, readability-identifier-naming)
 
-static bool jsb_ccx_empty_func(const se::State & /*s*/) { // NOLINT(readability-identifier-naming)
+static bool jsb_ccx_empty_func(se::State & /*s*/) { // NOLINT(readability-identifier-naming)
```
minggo, line 39: Why remove const?

native/cocos/bindings/jswrapper/State.cpp
```diff
@@ -27,24 +27,4 @@
...
-}
```
minggo, line 48: Any reason to implement it in .h?

native/cocos/bindings/manual/jsb_cocos_manual.cpp
PatriceJiang: All jsb function should be unified and meet the signature `using se_function_ptr = bool (*)(se::State &state);` as an parameter.

native/cocos/bindings/jswrapper/State.cpp
PatriceJiang: The code size of this function is somewhats small and `se::State` objects are only constructed and destructed within `jsb*Wrapper` functions, which are not frequently used.  By inlining the destructor fewer function calls will be generated within JSB functions.

minggo: Yep, the code size is small. But if there is not special reason, then should implement in .cp by default.

native/cocos/bindings/manual/jsb_cocos_manual.cpp
minggo: ok

native/cocos/bindings/jswrapper/State.cpp
PatriceJiang: Yes, for most cases destructors are implemented in source files. Here these `jsb*Wrapper` functions are hot and inlined.

minggo: Ok

minggo: Please do testing about it. We should do decision base on testing result. If it is true, then it is a new rule, we should check other source code too.

PatriceJiang: ### Code size
||final code size|
|--|----|
|inline| 21365912 B (+208B)|
|non-inline|21365704 B|

### Instructions

### non-inline
image
image

#### inline
image

less instructions are emitted, and should be a bit faster.

minggo (approved):

PR #12983, opened 2022-10-08
[optimization] Remove camera dependency

troublemaker52025:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

* [x] **introduces breaking changes**, please list all changes, affected features and the scope of violation.

### DIFF
--- a/cocos/core/geometry/frustum.ts
+++ b/cocos/core/geometry/frustum.ts
@@ -124,12 +124,11 @@ export class Frustum {
     /**
      * @en Calculate the splitted frustum.
      * @zh 创建一个新的截锥体。
-     * @param m @en The transform matrix @zh 变换矩阵
-     * @param fov @en Field of view of the camera @zh 相机的视角大小
-     * @param aspect @en The aspect ratio of the camera @zh 相机视图的长宽比
      * @param start @en The split start position @zh 分割开始位置
      * @param end @en The split end position @zh 分割末尾位置
-     * @return @en The out object @zh 返回新截锥体.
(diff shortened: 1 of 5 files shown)


troublemaker52025:
@cocos-robot run test cases

minggo (approved):

minggo:
>  list all changes, affected features and the scope of violation.

 Should describe it.

troublemaker52025:
# introduces breaking changes

affected features and the scope of violation

##  for js engine

* Change the way the split function is called, from a static member function to a member function, user scripts will not be able to use the static method directly.
* Change the form parameter of split function, user scripts need to change the parameter passed to it.
* Allow users to call functions of the same name in the native engine via scripts.
* Remove the dependency on the Camera class

## for native engine

* Allow users to call functions of the same name in the native engine via scripts.
* Change the formal arguments of the split function to be consistent with the js engine parameters.
* Remove the dependency on the scene::Camera class

cocos/core/geometry/frustum.ts
```diff
@@ -125,21 +124,20 @@ export class Frustum {
     /**
      * @en Calculate the splitted frustum.
      * @zh 创建一个新的截锥体。
```
minggo: It doesn't create a new frustum now. The comment should be changed too.

troublemaker52025: sure

jk20012001 (approved):

PR #12991, opened 2022-10-09
Added InputAssembler::update && Removed IA Pool 

bluesky013:
Re: #

### Changelog

* Added InputAssembler::update for buffer resizing.
* Removed IA Pool in batcher2D and used shared IA.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/2d/renderer/RenderDrawInfo.cpp
+++ b/native/cocos/2d/renderer/RenderDrawInfo.cpp
@@ -76,6 +76,10 @@ void RenderDrawInfo::uploadBuffers() {
     }
 }
 
+void RenderDrawInfo::resetMeshIA() const {
+    CC_ASSERT(_drawInfoAttrs._isMeshBuffer && _drawInfoAttrs._drawInfoType == RenderDrawInfoType::COMP);
+}
+
 void RenderDrawInfo::destroy() {
     CC_SAFE_DELETE(_iaInfo.indexBuffer);
     if (!_iaInfo.vertexBuffers.empty()) {
--- a/native/cocos/2d/renderer/RenderDrawInfo.cpp
+++ b/native/cocos/2d/renderer/RenderDrawInfo.cpp
@@ -52,55 +52,32 @@ void RenderDrawInfo::changeMeshBuffer() {
 
 gfx::InputAssembler* RenderDrawInfo::requestIA(gfx::Device* device) {
     CC_ASSERT(_drawInfoAttrs._isMeshBuffer && _drawInfoAttrs._drawInfoType == RenderDrawInfoType::COMP);
-    if (!_iaPool) {
-        _iaPool = ccnew ccstd::vector<gfx::InputAssembler*>;
-    }
-    if (_nextFreeIAHandle >= _iaPool->size()) {
-        initIAInfo(device);
-    }
-    auto* ia = (*_iaPool)[_nextFreeIAHandle++];
-    ia->setFirstIndex(getIndexOffset());
-    ia->setIndexCount(getIbCount());
-    return ia;
+    initIAInfo(device);
+    return _ia;
 }
 
 void RenderDrawInfo::uploadBuffers() {
     CC_ASSERT(_drawInfoAttrs._isMeshBuffer && _drawInfoAttrs._drawInfoType == RenderDrawInfoType::COMP);
     if (_drawInfoAttrs._vbCount == 0 || _drawInfoAttrs._ibCount == 0) return;
     uint32_t size = _drawInfoAttrs._vbCount * 9 * sizeof(float); // magic Number
-    gfx::Buffer* vBuffer = _iaInfo->vertexBuffers[0];
+    gfx::Buffer* vBuffer = _iaInfo.vertexBuffers[0];
     vBuffer->resize(size);
     vBuffer->update(_vDataBuffer);
-    gfx::Buffer* iBuffer = _iaInfo->indexBuffer;
+    gfx::Buffer* iBuffer = _iaInfo.indexBuffer;
     uint32_t iSize = _drawInfoAttrs._ibCount * 2;
     iBuffer->resize(iSize);
     iBuffer->update(_iDataBuffer);
 }
 
-void RenderDrawInfo::resetMeshIA() {
-    CC_ASSERT(_drawInfoAttrs._isMeshBuffer && _drawInfoAttrs._drawInfoType == RenderDrawInfoType::COMP);
-    _nextFreeIAHandle = 0;
-}
-
 void RenderDrawInfo::destroy() {
-    _nextFreeIAHandle = 0;
-    if (_iaInfo) {
-        CC_SAFE_DELETE(_iaInfo->indexBuffer);
-        if (!_iaInfo->vertexBuffers.empty()) {
-            // only one vb
-            CC_SAFE_DELETE(_iaInfo->vertexBuffers[0]);
-            _iaInfo->vertexBuffers.clear();
-        }
-        CC_SAFE_DELETE(_iaInfo);
+    CC_SAFE_DELETE(_iaInfo.indexBuffer);
+    if (!_iaInfo.vertexBuffers.empty()) {
+        // only one vb
+        CC_SAFE_DELETE(_iaInfo.vertexBuffers[0]);
+        _iaInfo.vertexBuffers.clear();
     }
 
-    if (_iaPool) {
-        for (auto* ia : *_iaPool) {
-            CC_SAFE_DELETE(ia);
-        }
-        _iaPool->clear();
-        CC_SAFE_DELETE(_iaPool);
-    }
+    CC_SAFE_DELETE(_ia);
 
     if (_localDSBF) {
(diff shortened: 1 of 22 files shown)


bluesky013:
@cocos-robot run test cases

minggo:
Should change ts too?

> Added InputAssembler::update for buffer resizing.
Removed IA Pool in batcher2D and used shared IA.

Can i know why do these change?

bluesky013:
> Should change ts too?
> 
> > Added InputAssembler::update for buffer resizing.
> > Removed IA Pool in batcher2D and used shared IA.
> 
> Can i know why do these change?

For the first change, I'm working on removing Hubs in gfx. Hubs record relations like Buffers to DescriptorSets. When buffer resizes, gfx automatically update related descriptorsets by traversing the relations form the Hub. However, Hubs introduces states between resources, which may cause impact to multi-thread support. GFX needs to be stateless to support multithreading capabilities. 

From another point of view，`buffer::resize()` or `image::resize()` is not a friendly interface for morden graphics api. GFX implementation needs to destroy the old gpuBuffer and create a new gpuBuffer. This is the root cause of Hubs design. Features do know which InputAssembler or DescriptorSet is using the Buffer without a global relation table.

For the second change, I found `InputAssemblers` in IAPool share the same vertexbuffers and indexbuffer. Use a shared `InputAssembler` and a set of `DrawInfos` may reduce the number of `VAOs`.

native/cocos/2d/renderer/RenderDrawInfo.h
```diff
@@ -289,8 +287,8 @@ class RenderDrawInfo final {
         scene::Model* _model;
         uint8_t* _sharedBuffer;
     };
-    gfx::InputAssemblerInfo* _iaInfo{nullptr};
-    ccstd::vector<gfx::InputAssembler*>* _iaPool{nullptr};
+    gfx::InputAssemblerInfo _iaInfo;
```
SantyWang: No need to cache iaInfo in RenderDrawInfo，just create it in `initIAInfo`

bluesky013: I tried, but I need a container to hold the pointer of vertexbuffers and indexbuffer.

bluesky013: Oh, use smart_ptr instead of cache iaInfo may be more appropriate.

native/cocos/renderer/gfx-vulkan/VKInputAssembler.cpp
```diff
@@ -39,24 +39,32 @@ CCVKInputAssembler::~CCVKInputAssembler() {
     destroy();
 }
 
-void CCVKInputAssembler::doInit(const InputAssemblerInfo &info) {
-    size_t vbCount = _vertexBuffers.size();
-
+void CCVKInputAssembler::doInit([[maybe_unused]] const InputAssemblerInfo &info) {
```
minggo, line 42: What's `[[maybe_unused]]`?

minggo:
It seems only vulkan implement `InputAssembler::doUpdate()`, so i have two questions:
- should other gfx backends add `InputAssembler::update` too? 
- the vulkan implementation seems do nothing special, why need this function?

native/cocos/renderer/gfx-vulkan/VKInputAssembler.cpp
bluesky013: [[mabe_unused]] is used to suppress `Clang-Tidy: Parameter 'info' is unused`.

bluesky013:
> It seems only vulkan implement `InputAssembler::doUpdate()`, so i have two questions:
> 
> * should other gfx backends add `InputAssembler::update` too?
> * the vulkan implementation seems do nothing special, why need this function?

Yes, metal and vulkan backends need to implement `doUpdate`. OpenGL ES backend supports resize without recreat glBuffer by `glBufferData`. This pr is a preparation for gfx smart_ptr modification and will not affect existing functionality.

#12732 is the draft implement. Without `InputAssembler::doUpdate`, gfx may need a IAHub.

native/cocos/renderer/gfx-vulkan/VKInputAssembler.cpp
minggo: Good to know, it is a new feature since c++17. We should make sure it works on iOS11 and NS.

native/cocos/2d/renderer/RenderDrawInfo.cpp
```diff
@@ -52,55 +52,50 @@ void RenderDrawInfo::changeMeshBuffer() {
...
+    if (needUpdateIA) {
+        _ia->update();
+    }
 }
 
-void RenderDrawInfo::resetMeshIA() {
+void RenderDrawInfo::resetMeshIA() const {
     CC_ASSERT(_drawInfoAttrs._isMeshBuffer && _drawInfoAttrs._drawInfoType == RenderDrawInfoType::COMP);
-    _nextFreeIAHandle = 0;
 }
```
minggo, line 86: It seems this function can be removed?

minggo:
> Yes, metal and vulkan backends need to implement doUpdate

But metal backend doesn't implement it.

> This pr is a preparation for gfx smart_ptr modification and will not affect existing functionality.

Don't know the relationship with this PR and smart_ptr. Why can use smart_ptr with this PR?

> Without InputAssembler::doUpdate, gfx may need a IAHub

Could you please describe it in detail?

native/cocos/renderer/gfx-vulkan/VKInputAssembler.cpp
bluesky013: I checked compiler support for `maybe_unused`.
MSVC: 2017 15.3+ meets our requirement(2017 15.7+)
Apple Clang: yes if c++ 17 is supported.
Clang：3.9 (Need a NS developer tools requiremens.)

I will replace it with another method.

native/cocos/2d/renderer/RenderDrawInfo.cpp
bluesky013: Yes, removing the interface will cause a compile error when running test cases. So I added back and ignore it in swig config. It could be removed after jsb automatically updated.

minggo: > removing the interface will cause a compile error when running test cases.

It is because it uses a precompiled editor to run test case. It is different from other CI. Other CI will use latest engine codes.

bluesky013:
> > Yes, metal and vulkan backends need to implement doUpdate
> 
> But metal backend doesn't implement it.
> 
> > This pr is a preparation for gfx smart_ptr modification and will not affect existing functionality.
> 
> Don't know the relationship with this PR and smart_ptr. Why can use smart_ptr with this PR?
> 
> > Without InputAssembler::doUpdate, gfx may need a IAHub
> 
> Could you please describe it in detail?

`doUpdate` is designed to get the latest `gpu::buffer` pointer from `gfx::buffer` and is a redundant operation for now. Vulkan implementation could also be empty here and move the implementation to the next pr.

* Why gpu objects use smart_ptr
Consider we have two gfx::Buffer and the second one is a bufferView of the first. In this situation, delete gfx::Buffer0 without delete gfx::Buffer1 and update gfx::DescriptorSet0 with new gfx::buffer would cause:
1. dangling pointer gpuBuffer0 in gfx::Buffer1
2. destroyed gpu device handle device::buffer0 in gfx::Buffer1
3. destroy gpu device buffer0 in gpu device descriptorset in use
企业微信截图_16653930604084

So use smart_ptr to manager gpuObjects would be more safe. It makes gpu objects live longer than gfx objects. But this also introduces a constraint, device handle and gpuObject must have the same lifetime. Device handle is released in destructor or `shutdown` of gpuObject.

* The impact of smart_ptr on resizing
In morden graphics APIs, a buffer resize operation means receating a device buffer. In the previous implementation, gfx::buffer will **replace** `vk::buffer` or `mtl::buffer` (device handle) in gpuBuffer during resizing, which results in the device buffer having a different lifetime than gpuBuffer. This breaks the rule above. We need to recreate the gpuBuffer and notify all related InputAssembler and DescriptorSets according to the relation table IAHub and DescriptorHub.

> Features do know which InputAssembler or DescriptorSet is using the Buffer without a global relation table.

We can call `Input Assembler::Update` or `Descriptor Set::Update` after resizing the buffer without maintain the relation table.

minggo:
> We can call Input Assembler::Update or Descriptor Set::Update after resizing the buffer without maintain the relation table.

Will it break compatibility?

bluesky013:
> > We can call Input Assembler::Update or Descriptor Set::Update after resizing the buffer without maintain the relation table.
> 
> Will it break compatibility?

Yes, it breaks compatibility.

native/cocos/renderer/gfx-vulkan/VKInputAssembler.cpp
minggo: Ok

bluesky013:
Close due to incompatibility resize behavior.

PR #13061, opened 2022-10-17
Add force stop and resume force stop application

qiuguohua:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

* [x] changes public API, and have ensured backward compatibility with [deprecated features](https://github.com/cocos/cocos-engine/blob/v3.5.0/docs/contribution/deprecated-features.md).

https://github.com/cocos/cocos-editor/pull/1016

### DIFF
--- a/cocos/game/game.ts
+++ b/cocos/game/game.ts
@@ -523,7 +523,8 @@ export class Game extends EventTarget {
      * @en Called by the engine to pause the game and will not be automatically resumed.
      * @zh 提供给引擎调用暂停游戏接口，不会主动恢复。
      */
-    public pauseByEngine () {
+    private pauseByEngine () {
+        if (this._paused) { return; }
         this._pausedByEngine = true;
         this.pause();
     }
@@ -532,9 +533,11 @@ export class Game extends EventTarget {
      * @en Resume paused game by engine call.
      * @zh 提供给引擎调用恢复暂停游戏接口。
      */
-    public resumeByEngine () {
-        this.resume();
-        this._pausedByEngine = false;
+    private resumeByEngine () {
+        if (this._pausedByEngine) {
+            this.resume();
+            this._pausedByEngine = false;
+        }
     }
 
     /**
@@ -564,10 +567,6 @@ export class Game extends EventTarget {
      * @zh 恢复游戏主循环。包含：游戏逻辑，渲染，事件处理，背景音乐和所有音效。
      */
     public resume () {
-        if (this._pausedByEngine) {
-            return;
-        }
-        this._pausedByEngine = false;
         if (!this._paused) { return; }
         // @ts-expect-error _clearEvents is a private method.
         input._clearEvents();
@@ -986,12 +985,12 @@ export class Game extends EventTarget {
 
     private _onHide () {
         this.emit(Game.EVENT_HIDE);
-        this.pause();
+        this.pauseByEngine();
     }
 
     private _onShow () {
         this.emit(Game.EVENT_SHOW);
-        this.resume();
+        this.resumeByEngine();
     }
 
     //  @ Persist root node section
--- a/cocos/game/game.ts
+++ b/cocos/game/game.ts
@@ -432,6 +432,7 @@ export class Game extends EventTarget {
     private _engineInited = false; // whether the engine has inited
     private _rendererInitialized = false;
     private _paused = true;
+    private _forcePaused = false;
     // frame control
     private _frameRate = 60;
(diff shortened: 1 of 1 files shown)


minggo:
Why modify like this?

qiuguohua:
> Why modify like this?

The message callbacks for receiving pauses and resumes are the same interface as the manual control pauses and resumes. A variable needs to be added to identify the different operations.

minggo:
Got it. There are two ways to pause/resume a game:
- by game codes
- by engine codes

So need to distinguish them.

`forcePaused ` is confused. Both paused by game codes or by engine, both want to force to pause. So need to use a more meaningful name.

Another way is to have different functions, such as pauseByEngine/resumeByEngine, to pause/resume a game by engine internally.

qiuguohua:
> Got it. There are two ways to pause/resume a game:
> 
> * by game codes
> * by engine codes
> 
> So need to distinguish them.
> 
> `forcePaused ` is confused. Both paused by game codes or by engine, both want to force to pause. So need to use a more meaningful name.
> 
> Another way is to have different functions, such as pauseByEngine/resumeByEngine, to pause/resume a game by engine internally.

Yes.

I've thought about using 'manual' versus 'automatic' before, and I've thought about adding an interface, but I'm not sure if adding an interface would affect compatibility issues.

I'm not using callers to distinguish them, what I understand is that when the app switches to the background, it pauses the game. When switching to the app from the background, it resumes the game. This is a natural behavior. And the user clicking pause does not want to resume the app when switching back from the background. So I think it's ok to use ```forcePaused```

minggo:
> And the user clicking pause does not want to resume the app when switching back from the background. So I think it's ok to use forcePaused

Yep. But game codes should manually pass true to pause a game according new design. It breaks compatibility.

qiuguohua:
Ok, I'll add new interface to implement it.

cocos/game/game.ts
```diff
@@ -518,6 +519,24 @@ export class Game extends EventTarget {
         director.tick(this.frameTime / 1000);
     }
 
+    /**
+     * @en Called by the engine to pause the game and will not be automatically resumed.
+     * @zh 提供给引擎调用暂停游戏接口，不会主动恢复。
+     */
+    public pauseByEngine () {
```
minggo: This function should be invoked by engine, which means, when app enter background, engine will pause it. But, currently, engine invokes pause when enter to foreground. And it will automatically resume by engine, not by game codes or editor.

qiuguohua: Yes, it's a difference between an active call and a passive call, it's a forced and non-forced operation

qiuguohua: Is 'ManualControlPause' or 'ForcedPause' used?

cocos/game/game.ts
```diff
@@ -518,6 +519,27 @@ export class Game extends EventTarget {
         director.tick(this.frameTime / 1000);
     }
 
+    /**
+     * @en Called by the engine to pause the game and will not be automatically resumed.
+     * @zh 提供给引擎调用暂停游戏接口，不会主动恢复。
```
minggo: The comment seems not correct. If paused by engine, then will be resumed by engine automatically.

cocos/game/game.ts
```diff
@@ -518,6 +519,27 @@ export class Game extends EventTarget {
         director.tick(this.frameTime / 1000);
     }
 
+    /**
+     * @en Called by the engine to pause the game.
+     * @zh 提供给引擎调用暂停游戏接口。
+     */
+    private pauseByEngine () {
+        if (this._paused) { return; }
```
minggo, line 527: Why not check this._pausedByEngine instead?

qiuguohua: This function is called by the engine, and the engine calls it every time it switches the background. The _paused check is to avoid restoring the 'pause' called by the user.

minggo: > The _paused check is to avoid restoring the 'pause' called by the user.

Could you please describe it in detail? What's error if not check it?

qiuguohua: It is not necessary to check _pausedByEngine , because it is called and resumed every time the background is switched, and checking _paused is necessary to distinguish it from the pause called by the user. If you use the _pausedByEngine check, there is no difference from the previous use of the pause interface

minggo: Seems reasonable.

minggo (approved):

Issue #12839, opened 2022-09-21
[3.4.2 ~ 3.6.0] After using game.restart(), the game's memory grows and a memory leak occurs


zhefengzhang:
### Cocos Creator version

3.4.2 ~ 3.6.0

### System information

ios 15.6.1

### Issue description

![图片](https://user-images.githubusercontent.com/35944775/191446489-75ea50c5-1e65-412e-8d6d-13cb26b4c669.png)

### Steps to reproduce

Build IOS projects and test them on IOS phones.

[momoryleak.zip](https://github.com/cocos/cocos-engine/files/9614240/momoryleak.zip)

novain1102:
follow up in internal issue: https://github.com/cocos/3d-tasks/issues/14034

novain1102:
Test for iOS/macOS(metal), Android/Windows (gles3, vulkan);
Memory do not grow significantly on Android/Windows when restart:
![wecom-temp-89702-c15d367762953b7963d0b68db337e133](https://user-images.githubusercontent.com/18634782/194695896-928c445d-2262-48ab-9149-9fd70a700207.png)
![image](https://user-images.githubusercontent.com/18634782/194697209-572b527d-b60b-48d8-9a03-6eb637be8c5c.png)
![image](https://user-images.githubusercontent.com/18634782/194697230-c12bfdc1-7f8c-453b-94aa-7d3ffc91dc15.png)

![image](https://user-images.githubusercontent.com/18634782/194697842-bb5f329b-cc1c-4600-80ea-fb259b769d39.png)

There is a gradually rising resource leakage only on iOS/macOS(metal) after each restart(although not as obvious as described in the issue). 
![image](https://user-images.githubusercontent.com/18634782/194701282-9adb6b30-0be8-4400-8b9a-12bd563b169d.png)
Please have a check with gfx related codes @hana-alice.

novain1102:
![企业微信截图_481f0b31-1da9-42e2-a70a-d94c474fa97e](https://user-images.githubusercontent.com/18634782/194735047-8868b33e-ef1c-4929-a73a-7651207151c4.png)

novain1102:
Resolved: https://github.com/cocos/cocos-engine/pull/12995

hana-alice:
Do not check memory allocation by this indicator, use Instruments->leaks->allocations instead. Focus on allocations, the anonymous VM part is not controlled by user but the system.

novain1102:
[...]

PR #13076, opened 2022-10-18
fix JSBLoadFont memory leak & refine implementation

novain1102:
Re: #https://github.com/cocos/3d-tasks/issues/14034
https://github.com/cocos/cocos-engine/issues/12839
### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/bindings/manual/jsb_platform_apple.mm
+++ b/native/cocos/bindings/manual/jsb_platform_apple.mm
@@ -42,53 +42,6 @@ const ccstd::unordered_map<ccstd::string, ccstd::string> &getFontFamilyNameMap()
     return _fontFamilyNameMap;
 }
 
-static ccstd::vector<ccstd::string> getAvailableFontFamilyNames() {
-    ccstd::vector<ccstd::string> ret;
-
-#if CC_PLATFORM == CC_PLATFORM_MACOS
-    CFArrayRef allFamilyNames = CTFontManagerCopyAvailableFontFamilyNames();
-#else
-    CFArrayRef allFamilyNames = (CFArrayRef)[[NSClassFromString(@"UIFont") familyNames] retain];
-#endif
-
-    char buf[256] = {0};
-    for (CFIndex i = 0; i < CFArrayGetCount(allFamilyNames); i++) {
-        CFStringRef fontName = (CFStringRef)CFArrayGetValueAtIndex(allFamilyNames, i);
-        if (CFStringGetCString(fontName, buf, sizeof(buf), kCFStringEncodingUTF8) != 0) {
-            ret.push_back(buf);
-        }
-    }
-
-    CFRelease(allFamilyNames);
-    return ret;
-}
-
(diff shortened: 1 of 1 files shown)


native/cocos/bindings/manual/jsb_platform_apple.mm
```diff
@@ -145,12 +95,14 @@ static bool JSB_loadFont(se::State &s) {
         }
 
         if (succeed) {
-            const auto &familyNamesAfterRegister = getAvailableFontFamilyNames();
-            ccstd::string familyName = getFontFamilyByCompareAvailableFontFamilyNames(familyNamesBeforeRegister, familyNamesAfterRegister);
+            CFStringRef fontName = CGFontCopyFullName(font);
+            ccstd::string familyName([(NSString *)fontName UTF8String]);
```
minggo, line 99: It is a new way to get family name. Could you please explain why it can work.

dumganhar: Also, is the familyName the same as old approach ? If it's the same, it's good for me of this change since it use less code to get the same result.

novain1102: The previous implementation uses `getFontFamilyByCompareAvailableFontFamilyNames()` to find the newly added FamilyName of `CTFontManagerRegisterGraphicsFont(font, &error)`. `CGFontCopyFullName` can achieve the same purpose and has the following advantages:
1. No errors in special cases: the previous implementation cannot recognize fonts with similar names such as `Open Sans Ragular/Open Sans Italic/Open Sans Bold/Open Sans BoldItalic` are passed in. Which will be recognize as Open Sans so not newly added one and run wrong logic in getFontFamilyByCompareAvailableFontFamilyNames(): afterLen == beforeLen
image
image
3. No memory leak: For the case of 1, the key-value pair cannot be set correctly for `Open Sans Ragular`/`Open Sans Italic`/`Open Sans Bold`/`Open Sans BoldItalic` to _fontFamilyNameMap. Assume _fontFamilyNameMap.size() == 4, but actually 1; So when restart engine, CTFontManagerRegisterGraphicsFont(font, &error) will be repeatedly called to cause memory leaks!
2. More efficient: the complexity of `getFontFamilyByCompareAvailableFontFamilyNames()` is O(n*n). `CGFontCopyFullName() may be O(1)?`

mmyduckx: > Yes, Verified.

novain1102: > Also, is the familyName the same as old approach ? If it's the same, it's good for me of this change since it use less code to get the same result.

Yes, Verified.

dumganhar: That's cool. Great work. 👍

dumganhar (approved):

bofeng-song (approved):

minggo (approved):

PR #13148, opened 2022-10-26
Fix core import issue

minggo:
### Changelog

* import `binarySearchEpsilon` and `easing` from core module
* remove unused file `motion-path-helper.ts`
* improve docs for `array.ts`
* use array module to access array related functions
* array is exported as a namespace instead of an object, the usage keeps the same

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


cocos/core/utils/js.ts
```diff
@@ -61,18 +60,21 @@ import {
...
 
 export * from './js-typed';
 export { default as IDGenerator } from './id-generator';
 export { default as Pool } from './pool';
-export const array = jsarray;
+export { array };
 
+/**
+ * @deprecated all functions expect deprecated function are exported by core.js
+ */
```
PPpro:
```suggestion
/**
 * @deprecated since v3.7.0, `js.js` is deprecated, please access `js` directly instead.
 */
```

PPpro: please add 
```
markAsWarning(js, 'js', [
    {
        name: 'js',
        suggest: `'js.js' is deprecated since v3.7.0, please access 'js' directly instead.`,
    },
]);
```
in a new script called `deprecated-3.7.0.ts` 
and import  `deprecated-3.7.0.ts`  in core/index.ts

this prints deprecated warning at engine running time

SantyWang (approved):

cocos/core/utils/js.ts
minggo: There is already a deprecated js file, why need to add a new one for new version? As not any deprecated API can be removed since v3.x.

PPpro (approved):

cocos/core/utils/js.ts
PPpro: > There is already a deprecated js file, why need to add a new one for new version? 

the versioned deprecate file is for deprecate module culling by version range , for detailed docs https://github.com/cocos/cocos-engine/blob/develop/docs/contribution/deprecated-features.md

shrinktofit (approved):

PR #13151, opened 2022-10-26
V2.4.11

wuweikd:
[Intervention] Ignored attempt to cancel a touchstart event with cancelable=false, for example because scrolling is in progress and cannot be interrupted.


wuweikd:
下面这种bug

https://forum.cocos.org/t/scrollview-web-mobile-bug/87749


https://forum.cocos.org/t/topic/141717

PR #13162, opened 2022-10-27
[Bug fix] Add detections for innerAudioContext.

timlyeee:
1. Pausing an innerAudioContext which is not fully initialized, the currentTime might be null.-https://github.com/cocos/3d-tasks/issues/14080
2. Pause means nothing for a stopped audio
3. InnerAudioContext.destroy behavior is insane that it might not stop the audio immediatly,

### DIFF
--- a/pal/audio/minigame/player-minigame.ts
+++ b/pal/audio/minigame/player-minigame.ts
@@ -85,13 +85,17 @@ export class AudioPlayerMinigame implements OperationQueueable {
             eventTarget.emit(AudioEvent.PLAYED);
             if (this._needSeek) {
                 this._needSeek = false;
+                console.log(`try seek`);
                 this.seek(this._cacheTime).catch((e) => {});
             }
         };
         innerAudioContext.onPlay(this._onPlay);
         this._onPause = () => {
             this._state = AudioState.PAUSED;
-            this._cacheTime = this._innerAudioContext.currentTime;
+            // NOTE: On Wechat, the current time of inner audio context might be null as it's not fully initilized.
+            if (this._innerAudioContext && this._innerAudioContext.currentTime !== null) {
+                this._cacheTime = this._innerAudioContext.currentTime;
+            }
             eventTarget.emit(AudioEvent.PAUSED);
         };
         innerAudioContext.onPause(this._onPause);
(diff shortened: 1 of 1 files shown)


pal/audio/minigame/player-minigame.ts
```diff
@@ -85,13 +85,17 @@ export class AudioPlayerMinigame implements OperationQueueable {
             eventTarget.emit(AudioEvent.PLAYED);
             if (this._needSeek) {
                 this._needSeek = false;
+                console.log(`try seek`);
```
minggo: This log is meaningless. Should describe where or why to try seek.

minggo (approved):

pal/audio/minigame/player-minigame.ts
```diff
@@ -91,7 +91,10 @@ export class AudioPlayerMinigame implements OperationQueueable {
         innerAudioContext.onPlay(this._onPlay);
         this._onPause = () => {
             this._state = AudioState.PAUSED;
-            this._cacheTime = this._innerAudioContext.currentTime;
+            // NOTE: On Wechat, the current time of inner audio context might be null as it's not fully initilized.
+            if (this._innerAudioContext && this._innerAudioContext.currentTime !== null) {
```
PPpro:
```suggestion
            if (WECHAT && this._innerAudioContext && this._innerAudioContext.currentTime !== null) {
```

PPpro: please add a `WECHAT` macro so that we can know this is a HACK for wechat platform

PPpro (changes requested):

pal/audio/minigame/player-minigame.ts
timlyeee: No, the update of the _cacheTime goes for all minigame platform, and we will never know if it will comes out for other new platform.

PPpro: so this fix should be for wechat only, because we haven't tested on the other platforms,
on the other platforms, `innerAudioContext.currentTime` could be a `number`, `null`, or even `undefined`

we can't cover all of these situations

PPpro (approved):

PR #13178, opened 2022-10-28
fix toggle err function name

kaxifakl:
fix toggle err function name `OnDestroy`. It will cause problem when destroy toggle component.


SantyWang (approved):

SantyWang:
Thanks for contribution

PR #13278, opened 2022-11-08
Fix MessageQueue deadlock.

bluesky013:
Re: #

### Changelog

* Fix MessageQueue deadlock.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/base/threading/MessageQueue.cpp
+++ b/native/cocos/base/threading/MessageQueue.cpp
@@ -113,7 +113,9 @@ MessageQueue::MessageQueue() {
 
 void MessageQueue::kick() noexcept {
     pushMessages();
-    _event.signal();
+
+    std::lock_guard<std::mutex> lock(_mutex);
+    _condVar.notify_all();
 }
 
 void MessageQueue::kickAndWait() noexcept {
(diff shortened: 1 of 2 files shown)


minggo:
@bluesky013 could you please describe the detail of this issue? Why it has problem with current logic.

bluesky013:
> @bluesky013 could you please describe the detail of this issue? Why it has problem with current logic.

There two shared variables in MessageQueue, the `(atomic_uint)writtenMessageCount` and the `(EventCV)event`.

`writtenMessageCount` is atomic, designed to to pass tail offset of message chunk memory.
`event` combind a `std::mutex` and a `std::condition_variable`, designed to block the consumer thread when MessageQueue is empty.

The issue occurs when the `kickAndWait` is called. The following is a flow chart of a deadlock scenario.
MessageQueue

The execution order is `Consumer::pullMsgs` -> `Main::pushMsgs` -> `Main::signal` -> `Consumer::hasMessage` -> `Consumer::wait`.

`consumer::hasMessage` uses the result of `Consumer::pullMsgs::writtenMessageCount::load` and misses the semaphore signal message. We must ensure the MessageQueue is real empty before thread blocks. So I used a lock to synchronize both `writtenMessageCount` and `event.wait`.

minggo:
Got it.
As using mutex and conditional variable, then does `writtenMessageCount` need to be atomic?

bluesky013:
> Got it. As using mutex and conditional variable, then does `writtenMessageCount` need to be atomic?

`writtenMessageCount` needs to be atomic. It is designed for the lock-free queue. 
* fetch_add in `pushMessages`
* load in `pullMessages`

The mutex is primarily designed for conditional variable.
* wait if the queue does not have new messages.
* signal when main thread kick

`pullMessages` and `executeMessage` can be out of the mutex scope technically. But the current `hasNewMessage` logic uses another `readerContext.newMessageCount`, so I have to put the `pullMessages` in the mutex scope. I'll do some more tests and try to find a better sullosion.

minggo:
Yep, i see the problem, write count may change before wait. It is better to lock as less codes as much. The best is to use lock free design. I am curious how other lock free thread pool work.

minggo (approved):

PR #13290, opened 2022-11-09
fix tween animation jitter

qinyuxiong:
Re: #

### Changelog

* the deltaTime is unstable which caused animation jitter, so we use requestAnimationFrame at some specific frame rate instead.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


pal/pacer/pacer-web.ts
```diff
@@ -1,5 +1,7 @@
 import { assertIsTrue } from '../../cocos/core/data/utils/asserts';
 
+const maxRafFrame = 60;
```
pandamicro, line 3: rAF may not be 60 per second, it could be 90 times or 120 times depends on the screen's refresh rate, we need to find a way to determine the frequency at runtime

pandamicro (changes requested):

pal/pacer/pacer-web.ts
pandamicro: Consider using the approach described here: https://gist.github.com/elundmark/38d3596a883521cb24f5#better-refined-approach

Also, we can merge the two branch in start function together using this approach, no need to use `_stTIme` anymore

PR #13503, opened 2022-11-23
Refactor animation graph evaluation

shrinktofit:
Re: #

### Changelog

* Refactor the evaluation of animation graph.

* Support additive layer, additive animation.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


cocos/animation/core/pose.ts
```diff
@@ -0,0 +1,202 @@
...
+
+export class Pose {
+    readonly transforms: TransformArray;
+
+    readonly metaValues: Float64Array;
+
+    protected constructor (transforms: TransformArray, metaValues: Float64Array) {
+        this.transforms = transforms;
+        this.metaValues = metaValues;
+    }
```
shrinktofit: Q(@holycanvas ): What if we want to support non-skeletal animations in animation graph?

A(@shrinktofit ):
1. We have to add more value types into this `Pose` class: `Vec2`, `Vec3`.... These values should all be interpolate-able.

2. We must find a way so that we can know two clip tracks animate the same target so we can connect them together.

cocos/animation/tracks/track.ts
```diff
@@ -379,6 +381,52 @@ export class TrackBinding {
...
+                    const transformHandle = bindContext.bindTransform(transformPath);
+                    if (!transformHandle) {
+                        return undefined;
+                    }
+                    return bindPoseTransform(transformHandle, lastPropertyKey);
+                }
+            }
+        }
+
+        // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```
shrinktofit: Q(@holycanvas ): Non-skeleton animations can not be used in animation graph?

A(@shrinktofit ): No, they can not be used as before.

cocos/animation/animation-clip.ts
```diff
@@ -384,6 +399,47 @@ export class AnimationClip extends Asset {
...
+        }
+
+        const trackEvaluations: AGTrackEvaluation<unknown>[] = [];
+        let exoticAnimationEvaluation: ExoticTrsAGEvaluation | undefined;
+
+        const { _tracks: tracks } = this;
+        const nTracks = tracks.length;
+        for (let iTrack = 0; iTrack < nTracks; ++iTrack) {
+            const track = tracks[iTrack];
+            if (track instanceof UntypedTrack) {
```
shrinktofit: This is not a breaking change: since the born of animation graph, it only accepts animations from model files.

cocos/animation/exotic-animation/exotic-animation.ts
```diff
@@ -680,6 +701,84 @@ interface ExoticTrackEvaluationRecord<TValue> {
...
+            const value = scale.evaluate(time);
+            poseTransforms.setScale(transformHandle.index, value);
+        }
+    }
+
+    private _position: ExoticTrackEvaluator<Vec3> | null = null;
+    private _rotation: ExoticTrackEvaluator<Quat> | null = null;
+    private _scale: ExoticTrackEvaluator<Vec3> | null = null;
+    private _transformHandle: TransformHandle;
+}
```
shrinktofit: Animation clip sampling is done in this way.

editor/inspector/assets/fbx/animation.js
```diff
@@ -71,6 +71,10 @@ exports.template = /* html */`
             <span slot="label">Speed</span>
             <ui-num-input slot="content" class="speed"></ui-num-input>
         </ui-prop>
+        <ui-prop>
```
arsen2010: 样式改下，改为： `<ui-prop ui="asset">`

cocos/animation/animation-clip.ts
```diff
@@ -384,6 +397,47 @@ export class AnimationClip extends Asset {
         return this._createEvalWithBinder(target, binder, context.rootMotion);
     }
 
+    public createAGEvaluation (context: AnimationClipGraphBindingContext) {
```
SantyWang: I think it is better not to put the method that creates the Evaluator in the AnimationClip and let the AnimationClip be pure data.

cocos/3d/skeletal-animation/skeletal-animation-blending.ts
```diff
@@ -354,261 +352,6 @@ export class LegacyBlendStateBuffer extends BlendStateBuffer<LegacyNodeBlendStat
...
-        Quat.copy(this.result, this._defaultValue);
-    }
-
-    private _defaultValue = new Quat();
-    private _clipBlendResult = new Quat();
-    private _accumulatedWeight = 0.0;
-}
-
-class LayeredNodeBlendState extends NodeBlendState<LayeredVec3PropertyBlendState, LayeredQuatPropertyBlendState> {
-    public setLayerMask (layerIndex: number) {
```
SantyWang, line 444: Why delete these class? What is the alternative solution

cocos/animation/exotic-animation/exotic-animation.ts
```diff
@@ -680,6 +701,86 @@ interface ExoticTrackEvaluationRecord<TValue> {
...
+    public evaluate (time: number, pose: Pose) {
+        this._nodeEvaluations.forEach((nodeEvaluator) => {
+            nodeEvaluator.evaluate(time, pose);
+        });
+    }
+
+    private _nodeEvaluations: ExoticNodeAnimationEvaluatorX[];
+}
+
+class ExoticNodeAnimationEvaluatorX {
```
SantyWang: Why is there an `X` at the end of the name?

cocos/animation/core/animation-handle.ts
```diff
@@ -0,0 +1,15 @@
+export interface TransformHandle {
```
SantyWang, line 6: Can Handle be used as a pure number? Using an object I think would be wasteful and cache unfriendly, and call xxx.destroy(handle) when you need to destroy it

cocos/animation/core/pose-allocator.ts
```diff
@@ -0,0 +1,148 @@
...
+        const pageIndex = this._pages.length;
+        this._pages.push(page);
+        const pose = page.tryAllocate();
+        assertIsTrue(pose); // Shall not fail
+        pose.pageIndex = pageIndex;
+        return pose;
+    }
+}
+
+class PagedPose extends Pose {
```
SantyWang: This type is a bit wired. Business class should not be coupled with allocator,  there may be various types of allocator in the future. I think it's better to keep a `memoryId` or `allocatorId` field in the `Pose` class to be set by the allocator, and all the rest of the logic is handled outside the pose.

cocos/animation/core/pose-allocator.ts
```diff
@@ -0,0 +1,148 @@
+import { assertIsTrue } from '../../core/data/utils/asserts';
+import { Pose } from './pose';
+import { TransformArray } from './transform-array';
+
+const MAX_POSE_PER_PAGE = 8;
+
+export class PoseAllocator {
```
SantyWang, line 6: The name `PosePageAllocator` may be more appropriate, and it should be noted that the allocator type is a page allocator

cocos/3d/skeletal-animation/skeletal-animation-blending.ts
shrinktofit: It's replaced by the pose stuff.

cocos/animation/core/animation-handle.ts
shrinktofit: I have documented the reason https://github.com/cocos/cocos-engine/pull/13503/commits/83e806201ba0a147e6bc78a5f2e26e088bd1a734

cocos/animation/core/pose-allocator.ts
shrinktofit: See https://github.com/cocos/cocos-engine/pull/13503/commits/c5e7a19125bc652c5b3f52d0552a9d8b64f48ac9 for my refactor

cocos/animation/exotic-animation/exotic-animation.ts
shrinktofit: https://github.com/cocos/cocos-engine/pull/13503/commits/2843818640584bc64db480c4889c4815bc20e505

cocos/animation/core/pose-allocator.ts
shrinktofit: This class in my opnion should not expose its implementation-no matter it use pool or anything else, it should be treated as a black-box from external's view. So I advise to keep the name.

cocos/animation/animation-clip.ts
shrinktofit: https://github.com/cocos/cocos-engine/pull/13503/commits/1c9898a4ede4b8c7bdfcc692368c08aa68a9b89c

cocos/animation/core/pose-allocator.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    public get allocatedCount () {
+        return this._allocatedCount;
+    }
+
+    public allocatePose (): Pose {
+        ++this._allocatedCount;
+        const { _pages: pages } = this;
+        const nPages = pages.length;
+        for (let iPage = 0; iPage < nPages; ++iPage) {
```
SantyWang: What is the frequency of allocation during the run? If frequency is high, you may need to design the `freeList` here to avoid iterating through all pages

cocos/animation/core/pose-allocator.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    private _poses: (PagedPose | null)[];
+
+    private _createPose (index: number) {
+        const transformsByteLength = TransformArray.BYTES_PER_ELEMENT * this._transformCount;
+        const baseOffset = (transformsByteLength
+            + Float64Array.BYTES_PER_ELEMENT * this._metaValueCount) * index;
+        const transforms = new TransformArray(this._buffer, baseOffset, this._transformCount);
+        const metaValues = new Float64Array(this._buffer, baseOffset + transformsByteLength, this._metaValueCount);
+        const pose = Pose.__create(transforms, metaValues);
```
SantyWang: why use `__create`

cocos/animation/core/pose.ts
```diff
@@ -0,0 +1,203 @@
...
+     * @internal
+     */
+    public static __create (transforms: TransformArray, metaValues: Float64Array) {
+        return new Pose(transforms, metaValues);
+    }
+}
+
+export class TransformFilter {
+    constructor (involvedTransforms: readonly number[]) {
+        this._involvedTransforms = new Uint32Array(involvedTransforms);
```
SantyWang: This value is a typedarray of transform index, UInt32 seems to be too large, Uint16 is enough?

cocos/animation/core/pose.ts
```diff
@@ -0,0 +1,203 @@
...
+            blendIntoTransformArrayAt(target, source, alpha, involvedTransformIndex);
+        }
+    }
+}
+
+function copyTransformsWithFilter (target: TransformArray, source: Readonly<TransformArray>, filter: TransformFilter) {
+    const nTransforms = target.length;
+    assertIsTrue(nTransforms === target.length);
+    for (const involvedTransformIndex of filter.involvedTransforms) {
+        target.setSpan(involvedTransformIndex, source, involvedTransformIndex, 1);
```
SantyWang: what does `setSpan` means?

cocos/animation/core/transform-array.ts
```diff
@@ -0,0 +1,227 @@
...
+        return TRANSFORM_STRIDE_IN_BYTES;
+    }
+
+    constructor (length?: number);
+
+    constructor (buffer: ArrayBuffer, byteOffset?: number, length?: number);
+
+    constructor (bufferOrLength?: number | ArrayBuffer, byteOffset?: number, length_?: number) {
+        if (typeof bufferOrLength === 'undefined') {
+            this._data = new Float64Array();
```
SantyWang, line 27: Is `Float64Array` necessary? It seems wasted

cocos/animation/core/transform-array.ts
```diff
@@ -0,0 +1,227 @@
...
+        );
+        return new TransformArray(dataSliced.buffer, dataSliced.byteOffset, dataSliced.length / TRANSFORM_STRIDE_IN_FLOATS);
+    }
+
+    /**
+     * Copy a span of `source` into `this`.
+     * Equivalent to `this.set(source.slice(sourceOffset, sourceOffset + size), targetOffset)`
+     * except without perform the slicing.
+     * */
+    public setSpan (
```
SantyWang: The name is not easy to understand，I think it's functionality is `copyInto` or `fastCopyInto` ?

cocos/animation/core/transform.ts
```diff
@@ -0,0 +1,228 @@
...
+        Vec3.copy(transform._scale, Vec3.ZERO);
+        return transform;
+    })());
+
+    get position (): Readonly<Vec3> {
+        return this._position;
+    }
+
+    set position (value) {
+        Vec3.copy(this._position, value);
```
SantyWang, line 30: I think use `this._position.set(value)` will get better readability.

cocos/animation/core/transform.ts
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    public static setIdentity (out: Transform) {
+        Vec3.copy(out._position, Vec3.ZERO);
+        Quat.copy(out._rotation, Quat.IDENTITY);
+        Vec3.copy(out._scale, Vec3.ONE);
+        return out;
+    }
+
+    public static copy (out: Transform, src: ReadonlyTransform) {
```
SantyWang, line 62: Why these function use `static` method? I think it looks a bit wired

cocos/animation/core/pose-allocator.ts
SantyWang: It's OK if we won't implement another allocator

cocos/animation/core/transform.ts
SantyWang: Assignment and copy operations are usually implemented with member methods

cocos/animation/core/pose-allocator.ts
shrinktofit: The allocation happens frequencely, for example:
- Every non-additive clip motion causes 1 allocation.
- Every additive clip motion causes 2 allocation.
- Every animation blend in theory causes 1 allocation, but I optimize it to no allocation.
-------
Nice idea for the "freeList", I commited another optmization: https://github.com/cocos/cocos-engine/pull/13503/commits/4ff004213607cda398465daabf42cf008a6ce1da .

SantyWang: > The allocation happens frequencely, for example:
Every non-additive clip motion causes 1 allocation.
Every additive clip motion causes 2 allocation.
Every animation blend in theory causes 1 allocation, but I optimize it to no allocation.

Perframe ?

shrinktofit:
I refactored the pose allocator at here: https://github.com/cocos/cocos-engine/pull/13503

The idea behind: https://www.figma.com/file/vJvssEvkKCVO0WSiqlA4Zn/%E5%A7%BF%E6%80%81%E5%88%86%E9%85%8D%E5%99%A8?node-id=0%3A1&t=eNNAgUlEIhalAghT-1

cocos/animation/animation-clip.ts
```diff
@@ -646,6 +658,16 @@ export class AnimationClip extends Asset {
         this._embeddedPlayers.length = 0;
     }
 
+    /**
+     * @internal
+     */
+    public __trySyncLegacyData () {
```
SantyWang: We didn't use the double underscore style, single underscore should work just fine.

cocos/animation/core/transform.ts
shrinktofit: Other value types in engine(for example https://github.com/cocos/cocos-engine/blob/v3.7.0/cocos/core/math/vec3.ts#L73) are implemented in this way.

cocos/animation/core/transform.ts
shrinktofit: [`Vec3.set`](https://github.com/cocos/cocos-engine/blob/v3.7.0/cocos/core/math/vec3.ts#L84) does not take a vector parameter.

cocos/animation/core/pose.ts
shrinktofit: Adjusted here: https://github.com/cocos/cocos-engine/pull/13503/commits/ad6a109de0e596abe1091f099c91f68e4dc00950

cocos/animation/core/pose-allocator.ts
shrinktofit: Yes per frame

cocos/animation/core/pose-allocator.ts
shrinktofit: Cause I want to hide the constructor: https://github.com/cocos/cocos-engine/pull/13503/commits/0869907fac28046d8dc607647d690aa3fc69c771 and discourage users to `new Pose`. 

The `__create` has now been renamed into `_create` https://github.com/cocos/cocos-engine/pull/13503/commits/49c47901bd82204d8385a45f984a57c0d50fa28c

cocos/animation/core/transform-array.ts
shrinktofit: float64 can be utmostly compatible with `number`, even I'm not sure if it's neccesary.

cocos/animation/core/shared-stack-based-allocator.ts
```diff
@@ -0,0 +1,299 @@
...
+        this._locking = false;
+    }
+
+    public getPageMemory (index: number) {
+        assertIsTrue(index >= 0 && index < this._pages.length, `Page index out of range`);
+        return this._pages[index].buffer;
+    }
+
+    public pushPage (allocator: SharedStackBasedAllocator): SharedMemoryPage {
+        const oldAllocatorPageCount = allocator[allocatorPageCountTag];
```
SantyWang, line 50: Why do you need to pass in an allocator? Shouldn't the allocator call `PagedStack.push` and get the `SharedMemoryPage`?

cocos/animation/core/shared-stack-based-allocator.ts
```diff
@@ -0,0 +1,299 @@
...
+
+    private _slices: SharedMemorySlice[] = [];
+}
+
+export class SharedStackBasedAllocatorManager {
+    constructor (private _thresholds: number[]) {
+        assertIsTrue(_thresholds.every((v, i, arr) => i === 0 || v > arr[i - 1]));
+    }
+
+    public get empty () {
```
SantyWang: `isEmpty` ?

cocos/animation/exotic-animation/exotic-animation.ts
```diff
@@ -680,6 +701,86 @@ interface ExoticTrackEvaluationRecord<TValue> {
...
+    }
+
+    public destroy () {
+        this._nodeEvaluations.forEach((nodeEvaluation) => {
+            nodeEvaluation.destroy();
+        });
+    }
+
+    public evaluate (time: number, pose: Pose) {
+        this._nodeEvaluations.forEach((nodeEvaluator) => {
```
SantyWang: Avoid using `forEach` in perframe operation. This might cause garbage colloction due to the lambda function object creation.

cocos/animation/marionette/animation-blend-1d.ts
```diff
@@ -53,10 +55,12 @@ export class AnimationBlend1D extends AnimationBlend {
...
 
-    public [createEval] (context: MotionEvalContext) {
-        const evaluation = new AnimationBlend1DEval(context, this, this._items, this._items.map(({ threshold }) => threshold), 0.0);
+    public [createEval] (context: AnimationGraphLayerWideBindingContext, clipOverrides: ReadonlyClipOverrideMap | null) {
+        const evaluation = new AnimationBlend1DEval(
+            context, clipOverrides, this, this._items, this._items.map(({ threshold }) => threshold), 0.0,
+        );
         const initialValue = bindOr(
-            context,
+            context.up,
```
SantyWang: `up` is not so easy to understand.

cocos/animation/core/transform-array.ts
shrinktofit: Within the name, `set` is referenced from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/set . `span` is a synonym of `slice`, taken from https://en.cppreference.com/w/cpp/container/span . It wants to express: copy a subrange of dest array into this array at specified offset.

How do you feel about:

- `setSlice` (according to document)
- `setRange`
- `copyRange`

cocos/animation/marionette/animation-graph-animation-clip-binding.ts
```diff
@@ -0,0 +1,349 @@
...
+ * The pose euler angles binding describes how to get/set the rotation(in euler angles) of a bound transform in animation graph.
+ */
+class PoseEulerAnglesBinding extends PoseBindingBase implements PoseBinding<Vec3> {
+    public setValue (value: Vec3, pose: Pose): void {
+        const quat = Quat.fromEuler(PoseEulerAnglesBinding._EULER_TO_QUAT_CACHE, value.x, value.y, value.z);
+        pose.transforms.setRotation(this._transformHandle.index, quat);
+    }
+
+    public getValue (pose: Pose) {
+        const q = pose.transforms.getRotation(this._transformHandle.index, CACHE_QUAT_GET_VALUE) as Readonly<Quat>;
```
SantyWang, line 136: This may cause some bugs! Because we are storing quaternions, so this will cause us to get different values for setValue and getValue when we set them using eulerAngles. In particular, when the incoming eulerAngles are Vec3(90, 90, 90). Need to be careful here.

cocos/animation/marionette/animation-graph-animation-clip-binding.ts
```diff
@@ -0,0 +1,349 @@
...
+     */
+    public evaluate (time: number, context: AnimationClipGraphEvaluationContext) {
+        const {
+            _trackEvaluations: trackEvaluations,
+            _exoticAnimationEvaluation: exoticAnimationEvaluation,
+        } = this;
+
+        const pose = context;
+
+        for (const trackEvaluation of trackEvaluations) {
```
SantyWang: Avoid using `for of` in perframe operetion.  The compiled code has additional overhead.
image

cocos/animation/core/transform-array.ts
SantyWang: `setRange` and `copyRange` are better for me.

cocos/animation/animation-clip.ts
shrinktofit: Tweaked: https://github.com/cocos/cocos-engine/pull/13503/commits/ebdf353008e6a08cc0dc90dbb791f7790259a913

cocos/animation/core/transform-array.ts
SantyWang: float32 precision should be enough, here you need to save a little memory, you can refer to the precision used by other engines.

cocos/animation/core/shared-stack-based-allocator.ts
shrinktofit: If it's `isEmpty`, it should be a function instead of a property right?

SantyWang: I think it's ok to use `isEmpty` in a property. why need a function ?

cocos/animation/marionette/animation-graph-animation-clip-binding.ts
shrinktofit: https://github.com/cocos/cocos-engine/pull/13503/commits/1780961fe0e8bd27cf0a5c145aecde1ef759ce97

cocos/animation/exotic-animation/exotic-animation.ts
shrinktofit: https://github.com/cocos/cocos-engine/pull/13503/commits/1780961fe0e8bd27cf0a5c145aecde1ef759ce97

cocos/animation/marionette/animation-graph-context.ts
```diff
@@ -0,0 +1,560 @@
...
+ *
+ * At each evaluation phase:
+ * - Creates a (or reuse a) `AnimationGraphEvaluationContext`.
+ * - Do the evaluation, generate a pose.
+ * - Call `PoseLayoutMaintainer.apply()` to apply the pose into scene graph.
+ *
+ * When an override-clip request is fired, the binding phase is performed again.
+ */
+
+function findBoneByNameRecurse (from: Node, name: string): Node | null {
```
SantyWang: -> `findBoneByNameRecursively`?

cocos/animation/marionette/animation-graph-context.ts
```diff
@@ -0,0 +1,560 @@
...
+export interface AnimationGraphLayerWideBindingContext {
+    /**
+     * Indicates if the current layer is an additive layer.
+     */
+    additive: boolean;
+
+    /**
+     * The upper binding context.
+     */
+    up: AnimationGraphBindingContext;
```
SantyWang: -> `owner` or `parent` ?

cocos/animation/marionette/animation-graph-context.ts
```diff
@@ -0,0 +1,560 @@
...
+        transformRecords.splice(newNodeIndex, 0, transformRecord);
+
+        return transformRecord.handle;
+    }
+
+    @checkBindStatus(true)
+    public getOrCreateMetaValueBinding (name: string) {
+        const { _metaValueRecords: metaValueRecords } = this;
+
+        const metaValueIndex = metaValueRecords.findIndex((record) => record.name === name);
```
SantyWang, line 227: What does the metaValue contain?

cocos/animation/marionette/animation-graph-context.ts
```diff
@@ -0,0 +1,560 @@
...
+        defaultTransform.position = node.position;
+        defaultTransform.rotation = node.rotation;
+        defaultTransform.scale = node.scale;
+        this.defaultTransform = defaultTransform;
+    }
+
+    /** The order of the transform. */
+    public order = -1;
+
+    public refs = 1;
```
SantyWang: -> refCount ?

cocos/animation/core/shared-stack-based-allocator.ts
shrinktofit: The `PagedStack` is shared by multiple allocators. A `pushPage()` request:

- may really push a page onto the stack - if current real page count is not sufficient to satisfy the requesting allocator.

- or may not involve page creation since current real page count is enough.

For example, at a time, the paged stack and allocators may be like(_`·` denotes a page_):

```
Stack      |  ·····>  (real stack,    5 pages)
Allocator1 |  ·····>  (virtual stack, 5 pages)
Allocator2 |  ···>    (virtual stack, 3 pages)
```

In `Stack`, page 0,1,2 has two allocator users, while page 3,4 has only one allocator user.

Now a `pushPage()` is requested from `Allocator2`. Since `Allocator2` needs 4 pages and the stack is sufficient, the stack does not grow. Only some flags are updated:

```
Stack      |  ·····>  (real stack,    5 pages)
Allocator1 |  ·····>  (virtual stack, 5 pages)
Allocator2 |  ····>   (virtual stack, 4 pages) // growed
```

Then, a `pushPage()` is requested from `Allocator1`. Since `Allocator1` needs 6 pages and the stack is insufficient, the stack needs to grow:

```
Stack      |  ······> (real stack,    6 pages) // growed
Allocator1 |  ······> (virtual stack, 6 pages) // growed
Allocator2 |  ····>   (virtual stack, 4 pages)
```

As you can see from above: **the stack has to know which allocator is firing the incoming `pushPage()` request**, and calculates its required pages(simply get its count and increase by 1) to determinate if a real stack grow is needed.

cocos/animation/marionette/animation-graph-animation-clip-binding.ts
shrinktofit: Seems like there's no way out. It's a pitfall of `Node` itself.

`getValue` is invoked in "component-wise animation" context. For example user has only `.x` animation in the node's rotation.

cocos/animation/marionette/animation-graph-context.ts
shrinktofit: It contains "animation curve" value, but we haven't implemented yet. Does it matter?

cocos/animation/marionette/animation-graph-context.ts
shrinktofit: https://github.com/cocos/cocos-engine/pull/13503/commits/2dee925357f662facd6a15792a45d4cdc9e46298

cocos/animation/marionette/animation-graph-context.ts
shrinktofit: https://github.com/cocos/cocos-engine/pull/13503/commits/0d8985ed9e979df3aee63cb6f9073e0b10a28366

cocos/animation/marionette/animation-graph-context.ts
shrinktofit: `outerContext`: https://github.com/cocos/cocos-engine/pull/13503/commits/a5080fb17a413877ebdb87748fd4bf1be66908fb

cocos/animation/marionette/animation-blend-1d.ts
shrinktofit: outerContext: https://github.com/cocos/cocos-engine/pull/13503/commits/a5080fb17a413877ebdb87748fd4bf1be66908fb

cocos/animation/marionette/animation-graph-context.ts
SantyWang: If the only data here is the animation curve, how about a more intuitive name, extraCurveData or something like that? `meta value` is too general.

shrinktofit: I'm not ready to rename it. I plan to rename it once the animation curve feature is implemented.

cocos/animation/core/shared-stack-based-allocator.ts
shrinktofit: Agreed.

shrinktofit: https://github.com/cocos/cocos-engine/pull/13503/commits/5c4b1ba7c485c466ba071aec373f55e87c813bb5

cocos/animation/marionette/animation-graph-context.ts
SantyWang: ok

cocos/animation/core/transform-array.ts
shrinktofit: https://github.com/cocos/cocos-engine/pull/13503/commits/38d60e6334156bccb6431b4be67d004266fa9bda

SantyWang (approved):

PR #13557, opened 2022-11-25
fix spine cached color format

zhakesi:


native/cocos/editor-support/spine-creator-support/SkeletonCache.cpp
```diff
@@ -317,15 +319,15 @@ void SkeletonCache::renderAnimationFrame(AnimationData *animationData) {
             _clipper->clipEnd(*slot);
             continue;
         }
-
+        spine::Color slotColor = slot->getColor();
```
minggo: It is better to use const reference.

native/cocos/editor-support/spine-creator-support/SkeletonCache.cpp
```diff
@@ -394,32 +388,42 @@ void SkeletonCache::renderAnimationFrame(AnimationData *animationData) {
...
+            darkColor.a = 0;
         } else {
             darkColor.r = 0;
             darkColor.g = 0;
             darkColor.b = 0;
+            darkColor.a = 0;
         }
-        darkColor.a = 0;
+
+        finalColor.r = static_cast<uint8_t>(round(color.r));
```
minggo: Use `std::round()` instead?

native/cocos/editor-support/spine-creator-support/SkeletonCache.cpp
zhakesi: done

native/cocos/editor-support/spine-creator-support/SkeletonCache.cpp
zhakesi: done

SantyWang (approved):

minggo (approved):

PR #13631, opened 2022-12-02
fix:android activity jni global reference leak

oahc09:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/platform/android/AndroidPlatform.cpp
+++ b/native/cocos/platform/android/AndroidPlatform.cpp
@@ -567,6 +567,7 @@ int AndroidPlatform::init() {
 void AndroidPlatform::onDestroy() {
     UniversalPlatform::onDestroy();
     unregisterAllInterfaces();
+    JniHelper::onDestroy();
     CC_SAFE_DELETE(_inputProxy)
 }
 
--- a/native/cocos/platform/java/jni/JniHelper.cpp
+++ b/native/cocos/platform/java/jni/JniHelper.cpp
@@ -144,12 +144,6 @@ void JniHelper::onDestroy() {
             cc::JniHelper::getEnv()->DeleteGlobalRef(JniHelper::sActivity);
             JniHelper::sActivity = nullptr;
         }
-
-        if(JniHelper::classloader) {
-            cc::JniHelper::getEnv()->DeleteGlobalRef(JniHelper::classloader);
-            JniHelper::classloader = nullptr;
-        }
-        JniHelper::classloaderCallback = nullptr;
         LOGD("JniHelper::onDestroy");
     }
 }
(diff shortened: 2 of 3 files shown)


minggo (approved):

native/cocos/platform/android/AndroidPlatform.cpp
```diff
@@ -567,6 +567,7 @@ int AndroidPlatform::init() {
 void AndroidPlatform::onDestroy() {
     UniversalPlatform::onDestroy();
     unregisterAllInterfaces();
+    JniHelper::onDestroy();
```
bofeng-song, line 570: It's too early to destroy jni env;  jni method after may be called another thread finish its task.

bofeng-song: Consider calling jni's destroy on the finalize interface of activity

oahc09: In JniHeper::onDestroy not destroyed jni env, just release activity's global reference

oahc09: why put JniHelper::onDestroy here, another reason is that 
    CC_SAFE_DELETE(_inputProxy) will trigger _androidPlatform->_app->activity->vm->DetachCurrentThread();

oahc09: > the finalize interface of activity

 APP_CMD_DESTROY has already been the finalize interface of activity
<html>
<body>


APP_CMD_DESTROY | Command from main thread: the app's activity is being destroyed, and waiting for the app thread to clean up and exit before proceeding.
-- | --



</body>
</html>

bofeng-song: > activity's

classloader may be used after calling destroy

oahc09: > activity's

if activity has destroyed, it should not has any jni call from engine api, because at this time engine has paused->destroyed. Developer has no chance call any engine api associated with jni.

oahc09: if engine internal has chance to use classloader  after destroyed, i think we should avoid or solve this condition.

bofeng-song: > if engine internal has chance to use classloader after destroyed, i think we should avoid or solve this condition.

after Activity's destroy called, download tasks may be need to call cancel, thus classloader need to be used

bofeng-song: We can release jniObjectActivity only here, classloader's lifecycle is the same as application

oahc09: > We can release jniObjectActivity only here, classloader's lifecycle is the same as application

if download tasks cannot cancel before destory, JniHelper::classloader must add double check to avoid NewGlobalRef many times.

oahc09: > We can release jniObjectActivity only here, classloader's lifecycle is the same as application

new commit
improve:only release jniObjectActivity,keep classloader
add nullptr check for classloader to avoid new global reference many times

oahc09:
@cocos-robot run test cases

bofeng-song (approved):

oahc09:
@cocos-robot run test cases

PR #13674, opened 2022-12-06
Fix material inspector missing elements

dumganhar:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/renderer/core/PassUtils.cpp
+++ b/native/cocos/renderer/core/PassUtils.cpp
@@ -362,4 +362,34 @@ MaterialProperty toMaterialProperty(gfx::Type type, const ccstd::vector<float> &
     return ret;
 }
 
+bool macroRecordAsBool(const MacroRecord::mapped_type &v) {
+    if (ccstd::holds_alternative<bool>(v)) {
+        return ccstd::get<bool>(v);
+    }
+
+    if (ccstd::holds_alternative<int32_t>(v)) {
+        return ccstd::get<int32_t>(v);
+    }
+
+    if (ccstd::holds_alternative<ccstd::string>(v)) {
+        return ccstd::get<ccstd::string>(v) == "true";
+    }
+    return false;
+}
+
+ccstd::string macroRecordAsString(const MacroRecord::mapped_type &v) {
+    if (ccstd::holds_alternative<ccstd::string>(v)) {
+        return ccstd::get<ccstd::string>(v);
+    }
+
+    if (ccstd::holds_alternative<bool>(v)) {
+        return ccstd::get<bool>(v) ? "1" : "0";
+    }
+
+    if (ccstd::holds_alternative<int32_t>(v)) {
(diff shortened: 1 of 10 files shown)


native/cocos/renderer/core/PassUtils.cpp
```diff
@@ -362,4 +362,34 @@ MaterialProperty toMaterialProperty(gfx::Type type, const ccstd::vector<float> &
     return ret;
 }
 
+bool macroRecordAsBool(const MacroRecord::mapped_type &v) {
+    if (ccstd::holds_alternative<bool>(v)) {
+        return ccstd::get<bool>(v);
+    }
+
+    if (ccstd::holds_alternative<int32_t>(v)) {
+        return ccstd::get<int32_t>(v);
```
minggo: It returns a bool value?

native/cocos/renderer/core/PassUtils.h
```diff
@@ -113,4 +113,7 @@ bool overrideMacros(MacroRecord &target, const MacroRecord &source);
 
 MaterialProperty toMaterialProperty(gfx::Type type, const ccstd::vector<float> &vec);
 
+bool macroRecordAsBool(const MacroRecord::mapped_type &v);
+ccstd::string macroRecordAsString(const MacroRecord::mapped_type &v);
```
minggo, line 117: How about rename to `macroRecordToBool` and `macroRecordToString `?

native/cocos/renderer/core/PassUtils.cpp
dumganhar: Yes, treat int as a bool, it's moved from ProgramLib.cpp.

Effect data may store `0` or `1` for bool type.

native/cocos/renderer/core/PassUtils.h
dumganhar: I think `As` means that it's a bool in general, but if it's not , convert it.
`To` means the source is not a `bool`, need to convert it.
@PatriceJiang What do you think? The function name was named by you originally.

native/cocos/renderer/core/PassUtils.cpp
minggo: I mean should change it explicitly, for example
```c++
return ccstd::get<int32_t>(v) != 0;
```

native/cocos/renderer/core/PassUtils.h
minggo: Why i suggest to use `to` is because it convert from `MacroRecord::mapped_type` to another type.

native/cocos/renderer/core/PassUtils.cpp
dumganhar: yep, it's better to . I just moved it.

dumganhar: DONE

minggo (approved):

PR #13740, opened 2022-12-08
fixed #12670: [doc] Swig documentation about %arg(...) and not need to add const prefix for attribute

### LINKED ISSUE #12670: [doc] Swig documentation about %arg(...) and not need to add const prefix for attribute

dumganhar:
Re: #12670

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/tools/swig-config/tutorial/index.md
+++ b/native/tools/swig-config/tutorial/index.md
@@ -748,13 +748,13 @@ Object.defineProperty(MyNewClass.prototype, 'width', {
 
 #### Reference type
 
-If C++ `get` and `set` function return reference type, don't forget to add `&` suffix in %attribute or %attribute_writeonly directives. The following `ccstd::string&` is an example.
+If C++ `get` function returns a reference data type or `set` function accesses a reference data type , don't forget to add `&` suffix in %attribute or %attribute_writeonly directives. The following `ccstd::string&` is an example.
 
 ```c++
 %attribute_writeonly(cc::ICanvasRenderingContext2D, ccstd::string&, fillStyle, setFillStyle);
 ```
 
-If `&` is missing, temporary `ccstd::string` objects will be created while binding functions are invoked.
+If `&` is missing, a temporary `ccstd::string` instance will be created while the binding function is invoked.
 
 #### %arg() directive
 
--- a/native/tools/swig-config/tutorial/index.md
+++ b/native/tools/swig-config/tutorial/index.md
@@ -1,16 +1,16 @@
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
-**Table of Contents** 
+**Table of Contents**
 
 - [The Tutorial of Swig Workflow in Cocos Creator](#the-tutorial-of-swig-workflow-in-cocos-creator)
-  - [How to Bind a New Module in Engine](#how-to-bind-a-new-module-in-engine)
-    - [Add a new module interface file](#add-a-new-module-interface-file)
-    - [Modify swig-config.js](#modify-swig-configjs)
-    - [Generate bindings](#generate-bindings)
-    - [Modify `engine/native/cocos/CMakeLists.txt`](#modify-enginenativecocoscmakeliststxt)
-    - [Register the new module to Script Engine](#register-the-new-module-to-script-engine)
-  - [How to Bind a New Module in Developer's Project](#how-to-bind-a-new-module-in-developers-project)
-    - [Bind a simple class](#bind-a-simple-class)
+  * [How to Bind a New Module in Engine](#how-to-bind-a-new-module-in-engine)
+    + [Add a new module interface file](#add-a-new-module-interface-file)
+    + [Modify swig-config.js](#modify-swig-configjs)
+    + [Generate bindings](#generate-bindings)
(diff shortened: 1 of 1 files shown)


native/tools/swig-config/tutorial/index.md
```diff
@@ -612,6 +612,28 @@ Build and run project, get log:
 
 ### Define attributes which bind C++ getter and setter as a JS property
 
+### Usage
+
+1. Define an attribute (JS property) without setter
+
+   ```c++
+   %attribute(your_namespace::your_class_name, cpp_member_variable_type, js_property_name, cpp_getter_name)
```
minggo: cpp_getter_name -> cpp_getter_function_name?

native/tools/swig-config/tutorial/index.md
```diff
@@ -689,6 +711,95 @@ Build and run project
...
+
+```c++
+%attribute_writeonly(cc::ICanvasRenderingContext2D, float, width, setWidth);
+%attribute_writeonly(cc::ICanvasRenderingContext2D, float, height, setHeight);
+%attribute_writeonly(cc::ICanvasRenderingContext2D, float, lineWidth, setLineWidth);
+%attribute_writeonly(cc::ICanvasRenderingContext2D, ccstd::string&, fillStyle, setFillStyle);
+%attribute_writeonly(cc::ICanvasRenderingContext2D, ccstd::string&, font, setFont);
+```
+
+This is the similar functionality like which in JS:
```
minggo: I think can remove `like which`. Just `This is the similar functionality in JS`

native/tools/swig-config/tutorial/index.md
```diff
@@ -689,6 +711,95 @@ Build and run project
...
+  set(v) {
+    this._width = v;
+  },
+	// No get() for property
+});
+```
+
+#### Reference type
+
+If C++ `get` and `set` function return reference type, don't forget to add `&` suffix in %attribute or %attribute_writeonly directives. The following `ccstd::string&` is an example.
```
minggo: return reference type -> return a reference data type?

native/tools/swig-config/tutorial/index.md
```diff
@@ -689,6 +711,95 @@ Build and run project
...
+
+```c++
+%attribute_writeonly(cc::ICanvasRenderingContext2D, ccstd::string&, fillStyle, setFillStyle);
+```
+
+If `&` is missing, temporary `ccstd::string` objects will be created while binding functions are invoked.
```
minggo: - temporary `ccstd::string` objects -> a temporary `ccstd::string` instance?
- while binding functions are invoked -> while the binding function is invoked

native/tools/swig-config/tutorial/index.md
```diff
@@ -689,6 +711,95 @@ Build and run project
...
+
+```c++
+%attribute(MyNewClass, %arg(std::map<std::string, std::string>&), config, getConfig, setConfig);
+```
+
+Re-run `node genbindings.js`, the error should have gone.
+
+#### Don't add `const` 
+
+In the previous sample, `%arg(std::map<std::string, std::string>&)` is used as C++ type in %attribue directive. You may consider to add a const prefix before `std::map` like `%arg(const std::map<std::string, std::string>&)`. If you do that, you will make a readyonly `config` property which only bind `MyNewClass::getConfig`. That's obviously not what we expect. If we need a readonly property, just don't assign a `set` function.
```
minggo: In the previous sample -> In above sample

minggo: is used as C++ type -> is used as a C++ data type

dumganhar: Any differences about `previous` and `above`?  I think they're both ok here.

dumganhar: And while we use `above`, should we drop `the` before it?
Is it `In the above sample` correct or `In above sample`?
I think  `the`  word should not be dropped since it describe the `exact` sample in the previous section.

native/tools/swig-config/tutorial/index.md
dumganhar: DONE

native/tools/swig-config/tutorial/index.md
dumganhar: DONE

native/tools/swig-config/tutorial/index.md
dumganhar: DONE

native/tools/swig-config/tutorial/index.md
dumganhar: DONE

native/tools/swig-config/tutorial/index.md
```diff
@@ -689,6 +719,95 @@ Build and run project
...
+
+```c++
+%attribute(MyNewClass, %arg(std::map<std::string, std::string>&), config, getConfig, setConfig);
+```
+
+Re-run `node genbindings.js`, the error should have gone.
+
+#### Don't add `const` 
+
+In the above sample, `%arg(std::map<std::string, std::string>&)` is used as a C++ type in %attribue directive. You may consider to add a `const` prefix before `std::map` like `%arg(const std::map<std::string, std::string>&)`. If you do that, you will make a readyonly `config` property which only binds `MyNewClass::getConfig`. That's obviously not what we expect. If we need a readonly property, just don't assign a `set` function.
```
minggo: a C++ type -> a C++ data type

dumganhar: DONE

minggo (approved):

PR #13743, opened 2022-12-08
fix the problem that finishCallback is lost after setCurrentTime

OnlyKoei:


cocos2d/audio/CCAudio.js
```diff
@@ -44,8 +44,11 @@ let Audio = function (src) {
 
     const self = this;
     this._onended = function () {
-        self._state = Audio.State.STOPPED;
-        self.emit('ended');
+        if(self.paused){
```
bofeng-song, line 47: If self.pused is true, it should need to be changed to false ?  Why it can be true, we may call stop firstly.

cocos2d/audio/CCAudio.js
```diff
@@ -248,7 +251,9 @@ Audio.State = {
             // setCurrentTime would fire 'ended' event
             // so we need to change the callback to rebind ended callback after setCurrentTime
             self._unbindEnded();
-            self._bindEnded(self._onendedSecond);
+            self._bindEnded(function(){
+                self._bindEnded();
```
bofeng-song, line 255: Does it need to bind _onendedSecond ?

minggo:
Close this PR as it is not responsible.

PR #13811, opened 2022-12-14
Fix lookat exception

GengineJS:


minggo (approved):

star-e:
Is Mat3 row-major or column-major？

star-e (approved):

PR #13960, opened 2022-12-29
TiledMap supports object and class type properties, and supports multi-line string properties.

biliskener:
TiledMap supports object and class type properties, and supports multi-line string properties.

### DIFF
--- a/native/external-config.json
+++ b/native/external-config.json
@@ -3,6 +3,6 @@
         "type": "github",
         "owner": "cocos-creator",
         "name": "engine-native-external",
-        "checkout": "v3.6.2-6"
+        "checkout": "v3.6.3-1"
     }
 }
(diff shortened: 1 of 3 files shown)


native/external-config.json
```diff
@@ -3,6 +3,6 @@
         "type": "github",
         "owner": "cocos-creator",
         "name": "engine-native-external",
-        "checkout": "v3.6.2-6"
+        "checkout": "v3.6.3-1"
```
zhakesi, line 6: should not change here?

zhakesi:
this PR will cause an error on minigame platform.
捕获
it seems need to fix with the dom adapter.

zhakesi:
please solve the conflict about external-config.json.
or you may cherry-pick the change to v3.7.2 and create a new pr.

PR #14073, opened 2023-01-16
Windows devices  input fix

LorenzoDezi:
We have found an issue with Windows devices, where some touch events like TOUCH_MOVE were not registering. We resolve the issue using [pointer events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events), which is the most modern API for events, addressing the need for a generic pointer that can handle different input devices.


pal/input/web/mouse-input.ts
```diff
@@ -93,17 +93,18 @@ export class MouseInputSource {
...
         // register mouse move event
-        this._canvas?.addEventListener('mousemove', this._handleMouseMove);
+        this._canvas?.addEventListener('pointermove', this._createCallback(InputEventType.MOUSE_MOVE));
 
         // register mouse up event
-        window.addEventListener('mouseup', this._handleMouseUp);
-        this._canvas?.addEventListener('mouseup', this._handleMouseUp);
+        const handleMouseUp = this._createCallback(InputEventType.MOUSE_UP);
+        window.addEventListener('pointerup', handleMouseUp);
+        this._canvas?.addEventListener('pointerup', handleMouseUp);
```
PPpro: we have created callbacks in the constructor, why should we create again here ?

pal/input/web/touch-input.ts
```diff
@@ -53,35 +53,26 @@ export class TouchInputSource {
...
+            if (touch) {
                 if (eventType === InputEventType.TOUCH_END || eventType === InputEventType.TOUCH_CANCEL) {
-                    touchManager.releaseTouch(touchID);
+                    touchManager.releaseTouch(event.pointerId);
                 }
                 handleTouches.push(touch);
             }
             event.stopPropagation();
-            if (event.target === this._canvas) {
+            if (event.target === this._canvas || InputEventType.TOUCH_MOVE) {
```
PPpro: Is this `InputEventType.TOUCH_MOVE` condition statement neccessary ?

pal/input/web/mouse-input.ts
PPpro: and can you fix the return type  of `_createCallback` method to `(pointerEvent: PointerEvent) => void`

pal/input/web/touch-input.ts
```diff
@@ -53,35 +53,26 @@ export class TouchInputSource {
...
-                }
+            const location = this._getLocation(event, canvasRect);
+            const touch = touchManager.getTouch(event.pointerId, location.x, location.y);
+            if (touch) {
                 if (eventType === InputEventType.TOUCH_END || eventType === InputEventType.TOUCH_CANCEL) {
-                    touchManager.releaseTouch(touchID);
+                    touchManager.releaseTouch(event.pointerId);
                 }
                 handleTouches.push(touch);
             }
```
PPpro, line 73: can pointer event support multi touch well ?

pal/input/web/touch-input.ts
LorenzoDezi: No, I fixed it in the new commit.

pal/input/web/touch-input.ts
LorenzoDezi: Yes, I tested it and it works.

LorenzoDezi:
@PPpro there is also another fix needed. In order for the pointer events to work correctly along with these changes, we found out that the document.documentElement.style.touchAction property must be set to 'none'. In our projects, we use a custom component to make sure that property gets set correctly at startup since we couldn't find a way to set it directly from the cocos engine. Do you know how to generate the aforementioned style just with a custom modification directly in the cocos engine?

pal/input/web/mouse-input.ts
LorenzoDezi: I fixed it in the new commit.

PPpro (approved):

minggo:
@LorenzoDezi could you please fix conflict? Thanks.

pandamicro:
Fixed

minggo (approved):

PR #14235, opened 2023-02-14
Fix the problem that pointerlock cannot move infinitely

qiuguohua:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/@types/jsb.d.ts
+++ b/@types/jsb.d.ts
@@ -54,8 +54,8 @@ declare namespace jsb {
     export interface MouseEvent {
         x: number,
         y: number,
-        xref: number | undefined,
-        yref: number | undefined,
+        xDelta: number | undefined,
+        yDelta: number | undefined,
         button: number,
         windowId: number,
     }
--- a/@types/jsb.d.ts
+++ b/@types/jsb.d.ts
@@ -54,6 +54,8 @@ declare namespace jsb {
     export interface MouseEvent {
         x: number,
         y: number,
+        xref: number | undefined,
+        yref: number | undefined,
         button: number,
         windowId: number,
     }
--- a/native/cocos/bindings/event/EventDispatcher.cpp
+++ b/native/cocos/bindings/event/EventDispatcher.cpp
@@ -94,7 +94,7 @@ void EventDispatcher::init() {
         listenerLowMemory.bind(&dispatchMemoryWarningEvent);
         listenerClose.bind(&dispatchCloseEvent);
         listenerRestartVM.bind(&dispatchRestartVM);
-        listenerPointerLock.bind(&dispatchPointerLockEvent);
+        listenerPointerLock.bind(&dispatchPointerlockChangeEvent);
         busListenerInited = true;
     }
 }
@@ -394,7 +394,7 @@ void EventDispatcher::dispatchCloseEvent() {
     EventDispatcher::doDispatchJsEvent("onClose", se::EmptyValueArray);
 }
 
-void EventDispatcher::dispatchPointerLockEvent(bool value) {
+void EventDispatcher::dispatchPointerlockChangeEvent(bool value) {
     se::ValueArray args;
     args.emplace_back(se::Value(value));
     EventDispatcher::doDispatchJsEvent("onPointerlockChange", args);
--- a/native/cocos/bindings/event/EventDispatcher.cpp
+++ b/native/cocos/bindings/event/EventDispatcher.cpp
@@ -215,10 +215,10 @@ void EventDispatcher::dispatchMouseEvent(const MouseEvent &mouseEvent) {
         jsMouseEventObj->setProperty("x", xVal);
         jsMouseEventObj->setProperty("y", yVal);
         if (type == MouseEvent::Type::MOVE) {
-            const auto &xrefVal = se::Value(mouseEvent.xref);
-            const auto &yrefVal = se::Value(mouseEvent.yref);
-            jsMouseEventObj->setProperty("xref", xrefVal);
-            jsMouseEventObj->setProperty("yref", yrefVal);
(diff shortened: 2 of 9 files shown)


native/cocos/bindings/event/EventDispatcher.cpp
```diff
@@ -212,6 +214,12 @@ void EventDispatcher::dispatchMouseEvent(const MouseEvent &mouseEvent) {
         }
         jsMouseEventObj->setProperty("x", xVal);
         jsMouseEventObj->setProperty("y", yVal);
+        if (type == MouseEvent::Type::MOVE) {
+            const auto &xrefVal = se::Value(mouseEvent.xref);
+            const auto &yrefVal = se::Value(mouseEvent.yref);
+            jsMouseEventObj->setProperty("xref", xrefVal);
+            jsMouseEventObj->setProperty("yref", yrefVal);
+        }
```
minggo, line 222: I think only pointerlock mode need to pass delta value.

minggo: We should pass as less data as possible.

qiuguohua: If the event is a click, it will cause a coordinate exception

pal/input/native/mouse-input.ts
```diff
@@ -95,9 +99,14 @@ export class MouseInputSource {
             const eventMouse = new EventMouse(eventType, false, this._preMousePos, mouseEvent.windowId);
             eventMouse.setLocation(location.x, location.y);
             eventMouse.setButton(button);
-            eventMouse.movementX = location.x - this._preMousePos.x;
-            eventMouse.movementY = this._preMousePos.y - location.y;
-
+            if (this._pointLocked && eventType === InputEventType.MOUSE_MOVE) {
```
PPpro: should we ignore the other mouse event, like mouse down, mouse up ?

pal/input/native/mouse-input.ts
```diff
@@ -95,9 +99,14 @@ export class MouseInputSource {
...
+                const dpr = screenAdapter.devicePixelRatio;
+                eventMouse.movementX = mouseEvent.xref === undefined ? 0 : mouseEvent.xref * dpr;
+                eventMouse.movementY = mouseEvent.yref === undefined ? 0 : mouseEvent.yref * dpr;
```
PPpro: it's safer to use `typeof mouseEvent.xref === 'undefined'` instead of `mouseEvent.xref === undefined`
because the keyword `undefined` can also be a variable in JS like
```
(function() {
  var undefined = "foo";
  var variable = "foo";
  if (variable === undefined)
    console.log("eh, what?!");  
})()
```

native/cocos/bindings/event/EventDispatcher.cpp
minggo: What i mean is that, should pass different data to TS in different mode. Passing delta value in not pointerlock mode is meaningless.

qiuguohua: I think it would be better to be consistent. In fact, on the ts is required to calculate the delta data, if passed up through the native, we do not need to calculate

pal/input/native/mouse-input.ts
qiuguohua: Other events do not need to use delta directly.

native/cocos/bindings/event/EventDispatcher.cpp
minggo: As passing more data will affect performance. That's why we should pass as less data as possible, especially the move event.

minggo: As passing more data will affect performance. That's why we should pass as less data as possible, especially the move event.

minggo: Did you mean delta value is needed in non pointerlock mode too?

qiuguohua: Yes, it is need to calculate.
image

pal/input/native/mouse-input.ts
```diff
@@ -95,11 +99,10 @@ export class MouseInputSource {
...
             eventMouse.setLocation(location.x, location.y);
             eventMouse.setButton(button);
-            eventMouse.movementX = location.x - this._preMousePos.x;
-            eventMouse.movementY = this._preMousePos.y - location.y;
-
+            const dpr = screenAdapter.devicePixelRatio;
+            eventMouse.movementX = typeof mouseEvent.xref === 'undefined' ? 0 : mouseEvent.xref * dpr;
+            eventMouse.movementY = typeof mouseEvent.yref === 'undefined' ? 0 : mouseEvent.yref * dpr;
             // update previous mouse position.
-            this._preMousePos.set(location.x, location.y);
```
PPpro: we still need the `_preMousePos` property for EventMouse constructor

qiuguohua: If we use pass delta values directly, there is no need to calculate them.

@types/jsb.d.ts
```diff
@@ -54,6 +54,8 @@ declare namespace jsb {
     export interface MouseEvent {
         x: number,
         y: number,
+        xref: number | undefined,
+        yref: number | undefined,
```
minggo: Why not `xDelta` instead of `xref`?

pal/input/native/mouse-input.ts
PPpro: class EventMouse still need the previous location property

qiuguohua: ok

minggo (approved):

PPpro (approved):

Issue #12992, opened 2022-10-09
[3.6.x] （motionstreak + rigidbody2d） does not display correctly on native platforms


OnlyKoei:
### Cocos Creator version

3.6.0，3.3.2，3.6.1

### System information

window10

### Issue description

https://forum.cocos.org/t/topic/140885
![image](https://user-images.githubusercontent.com/40415168/194740746-00dd6549-5745-4d45-8962-8ddc13c95d25.png)
![image](https://user-images.githubusercontent.com/40415168/194740764-6265e54d-5f19-4d07-b8f8-4266bc345773.png)

### Steps to reproduce

demo:
[NewProject_23.zip](https://github.com/cocos/cocos-engine/files/9740852/NewProject_23.zip)

PR #14392, opened 2023-02-27
v3.7.2 Fix the rendering issue with MotionStreak + RigidBody2D on native platforms

### LINKED ISSUE #14300: [v3.7.0] (Regression) MotionStreak with RigidBody2D does not move correctly on Android
### Cocos Creator version

3.7.0, 3.6.3

### System information

Windows 10 Editor

### Issue description

I have a node with a MotionStreak and RigidBody2D components. When I try to move the node it does not render properly on Android devices (seems like it has wrong position and it goes offscreen).
This is the code for the demo:
```
export class Movable extends Component {
    private time: number = 0;
    private direction: number = 1;
    private pos: Vec3 = new Vec3();

    update(deltaTime: number) {
        this.time += deltaTime;
        if (this.time > 3) {
            this.time = 0;
            this.direction *= -1;
        }
        this.node.getPosition(this.pos);
        this.node.setPosition(this.pos.x, this.pos.y + this.direction * 80 * deltaTime, 0);
    }
}
```
This is how it looks when this demo is running in a browser (which is correct):

https://user-images.githubusercontent.com/82378750/219465509-6ea2f380-fbbf-460b-9b72-9efebaf834a4.mp4

But running the demo on Android devices the motionstreak does not appear and the screen is plain black.

This issue was not present in v3.5.2.

### Relevant error log output

_No response_

### Steps to reproduce

Run the attached demo project on Android to see the issue.

[Demo2D.zip](https://github.com/cocos/cocos-engine/files/10759706/Demo2D.zip)

mrtzd:
Re: # This pr fixes #14300 and #12992. 

### Changelog

* Fix the issue with MotionStreak plus RigidBody2D not rendering correctly on native platforms

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/particle-2d/motion-streak-2d-assembler.ts
+++ b/cocos/particle-2d/motion-streak-2d-assembler.ts
@@ -187,8 +187,7 @@ export const MotionStreakAssembler: IAssembler = {
             renderData.updateRenderData(comp, comp.texture!);
             // No need update WorldMatrix, so change dirty flag
             // A dirty hack
-            renderData.renderDrawInfo.setVertDirty(false);
-            comp.node.hasChangedFlags = 0;
+            comp.markForUpdateRenderData();
         }
     },
 
@@ -244,6 +243,10 @@ export const MotionStreakAssembler: IAssembler = {
     },
 
     updateRenderData (comp: MotionStreak) {
+        if (JSB) {
+            comp.renderData!.renderDrawInfo.setVertDirty(false);
+            comp.node.hasChangedFlags = 0;
+        }
     },
 
     updateColor (comp: MotionStreak) {
--- a/cocos/particle-2d/motion-streak-2d-assembler.ts
+++ b/cocos/particle-2d/motion-streak-2d-assembler.ts
@@ -243,10 +243,6 @@ export const MotionStreakAssembler: IAssembler = {
     },
 
     updateRenderData (comp: MotionStreak) {
-        if (JSB) {
-            comp.renderData!.renderDrawInfo.setVertDirty(false);
-            comp.node.hasChangedFlags = 0;
-        }
     },
 
     updateColor (comp: MotionStreak) {
--- a/cocos/particle-2d/motion-streak-2d-assembler.ts
+++ b/cocos/particle-2d/motion-streak-2d-assembler.ts
@@ -185,8 +185,6 @@ export const MotionStreakAssembler: IAssembler = {
             this.updateWorldVertexAllData(comp);
 
             renderData.updateRenderData(comp, comp.texture!);
-            // No need update WorldMatrix, so change dirty flag
-            // A dirty hack
             comp.markForUpdateRenderData();
         }
     },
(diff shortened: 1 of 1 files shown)


mrtzd:
@minggo I can't ask for a review but I'd really like to see this get merged for the v3.7.2 release, thanks!

minggo:
@mrtzd thanks for your contribution. We will review it ASAP.

SantyWang:
Thanks for the contribution, I think the previous code is very strange, we will check it out and try to fix it in 3.7.2

SantyWang:
```
comp.node.hasChangedFlags = 0;
```
is INCORRECT, should not be invoked anywhere anytime.

mrtzd:
> ```
> comp.node.hasChangedFlags = 0;
> ```
> 
> is INCORRECT, should not be invoked anywhere anytime.

Yeah since I'm not very familiar with this codebase I didn't know why this line was there too. I removed the whole 'if' block and the code seems to be working fine with my testing although I'm not sure if this is the proper solution.

SantyWang:
> Yeah since I'm not very familiar with this codebase I didn't know why this line was there too. I removed the whole 'if' block and the code seems to be working fine with my testing although I'm not sure if this is the proper solution.

Yeah, I see. Thanks, we will figure it out.

LinYunMo (commented):
Thank you for your PR. @mrtzd 

Regarding this issue, **the strange code here is a dirty hack**, due to the special nature of this MotionStreak component, we need to perform world coordinate transformations at the component level, so at the cpp level we want to **avoid repeated world coordinate transformations**, it will cause rendering position exceptions, limited by the current design, we don't have an explicit tag to handle this behavior, so we use the hasChangeFlags to forcibly prevent updates.

Regarding physics, since RigidBody2D have to use hasChangedFlags for position updates, this hack is **moved to after physics uses this tag**, so the problem is avoided.

We will make structural changes in a later version to ensure a more logical flow.

Thanks again for the PR!

cocos/particle-2d/motion-streak-2d-assembler.ts
```diff
@@ -244,6 +243,10 @@ export const MotionStreakAssembler: IAssembler = {
     },
 
     updateRenderData (comp: MotionStreak) {
+        if (JSB) {
```
LinYunMo, line 244: Could you please add some comments to this code so that it can be identified for subsequent maintenance: like this

```
// A dirty hack
// The world matrix was updated in advance and needs to be avoided at the cpp level
// Need a flag to explicitly not update the world transform to solve this problem
```

Thanks a lot

LinYunMo (changes requested):

cocos/particle-2d/motion-streak-2d-assembler.ts
mrtzd: Thanks for the explanation. I added the comments.

SantyWang (approved):

PR #14402, opened 2023-02-28
fix spine/dragonbones error log

zhakesi:


cocos/dragon-bones/assembler/simple.ts
```diff
@@ -409,7 +409,7 @@ function updateComponentRenderData (comp: ArmatureDisplay, batcher: Batcher2D) {
     // comp.node._renderFlag |= RenderFlow.FLAG_UPDATE_RENDER_DATA;
 
     const armature = comp._armature;
-    if (!armature) return;
+    if (!armature || comp.renderData === null) return;
```
SantyWang, line 412: I do not thick checking renderData is necessary here. if there is no renderData, should return false when called `canRender()`, then this function will not be invoked.

zhakesi: > I do not thick checking renderData is necessary here. if there is no renderData, should return false when called `canRender()`, then this function will not be invoked.

_canRender() function do not check renderData.
and it be called after than assembler.updateRenderData.
```
    public updateRenderer () {
        if (this._assembler) {
            this._assembler.updateRenderData(this);
        }
        this._renderFlag = this._canRender();
        this._renderEntity.enabled = this._renderFlag;
    }
```

SantyWang (approved):

PR #14403, opened 2023-02-28
jsb classes apply decorator from .ts source files

PatriceJiang:
Re: https://github.com/cocos/cocos-engine/issues/14426

The current jsb.ts requires manual invocation of methods in _decorators, which is difficult to maintain. Here, a Babel plugin is used to parse the ts source code, extract decorators and encapsulate them into separate methods for jsb.ts to use.

### Changelog

* Use babel plugin to record decorators and apply to native binding classes.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


PatriceJiang:
@cocos-robot run test cases

scripts/build-engine/package-lock.json
```diff
@@ -259,24 +259,45 @@
...
     },
     "@babel/generator": {
-      "version": "7.9.4",
-      "resolved": "https://registry.npm.taobao.org/@babel/generator/download/@babel/generator-7.9.4.tgz?cache=0&sync_timestamp=1585038762678&other_urls=https%3A%2F%2Fregistry.npm.taobao.org%2F%40babel%2Fgenerator%2Fdownload%2F%40babel%2Fgenerator-7.9.4.tgz",
-      "integrity": "sha1-EkQekMOzxBWc3s8xIHW/Gozi284=",
-      "requires": {
-        "@babel/types": "^7.9.0",
-        "jsesc": "^2.5.1",
-        "lodash": "^4.17.13",
-        "source-map": "^0.5.0"
```
minggo: Why need to update this file?

minggo:
Please create an issue for PR in future.

scripts/build-engine/package-lock.json
PatriceJiang: I checked with @PPpro, this change will cause compatibility issues, and have rolled back.

minggo (approved):

dumganhar (approved):

PatriceJiang:
> Please create an issue for PR in future.

Issue is created here https://github.com/cocos/cocos-engine/issues/14426

.github/workflows/gen_decorators.js
```diff
@@ -0,0 +1,105 @@
...
+            'intersection-2d', 'light-probe',
+            'marionette', 'particle',
+            'particle-2d', 'physics-2d-box2d',
+            'physics-ammo', 'primitive',
+            'profiler', 'skeletal-animation',
+            'spine', 'terrain',
+            'tiled-map', 'tween',
+            'ui', 'video',
+            'websocket', 'webview',
+        ],
```
PPpro, line 47: waht if we add new feature in the future ? Do we need to update this workflow ?
can we get all the features from `cc.config.json` ?

PatriceJiang: Great idea! However, there is still an issue that needs to be addressed, which is the potential conflicts and dependencies between modules. Dependencies can be managed easily, but what about the case where multiple backends, such as physical modules, need to be selected? It seems like we might need to hardcode the solution for this scenario.

PatriceJiang: I suggest that hardcoding the solution for the current scenario is a feasible approach. In case of a missing module, it should be relatively easy to locate and modify the configuration.

PPpro: got it

PPpro (approved):

PR #14457, opened 2023-03-03
Fix console.log output for objects by calling toString() method

PatriceJiang:
Re: https://github.com/cocos/3d-tasks/issues/7908

Fixed console.log output for Object parameters. Previously, the output for Objects was [object Object]. With this fix, the toString method of the Object will be called instead, resulting in more consistent output with devtools

- web
image
- visual studio
image

### Changelog

* Modified the implementation of console.log to call the toString method on objects, arrays, and other non-scalar types when printing them to the console.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/bindings/jswrapper/v8/Object.cpp
+++ b/native/cocos/bindings/jswrapper/v8/Object.cpp
@@ -1101,16 +1101,23 @@ bool Object::detachObject(Object *obj) {
 }
 
 ccstd::string Object::toString() const {
-    ccstd::string ret;
+    v8::Local<v8::Object> v8Obj = const_cast<Object *>(this)->_obj.handle(__isolate);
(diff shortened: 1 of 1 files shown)


dumganhar:
If se::Object represents a `plane object`, `toString` & `console.log` will output different result.
Screenshot 2023-03-06 at 09 56 49

What' the result after applying this patch?

minggo (approved):

PatriceJiang:
> If se::Object represent a `plane object`, `toString` & `console.log` will output different result. Screenshot 2023-03-06 at 09 56 49
> 
> What' the result after applying this patch?

It will call the toString method to output the result. Devtools is quite special, it is interactive and can dynamically query, which is something that native platforms cannot achieve.

dumganhar (approved):

PR #14468, opened 2023-03-03
Fix pass properties order in native editor

PatriceJiang:
Re: https://github.com/cocos/3d-tasks/issues/15277

Add a new function to retrieve the value of a specific property from a container, where the type of the container depends on whether the CC_EDITOR macro is enabled or not. If the macro is enabled, the container is a list, otherwise it is an unordered map.

The associated issue states that due to the use of an `unordered_map` to store effects data in C++, the traversal order cannot be guaranteed. To maintain order, a `vector<pair>` is being used instead. However, to ensure performance, the `unordered_map` will still be used in non-Editor mode.

### Changelog

* [bugfix] Fix the property order of Material/Pass in native editor

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/core/assets/EffectAsset.h
+++ b/native/cocos/core/assets/EffectAsset.h
@@ -39,35 +39,26 @@
 #include "renderer/pipeline/Define.h"
 namespace cc {
 
+
 #if !SWIGCOCOS
-
-template <typename T>
-struct IsVectorContainer : std::false_type {};
-
-template <typename E, typename A>
-struct IsVectorContainer<ccstd::vector<E, A>> : std::true_type {};
-
-/**
- * This function retrieves the value of a specific property from an attribute container. 
- * The type of container used for storing the properties depends on whether the CC_EDITOR macro is enabled or not.
- * If the macro is enabled, the container is a list, otherwise it is an unordered map.
- */
-template <typename T, typename K>
-inline auto retrieveProperty(T &collection, K &key) {
-    if constexpr (IsVectorContainer<std::remove_cv_t<T>>::value) {
-        return std::find_if(collection.begin(), collection.end(), [&](auto &e) { return e.first == key; });
-    } else {
-        return collection.find(key);
+template <typename K, typename V>
+class StablePropertyMap : public ccstd::vector<std::pair<K, V>> { // NOLINT
+    using Super = ccstd::vector<std::pair<K, V>>;
+public:
+    auto find(const K &key) const {
+        auto *self = static_cast<const Super*>(this);
+        return std::find_if(self->begin(), self->end(), [&](auto &ele) {
+            return ele.first == key;
+        });
     }
-}
-
+};
 #endif
 
 template <typename K, typename V>
 using UnstablePropertyContainer = ccstd::unordered_map<K, V>;
 
 template <typename K, typename V>
-using StablePropertyContainer = ccstd::vector<std::pair<K, V>>;
+using StablePropertyContainer = StablePropertyMap<K, V>;
 
 #if CC_EDITOR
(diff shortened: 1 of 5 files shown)


native/cocos/core/assets/EffectAsset.h
```diff
@@ -36,15 +37,52 @@
...
 #include "renderer/pipeline/Define.h"
-
 namespace cc {
 
+#if !SWIGCOCOS
+
+template <typename T>
+struct IsVectorContainer : std::false_type {};
+
+template <typename E, typename A>
```
minggo: It is better to have a new class to wrap these codes.

PatriceJiang: Fixed

dumganhar (approved):

native/cocos/scene/Pass.h
```diff
@@ -342,7 +342,7 @@ class Pass : public RefCounted {
 
     const IProgramInfo *_shaderInfo; // weakref to template of ProgramLib
     MacroRecord _defines;
-    Record<ccstd::string, IPropertyInfo> _properties;
+    PassPropertyInfoMap _properties;
```
minggo, line 345: Please add comment to describe why use `PassPropertyInfoMap `.

native/cocos/core/assets/EffectAsset.h
```diff
@@ -36,15 +37,43 @@
 #include "renderer/core/PassUtils.h"
 #include "renderer/gfx-base/GFXDef.h"
 #include "renderer/pipeline/Define.h"
-
 namespace cc {
 
+
+#if !SWIGCOCOS
+template <typename K, typename V>
```
minggo, line 49: Please add comments about:
- why use `!SWIGCOCOS`
- comment about `StablePropertyMap `

native/cocos/core/assets/EffectAsset.h
```diff
@@ -36,15 +37,43 @@
 #include "renderer/core/PassUtils.h"
 #include "renderer/gfx-base/GFXDef.h"
 #include "renderer/pipeline/Define.h"
-
 namespace cc {
 
+
```
PatriceJiang:
```suggestion
// To avoid errors when generating code using SWIG.
```

native/cocos/core/assets/EffectAsset.h
```diff
@@ -36,15 +37,43 @@
...
+#if !SWIGCOCOS
+template <typename K, typename V>
+class StablePropertyMap : public ccstd::vector<std::pair<K, V>> { // NOLINT
```
PatriceJiang, line 50:
```suggestion
// The properties in Pass are obtained from an asset file. If they are directly stored in an unordered_map, the order of these properties may become scrambled. To maintain their order, a vector<pair> is used instead. Currently, there is no handling of randomly inserting these data objects.
class StablePropertyMap : public ccstd::vector<std::pair<K, V>> { // NOLINT
```

native/cocos/core/assets/EffectAsset.h
PatriceJiang: Comments have been added.

minggo (approved):

native/cocos/scene/Pass.h
PatriceJiang: The definitions of the types `Record<ccstd::string, IPropertyInfo>` and `PassPropertyInfoMap` (`ccstd::unordered_map<ccstd::string, IPropertyInfo>`) were exactly the same, so the same literal type can be used directly to simplify the expression.

Issue #14273, opened 2023-02-16
Reduce build warnning 


PatriceJiang:
![企业微信截图_16765170854305](https://user-images.githubusercontent.com/40414978/219262063-f7e97209-ac5a-4a8c-b933-7c4f719ebf6f.png)

Need to upgrade external libraries

PR #14527, opened 2023-03-09
Resolve certain warning messages on Visual Studio

PatriceJiang:
Re: https://github.com/cocos/cocos-engine/issues/14273

CMake warnings will be redirected to the editor in a `console.log` style to minimize distraction to the user. However, this approach also poses the risk of missing important information. Nevertheless, we can analyze the build logs to mitigate this potential issue.

### Changelog
- Lower the level of CMake warnings.
- Remove empty files and resolve `LNK4221` warning message, which reads: 
``` 
warning LNK4221: This object file does not define any previously undefined public symbols, so it will not be used by any link operation that consumes this library.
```

  *This doesn't address all the similar errors. Some files are empty and were generated automatically, and they cannot be removed in this change.*
- Simultaneously remove empty files from CMakeLists.txt.
- Fix a few warnings concerning integer type conversions.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


native/CMakeLists.txt
```diff
@@ -2033,8 +1940,8 @@ cocos_source_files(
     cocos/core/utils/IDGenerator.h
     cocos/core/utils/Path.cpp
     cocos/core/utils/Path.h
-    cocos/core/utils/MutableForwardIterator.cpp
-    cocos/core/utils/MutableForwardIterator.h
+    # cocos/core/utils/MutableForwardIterator.cpp
+    # cocos/core/utils/MutableForwardIterator.h
```
minggo: Why comment it?

native/cocos/audio/common/utils/primitives.cpp
```diff
@@ -317,7 +317,7 @@ size_t nonZeroStereo16(const int16_t *frames, size_t count) {
  */
 #define COPY_FRAME_BY_MASK(dst, dmask, src, smask, count, zero) \
     {                                                           \
-        uint32_t bit, ormask;                                   \
+        int32_t bit, ormask;                                    \
```
minggo, line 320: Why need to change to int32_t?

native/cocos/audio/common/utils/primitives.cpp
```diff
@@ -437,8 +437,8 @@ size_t memcpy_by_index_array_initialization(int8_t *idxary, size_t idxcount,
                                             uint32_t dstMask, uint32_t srcMask) {
     size_t n = 0;
     int srcidx = 0;
-    uint32_t bit;
-    uint32_t ormask = srcMask | dstMask;
+    int32_t bit;
+    int32_t ormask = srcMask | dstMask;
```
minggo, line 441: The same as above.

native/cocos/audio/common/utils/primitives.cpp
PatriceJiang: A warning is generated when taking the negative of an unsigned int variable, but in this case, both variables are used for bitwise operations and do not impact the result.

native/CMakeLists.txt
PatriceJiang: These two files did not participate in the compilation.

minggo: If so, why not delete these files too?

minggo (approved):

PR #14595, opened 2023-03-16
support SafeArray and SafeRecord to optimize type checking

PPpro:
Re: https://github.com/cocos/3d-tasks/issues/15913

### Changelog

* support SafeArray and SafeRecord to optimize type checking 

#### before
```ts
interface A {
    a: number;
}

const aList: A[] = []

aList[0] = aList[0] || {}  // ts doesn't emit error
```
#### after
```ts
interface A {
    a: number;
}

type SafeArray<T> = Array<T | undefined>;

const aList: SafeArray<A> = []

aList[0] = aList[0] || {}  // ts emit that  Property 'a' is missing in type '{}' but required in type 'A'.
```

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


cocos/dragon-bones/ArmatureDisplay.ts
```diff
@@ -895,7 +895,7 @@ export class ArmatureDisplay extends UIRenderer {
         if (!this._playing) {
             if (frameCache.isInvalid()) {
                 frameCache.updateToFrame();
-                this._curFrame = frames[frames.length - 1];
+                this._curFrame = frames[frames.length - 1]!;
```
minggo, line 901: Why need to add `!` now?

PPpro: because ts always take `frames[frames.length - 1]` to be a defined value before, while `SafeArray` type says that `frames[frames.length - 1]` cound be `undefined`

here we know it is exactly not `undefined`, we don't need the TSC to tell us this could be error, for example
```ts
for (let i = 0 ; i < aList.length; ++i) {
   aList[i]!  // I know it's not undefined, so ts don't need to emit error
}
```
or it could be 

```ts
for (let i = 0 ; i < aList.length; ++i) {
   const a = aList[i];
    if (!a) {
         continue;
    }
}
```

minggo: Got it. It seems SafeArray brings some inconvenient.

PPpro: yes, but it's type safer, and it should be optional

cocos/dragon-bones/ArmatureCache.ts
```diff
@@ -286,11 +286,14 @@ export class AnimationCache {
...
             boneInfos: [],
-            vertices: null,
-            uintVert: null,
-            indices: null,
+            vertices: new Float32Array(),
+            uintVert: new Uint32Array(),
+            indices: new Uint16Array(),
         };
         const frame = this.frames[index];
+        if (!frame) {
```
zhakesi: it seems no need check frame.
```
        this.frames[index] = this.frames[index] || {
            segments: [],
            colors: [],
            boneInfos: [],
            vertices: null,
            uintVert: null,
            indices: null,
        };
        const frame = this.frames[index];
```

cocos/spine/skeleton-cache.ts
```diff
@@ -366,11 +366,13 @@ export class AnimationCache {
...
             colors: [],
             boneInfos: [],
-            vertices: null,
-            uintVert: null,
-            indices: null,
+            vertices: new Float32Array(),
+            indices: new Uint16Array(),
         };
         const frame = this.frames[index];
+        if (!frame) {
```
zhakesi: no need check frame.

cocos/tiledmap/tiled-layer.ts
```diff
@@ -352,7 +352,7 @@ export class TiledLayer extends UIRenderer {
 
         const rowData = this._userNodeGrid[row];
         const colData = rowData && rowData[col];
-        if (colData) {
+        if (rowData && colData) {
```
zhakesi: no need add rowData here?

zhakesi (approved):

minggo (approved):

PR #14624, opened 2023-03-21
fix JS error after 3d models open USE_INSTANCING on Android platform …

foupwang:
fix JS error after 3d models open USE_INSTANCING on Android platform since v3.7.1

### DIFF
--- a/cocos/rendering/render-instanced-queue.ts
+++ b/cocos/rendering/render-instanced-queue.ts
@@ -88,7 +88,7 @@ export class RenderInstancedQueue {
      */
     public recordCommandBuffer (device: Device, renderPass: RenderPass, cmdBuff: CommandBuffer,
         descriptorSet: DescriptorSet | null = null, dynamicOffsets?: Readonly<number[]>) {
-        const it = this._renderQueue.length === 0 ? this.queue.values() : this._renderQueue.values();
+        const it = this._renderQueue.length === 0 ? this.queue.values() : this._renderQueue[Symbol.iterator]();
         let res = it.next();
 
         while (!res.done) {


cocos/rendering/render-instanced-queue.ts
```diff
@@ -88,7 +88,7 @@ export class RenderInstancedQueue {
      */
     public recordCommandBuffer (device: Device, renderPass: RenderPass, cmdBuff: CommandBuffer,
         descriptorSet: DescriptorSet | null = null, dynamicOffsets?: Readonly<number[]>) {
-        const it = this._renderQueue.length === 0 ? this.queue.values() : this._renderQueue.values();
+        const it = this._renderQueue.length === 0 ? this.queue.values() : this._renderQueue[Symbol.iterator]();
```
dumganhar, line 91: What's the difference between `this._renderQueue.values()` and `this._renderQueue[Symbol.iterator]()` ?

foupwang: _renderQueue is a JS Array. In my mini game, if using the former, there will be a persistent error message "TypeError: this._renderQueue.values() is not a function" on Android phone, but there is no such problem on iOS phones. If using the latter, there are no issues on either platform.

dumganhar: Got it. May I ask which mini-game platform did you test on?

dumganhar: If this is an issue, all places where using `.values()` needs to be modified too.

foupwang: Wechat on XiaoMI 8

foupwang: The scope of impact should not be large. As in the example above, this.queue is a JS Set and it is normal to use values to get an iterator. However, this._renderQueue is an JS Array and using values is a bit uncommon.

shrinktofit (approved):

PPpro (approved):

pandamicro (approved):

PR #14633, opened 2023-03-23
Add HPE Feature Detetion and JNI Method Refactoring

PatriceJiang:
Re: https://github.com/cocos/3d-tasks/issues/16000

- Add the HPE (High Performance Emulator) feature for the Google Play for Games PC platform,

- Some functions related to JNI are global, but these symbols can be restricted to within `cc` to avoid polluting the global namespace.

### Changelog

* Add feature detection for Android Google Play For Game Emulator
* Moved certain global JNI-related methods into the `cc` namespace

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


minggo (approved):

native/cocos/platform/android/java/src/com/cocos/lib/CocosHelper.java
```diff
@@ -427,4 +428,8 @@ public void run() {
             }
         });
     }
+    public static boolean supportHPE() {
+        PackageManager pm = GlobalObject.getActivity().getPackageManager();
```
dumganhar: Why not use `GlobalObject.getContext()` ?
We need to use `getContext()` if we can.

PatriceJiang: The method is not present in the `CocosHelper` class and can be accessed through the `GlobalObject`.

PatriceJiang: While both `getContext` and `getActivity` can be used interchangeably here, using `getContext` would be a better option if we plan to support `Activityless` mode in the future.

PPpro (approved):

dumganhar (approved):

PR #14710, opened 2023-04-03
[3.7.3] Fix build windows error when editor path contains spaces

yanOO1497:
RT

### DIFF
--- a/scripts/native-pack-tool/source/utils.ts
+++ b/scripts/native-pack-tool/source/utils.ts
@@ -428,11 +428,11 @@ export const toolHelper = {
     },
 
     runCmake(args: string[]) {
-        let cmakeCommand = Paths.cmakePath;
-        if (process.platform === 'win32' && cmakeCommand.indexOf(' ') > -1) {
-            cmakeCommand = `"${cmakeCommand}"`;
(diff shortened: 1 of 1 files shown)


scripts/native-pack-tool/source/utils.ts
```diff
@@ -428,15 +428,20 @@ export const toolHelper = {
...
+        }
         // Delete environment variables start with `npm_`, which may cause compile error on windows
         const newEnv: any = {};
         Object.assign(newEnv, process.env);
         Object.keys(newEnv).filter(x => x.toLowerCase().startsWith("npm_")).forEach(e => delete newEnv[e]);
 
         return new Promise<void>((resolve, reject) => {
             console.log(`run ${cmakePath} ${args.join(' ')}`);
-            const cp = spawn(cmakePath, args, {
+            const cp = spawn(`${cmakePath}`, args, {
```
PatriceJiang: There's no need to use string templates here.

PatriceJiang (approved):

scripts/native-pack-tool/source/utils.ts
yanOO1497: OK

minggo (approved):

PR #14782, opened 2023-04-12
Fix Richtext miss string bug

LinYunMo:
Re: https://discord.com/channels/634690911896469525/634690911896469527/1091292440058675250

Fix RichText misc bug:

- Line feed error when maxWidth = 0 and the text is very long (width greater than 2048)
- When the text is very long (width greater than 2048), the text display is missing the penultimate line

[RichTextBug.zip](https://github.com/cocos/cocos-engine/files/11208975/RichTextBug.zip)

### DIFF
--- a/cocos/2d/components/rich-text.ts
+++ b/cocos/2d/components/rich-text.ts
@@ -27,7 +27,7 @@ import { ccclass, executeInEditMode, executionOrder, help, menu, tooltip, multil
 import { DEBUG, DEV, EDITOR } from 'internal:constants';
 import { Font, SpriteAtlas, TTFFont, SpriteFrame } from '../assets';
 import { EventTouch } from '../../input/types';
-import { assert, warnID, Color, Vec2, CCObject, cclegacy, js } from '../../core';
+import { assert, warnID, Color, Vec2, CCObject, cclegacy, js, Size } from '../../core';
 import { BASELINE_RATIO, fragmentText, isUnicodeCJK, isUnicodeSpace, getEnglishWordPartAtFirst, getEnglishWordPartAtLast } from '../utils/text-utils';
 import { HtmlTextParser, IHtmlTextParserResultObj, IHtmlTextParserStack } from '../utils/html-text-parser';
 import { Node } from '../../scene-graph';
@@ -478,6 +478,9 @@ export class RichText extends Component {
     protected _updateRichTextStatus: () => void;
     protected _labelChildrenNum = 0; // only ISegment
 
+    private _tempSize = new Size();
+    private _tempSizeLeft = new Size();
+
     constructor () {
         super();
         if (EDITOR) {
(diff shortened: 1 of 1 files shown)


cocos/2d/components/rich-text.ts
```diff
@@ -582,14 +582,14 @@ export class RichText extends Component {
             return partStringArr;
         }
 
-        const labelSize = this._calculateSize(styleIndex, text);
+        const labelSize = this._calculateSize(styleIndex, text).clone();
```
SantyWang: Why needs a new vec2?

cocos/2d/components/rich-text.ts
```diff
@@ -611,12 +611,16 @@ export class RichText extends Component {
         let curEnd = longStr.length / 2;
         let curString = longStr.substring(curStart, curEnd);
         let leftString = longStr.substring(curEnd);
-        let curStringSize = this._calculateSize(styleIndex, curString);
-        let leftStringSize = this._calculateSize(styleIndex, leftString);
+        const curStringSize = this._calculateSize(styleIndex, curString).clone();
+        const leftStringSize = this._calculateSize(styleIndex, leftString).clone();
+        let maxWidth = this._maxWidth;
+        if (this._maxWidth === 0) {
```
SantyWang, line 620: Why needs this?

cocos/2d/components/rich-text.ts
```diff
@@ -684,19 +688,21 @@ export class RichText extends Component {
...
             // Exit: If the left part string size is less than 2048, the method will finish.
-            if (leftStringSize.x < 2048) {
+            if (leftStringSize.x < 2048 && curStringSize.x < sizeForOnePart) {
+                partStringArr.push(curString);
                 curStart = text.length;
                 curEnd = text.length;
                 curString = leftString;
-                partStringArr.push(curString);
+                if (leftString !== '') {
+                    partStringArr.push(curString);
```
SantyWang, line 706: why push curString twice?

cocos/2d/components/rich-text.ts
LinYunMo: Because the function returns an object, it is not safe to use it directly

cocos/2d/components/rich-text.ts
LinYunMo: If the user does not enter maxWidth, the original algorithm does not take this case into account, resulting in incorrect typesetting

cocos/2d/components/rich-text.ts
LinYunMo: The logic here is to jump out of the loop when the remaining width is less than 2048, at this time, the penultimate line and the last line should be pushed in, the previous bug is missing the penultimate line

SantyWang (approved):

PR #14779, opened 2023-04-12
Introduce pose graph, pose node and pose state

shrinktofit:
Re: #

### Changelog

* Introduce pose graph, pose node and pose states. Detailed as below.

#### Pose State

The state machine now welcomed a new kind of state. Say: **pose state**. It has the following characteristics:

- A pose state contains a pose graph (see below).

- In initialization stage of animation graph, the pose graph gets into instantiation, then the root node of the (instantiated) graph starts its bind stage. After initialization stage, the root node starts its settling stage.

- If a state machine runs into a pose state, it will fire up the root pose node's re-entering, and then keep updating and evaluating it to get a pose object. Just like motion state, if the pose state is running standalone, the result pose would be the result of state machine. Otherwise, the result pose would be blended with other transitioning states' poses.

- In any place that a "state duration" is needed, the duration of pose states is defined as 1s.

- Transitions outgoing from a pose state called **pose transitions**. Pose transitions have duration, in unit of seconds. They do not have "relative duration".

- If a pose state is transitioned from a motion state/empty state/pose state, the "destination start time"/"relative destination start time" attribute on that transition is ignored.

#### Pose Graph and Pose Node

- A **pose graph** is composed by a set of nodes called **pose nodes** and their input/output links(in future). There could be and could only be one pose node marked as "root node".

- Pose nodes generate pose objects, it has the following behaviours:

  - bindings - Perform some bind operations such as grabbing transform handle for bones.
  - settling - Perform some later-bind operations such create transform filters for animation mask.
  - re-entering - This will be called if this node's master needs a clear state. For example, in state machine, the activation of a pose state will cause the root pose node to be re-entered.
  - updating - Perform the well-known updating operation.
  - evaluating - Evaluates and results a pose object.

-------

### Continuous Integration

This pull request:

* [ ] needs automatic test cases check.
  > Manual trigger with `@cocos-robot run test cases` afterward.
* [ ] does not change any runtime related code or build configuration
  > If any reviewer thinks the CI checks are needed, please uncheck this option, then close and reopen the issue.

-------

### Compatibility Check

This pull request:

* [ ] changes public API, and have ensured backward compatibility with [deprecated features](https://github.com/cocos/cocos-engine/blob/v3.5.0/docs/contribution/deprecated-features.md).
* [ ] affects platform compatibility, e.g. system version, browser version, platform sdk version, platform toolchain, language version, hardware compatibility etc.
* [ ] affects file structure of the build package or build configuration which requires user project upgrade.
[...]

PR #14952, opened 2023-04-28
Complementarily implement pose graph

shrinktofit:
Re: #

### Changelog

* This PR complementarily implements the pose graph introduced in PR #14779 .

### Design Detail

- There're values flowing in pose graph. The type of value can be pose, floating point number, integer, boolean, vector 3, quaternion. The type is represented by enumeration `PoseGraphType`.

- Pose graphs contains nodes, referenced as **pose graph node**, and is represented by class `PoseGraphNode`.

- Pose graph nodes, decided by its node class, can have zero or more inputs and zero or more outputs. Each input or output has definitely a type associated. Values flow between inputs/outputs:

  - An input receives value of that type from other nodes' output, or from constant;

  - An output sends value of that type to other nodes' input.

  - Nodes having outputs usually take their possible inputs and generate outputs.
 
  If such a flowing is established between a output and a input, say that a **binding** is formed between them.

- This PR introduces only 3 kinds of pose graph nodes (as subclass of `PoseGraphNode`):

  - **pose node**, represented by class `PoseNode`, as defined in PR #14779 .

    Pose nodes have **exactly one** output of pose type, and zero or more inputs of any type.

  - **x-node**, represented by class `XNode`.

    X-nodes have **one or more** outputs, and have zero or more inputs. **Neither** of input and output of a x-node can have pose type. 

  - **pose graph output node**, represented by class `PoseGraphOutputNode`.

    The pose graph output node is unique (per pose graph), built-in and undeletable. It has exactly one input of pose type and has no output. Pose flowed to its input would be served as the output pose of the animation graph.

#### Internal Concept: Node Shell

A pose graph node is a freestanding object, meaning it does not belong to any pose graph yet, can be directly created by `new` operator. When a pose graph node is added into graph, say it's non-freestanding. A pose graph node can not be added into more than one pose graph.

Internally, when a pose graph node is added into graph, a **node shell** is automatically created and associated to it. The shell object records the "incoming" bindings of the node. The reason of such design is:

- In general, in an abstract graph data structure, nodes and edges are stored separately. The shell objects are essentially "single forward edge".

- Pose graph nodes themselves don't have to concern about their input/output connections. They just focus on computation logic. Value flowing should be handled by binding system.

- This makes duplication easy:

  - Make duplicating a single node easy: only duplicate the node itself.

  - Make duplicating a subgraph easy: duplicate the subgraph nodes and the bindings limited to the subgraph.

Node shells are never known by external from graph.

#### Input declaration

- Inputs of a node are _mapped_ to the node's properties by applying decorator `@input` to the properties(both data field and accessor is ok):

  - `@input` can only be applied to properties of subclasses of `PoseNode` and `XNode`.

  - `@input` requires a type to be specified. `@input` specifying pose type can only be applied to properties of subclasses of `PoseNode`.

  - If `@input` is well applied to a property, the property is mapped as one or more inputs:

    - If the property is **NOT** an array, it's mapped as a single input with specified type.

    - If the property is an array, each element of it is mapped as an input with specified type.

  If a property or its element is mapped as input, value received from the input will be set onto that property or to that element.

- 

#### Output declaration

- Pose nodes don't have to declare their outputs.

- X-nodes declares their output count, type by specifying them to the base class `XNode`.

#### Instantiation

When a animation graph is instantiated(in per animation controller), the within pose graphs will be instantiated. The instantiation starts from the graph's output node and go recursively: each reached node is instantiated, then their input/output is connected.

The instantiation of a pose graph node(note: only pose node and x-node can be and will be instantiated) is call `cc.instantiate()` on it, that's, duplicate a copy of it. The evaluation status is recorded directly in the copy. This method very differs from previous: the instantiating stuff has a `createEval()` which yields another type of object "xx-evaluator", then pass data from original stuff to the evaluator. The reason is:

- It's more convenient. Don't have to create another "xx-evaluator" and pass data. Evaluation is done in place.

- It does not waste memory in most time. Usually, all non-evaluator-specific data should be passed to the "xx-evaluator". So in memory usage aspect, they have a flat hand.

But this manner has a little cons: a node has two states: **being evaluated** and **not being evaluated**!

Let's talk about how the binding system work now. Binding system treats the following kinds of binding in different manner:

- The binding between pose nodes(_a_ --> _b_).

  When encountered such binding type, binding system directly assign pose node _a_ to _b_'s property or _b_'s property's element.

- The binding comes from X-Node _a_ and flows into X-Node/Pose Node(b)

  When encountered such binding type, binding system creates a `DependencyEvaluation` record object for _b_. When a `DependencyEvaluation` executes, it:

  - evaluates _a_ and store the result into array `outputs`.

  - fetch the corresponding result in `outputs` then assign to _b_'s property or _b_'s property's element.

  Binding system roughly injects `DependencyEvaluation` object into `PoseNode`. Then it will be called by `PoseNode`'s update() method. For recursive x-nodes connection, recursive `DependencyEvaluation`s are created for them and connected.

#### Node input identity representation

The node input requires to be uniquely referenced by binding system and in pose graph authoring(especially in editor), then we need something like "id" of node input.

The "id" of a node input is internally called **input path**(represented by `InputPath`) represented as a tuple `[string]` or `[string, number]`. The first element is the name of the class property mapped as input(s). The possible second element is the element index in case of array element input mapping.

The detail construction of a input path is opacity to external(for example, editor), but it's transparent within engine. To encapsulate, we used **input key**(represented by `InputKey`) as an alias of input path and expose it to external. The input key, for external, is only guaranteed to be a value which can be `JSON.stringfy`.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/animation/marionette/pose-graph/foundation/authoring/input-authoring.ts
+++ b/cocos/animation/marionette/pose-graph/foundation/authoring/input-authoring.ts
@@ -1,8 +1,9 @@
 import { Quat, Vec3, assertIsTrue, js } from '../../../../../core';
-import { PoseGraphNode, shellTag } from '../pose-graph-node';
+import { PoseGraphNode } from '../pose-graph-node';
 import { PoseGraphType } from '../type-system';
 import { NodeInputPath } from '../node-shell';
 import { OperationOnFreestandingNodeError } from '../errors';
+import type { PoseGraph } from '../../pose-graph';
 
 export type PoseGraphInputKey = NodeInputPath;
 
@@ -243,7 +244,7 @@ class NodeInputManager {
         return result;
     }
 
-    public deleteInput (node: PoseGraphNode, key: PoseGraphInputKey) {
+    public deleteInput (graph: PoseGraph, node: PoseGraphNode, key: PoseGraphInputKey) {
         const [
             propertyKey,
             elementIndex = -1,
@@ -268,6 +269,7 @@ class NodeInputManager {
             const { arraySyncGroup } = propertyInputRecord;
             if (arraySyncGroup) {
                 this._deleteInputInArraySyncGroup(
+                    graph,
                     node,
                     arraySyncGroup,
                     property.length,
@@ -277,10 +279,10 @@ class NodeInputManager {
             }
         }
 
-        deletePoseGraphNodeArrayElement(node, key);
+        deletePoseGraphNodeArrayElement(graph, node, key);
     }
 
-    public insertInput (node: PoseGraphNode, insertId: PoseGraphNodeInputInsertId) {
+    public insertInput (graph: PoseGraph, node: PoseGraphNode, insertId: PoseGraphNodeInputInsertId) {
         const propertyKey = insertId;
         const propertyInputRecord = this._getPropertyNodeInputRecord(node.constructor, propertyKey);
         if (!propertyInputRecord) {
@@ -301,6 +303,7 @@ class NodeInputManager {
             const { arraySyncGroup } = propertyInputRecord;
             if (arraySyncGroup) {
                 this._insertInputInArraySyncGroup(
+                    graph,
                     node,
                     arraySyncGroup,
                     property.length,
@@ -311,6 +314,7 @@ class NodeInputManager {
         }
 
         insertPoseGraphNodeArrayElement(
+            graph,
             node,
             [propertyKey, hint],
             createDefaultInputValueByType(propertyInputRecord.type),
@@ -337,6 +341,7 @@ class NodeInputManager {
     }
 
     private _insertInputInArraySyncGroup (
+        graph: PoseGraph,
         node: PoseGraphNode,
         syncGroup: ArrayPropertySyncGroup,
         expectedOriginalSyncLength: number,
@@ -354,6 +359,7 @@ class NodeInputManager {
                 continue;
             }
             insertPoseGraphNodeArrayElement(
+                graph,
                 node,
                 [syncedPropertyKey, insertHint],
                 createDefaultInputValueByType(syncedPropertyInputRecord.type),
@@ -362,6 +368,7 @@ class NodeInputManager {
     }
 
     private _deleteInputInArraySyncGroup (
+        graph: PoseGraph,
         node: PoseGraphNode,
         syncGroup: ArrayPropertySyncGroup,
         expectedOriginalSyncLength: number,
@@ -378,13 +385,13 @@ class NodeInputManager {
                 // To avoid un-expectations, interrupt.
                 continue;
             }
-            deletePoseGraphNodeArrayElement(node, [syncedPropertyKey, index]);
+            deletePoseGraphNodeArrayElement(graph, node, [syncedPropertyKey, index]);
         }
     }
 }
 
-function insertPoseGraphNodeArrayElement (node: PoseGraphNode, inputKey: PoseGraphInputKey, value: unknown) {
-    const shell = node[shellTag];
+function insertPoseGraphNodeArrayElement (graph: PoseGraph, node: PoseGraphNode, inputKey: PoseGraphInputKey, value: unknown) {
+    const shell = graph.getShell(node);
     if (!shell) {
         throw new OperationOnFreestandingNodeError(node);
     }
@@ -405,8 +412,8 @@ function insertPoseGraphNodeArrayElement (node: PoseGraphNode, inputKey: PoseGra
     shell.moveArrayElementBindingForward(propertyKey, elementIndex + 1, false);
 }
 
-function deletePoseGraphNodeArrayElement (node: PoseGraphNode, inputKey: PoseGraphInputKey) {
-    const shell = node[shellTag];
+function deletePoseGraphNodeArrayElement (graph: PoseGraph, node: PoseGraphNode, inputKey: PoseGraphInputKey) {
+    const shell = graph.getShell(node);
     if (!shell) {
         throw new OperationOnFreestandingNodeError(node);
     }
--- a/cocos/animation/marionette/pose-graph/foundation/authoring/input-authoring.ts
+++ b/cocos/animation/marionette/pose-graph/foundation/authoring/input-authoring.ts
@@ -42,7 +42,7 @@ export type PoseGraphNodeInputInsertId = string;
  * @en Describes the options used
  * when a pose node class property is going to be mapped as node input(s).
  */
-export interface NodeInputMappingOptions {
+export interface PoseGraphNodeInputMappingOptions {
     /**
      * @zh 此属性或属性的元素映射的输入类型。
      * 输入的类型必须和属性（的元素）的实际值类型匹配。
@@ -110,11 +110,11 @@ export interface NodeInputMappingOptions {
 }
 
 // eslint-disable-next-line @typescript-eslint/ban-types
-class NodeInputManager {
+class PoseGraphNodeInputManager {
     public setPropertyNodeInputRecord (
         constructor: Constructor,
         propertyKey: string,
-        options: NodeInputMappingOptions,
+        options: PoseGraphNodeInputMappingOptions,
     ) {
         let classInputRecord = this._classInputMap.get(constructor);
         if (!classInputRecord) {
@@ -457,4 +457,4 @@ function createDefaultInputValueByType (type: PoseGraphType) {
     }
 }
 
-export const globalNodeInputManager = new NodeInputManager();
+export const globalPoseGraphNodeInputManager = new PoseGraphNodeInputManager();
--- a/cocos/animation/marionette/pose-graph/foundation/authoring/input-authoring.ts
+++ b/cocos/animation/marionette/pose-graph/foundation/authoring/input-authoring.ts
@@ -1,0 +1,453 @@
+import { Quat, Vec3, assertIsTrue, js } from '../../../../../core';
+import { PoseGraphNode, shellTag } from '../pose-graph-node';
+import { PoseGraphType } from '../type-system';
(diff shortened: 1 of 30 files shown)


cocos/animation/marionette/pose-graph/foundation/authoring/node-authoring.ts
```diff
@@ -0,0 +1,97 @@
+import type { AnimationGraph } from '../../../animation-graph';
+import { PoseGraphNode } from '../pose-graph-node';
+
+export {};
+
+export type PoseGraphCreateNodeContext = {
```
SantyWang: Why don't use interface?

cocos/animation/marionette/pose-graph/foundation/pose-graph-node.ts
```diff
@@ -0,0 +1,67 @@
...
+     * 在结点添加到姿势图中时，该字段被置为相应的外壳对象；
+     * 当结点被移出姿势图时，该字段被置空。
+     * @en
+     * The node's shell in pose graph.
+     * This field is set to the corresponding shell object when this node is added into pose graph,
+     * is set to `undefined` when this node is removed from pose graph.
+     * @internal
+     */
+    get [shellTag] () {
+        return this._shell;
```
SantyWang: If the PoseGraphNode reference its shell, I don't think it can be called 'imperceptible'. And is there a possibility that a node needs to keep its shell even after it is moved out of the pose graph. For example, if 3 nodes and their bindings are copied to another PoseGraph at the same time.

cocos/animation/marionette/pose-graph/x-node.ts
```diff
@@ -0,0 +1,43 @@
+import { VarInstance } from '../variable';
+import { PoseGraphNode } from './foundation/pose-graph-node';
+import { PoseGraphType } from './foundation/type-system';
+
+type Outputs = unknown[];
+
+export abstract class XNode extends PoseGraphNode {
```
SantyWang: it's better named `ValueNode` .

cocos/animation/marionette/pose-graph/foundation/authoring/input-authoring.ts
```diff
@@ -0,0 +1,453 @@
...
+     *     poses: PoseNode[] = [];
+     *
+     *     \@input({ type: PoseGraphType.FLOAT, syncGroup: 'blend-item' })
+     *     weights: number[] = [];
+     * }
+     * ```
+     *
+     * If a new input is inserted into `poses`, there will be also a new input inserted into weights.
+     */
+    arraySyncGroup?: string;
```
SantyWang, line 109: What is this used for?

cocos/animation/marionette/pose-graph/foundation/authoring/input-authoring.ts
```diff
@@ -0,0 +1,453 @@
...
+}
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+class NodeInputManager {
```
SantyWang: PoseGraphNodeInputManager seems more clear.

cocos/animation/marionette/pose-graph/foundation/pose-graph-node.ts
shrinktofit: The node has its shell referenced here since no matter in runtime or during authoring, the binding system needs to be able obtain the shell of a node. Currently they achieve this through `PoseGraphNode[shellTag]`.

I know your question. You mean that if the shell is imperceptible to node, then node should not have the reference there. I once thought about this too. If we want to complete decouple node from shell, we have to work in another way, that's: let the graph holding the relationship between nodes and shells. In further, all methods query bindings should also requiring a "belonging pose graph object" passed to them. For example: querying a binding currently can be accomplished by `getInputBinding(node: PoseNode)`. If we change, the method should be changed as `getInputBinding(poseGraph: PoseGraph, node: PoseNode)`. And its implementation should be adjusted as something like:

```ts
function getInputBinding(poseGraph: PoseGraph, node: PoseNode) {
    const shell = poseGraph.getShell(node); // Query the shell of the node
    // Return bindings recorded in shell...
}
```

How do you feel on this?


---------------- 


> And is there a possibility that a node needs to keep its shell even after it is moved out of the pose graph.

By current design:

- a node after `new()` is "freestanding", meaning it does not belong to any graph yet and therefor doesn't have a shell.

- Once it's added into graph, it's no longer freestanding and shell is attached to them.  See: https://github.com/cocos/cocos-engine/pull/14952/files#diff-590888e6d619412a6cd2c0578d431fbb3340a122bb37bead1d6588764f4ceeecR64

- No node can be added to multiple graphs. See: https://github.com/cocos/cocos-engine/pull/14952/files#diff-590888e6d619412a6cd2c0578d431fbb3340a122bb37bead1d6588764f4ceeecR65-R67

- Once a node is removed from graph, its shell is then destroyed by the graph, it becomes freestanding again. See: https://github.com/cocos/cocos-engine/pull/14952/files#diff-590888e6d619412a6cd2c0578d431fbb3340a122bb37bead1d6588764f4ceeecR85

cocos/animation/marionette/pose-graph/x-node.ts
shrinktofit: The name `ValueNode` does not distinguish these nodes from pose nodes: pose nodes also yield values.

cocos/animation/marionette/pose-graph/foundation/authoring/input-authoring.ts
shrinktofit: The API document has illustrated a use case.

cocos/animation/marionette/pose-graph/foundation/authoring/input-authoring.ts
shrinktofit: Should we prefix all node related a `PoseGraphNode`? For example exported methods in https://github.com/cocos/cocos-engine/pull/14952/files/4d734418971df040a536acf0e30698b749e76d8e#diff-fef71171b3da4eb1de158627c4eaca12a9f456a43417c06075f2bb27814ad755 .

cocos/animation/marionette/pose-graph/foundation/authoring/node-authoring.ts
shrinktofit: It's equivalent. I could change, what about also employ it as a lint rule? https://typescript-eslint.io/rules/consistent-type-definitions

shrinktofit: Solved here: https://github.com/cocos/cocos-engine/pull/14952/commits/a0d0a5e38f616e2691f0dc57de3cdf7d22cec16e

cocos/animation/marionette/pose-graph/foundation/pose-graph-node.ts
SantyWang: Yeah, I know that binding and node usually need to reference to each other for quick and easy lookups, and if they were separated, each lookup might require a global table, which would be somewhat wasteful. So my point is that instead of emphasizing that the binding is not imperceptible to the node, which doesn't fit well with our implementation, it might be more intuitive to just state that the node will hold all of its bindings.

cocos/animation/marionette/pose-graph/x-node.ts
SantyWang: But PoseNode returns a Pose, doesn't it? The return values of Pose node and these value nodes don't feel like they have anything in common. And if Pose Node does have any similarity to XNode, maybe it would be better to have PoseNode inherit from ValueNode ?

cocos/animation/marionette/pose-graph/foundation/authoring/input-authoring.ts
SantyWang: Since I noticed that some other types have added prefixes, but this NodeInputManager has no prefix, so it's better to unify them. Secondly, there is already a Node in the engine, and without prefixing it, it is not possible to understand exactly what it is intended to do by looking at the name alone. 
image

shrinktofit: These exports are public, whereas `NodeInputManager` is interal implementation

cocos/animation/marionette/pose-graph/x-node.ts
shrinktofit: Now it's changed to be non-intrusive: https://github.com/cocos/cocos-engine/pull/14952/commits/31291aa8b9a7a62d1181621d4ae746a474b0b8a7

cocos/animation/marionette/pose-graph/foundation/pose-graph-node.ts
shrinktofit: Now it's changed to be non-intrusive: https://github.com/cocos/cocos-engine/pull/14952/commits/31291aa8b9a7a62d1181621d4ae746a474b0b8a7

cocos/animation/marionette/pose-graph/x-node.ts
shrinktofit: The term "x node" is now renamed as "pure value node": https://github.com/cocos/cocos-engine/pull/14952/commits/f3b5c17bb22c0fd5b63e9a20b010b7cdb78262e0

SantyWang (approved):

cocos/animation/marionette/pose-graph/foundation/authoring/input-authoring.ts
shrinktofit: Have prefixed them with "PoseGraph": https://github.com/cocos/cocos-engine/pull/14952/commits/9cc3c274925e8010d3f64fd5822523482d2cb14c

Issue #14959, opened 2023-05-04
@ccclass won't take effect if there're another class decorators returning new classes


shrinktofit:
Testcase files stub:

[Testcase.zip](https://github.com/cocos/cocos-engine/files/11396574/Testcase.zip)

shrinktofit:
Should be solved be https://github.com/cocos/cocos-engine/pull/14961 . Here's my feedback stub to issue emitter:

------------------

# Updated Reply

Now #14961 gets updated. We turn to provide the original "transferring" functionality in form of a wrapping decorator instead:

```ts
@ccclass('CardGameMode')
@ccclass.forward(hasCGEvents()) // Forward the decorator execution to `hasCGEvents()` as if there's `@hasCGEvents()`, but with bug solved
export class CardGameMode extends GameMode {
```

Now you don't have to manually call `_decorator.transferCCClass`.


---------------------

## First Reply

Hi.

We have looked through your code and had some conclusions:

1. Your provided decorator `@hasCGEvent()` does substitute the original decorating class with a brand new class.

    Cocos Creator currently does not handle this case. Since properties are executed before class decorators, Cocos Creator  will temporally stash the properties decorating by `@property` into "class", which is the class you will in later substitute. This make all `@property` loose their effects.

    To strike this, Creator in this PR provides a method, https://github.com/cocos/cocos-engine/pull/14961 , <del>You can call it after you substitute the class, that's, change here:
    ```diff
      newConstructorFunction.prototype = constructorFunction.prototype;
    ++ _decorator.transferCCClass(constructorFunction, newConstructorFunction);
      return newConstructorFunction;
    ```
    </del>

    > You should do it in another form. See update above.

2. Even you applied the change suggested above, you may noticed that you still can not get the properties to be serialized or shown in editor. That's a story related to how Creator find the constructor of an object: it simply take the `.constructor` of that object.

    However, objects created from `new CardGameMode()` never have `.constructor === CardGameMode`. Their `.constructor` would point to the `func` you wrapped. You can try to verify it.

    So, we're here suggest you that make the following change:

    ```diff
      if (result.__eventList) {
        // ...
  
    ++  result.constructor = newConstructorFunction;

        return result;
      }

    return result;
    ```

mikecoker:
This looks like a good solution. I'm assuming that I will still need to set the constructor function as stated in 2 of the first reply, but can now use the additional decorator instead of calling the transferCCClass method.

PR #14961, opened 2023-05-04
[DON'T MERGE] Add `_decorator.ccclass.forward()` helper decorator to allow transfer cc-class information

shrinktofit:
Re: #14959

### Changelog

* Add a helper decorator `_decorator.ccclass.forward()` to allow transfer cc-class information from one class to another. This function is introduced for solving problems similar to #14959 : user may return a different class in class decorator.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/core/utils/js-typed.ts
+++ b/cocos/core/utils/js-typed.ts
@@ -231,6 +231,7 @@ export function createMap (forceDictMode?: boolean): any {
 
 /**
  * @internal
+ * @engineInternal
  */
 export function transferCCClassIdAndName (
     // eslint-disable-next-line @typescript-eslint/ban-types
--- a/cocos/core/utils/js-typed.ts
+++ b/cocos/core/utils/js-typed.ts
@@ -230,6 +230,29 @@ export function createMap (forceDictMode?: boolean): any {
 }
 
 /**
+ * @internal
+ */
+export function transferCCClassIdAndName (
+    // eslint-disable-next-line @typescript-eslint/ban-types
+    originalConstructor: Function,
+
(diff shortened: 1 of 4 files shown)


cocos/core/utils/js-typed.ts
```diff
@@ -229,6 +229,29 @@ export function createMap (forceDictMode?: boolean): any {
     return map;
 }
 
+/**
+ * @internal
```
PPpro, line 233: for now, use engineInternal instead

shrinktofit:
As suggested by @jareguo  , we turn to provide the functionality in a wrapping decorator form, see: https://github.com/cocos/cocos-engine/pull/14961/commits/d078e6e69e77482104c6c7d9721c2bc33621edd8

cocos/core/utils/js-typed.ts
```diff
@@ -719,10 +743,12 @@ export function unregisterClass (...constructors: Function[]) {
         const p = constructor.prototype;
         const classId = p[classIdTag];
         if (classId) {
+            delete p[classIdTag];
```
jareguo, line 746: If we have a registerClass method, then it should satisfy the following:
unregister(A);
register(A);
// At this point, there should be no changes.

So I don't think it's very necessary to reset a class's own properties. Literally speaking, unregistering a class doesn't necessarily have to include resetting or clearing the target class. When a class needs to be unregistered, it's often no longer needed, so its own data can be safely garbage collected without the need for a reset.

jareguo (approved):

shrinktofit:
Labelled as "Don't Merge" since there's no further feedback from user.

PR #14990, opened 2023-05-06
Animation graph: state machine adjustment

shrinktofit:
Re: #

### Changelog

This PR adjusts some aspects related animation graph state machine:

- Remove the experimental transition interruption mechanism.

- Adds the support for multiple transitions.

- Redefine the evaluation behaviour of a state machine, especially related to transition exit condition.

- Optimize implementation of self-transitioning motion states.

Details are illustrated as following.

#### Remove Existing Transition Interruption Mechanism

The existing transition interruption mechanism was introduced in 3.6, in PR #11158 . It's marked as experimental until now. In practise, animation team found it's not worth maintaining:

- Users don't have requirements on it or give it a feed back. 

  Search in [discussion](https://forum.cocos.org/search?q=interruption), it appears that users don't actually talk about it. We're not able to collect much feedbacks.

- The mechanism is not able to satisfy the new requirement of pose graph, introduced in this version.

  The existing mechanism has two pitfalls make it becoming a obstacle to pose graph:

  - It stops the interrupted state from updating. This in practise make it unable to implement some effects about for example states in transition sequence `A --> B --> C` can update parallelly.

  - Transition `A --> B` can be interrupted by `A --> C`. This makes the transition relationship more complex and the definition of "current state" being vague. For example, given `A-->B-->C`, should we match interruptions from the middle state `B`?

  - We will have a similar but more nice alternative "multiple transitions" support.

#### Support multiple transitions

The multiple transitions is now supported to overcome pitfalls of mentioned existing interruption:

- If there's already been a transition sequence $[S_0, S_1, ..., S_n]$ , the state machine will still attempt to match a transition outgoing from $S_n$, if there's a match hit, the matched transition would be appended to the transition sequence and hence formed $[S_0, S_1, ..., S_n, S_{n+1}]$.

- In transition sequence $[S_0, S_1, ..., S_n]$ , all transitions are subjected to update, so as the involved states. As for the state weight allocation, the $S_n$ will have its weight **accumulated by** $t$, where $t$ is the transition ratio, and states in $[S_0, S_{n-1}]$ will recursively have their respective weights **accumulated by** "corresponding transition ratio but multiplied by $1 - t$".

- The update of transition sequence $[S_0, S_1, ..., S_n]$ is begin from last to first. If during update, found that the $i$-th transition has finished, then the prefix transitions $[S_0, S_1, ..., S_i]$ will be dropped, and the result transition sequence becomes $[S_i, ...,  S_n]$.

### Redefine the evaluation behaviour

Prior to this PR, a state machine has the following evaluation flow:

0. Let `deltaTime` be the update delta time passed to the state machine.
1. Loop util `deltaTime == 0`:
  1.1 If there's a transition:
     - Step the transition for `t` time, the `t` is no more than `deltaTime`.
     - `deltaTime -= t`
     - Go to 1 .
  1.2 Try to match a transition:
     - If matched:
       - `deltaTime -= t`, where `t` is the cost of the match, especially the required time to satsify the exit condition of the match transition.
       - Go to 1.
     - Otherwise, update current state, and break.

Look through the algorithm, you can find that the algorithm is complex. It's complex since it want to perform a **steady** evaluation, that's:

For delta time `t`, if after updating current state `s` (no more than `t`), an exit condition can be satisfied so that a transition can be matched, then it will first update `s`, then `s` will be subtracted from the delta time.

This is certainly more steady behaviour. But it increase the implementation complexity and sometimes "steady" can not be actually achieved:

- The steady behaviour requires the transition matching system not only returns the matched transition, but also returns a "cost time". And internally, the matching system have to confirm the "minimal cost time".

- The problem is actually a race between "update the state" and "update the state then match transition then update the transition". This solution does solve. But, there's another more complex race: "update the transition" and "update the transition then interruption", which is hard to solve.

After all, animation team decides to change the behaviour to simplify the evaluation. In a frame:

- First, loop try to match transitions, util no transition can be matched. During matching, no update will be performed. If the exit condition is not satisfied before update, the belonging transition won't be matched.

- Then, update the formed transition sequence, as well as the states.

The new mechanism has no other pitfalls if the FPS is not too small(ie. the detal time is too big).

#### Optimize implementation of self-transitioning motion states

Prior to this PR, a motion state will definitely have a `MotionStateEval` record. In the record, it however has two "ports". One is called "from port", another is called "to port". The reason it's implemented as such is to support self transition of a motion state:

- When there's a transition to the motion state, its "to port" will be activated.

- When the motion state becomes "current" state, its "from port" will be activated, with inherited animation progress from "to port".

- So, motion state self transition `A --> A` is actually the transition from "from port" to "to port".

To implement this, the code use many `if-else` to identify different ports. For example:

- When sampling a "current state", if it's a motion state, its "from port" will be sampled; Instead when sampling a "destination state", if it's a motion state, its "to port" will be sampled.

- When a state becomes "current state" and if the state is a motion state, then firstly asserts the state's "to port" is activated. Then transfer the animation progress of "to port" to "from port".

- When deciding if an exit condition holding, the transition matching system must know if the matching source state is "from port" or "to port".

- etc ...(you can view the code)

This PR after all optimized the implementation:

- A motion state, during evaluation, will be treated as a "virtual state machine". In code, it's called `VMSM`(Virtual Motion State Machine). This virtual state machine has two states: one is public to external(outsides the virtual state machine), one is private. Both the states has state kind `animation`. Entry of the virtual state machine is the public state. Both state sampling the same motion but has independent animation time.

- If there's a transition into the motion state, the transition will be redirected to the public state. But if there's a self-transition, the transition will be converted into a `public-state --> private-state` transition and reverse `private-state --> public-state`. Note, no other transitions can directly transition into the private state.

- Outgoing transitions(from the motion state), speed, state-machine-components will all be inherited by both the public and private state.

Here's a diagram illustrate this:

```mermaid
---
title: the internal illustration of a Motion State
---
stateDiagram-v2
    SomeStateA --> VMSM
   state VMSM {
       [*] --> public
       public --> [*]
       private --> [*]
       public --> private
       private --> public
   }
   VMSM --> SomeStateB
```

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/animation/marionette/state-machine/state-machine-eval.ts
+++ b/cocos/animation/marionette/state-machine/state-machine-eval.ts
@@ -443,8 +443,6 @@ class TopLevelStateMachineEvaluation {
                 break;
             }
 
-            ++iterations;
-
             const transition = this._matchNextTransition(matchingSource);
             if (!transition) {
                 break;
--- a/cocos/animation/marionette/state-machine/state-machine-eval.ts
+++ b/cocos/animation/marionette/state-machine/state-machine-eval.ts
@@ -1,7 +1,7 @@
 import { DEBUG } from 'internal:constants';
 import {
     StateMachine, State, isAnimationTransition,
-    SubStateMachine, EmptyState, EmptyStateTransition, TransitionInterruptionSource,
+    SubStateMachine, EmptyState, EmptyStateTransition,
     PoseState, PoseTransition,
 } from '../animation-graph';
 import { MotionEval, MotionEvalContext, MotionPort } from '../motion';
@@ -9,7 +9,7 @@ import { createEval } from '../create-eval';
 import { BindContext, validateVariableExistence, validateVariableType, VariableType } from '../parametric';
 import { ConditionEval, TriggerCondition } from './condition';
 import { MotionState } from './motion-state';
-import { warnID, assertIsTrue, assertIsNonNullable } from '../../../core';
+import { warnID, assertIsTrue, assertIsNonNullable, Pool, approx, clamp01 } from '../../../core';
 import { AnimationClip } from '../../animation-clip';
 import type { AnimationController } from '../animation-controller';
 import { StateMachineComponent } from './state-machine-component';
@@ -125,6 +125,7 @@ class TopLevelStateMachineEvaluation {
         this._topLevelEntry = entry;
         this._topLevelExit = exit;
         this._currentNode = entry;
+        entry.increaseActiveReference();
         this._resetTrigger = context.triggerResetter;
     }
 
@@ -145,10 +146,12 @@ class TopLevelStateMachineEvaluation {
     }
 
     public update (context: AnimationGraphUpdateContext) {
-        this._transitionAlpha = 0.0;
-        if (!this.exited) {
-            this._eval(context);
-        }
+        assertIsTrue(!this.exited);
+
+        this._loopMatchTransitions();
+        this._resetStateTickFlagsAndWeights();
+        this._updateActivatedTransitions(context.deltaTime);
+        this._commitStateUpdates(context);
     }
 
     public evaluate (context: AnimationGraphEvaluationContext): Pose {
@@ -162,9 +165,7 @@ class TopLevelStateMachineEvaluation {
     public getCurrentStateStatus (): Readonly<MotionStateStatus> | null {
         const { _currentNode: currentNode } = this;
         if (currentNode.kind === NodeKind.animation) {
-            return currentNode.getFromPortStatus();
-        } else if (currentNode.kind === NodeKind.transitionSnapshot) {
-            return currentNode.first.getFromPortStatus();
+            return currentNode.getStatus();
         } else if (currentNode.kind === NodeKind.pose) {
             return currentNode.status;
         } else {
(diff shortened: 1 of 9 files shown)


cocos/animation/marionette/state-machine/state-machine-eval.ts
```diff
@@ -175,107 +176,84 @@ class TopLevelStateMachineEvaluation {
...
-            return true;
-        } else {
+        const { _activatedTransitions: activatedTransitions } = this;
+        if (activatedTransitions.length === 0) {
             return false;
         }
+        const lastActivatedTransition = activatedTransitions[activatedTransitions.length - 1];
+        const baseDurationState = activatedTransitions.length === 1
+            ? this._currentNode
+            : activatedTransitions[activatedTransitions.length - 2].destination;
```
SantyWang: Why use the last two activated transition?

cocos/animation/marionette/state-machine/state-machine-eval.ts
```diff
@@ -626,95 +589,64 @@ class TopLevelStateMachineEvaluation {
     }
 
     /**
-     * Searches for a transition which should be performed
-     * if current node update for no more than `deltaTime`.
-     * @param deltaTime
+     * Searches for a transition which should be performed.
+     * @param sourceState The transition source state.
+     * @param useFromPort See `this._matchTransition`.
```
SantyWang: There is only one parameter.

cocos/animation/marionette/state-machine/state-machine-eval.ts
```diff
@@ -626,95 +589,64 @@ class TopLevelStateMachineEvaluation {
...
-
         return null;
     }
 
     /**
      * Notes the real node is used:
      * - to determinate the starting state machine from where the any states are matched;
      * - so we can solve transitions' relative durations.
-     * @param isCurrentState See `_matchTransition`.
+     * @param useFromPort See `_matchTransition`.
```
SantyWang: No useFromPort parameter.

cocos/animation/marionette/state-machine/state-machine-eval.ts
```diff
@@ -762,464 +688,168 @@ class TopLevelStateMachineEvaluation {
...
+            _activatedTransitions: activatedTransition,
+            _activatedTransitionPool: activatedTransitionPool,
         } = this;
+        assertIsTrue(lastTransitionIndex >= 0 && lastTransitionIndex < activatedTransition.length);
 
-        if (currentNode.kind !== NodeKind.animation
-            && currentNode.kind !== NodeKind.transitionSnapshot) {
-            return null;
-        }
+        const lenSubpath = (lastTransitionIndex - 0) + 1;
```
SantyWang: Why substract 0?

cocos/animation/marionette/state-machine/state-machine-eval.ts
shrinktofit: The `baseDurationState` is the state who fired the `lastActivatedTransition`, in other word, it's the "previous transition"'s destination, or current state.

For example:

- `A -> B`   ---- Would be `A`
- `(A -> B) -> C` ---- Would be `B`.

cocos/animation/marionette/state-machine/state-machine-eval.ts
shrinktofit: Solved https://github.com/cocos/cocos-engine/pull/14990/commits/1014b217fea545b78a27b644d43bfa3aab642536

cocos/animation/marionette/state-machine/state-machine-eval.ts
shrinktofit: Solved https://github.com/cocos/cocos-engine/pull/14990/commits/1014b217fea545b78a27b644d43bfa3aab642536

cocos/animation/marionette/state-machine/state-machine-eval.ts
shrinktofit: Solved https://github.com/cocos/cocos-engine/pull/14990/commits/1014b217fea545b78a27b644d43bfa3aab642536

SantyWang (approved):

PR #15029, opened 2023-05-11
[DON'T MERGE] Implement enum types as non-intrusive, support named enum

shrinktofit:
Re: #

### Changelog

* Implement enumeration types as non-intrusive.

* Support named enumerations:

  - Registering(unregister) an enumeration with name.

  - Query the name of an enumeration. (in editor)

  - Retrieving enumeration by name. (in editor)

* In `@type(enumType)`, instead of storing the "enumList", store the enumeration type itself.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/core/data/class.ts
+++ b/cocos/core/data/class.ts
@@ -28,7 +28,7 @@ import { errorID, warnID, error } from '../platform/debug';
 import * as js from '../utils/js';
 import { getSuper } from '../utils/js';
 import { BitMask } from '../value-types';
-import { Enum } from '../value-types/enum';
+import { Enum, EnumType } from '../value-types/enum';
 import * as attributeUtils from './utils/attribute';
 import { IAcceptableAttributes } from './utils/attribute-defines';
 import { preprocessAttrs } from './utils/preprocess-class';
@@ -406,6 +406,7 @@ const PrimitiveTypes = {
 interface IParsedAttribute extends IAcceptableAttributes {
     ctor?: Function;
     enumList?: readonly any[];
+    enumType?: EnumType;
     bitmaskList?: any[];
 }
 type OnAfterProp = (constructor: Function, mainPropertyName: string) => void;
--- a/cocos/core/value-types/enum.ts
+++ b/cocos/core/value-types/enum.ts
@@ -31,6 +31,22 @@ import { assertIsTrue } from '../data/utils/asserts';
 
 export type EnumType = Record<string, string | number>;
 
+interface EnumRegister {
+    /**
+     * `string` indicates the enumeration's name. `undefined` indicates the enumeration has no name.
+     */
+    name: string | undefined;
+
+    /**
+     * `null` if haven't not computed.
+     */
+    members: null | Enum.Member<EnumType>[];
+}
+
+const enumTypeRegistry = new WeakMap<EnumType, EnumRegister | null>();
+
+const enumNameLookupMap = new Map<string, EnumType>();
+
 /**
  * @en
  * Define an enum type. <br/>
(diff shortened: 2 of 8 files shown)


cocos/core/data/class.ts
```diff
@@ -455,7 +455,7 @@ function parseAttributes (constructor: Function, attributes: PropertyStash, clas
         else if (typeof type === 'object') {
             if (Enum.isEnum(type)) {
                 (attrs || initAttrs())[`${propertyNamePrefix}type`] = ENUM_TAG;
-                attrs![`${propertyNamePrefix}enumList`] = Enum.getList(type);
+                attrs![`${propertyNamePrefix}enumType`] = type;
```
jareguo: Currently, many developers, especially plugin authors, directly use the `enumList` variable to control the dropdown list in the inspector. How can we ensure compatibility after making changes?

cocos/core/value-types/enum.ts
```diff
@@ -29,6 +29,24 @@ import { legacyCC } from '../global-exports';
 import { errorID } from '../platform/debug';
 import { assertIsTrue } from '../data/utils/asserts';
 
+type EnumType = Record<string, string | number>;
+
+interface EnumRegister {
```
jareguo, line 34: enumData? or CCEnum?

cocos/core/value-types/enum.ts
```diff
@@ -29,6 +29,24 @@ import { legacyCC } from '../global-exports';
...
+
+    /**
+     * `null` if haven't not computed.
+     */
+    enumerators: null | Enum.Enumerator<EnumType>[];
+}
+
+const enumTypeRegistry = new WeakMap<EnumType, EnumRegister | null>();
+
+const enumNameReverseLookupMap = new Map<string, EnumType>();
```
jareguo: I don't think there's a need to emphasize "reverse" here. Since it's a register, it's obviously mapping from name to type.

cocos/core/value-types/enum.ts
```diff
@@ -29,6 +29,24 @@ import { legacyCC } from '../global-exports';
...
+     */
+    name: string | undefined;
+
+    /**
+     * `null` if haven't not computed.
+     */
+    enumerators: null | Enum.Enumerator<EnumType>[];
+}
+
+const enumTypeRegistry = new WeakMap<EnumType, EnumRegister | null>();
```
jareguo, line 46: Why not just add a Symbol as an attribute directly? Like `enumTypeObj[Symbol("EnumRegister")] = registerObj`;

PPpro (approved):

cocos/spine/skeleton.ts
```diff
@@ -94,7 +94,7 @@ ccenum(AnimationCacheMode);
 
 function setEnumAttr (obj, propName, enumDef) {
     CCClass.Attr.setClassAttr(obj, propName, 'type', 'Enum');
-    CCClass.Attr.setClassAttr(obj, propName, 'enumList', Enum.getList(enumDef));
+    CCClass.Attr.setClassAttr(obj, propName, 'enumType', enumDef);
```
arsen2010: It can not remove 'enumList' which be used so many places in Editor

arsen2010: Maybe the ‘enumType‘ can be an additional prop name，detail with 'type ( Enum) '.

cocos/core/data/class.ts
shrinktofit: How about add a new record here instead of replace it?

cocos/core/value-types/enum.ts
shrinktofit: It's not intrusive and it make it possible to `const enum E{}; ccenum(E)` or `const E = Object.freeze({}); ccenum(E)`

cocos/core/value-types/enum.ts
shrinktofit: Fixed in d209f889ed711c60c2ab5c84ba899347aaadb5d8

cocos/spine/skeleton.ts
shrinktofit: I have fixed all editor references in here: https://github.com/cocos/cocos-editor/pull/1959

cocos/core/data/class.ts
shrinktofit: Should have been fixed here: https://github.com/cocos/cocos-engine/pull/15029/commits/59666a14b5c643b105827f1022d24078e3654731

shrinktofit:
I will convert this into draft temporarily.

cocos/spine/skeleton.ts
shrinktofit: Update: I've added enumList back

PR #15032, opened 2023-05-11
gradient misc optimization and add some helpful function

SantyWang:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/core/curves/gradient.ts
+++ b/cocos/core/curves/gradient.ts
@@ -133,11 +133,23 @@ export class Gradient {
     /**
      * @en Interpolate color and alpha from color and alpha keys.
      * @zh 根据颜色列表插值计算颜色和透明度。
+     * @param time @en Normalized time to interpolate. @zh 用于插值的归一化时间。
+     * @returns @en Interpolated color value. @zh 插值过后的颜色值。
+     *
+     * @deprecated since v3.8 please use [[evaluateFast]] instead
+     */
+    public evaluate (time: number) {
+        return this.evaluateFast(new Color(), time);
+    }
+
+    /**
+     * @en Interpolate color and alpha from color and alpha keys.
+     * @zh 根据颜色列表插值计算颜色和透明度。
      * @param out @en Interpolated color value. @zh 插值过后的颜色值。
      * @param time @en Normalized time to interpolate. @zh 用于插值的归一化时间。
      * @returns @en Interpolated color value. @zh 插值过后的颜色值。
      */
-    public evaluate (out: Color, time: number) {
+    public evaluateFast (out: Color, time: number) {
         this.getRGB(out, time);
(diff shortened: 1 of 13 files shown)


zxx43 (approved):

cocos/core/curves/gradient.ts
```diff
@@ -92,24 +94,18 @@ export default class Gradient {
      * @en Array of color key.
      * @zh 颜色关键帧列表。
      */
-    public colorKeys = new Array<ColorKey>();
+    public colorKeys: ColorKey[] = [];
```
minggo, line 97: Is there any reason to change it?

cocos/scene-graph/node.ts
```diff
@@ -1749,6 +1749,14 @@ export class Node extends CCObject implements ISchedulable, CustomSerializable {
         return this._layer;
     }
 
+    /**
+     * @zh 节点的变换改动版本号。
+     * @en The transformation change version number of the node.
+     */
+    get flagChangedVersion () {
```
minggo, line 1758: Why need this value? As it is used internally, if the value is returned, then can not change to mechanism in future.

cocos/core/curves/gradient.ts
SantyWang: I think this looks more clearer

cocos/scene-graph/node.ts
SantyWang: In some scenarios we need to fetch the value and cache it, and compare the current value with the cached value when needed to calculate whether there has been a change in that time.

minggo: Could you please describe the scenarios that need it as there is `Node.hasChangedFlags()`;

SantyWang: `Node.hasChangedFlags` can only be used to distinguish if there is a change in this frame, if I need to know if a change has occurred in a certain period of time, I need to call Node.hasChangedFlags once for each frame in that period of time. 
With this new interface, I only need to record version once before that time, and again after that time, and compare the following two, if they don't match, then a change has occurred, which is much better and avoids many special cases

minggo: Got it, can i know what module need it? As it is implemented depended, so it is better to public it as possible.

SantyWang: UI and particle will need it. ui can use it to determine when should transform vertices position into world space. particle use it to recalculate the matrix when the version changed.

minggo: If so, it is better to mark it as engine internal.

SantyWang: marked as engine internal

minggo (approved):

PR #15064, opened 2023-05-16
fix wrong orientation set

bofeng-song:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/scripts/native-pack-tool/source/platforms/android.ts
+++ b/scripts/native-pack-tool/source/platforms/android.ts
@@ -129,8 +129,10 @@ export class AndroidPackTool extends NativePackTool {
             const pattern = /android:screenOrientation="[^"]*"/;
             let replaceString = 'android:screenOrientation="unspecified"';
 
-            if (cfg.landscapeRight && cfg.landscapeLeft && cfg.portrait && cfg.upsideDown) {
+            if (cfg.landscapeRight && cfg.landscapeLeft && (cfg.portrait || cfg.upsideDown)) {
                 replaceString = 'android:screenOrientation="fullSensor"';
+            } else if ((cfg.landscapeRight || cfg.landscapeLeft) && (cfg.portrait || cfg.upsideDown)) {
+                replaceString = 'android:screenOrientation="unspecified"';
             } else if (cfg.landscapeRight && !cfg.landscapeLeft) {
                 replaceString = 'android:screenOrientation="landscape"';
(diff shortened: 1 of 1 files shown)


scripts/native-pack-tool/source/platforms/android.ts
```diff
@@ -129,8 +129,10 @@ export class AndroidPackTool extends NativePackTool {
             const pattern = /android:screenOrientation="[^"]*"/;
             let replaceString = 'android:screenOrientation="unspecified"';
 
-            if (cfg.landscapeRight && cfg.landscapeLeft && cfg.portrait && cfg.upsideDown) {
+            if (cfg.landscapeRight && cfg.landscapeLeft && (cfg.portrait || cfg.upsideDown)) {
```
bofeng-song, line 132: image
Currently, orientation only have three options

scripts/native-pack-tool/source/platforms/android.ts
```diff
@@ -129,8 +129,10 @@ export class AndroidPackTool extends NativePackTool {
...
                 replaceString = 'android:screenOrientation="fullSensor"';
+            } else if ((cfg.landscapeRight || cfg.landscapeLeft) && (cfg.portrait || cfg.upsideDown)) {
+                replaceString = 'android:screenOrientation="unspecified"';
```
minggo, line 135: What does `unspecified ` mean?

PatriceJiang (approved):

scripts/native-pack-tool/source/platforms/android.ts
bofeng-song: In Android, if the value of the android:screenOrientation attribute is set to unspecified, it means that the screen orientation is not specified. This means that the system will automatically choose the screen orientation based on the device's sensors and other conditions.

Specifically, if the android:screenOrientation attribute is set to unspecified, the system will handle the screen orientation as follows:

If the device's orientation is locked, the system will use the locked orientation.
If the device's orientation is not locked and the device is in landscape mode, the system will use the landscape orientation.
If the device's orientation is not locked and the device is in portrait mode, the system will use the portrait orientation.
In some cases, using unspecified screen orientation may result in unexpected behavior, such as when the application needs to run in a fixed orientation. In such cases, the screen orientation can be explicitly specified by setting the android:screenOrientation attribute to an appropriate value.

minggo (approved):

PR #15090, opened 2023-05-17
Fix the issue of unable to click on QNX platform.

wangzhenGitHup:
Fix the issue of unable to click on QNX platform.

### DIFF
--- a/pal/system-info/native/system-info.ts
+++ b/pal/system-info/native/system-info.ts
@@ -56,7 +56,8 @@ const platformMap: Record<number, Platform> = {
     // 5 is IPAD
     5: Platform.IOS,
     6: Platform.OHOS,
-    7: Platform.OPENHARMONY,
+	7: Platform.QNX,
+    10: Platform.OPENHARMONY,
 };
 
 class SystemInfo extends EventTarget {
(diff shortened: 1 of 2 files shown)


pal/system-info/native/system-info.ts
```diff
@@ -88,7 +89,7 @@ class SystemInfo extends EventTarget {
 
         this.platform = platformMap[__getPlatform()];
         // eslint-disable-next-line max-len
-        this.isMobile = this.platform === Platform.ANDROID || this.platform === Platform.IOS || this.platform === Platform.OHOS || this.platform === Platform.OPENHARMONY;
+        this.isMobile = this.platform === Platform.ANDROID || this.platform === Platform.IOS || this.platform === Platform.OHOS || this.platform === Platform.OPENHARMONY || this.platform === Platform.QNX;
```
minggo, line 92: QNX is not a mobile platform.

pal/system-info/native/system-info.ts
```diff
@@ -56,7 +56,8 @@ const platformMap: Record<number, Platform> = {
     // 5 is IPAD
     5: Platform.IOS,
     6: Platform.OHOS,
-    7: Platform.OPENHARMONY,
+	7: Platform.QNX,
+    10: Platform.OPENHARMONY,
```
qiuguohua: image
The value of qnx is 8?

pal/system-info/native/system-info.ts
qiuguohua: 'Platform.QNX' is a string, defined in pal/system-info/enum-type/platform.ts.

wangzhenGitHup: The QNX device being adapted has some mobile platform properties, such as touchscreen capabilities.

minggo: Touch is not a criterion for judging mobile devices. Does it not work properly without being set as a mobile device?

wangzhenGitHup:
Currently, QNX is adapted to version3.5.0 of the engine. If there is a need to adapt to a higher version of the engine in the future, we will consider it then.

minggo:
Why close it?

wangzhenGitHup:
> Why close it?

Currently, the QNX adaptation engine version is 3.5.0. According to Guohua, this test is using version 3.8.Therefore, we will consider adapting to higher versions of QNX in the future.

PR #15129, opened 2023-05-21
remove `rangeMin` decorator from camera near

mvrlin:
Re: When using Ortographic Projection it doesn't allow to set `near` less than zero and because of that, we can see through the mesh when it's close to the bottom of camera

As a temporary fix you can set `_near` property inside Camera component in scene
image

### Changelog

* remove `rangeMin` decorator from camera near

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/misc/camera-component.ts
+++ b/cocos/misc/camera-component.ts
@@ -372,7 +372,6 @@ export class Camera extends Component {
      * @zh 相机的近裁剪距离，应在可接受范围内尽量取最大。
      */
     @displayOrder(10)
-    @rangeMin(0)
     @tooltip('i18n:camera.near')
     get near () {
         return this._near;


mvrlin:
Attaching a demo project
[near-demo.zip](https://github.com/cocos/cocos-engine/files/11551384/near-demo.zip)

mvrlin:
@pandamicro any update on this?

minggo:
It sounds reasonable.
@GengineJS please review it.

GengineJS:
In theory, setting the value of the near clipping plane to less than zero may lead to issues. This is because the near clipping plane is typically used to determine objects within the camera's field of view and is employed in depth testing. If the near clipping plane is set to a negative value, it may result in incorrect depth testing, potentially causing rendering anomalies or unpredictable outcomes.

star-e:
We are fixing this in cocos creator.

PR #15179, opened 2023-05-25
Fix buttons animation not always working on Android devices

### LINKED ISSUE #15178: TOUCH_MOVE events does not trigger properly at the bottom corners of the screen (on Android devices)
### Cocos Creator version

3.6.3, 3.7.3

### System information

Windows 10 Editor

### Issue description

When buttons are placed at the bottom corners of the screen (like the below image), their animations won't play when running the game on Android device. 

image

Tapping the button on Android does not trigger the button animation. Only if I hold my finger on the button for 1 second the animation will play. This is not an issue when buttons are not placed very close to the bottom corners of the screen.

After some digging in the source code it seems like that for some reason the "TOUCH_MOVE" event does not trigger properly in the bottom corners of the screen.

I suspect that this bug is introduced by #12243 but I haven't tested earlier versions of the engine to be sure.

### Relevant error log output

_No response_

### Steps to reproduce

- Create a new empty project
- Put a new button at the bottom corner of the screen (either left or right). 
- Run the game on an Android device and tap the buttons to see if they're animations are working.

_No response_

mrtzd:
Re: # This pr fixes #15178, although the underlying issue that causes this bug is not fixed in this patch.

### Changelog

* Fix buttons animation not always playing on Android devices

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/ui/button.ts
+++ b/cocos/ui/button.ts
@@ -577,6 +577,8 @@ export class Button extends Component {
     private _originalScale: Vec3 | null = null;
     private _sprite: Sprite | null = null;
     private _targetScale: Vec3 = new Vec3();
+    private _pendingStartAnim = false;
+    private _pendingEndAnim = false;
 
     public __preload () {
         if (!this.target) {
@@ -635,6 +637,7 @@ export class Button extends Component {
         }
 
         if (this._transition !== Transition.COLOR && this._transition !== Transition.SCALE) {
+            this._updateTransitionState();
             return;
         }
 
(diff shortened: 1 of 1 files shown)


mrtzd:
A bit more explanation about this pull request:
Right now if you tap a button on an Android device normally the chain of events that would occur is like this:

`
one TOUCH_START event ---> at least one or more TOUCH_MOVE events ---> one TOUCH_END event
`

This way the button animation plays correctly and there's no problem.

But if you put a button at the bottom corners of the screen, the chain of events after tapping the button is like this:

`one TOUCH_START event ---> one TOUCH_END event`

This way the button animation won't play at all because with the current implementation it requires that at least one TOUCH_MOVE event is triggered so that the button animation is played properly.

What I changed in this pr is that button animations will play correctly even when there are no TOUCH_MOVE events.

Now I don't know the reason for why TOUCH_MOVE events don't trigger properly when buttons are at the bottom corners of the screen. Maybe there's a bug with how touch input is being handled on Android. I looked at the relevant code for a bit but I wasn't able to identify the issue so I made this patch as a workaround.

LinYunMo:
Reasons for closing see： https://github.com/cocos/cocos-engine/issues/15178#issuecomment-1635384785

PR #13720, opened 2022-12-07
synchronize spine anchor to UITransform

zvars:
### Changelog

* [spine] calculate anchor point of UITransform by skeleton offset in the bounding box.

-------

### Continuous Integration

This pull request:

* [x] needs automatic test cases check.
  > Manual trigger with `@cocos-robot run test cases` afterward.
* [x] does not change any runtime related code or build configuration
  > If any reviewer thinks the CI checks are needed, please uncheck this option, then close and reopen the issue.

-------

PR #15215, opened 2023-05-29
revert to set anchor Point

zhakesi:


zhakesi:
* https://github.com/cocos/cocos-engine/pull/13720

PR #15279, opened 2023-06-01
onMacroPatchesStateChanged func performance optimization

troublemaker52025:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/scene/SubModel.cpp
+++ b/native/cocos/scene/SubModel.cpp
@@ -220,7 +220,7 @@ void SubModel::onPipelineStateChanged() {
 }
 
 void SubModel::onMacroPatchesStateChanged(const ccstd::vector<IMacroPatch> &patches) {
-    if (patches.empty() && patches.empty()) {
+    if (patches.empty() && _patches.empty()) {
         return;
     }
 
--- a/native/cocos/scene/SubModel.cpp
+++ b/native/cocos/scene/SubModel.cpp
@@ -145,7 +145,9 @@ void SubModel::initialize(RenderingSubMesh *subMesh, const SharedPassArray &pPas
     }
 
     _subMesh = subMesh;
-    _patches = patches;
+    ccstd::vector<IMacroPatch> tmp = patches;
+    std::sort(tmp.begin(), tmp.end(), IMacroPatch::compare);
+    _patches = tmp;
     _passes = pPasses;
 
     flushPassInfo();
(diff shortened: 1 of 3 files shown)


troublemaker52025:
@cocos-robot run test cases

shaoqiangcai:
@cocos-robot run test cases

jk20012001 (approved):

native/cocos/scene/SubModel.cpp
```diff
@@ -218,7 +220,16 @@ void SubModel::onPipelineStateChanged() {
 }
 
 void SubModel::onMacroPatchesStateChanged(const ccstd::vector<IMacroPatch> &patches) {
-    _patches = patches;
+    if (!patches.empty() && !patches.empty()) {
```
minggo: !patches.empty() && !patches.empty() -> patches.empty() && _patches.empty()

native/cocos/scene/SubModel.cpp
```diff
@@ -218,7 +220,16 @@ void SubModel::onPipelineStateChanged() {
...
+        return;
+    }
+
+    ccstd::vector<IMacroPatch> tmp = patches;
+    std::sort(tmp.begin(), tmp.end(), IMacroPatch::compare);
+    if (std::equal(std::begin(tmp), std::end(tmp), std::begin(_patches))) {
```
minggo: I think this feature is c++17 depend, it is better to pass `std::end(_patches)` too. For example
```c++
std::equal(std::begin(tmp), std::end(tmp), std::begin(_patches), std::end(_patches))
```

native/cocos/scene/SubModel.cpp
minggo: And i think it is better to change like this
```c++
if (_patches.empty()) _patches = patches;
else if (patches.empty()) _patches.clear();
else {
    // other logic;
}
```

minggo: And may be doesn't have to check it. If both of `patches` and `_patches` are empty, they can also use the same logic below. As empty equals empty.

native/cocos/scene/SubModel.cpp
troublemaker52025: > ```c++
> , std::end(_patches)
> ```

sure

native/cocos/scene/SubModel.cpp
```diff
@@ -218,7 +220,16 @@ void SubModel::onPipelineStateChanged() {
 }
 
 void SubModel::onMacroPatchesStateChanged(const ccstd::vector<IMacroPatch> &patches) {
-    _patches = patches;
+    if (patches.empty() && patches.empty()) {
```
minggo: patches.empty() -> _patches.empty()

minggo: Current code check `patches.empty()` twice.

minggo (approved):

PR #15330, opened 2023-06-06
Refactor animation graph event dispatching mechanism

shrinktofit:
Re: #

### Changelog

* This PR refactors the animation graph event dispatching mechanism that was initially introduced in PR #14911  .

During discussion with @jareguo , @holycanvas , @AndyYuTian  and @VisualSJ . We had agreed that the mechanism used to dispatch state machine event should match the mechanism how animation clip frame events are dispatched. But we had a disagreement on which mechanism should be selected: @jareguo and @holycanvas presented that they prefer the original mechanism for the following reasons:

- Easier to use. Users no need to explicitly do the registration. Since no registration is needed, no deregistration is needed too.

However @shrinktofit presented an opposite view:

- The original mechanism may bring perform issue: there're iteration needed to finish the invocation.

- And the name problem: a component may just right have a method with name specified by the event, but it's not designed to be used as animation event handler.

After all, the minority should be subordinate to the majority. In this PR, the animation graph event dispatching mechanism has been refactored:

- Users should specify a "method name" to event binding on state machine events, instead of "event name".

- At runtime, when the event is triggered, user handler code are found in a way identical to animation clip frame events: each component of the node attaching the animation controller would be inspected to search for a method with name "method name".

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


cocos/animation/event/event-emitter.ts
```diff
@@ -0,0 +1,24 @@
...
+ * @param node The node to search components with.
+ * @param methodName The method name to invoke.
+ * @param args The arguments passed to the method.
+ */
+export function invokeComponentMethodsEngagedInAnimationEvent (node: Node, methodName: string, ...args: unknown[]) {
+    const components = node.components;
+    const nComponents = components.length;
+    for (let iComponent = 0; iComponent < nComponents; ++iComponent) {
+        const component = components[iComponent];
+        if (!component.enabled) {
```
jareguo: Could @AndyYuTian  please help confirm whether the compatibility break here is worth it?

AndyYuTian: In our competitor's similar feature,  it does not check for the component enable/disable status. In addition, since our embedded player does not check for the component status as well, i dont think we need to break the compatibility with this change.

SantyWang (approved):

jareguo (approved):

AndyYuTian:
approved this feature update.

PR #15350, opened 2023-06-07
support Pacer to set specific frameRate without frame lost

novain1102:
Re: #https://github.com/cocos/3d-tasks/issues/16538

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/pal/pacer/pacer-web.ts
+++ b/pal/pacer/pacer-web.ts
@@ -25,6 +25,8 @@
 import { EDITOR } from 'internal:constants';
 import { assertIsTrue } from '../../cocos/core/data/utils/asserts';
 
+const FRAME_RESET_TIME = 2000;
+
 export class Pacer {
     private _stHandle = 0;
     private _onTick: (() => void) | null = null;
@@ -32,13 +34,13 @@ export class Pacer {
     private _frameTime = 0;
     private _startTime = 0;
     private _isPlaying = false;
+    private _frameCount = 0;
     private _callback: (() => void) | null = null;
-    private _delay = 0;
-    private _start = 0;
     private _rAF: typeof requestAnimationFrame;
     private _cAF: typeof cancelAnimationFrame;
 
     constructor () {
+        this._frameTime = 1000 / this._targetFrameRate;
         this._rAF = window.requestAnimationFrame
         || window.webkitRequestAnimationFrame
         || window.mozRequestAnimationFrame
@@ -84,7 +86,6 @@ export class Pacer {
         if (this._isPlaying) return;
 
         const updateCallback = () => {
-            this._startTime = performance.now();
             if (this._isPlaying) {
                 this._stHandle = this._stTime(updateCallback);
             }
@@ -96,6 +97,7 @@ export class Pacer {
         this._stHandle = this._stTime(updateCallback);
 
         this._isPlaying = true;
+        this._frameCount = 0;
     }
 
     stop (): void {
@@ -103,25 +105,33 @@ export class Pacer {
         this._ctTime(this._stHandle);
         this._stHandle = 0;
         this._isPlaying = false;
+        this._frameCount = 0;
     }
 
     _handleRAF = () => {
-        if (performance.now() - this._start < this._delay) {
+        const elapseTime = performance.now() - this._startTime;
+        const elapseFrame = Math.floor(elapseTime / this._frameTime);
+        if (elapseFrame < this._frameCount) {
             this._rAF.call(window, this._handleRAF);
-        } else if (this._callback) {
-            this._callback();
+        } else {
+            this._frameCount++;
+            if (this._callback) {
+                this._callback();
+            }
+        }
+        if (performance.now() - this._startTime > FRAME_RESET_TIME) {
+            this._startTime = performance.now();
+            this._frameCount = 0;
         }
     };
 
     private _stTime (callback: () => void) {
-        const currTime = performance.now();
-        const elapseTime = Math.max(0, (currTime - this._startTime));
-        const timeToCall = Math.max(0, this._frameTime - elapseTime);
         if (EDITOR || this._rAF === undefined || globalThis.__globalXR?.isWebXR) {
+            const currTime = performance.now();
+            const elapseTime = Math.max(0, currTime - this._startTime);
+            const timeToCall = Math.max(0, this._frameTime - elapseTime);
             return setTimeout(callback, timeToCall);
         }
-        this._start = currTime;
-        this._delay = timeToCall;
         this._callback = callback;
         return this._rAF.call(window, this._handleRAF);
     }


pal/pacer/pacer-web.ts
```diff
@@ -25,20 +25,22 @@
 import { EDITOR } from 'internal:constants';
 import { assertIsTrue } from '../../cocos/core/data/utils/asserts';
 
+const FRAME_RESET_TIME = 2000;
```
minggo, line 28: What does this variable mean?

pal/pacer/pacer-web.ts
```diff
@@ -96,32 +97,41 @@ export class Pacer {
         this._stHandle = this._stTime(updateCallback);
 
         this._isPlaying = true;
+        this._frameCount = 0;
```
minggo, line 100: What's frame count?

pal/pacer/pacer-web.ts
```diff
@@ -96,32 +97,41 @@ export class Pacer {
...
             this._rAF.call(window, this._handleRAF);
-        } else if (this._callback) {
-            this._callback();
+        } else {
+            this._frameCount++;
+            if (this._callback) {
+                this._callback();
+            }
+        }
+        if (performance.now() - this._startTime > FRAME_RESET_TIME) {
```
minggo, line 122: Why need to reset?

novain1102: The previous method was to calculate the delay frame by frame and use raf to run each delay frame. For the problems caused, see issue: https://github.com/cocos/3d-tasks/issues/16538#issuecomment-1575960596. Now it is to accumulate the delay for a period of time, tick one frame when it is longer than `_frameTime`, otherwise just delay to next frame.
Because `_frameCount` cannot be accumulated infinitely, it needs to be reset at a certain event point.

pal/pacer/pacer-web.ts
novain1102: _frameCount is used as a counter to determine whether a tick is needed in current frame or delayed until the next frame to tick.

pal/pacer/pacer-web.ts
minggo: Got it. How you determine the value of `FRAME_RESET_TIME `? It seems a magic number. Have you tested with different fps setting?

pal/pacer/pacer-web.ts
novain1102: 2000 represents do reset each 2 seconds, that is, a reset is performed every 120 ticks on a 60hz screen. I'm not sure whether 2000 is the most suitable value. However, there is no problem with running, so it may be an appropriate value

pal/pacer/pacer-web.ts
novain1102: > 2000 represents do reset each 2 seconds, that is, a reset is performed every 120 ticks on a 60hz screen. I'm not sure whether 2000 is the most suitable value. However, there is no problem with running, so it may be an appropriate value

In my opinion, as long as the value is not too small (does not cause calculation errors), nor too large (does not cause overflow).

> Have you tested with different fps setting?

Yes. I've different fps on mac and windows. This value seems run well.

pal/pacer/pacer-web.ts
```diff
@@ -96,32 +97,41 @@ export class Pacer {
...
+        const elapseFrame = Math.floor(elapseTime / this._frameTime);
+        if (elapseFrame < this._frameCount) {
             this._rAF.call(window, this._handleRAF);
-        } else if (this._callback) {
-            this._callback();
+        } else {
+            this._frameCount++;
+            if (this._callback) {
+                this._callback();
+            }
```
minggo, line 120: If the game logic is very complex, then `this._frameCount < elapseFrame ` is always true. In this situation, should invoke `this._callbak()` more times?

May be you can run a test case that is very complex, and set target frame to 60 to have a try.

novain1102: > If the game logic is very complex, then this._frameCount < elapseFrame  is always true. In this situation, should invoke this._callbak() more times?

As far as  this._frameCount < elapseFrame  is always true, just invoke `this._callback()` this frame and wait next frame to do next tick, which is as expected. 

The execution frame rate will always be the same as the set frame rate (if the default is 60, it will be executed as 60). Because this way doesn't mean to make up all the frame before the delay, it just to determine whether the current frame is to be ticked or skipped. So do not need to invoke this._callbak() more times. If `large game logic` happened, I think it will return to normal in subsequent cycles over time.

minggo (approved):

pal/pacer/pacer-web.ts
```diff
@@ -96,32 +97,41 @@ export class Pacer {
...
     };
 
     private _stTime (callback: () => void) {
-        const currTime = performance.now();
-        const elapseTime = Math.max(0, (currTime - this._startTime));
-        const timeToCall = Math.max(0, this._frameTime - elapseTime);
         if (EDITOR || this._rAF === undefined || globalThis.__globalXR?.isWebXR) {
+            const currTime = performance.now();
+            const elapseTime = Math.max(0, currTime - this._startTime);
+            const timeToCall = Math.max(0, this._frameTime - elapseTime);
```
pandamicro, line 132: Here, the timeToCall is not correct, that's also the reason why previous code doesn't work.

`const timeToCall = Math.max(0, this._frameTime - (elapseTime - this._frameTime));`

`elapseTime - this._frameTime` means the remaining elapsed time after the current tick.
`this._frameTime - (elapseTime - this._frameTime)` means how much time still need to be wait for the next tick.

pal/pacer/pacer-web.ts
```diff
@@ -96,32 +97,41 @@ export class Pacer {
...
             return setTimeout(callback, timeToCall);
```
pandamicro, line 133: As we are using the previous code only for setTimeout branch, I think it's safe to use directly `this._frameTime` here to setTimeout, no need to do any timeToCall calculations

pandamicro (approved):

pal/pacer/pacer-web.ts
```diff
@@ -96,32 +97,41 @@ export class Pacer {
...
         this._stHandle = 0;
         this._isPlaying = false;
+        this._frameCount = 0;
     }
 
     _handleRAF = () => {
-        if (performance.now() - this._start < this._delay) {
+        const elapseTime = performance.now() - this._startTime;
+        const elapseFrame = Math.floor(elapseTime / this._frameTime);
+        if (elapseFrame < this._frameCount) {
```
jareguo, line 114: Here the elapseFrame is only used to determine if the number of frames has changed. Then why not just
if (this._frameTime * this._frameCount > elapseTime) {

novain1102: It is also correct. Diffenert code means different perspective of understanding.

PR #15392, opened 2023-06-08
fix V3.8.0 performance unexpected downgrade

SantyWang:
Re: #

### Changelog

* This member does not exist at runtime, but because the attribute is registered, the member will be copied when instantiated, and the copying operation will cause the jit of the type to be corrupted for unknown reasons, resulting in a significant performance degradation

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


cocos/core/data/object.ts
```diff
@@ -452,9 +452,12 @@ if (EDITOR) {
  * issue: https://github.com/cocos/cocos-engine/issues/14643
  */
 (prototype as any)._deserialize = null;
-
-CCClass.fastDefine('cc.Object', CCObject, { _name: '', _objFlags: 0, [editorExtrasTag]: {} });
-CCClass.Attr.setClassAttr(CCObject, editorExtrasTag, 'editorOnly', true);
+if (EDITOR_NOT_IN_PREVIEW) {
```
jareguo: It is possible to pause editing while previewing, so all data should not be previewed, just use the EDITOR.
EDITOR_NOT_IN_PREVIEW should be used logically and can be executed differently depending on the environment

SantyWang: OK, I was using `EDITOR` before, I think the performance of gameview might be affected too.

SantyWang: > so all data should not be previewed

Not so serious, editorExtras is currently used mainly in the animation module editor interaction

pandamicro (approved):

jareguo (approved):

jareguo (approved):

PR #15426, opened 2023-06-13
[AOT] fix type in debug.ts

PPpro:

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/2d/components/ui-mesh-renderer.ts
+++ b/cocos/2d/components/ui-mesh-renderer.ts
@@ -268,7 +268,7 @@ export class UIMeshRenderer extends Component {
      */
     get renderEntity (): RenderEntity {
         if (DEBUG) {
-            assert(this._renderEntity, 'this._renderEntity should not be invalid');
+            assert(!!this._renderEntity, 'this._renderEntity should not be invalid');
         }
         return this._renderEntity;
     }
--- a/cocos/2d/components/ui-mesh-renderer.ts
+++ b/cocos/2d/components/ui-mesh-renderer.ts
@@ -268,7 +268,7 @@ export class UIMeshRenderer extends Component {
      */
     get renderEntity (): RenderEntity {
         if (DEBUG) {
-            assert(!!this._renderEntity, 'this._renderEntity should not be invalid');
+            assert(Boolean(this._renderEntity), 'this._renderEntity should not be invalid');
         }
         return this._renderEntity;
     }
--- a/cocos/core/platform/debug.ts
+++ b/cocos/core/platform/debug.ts
@@ -22,7 +22,6 @@
  THE SOFTWARE.
 */
 
-/* eslint-disable no-console */
 import { EDITOR, JSB, DEV, DEBUG } from 'internal:constants';
 import debugInfos from '../../../DebugInfos';
 import { legacyCC, ccwindow } from '../global-exports';
@@ -30,6 +29,8 @@ import { legacyCC, ccwindow } from '../global-exports';
 const ccdocument = ccwindow.document;
 
 const ERROR_MAP_URL = 'https://github.com/cocos-creator/engine/blob/develop/EngineErrorMap.md';
+
+export type StringSubstitution = number | string;
 
 // The html element displays log in web page (DebugMode.INFO_FOR_WEB_PAGE)
 let logList: HTMLTextAreaElement | null = null;
@@ -40,7 +41,7 @@ let ccWarn = ccLog;
 
 let ccError = ccLog;
 
-let ccAssert = (condition: any, message?: any, ...optionalParams: any[]): void => {
+let ccAssert = (condition: boolean, message?: string, ...optionalParams: StringSubstitution[]): void => {
     if (!condition) {
         console.log(`ASSERT: ${formatString(message, ...optionalParams)}`);
     }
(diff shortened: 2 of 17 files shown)


cocos/core/platform/debug.ts
```diff
@@ -49,80 +50,68 @@ let ccAssert = (condition: any, message?: any, ...optionalParams: any[]): void =
...
- * @zh 输出一条消息到 Cocos Creator 编辑器的 Console 或运行时 Web 端的 Console 中。这为你提供了对输出格式的额外控制。
- * @param message @zh 包含零个或多个需要替换的JavaScript字符串。@en JavaScript objects to replace substitution strings in msg.
- * @param optionalParams  @zh 用来替换在message中需要替换的JavaScript对象。@en JavaScript objects with which to replace substitution strings within msg.
+ * @en Outputs a log message to the console. The message may be a single string (with optional substitution values), or it may be any one or more JavaScript objects.
+ * @zh 向控制台输出一条日志信息。这条信息可能是单个字符串（包括可选的替代字符串），也可能是一个或多个对象。
  */
-export function log (message?: any, ...optionalParams: any[]): void {
-    return ccLog(message, ...optionalParams);
+export function log (...data: unknown[]): void {
+    return ccLog(...data);
```
PPpro, line 64: Our implementation is aligned to web standards, so the `log` method should be defined as `function log (...data: unknown[]): void`, the same goes for other methods like `warn error debug`

cocos/core/utils/js-typed.ts
```diff
@@ -348,7 +348,21 @@ const REGEXP_STR = /%s/;
...
+ * ```
+ * import { js } from 'cc';
+ * js.formatStr({}, null, undefined);  // [object Object] null undefined
+ * ```
+ */
+export function formatStr (...data: unknown[]): string;
+export function formatStr (msg: unknown, ...subst: unknown[]): string {
     if (arguments.length === 0) {
         return '';
     }
```
PPpro, line 368: we should support another formatStr signature, see this use case in unit test:
https://github.com/cocos/cocos-engine/blob/2a5b5711c1692c235bf05275b73cabacc9b15503/tests/core/js.test.ts#L41

cocos/2d/components/ui-mesh-renderer.ts
```diff
@@ -268,7 +268,7 @@ export class UIMeshRenderer extends Component {
      */
     get renderEntity (): RenderEntity {
         if (DEBUG) {
-            assert(this._renderEntity, 'this._renderEntity should not be invalid');
+            assert(!!this._renderEntity, 'this._renderEntity should not be invalid');
```
minggo: Why use `!!` as it is not clear to be understood.

PPpro: it means we convert this object to a boolean value

minggo: It is better to make it more clear, for example: `this._renderEntity != null`

cocos/core/utils/js-typed.ts
minggo: I don't quite understand this change. What problem does it fix?

cocos/core/platform/debug.ts
minggo: Will it break compatibility?

cocos/core/utils/js-typed.ts
PPpro: Before, we only supported `formatStr (msg: string, ...subst: StringSubstitution[]): string;`, in reality, the `log` method can pass in any js object, not just a string, such as the following :

```ts
log(null)
log(undefined)
log({}, 1, 2, 3)
```

In this situation, `function formatStr (msg: string, ...subst: StringSubstitution[]): string` can no longer meet the demand, we need a `function formatStr (...data: unknown[]): string` type declaration to support constructing any js object sequence into a string, in fact, the implementation of `formatStr` itself can support this situation, it just lacks a type declaration

PPpro: In addition, the definition `formatStr (msg: string, ...subst: StringSubstitution[]): string` does not satisfy the usage of this test case:
https://github.com/cocos/cocos-engine/blob/2a5b5711c1692c235bf05275b73cabacc9b15503/tests/core/js.test.ts#L41

cocos/2d/components/ui-mesh-renderer.ts
PPpro: We may also need to judge the situation of `undefined`, which will be more verbose.
I'll consider to use `Boolean(this._renderEntity)`

cocos/core/utils/js-typed.ts
PPpro: With the declaration before we can only support the following use cases:
```ts
log('hello world')
// or
log('hello %s', 'world')
```

cocos/core/platform/debug.ts
PPpro: no, it won't, any variable can be assigned to a variable of type `unknown`

and `unknown` is safer than `any`

cocos/2d/components/ui-mesh-renderer.ts
PPpro: updated https://github.com/cocos/cocos-engine/pull/15426/commits/52ab100e94922a98c6d05dab24257f1faedea506

minggo (approved):

PR #15434, opened 2023-06-14
Repair the problem of invalid game.end on web platform and minigame platform.

qiuguohua:
Re: #

### Changelog

https://github.com/cocos/3d-tasks/issues/16294

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/pal/system-info/minigame/system-info.ts
+++ b/pal/system-info/minigame/system-info.ts
@@ -243,8 +243,15 @@ class SystemInfo extends EventTarget {
         }
     }
 
-    public close (): void {
+    public exit () {
         minigame.exitMiniProgram?.();
+    }
+
+    public close () {
+        // TODO(qgh):The minigame platform does not have an exit interface,
+        // so there is no need to send a close message to the engine to release resources.
+        // this.emit('close');
+        this.exit();
     }
 }
 
--- a/pal/system-info/minigame/system-info.ts
+++ b/pal/system-info/minigame/system-info.ts
@@ -243,11 +243,11 @@ class SystemInfo extends EventTarget {
         }
     }
 
-    public exit () {
+    public exit (): void {
         minigame.exitMiniProgram?.();
     }
 
-    public close () {
+    public close (): void {
         // TODO(qgh):The minigame platform does not have an exit interface,
         // so there is no need to send a close message to the engine to release resources.
         // this.emit('close');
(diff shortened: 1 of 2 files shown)


pal/system-info/minigame/system-info.ts
```diff
@@ -243,9 +243,16 @@ class SystemInfo extends EventTarget {
         }
     }
 
-    public close (): void {
+    public exit (): void {
         minigame.exitMiniProgram?.();
     }
```
PPpro, line 248: please don't add new public interface for `pal/system-info` module

what's the difference between close and exit, can they merge into one interface?

minggo: Yep, it is confused.

qiuguohua: It cannot be divided into two interfaces.
Both interfaces should be public.

There are several processes for destruction:
1. Triggering the start of destruction, which is the role of 'close';
2, Ts engine to destroy resources, that is, the need to send the message close;
3, After the engine is destroyed, the active call to exit the program.

minggo: I think only engine internally need to care about these processes. From the view of developer, just need one interface. Is there any situation that developer should care about these processes?

qiuguohua: Yes, it is an internal engine interface. Is 'systemInfo' not exported to the user?

minggo: @PPpro please check if `systemInfo` is exported to developers.

PPpro: no, we only export `sys` to developers, which is a wrapper of `systemInfo`

PPpro: Is it more meaningful to name `close()` as `clearMemory()` ?

and the message `'close'` should be `'memory-clear'` ?

qiuguohua: It doesn't feel right because it's a process that closes.
We can change 'exit' to 'exitInternal' or 'closeInternal'？

minggo: Can merge them into on interface and do close and exit in this interface?

qiuguohua: We can do this, but there will be multiple implementations for different platforms

minggo: Yep, can do it in another PR.

minggo (approved):

PPpro (approved):

PR #15444, opened 2023-06-14
[AOT] fix type in node-activator.ts && component-scheduler.ts

PPpro:

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/scene-graph/component-scheduler.ts
+++ b/cocos/scene-graph/component-scheduler.ts
@@ -548,9 +548,10 @@ export class ComponentScheduler {
 
 if (EDITOR) {
     ComponentScheduler.prototype.enableComp = function (comp, invoker): void {
-        if (legacyCC.GAME_VIEW || comp.constructor._executeInEditMode) {
+        // NOTE: _executeInEditMode is dynamically injected on Editor environment
+        if (legacyCC.GAME_VIEW || (comp.constructor as any)._executeInEditMode) {
             if (!(comp._objFlags & IsOnEnableCalled)) {
-                if (comp.onEnable) {
+                if (comp._internal_onEnable) {
                     if (invoker) {
                         invoker.add(comp);
                         enableInEditor(comp);
@@ -558,7 +559,7 @@ if (EDITOR) {
                     } else {
                         callOnEnableInTryCatch(comp);
 
-                        const deactivatedDuringOnEnable = !comp.node._activeInHierarchy;
+                        const deactivatedDuringOnEnable = !comp.node.activeInHierarchy;
                         if (deactivatedDuringOnEnable) {
                             return;
                         }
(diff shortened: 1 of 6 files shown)


cocos/scene-graph/component-scheduler.ts
```diff
@@ -84,15 +85,35 @@ function stableRemoveInactive (iterator, flagToClear): void {
...
+type InvokeFunc = (...args: unknown[]) => void;
+
 // This class contains some queues used to invoke life-cycle methods by script execution order
 export class LifeCycleInvoker {
     public static stableRemoveInactive = stableRemoveInactive;
 
+    /**
+     * @engineInternal please access `_zero` instead.
+     */
+    public get _internal_zero (): js.array.MutableForwardIterator<any> {
```
minggo: If suggest to assess property instead, than why add this function?

PPpro: because _zero is a protected property，and NodeActivator cannot access it

minggo: 1. The comment is confused.
2. The property name is not good, why just use `zero` for property name?

PPpro: here is the comment for zero neg pos property

https://github.com/cocos/cocos-engine/blob/2a5b5711c1692c235bf05275b73cabacc9b15503/cocos/scene-graph/component-scheduler.ts#L91C58-L102

I'll optimize the comment for `_internal_zero _internal_neg _internal_pos`

PPpro: updated here https://github.com/cocos/cocos-engine/pull/15444/commits/881b44dcd45ef8b510b7ae2e31a9d8d206b15356

minggo: `_internal_zero` is not a good name:
- we don't use `_` to connect words
- why not just use `zero`?

PPpro: though it's a public property, but I prefer to use `_` at the beginning of the name, so that even if the developers debug at runtime, they can see that this is a js-style private interface

can we just name it as `_internalZero` ?

PPpro: I've try updating here 
https://github.com/cocos/cocos-engine/pull/15444/commits/06f480c7be8f44e799d76cf2fe15500cf8b69934

minggo: Why can not just use `zero`? I think it is more suitable in script. As it is engine internal, so i don't think need `_` prefix.

PPpro: update https://github.com/cocos/cocos-engine/pull/15444/commits/1c0cd24d40c4d1928afedad34dfe7edc4aee9faa

minggo (approved):

Issue #14646, opened 2023-03-23
support pal/image


PPpro:
relative issue https://github.com/cocos-creator/3d-tasks/issues/4452 https://github.com/cocos/3d-tasks/issues/7519

We need an image platform abstraction layer to smooth out platform interface differences and provide it to the engine framework layer

also issue from https://github.com/cocos/cocos-engine/pull/14638, see this commit: https://github.com/cocos/cocos-engine/pull/14638/commits/c268b8982cb9975e95b5f1ef855a681eff82b059

Issue #15324, opened 2023-06-05
'doCleanupImageCache' requires a texture type variable to be passed in.


qiuguohua:
### Cocos Creator version

3.8.0

### System information

windows

### Issue description

'doCleanupImageCache' requires a texture type variable to be passed in, but currently 'SpritFrame' is passed in.
![image](https://github.com/cocos/cocos-engine/assets/5205637/5a2d1711-5eac-4f91-8a49-4e0d0767641b)

### Steps to reproduce

none

PR #15485, opened 2023-06-20
Encapsulating PAL/ImageData

qiuguohua:
Re: #
https://github.com/cocos/cocos-engine/issues/14646

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


qiuguohua:
@cocos-robot run test cases

qiuguohua:
@cocos-robot run test cases

cocos/asset/assets/simple-texture.ts
```diff
@@ -140,11 +141,11 @@ export class SimpleTexture extends TextureBase {
      * @param level @en Mipmap level to upload the image to. @zh 要上传的 mipmap 层级。
      * @param arrayIndex @en The array index. @zh 要上传的数组索引。
      */
-    public uploadData (source: HTMLCanvasElement | HTMLImageElement | ArrayBufferView | ImageBitmap, level = 0, arrayIndex = 0) {
+    public uploadData (source: HTMLCanvasElement | HTMLImageElement | ArrayBufferView | ImageBitmap | IMemoryImageSource, level = 0, arrayIndex = 0) {
```
minggo: As the purpose is not use `HTMLCanvasElement `, `HTMLImageElement `, `ArrayBufferView ` and `ImageBitmap ` in engine, pal/image should handle it internally.

pal/image/base-image-data.ts
```diff
@@ -0,0 +1,166 @@
...
+        return ArrayBuffer.isView((this._imageSource as IMemoryImageSource)._data);
+    }
+    public isHtmlElement () {
+        return this._imageSource instanceof HTMLElement;
+    }
+    public isImageBitmap (imageSource: any): imageSource is ImageBitmap {
+        return !!(sys.hasFeature(sys.Feature.IMAGE_BITMAP) && imageSource instanceof ImageBitmap);
+    }
+
+    // 返回该图像源是否是平台提供的图像对象。
```
minggo: Please don't use chinese.

pal/image/base-image-data.ts
```diff
@@ -0,0 +1,166 @@
...
+        return '_data' in imageSource ? imageSource._data : imageSource;
+    }
+
+    public addEventListener (name, cb) {
+        if (this.isHtmlElement()) {
+            (this.data as HTMLImageElement).addEventListener(name, cb);
+        }
+    }
+
+    public removeEventListener (name, cb) {
```
minggo: Should add return type for every function.

cocos/asset/assets/simple-texture.ts
minggo: I found there is ImageSource in pal/image/types, why not use it?

pal/image/minigame/image-data.ts
```diff
@@ -0,0 +1,70 @@
...
+        return super.isNativeImage(imageSource);
+    }
+
+    set src (url: string) {
+        (this._imageSource as HTMLImageElement).src = url;
+    }
+
+    static downloadImage (url: string,
+        options: Record<string, any>,
+        onComplete: ((err: Error | null, data?: HTMLImageElement | null) => void)): HTMLImageElement {
```
minggo: If return `HTMLImageElement `, then the invoker will use `HTMLImageElement`. And we want to avoid using it in engine codes except pal.

cocos/asset/assets/simple-texture.ts
qiuguohua: This is an exported interface.' ImageSource' is not exported to the developer.

pal/image/base-image-data.ts
qiuguohua: OK

pal/image/minigame/image-data.ts
qiuguohua: This is designed for compatibility.

@types/pal/image.d.ts
```diff
@@ -0,0 +1,30 @@
...
+    type IMemoryImageSource = import('pal/image/types').IMemoryImageSource;
+    export class ImageData {
+        constructor (imageAsset?: ImageSource | ArrayBufferView);
+        //constructor (content: string, manifestRoot: string);
+        destroy (): void;
+        get src (): string;
+        set src(url: string);
+        set data(value: any);
+        get data(): ImageSource | ArrayBufferView;
+        set width(value: any);
```
PPpro:
```suggestion
        set width(value: number);
```

@types/pal/image.d.ts
```diff
@@ -0,0 +1,30 @@
+declare module 'pal/image' {
+    type ImageSource = import('pal/image/types').ImageSource;
+    type IMemoryImageSource = import('pal/image/types').IMemoryImageSource;
+    export class ImageData {
+        constructor (imageAsset?: ImageSource | ArrayBufferView);
+        //constructor (content: string, manifestRoot: string);
+        destroy (): void;
+        get src (): string;
+        set src(url: string);
+        set data(value: any);
```
PPpro:
```suggestion
        set data(value: ImageSource | ArrayBufferView);
```

@types/pal/image.d.ts
```diff
@@ -0,0 +1,30 @@
...
+        get data(): ImageSource | ArrayBufferView;
+        set width(value: any);
+        get width(): number;
+        get height(): number;
+        set height(value: any);
```
PPpro:
```suggestion
        set height(value: number);
```

@types/pal/image.d.ts
```diff
@@ -0,0 +1,30 @@
...
+        get format(): number;
+        get compressed(): number;
+        get mipmapLevelDataSize(): number[] | undefined;
+        static downloadImage (
+            url: string,
+            options: Record<string, any>,
+            onComplete: ((err: Error | null, data?: HTMLImageElement | null) => void),
+        ): HTMLImageElement;
+        onload: ((ev: Event) => any) | null;
+        onerror: ((ev: Event) => any) | null;
```
PPpro:
```suggestion
        onload: ((ev: Event) => void) | null;
        onerror: ((ev: Event) => void) | null;
```

@types/pal/image.d.ts
```diff
@@ -0,0 +1,30 @@
+declare module 'pal/image' {
```
PPpro, line 1: it's better we add more comment on every 'pal/image' public interface
it's OK to only have en version of comment

the comment sometimes is necessary like what's the difference between `data` and `nativeData`
what is `format`, what number it may returns, what each number represents, is it better to be an enum ?
when to use static method `downloadImage`, why not to set `src` directly, or why not use `reset` instead of setting `src`

`image.d.ts` is a design about `pal/image`, some usage and behavior of interfaces can be fixed in the form of documents to avoid abuse.

@types/pal/image.d.ts
```diff
@@ -0,0 +1,30 @@
...
+        get src (): string;
+        set src(url: string);
+        set data(value: any);
+        get data(): ImageSource | ArrayBufferView;
+        set width(value: any);
+        get width(): number;
+        get height(): number;
+        set height(value: any);
+        get format(): number;
+        get compressed(): number;
```
PPpro:
```suggestion
        get compressed(): boolean;
```

this seems should return a boolean

cocos/asset/assets/simple-texture.ts
minggo: But how to fix the issue that, engine internally don't use web standard class?

pal/image/minigame/image-data.ts
minggo: Does it exported to developers? If not, then just modify engine codes.

@types/pal/image.d.ts
```diff
@@ -0,0 +1,30 @@
...
+        destroy (): void;
+        get src (): string;
+        set src(url: string);
+        set data(value: any);
+        get data(): ImageSource | ArrayBufferView;
+        set width(value: any);
+        get width(): number;
+        get height(): number;
+        set height(value: any);
+        get format(): number;
```
PPpro: should this return an enum ?

cocos/2d/assets/sprite-frame.ts
```diff
@@ -27,10 +27,11 @@
 
 import { ccclass } from 'cc.decorator';
 import { EDITOR, TEST, BUILD } from 'internal:constants';
+import { ImageSource } from 'pal/image/types';
```
PPpro: please use relative path, we don't have module called 'pal/image/types'

this is an absolute path, which is not encouraged

cocos/3d/skinned-mesh-renderer/skinned-mesh-batch-renderer.ts
```diff
@@ -469,7 +469,7 @@ export class SkinnedMeshBatchRenderer extends SkinnedMeshRenderer {
                 region.texExtent.height = unit.size.y * this.atlasSize;
                 const { data } = partial.image;
                 if (!ArrayBuffer.isView(data)) {
-                    texImages.push(data);
+                    texImages.push(data as TexImageSource);
```
PPpro: does `pal/image` module provides a type called `TexImageSource` ? 

It's currently defined in lib.dom.d.ts

cocos/asset/assets/simple-texture.ts
```diff
@@ -157,18 +158,18 @@ export class SimpleTexture extends TextureBase {
...
             }
         }
 
-        if (ArrayBuffer.isView(source)) {
-            gfxDevice.copyBuffersToTexture([source], this._gfxTexture, _regions);
+        if (ArrayBuffer.isView(imageData.data)) {
+            gfxDevice.copyBuffersToTexture([imageData.data], this._gfxTexture, _regions);
         } else {
-            gfxDevice.copyTexImagesToTexture([source], this._gfxTexture, _regions);
+            gfxDevice.copyTexImagesToTexture([imageData.data as TexImageSource], this._gfxTexture, _regions);
```
PPpro: the same above

cocos/game/splash-screen.ts
```diff
@@ -340,7 +341,7 @@ export class SplashScreen {
         region.texExtent.width = this.bgImage.width;
         region.texExtent.height = this.bgImage.height;
         region.texExtent.depth = 1;
-        device.copyTexImagesToTexture([this.bgImage], this.bgTexture, [region]);
+        device.copyTexImagesToTexture([this.bgImage.data as TexImageSource], this.bgTexture, [region]);
```
PPpro: the same above

cocos/game/splash-screen.ts
```diff
@@ -375,7 +376,7 @@ export class SplashScreen {
         region.texExtent.width = this.logoImage.width;
         region.texExtent.height = this.logoImage.height;
         region.texExtent.depth = 1;
-        device.copyTexImagesToTexture([this.logoImage], this.logoTexture, [region]);
+        device.copyTexImagesToTexture([this.logoImage.data as TexImageSource], this.logoTexture, [region]);
```
PPpro: the same above

@types/pal/image.d.ts
```diff
@@ -0,0 +1,30 @@
...
+        get compressed(): number;
+        get mipmapLevelDataSize(): number[] | undefined;
+        static downloadImage (
```
PPpro: `download` is a concept on web platforms, on many other platforms, we call it `load` instead.

could this be named as `loadImage` ?

@types/pal/image.d.ts
```diff
@@ -0,0 +1,30 @@
+declare module 'pal/image' {
+    type ImageSource = import('pal/image/types').ImageSource;
+    type IMemoryImageSource = import('pal/image/types').IMemoryImageSource;
```
PPpro:
```suggestion
    export type ImageSource = import('pal/image/types').ImageSource;
    export type IMemoryImageSource = import('pal/image/types').IMemoryImageSource;
```

cocos/asset/assets/image-asset.jsb.ts
```diff
@@ -24,78 +24,39 @@
...
     enumerable: true,
     get () {
-        return this._nativeData;
+        return this._imageData;
     },
-    set (value: ImageSource) {
-        if (!(value instanceof jsbWindow.HTMLElement) && !isImageBitmap(value)) {
-            (value as IMemoryImageSource).format = (value as IMemoryImageSource).format || this.format;
-        }
+    set (value: any) {
```
PPpro: It's better not to use any type

@types/pal/image.d.ts
```diff
@@ -0,0 +1,30 @@
...
+        get compressed(): number;
+        get mipmapLevelDataSize(): number[] | undefined;
+        static downloadImage (
+            url: string,
+            options: Record<string, any>,
+            onComplete: ((err: Error | null, data?: HTMLImageElement | null) => void),
+        ): HTMLImageElement;
+        onload: ((ev: Event) => any) | null;
+        onerror: ((ev: Event) => any) | null;
+        nativeData (): unknown;
```
PPpro: can this be a getter like `get data()` `get nativeData ()`

cocos/tiledmap/tiled-map.ts
```diff
@@ -598,19 +598,15 @@ export class TiledMap extends Component {
...
-        if (texture._image instanceof HTMLImageElement) {
-            texture._image.src = '';
-            if (JSB) texture._image.destroy();
-        } else if (sys.hasFeature(sys.Feature.IMAGE_BITMAP) && texture._image instanceof ImageBitmap) {
-            if (texture._image.close) texture._image.close();
-        }
-        texture._image = null;
+        // TODO:There is a bug here, if you uncomment, it will cause the display result to be incorrect.
+        // texture.image.destroy();
+        // texture.image = null;
```
PPpro: does it have any tracking issue ?

pal/image/base-image-data.ts
```diff
@@ -0,0 +1,166 @@
...
+        return this._imageSource && this._imageSource._data;
+    }
+
+    public nativeData (): unknown {
+        return this.data as any;
+    }
+
+    set crossOrigin (string) {
+        (this._imageSource as HTMLImageElement).crossOrigin = 'anonymous';
+    }
```
PPpro:
```suggestion
    public acceptCORS (): void {
        (this._imageSource as HTMLImageElement).crossOrigin = 'anonymous';
    }
```

is this a public interface ?
I don't see it declared in pal/image module

this interface seems to solve the CORS issues, is it better to delcare it as `public acceptCORS(): void`
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin

pal/image/base-image-data.ts
```diff
@@ -0,0 +1,166 @@
...
+    get src (): string {
+        return (this._imageSource as HTMLImageElement).src;
+    }
+
+    get width () {
+        return this._imageSource.width;
+    }
+
+    set width (value) {
+        // @ts-ignore
```
PPpro: please don't use `ts-ignore`

pal/image/base-image-data.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    get height (): number {
+        if (!(this._imageSource instanceof ArrayBuffer)) {
+            return this._imageSource.height;
+        }
+        return 0;
+    }
+
+    set height (value) {
+        // @ts-ignore
```
PPpro: the same above

pal/image/base-image-data.ts
PPpro: are they private interfaces? because they are not declared in `pal/image` module

@types/pal/image.d.ts
```diff
@@ -0,0 +1,30 @@
...
+        get height(): number;
+        set height(value: any);
+        get format(): number;
+        get compressed(): number;
+        get mipmapLevelDataSize(): number[] | undefined;
+        static downloadImage (
+            url: string,
+            options: Record<string, any>,
+            onComplete: ((err: Error | null, data?: HTMLImageElement | null) => void),
+        ): HTMLImageElement;
```
PPpro:
```suggestion
            onComplete: ((err: Error | null, data?: ImageData | null) => void),
        ): ImageData ;
```

we should not design the interface to rely on web type, even though we don't publish `pal/image` module to developpers

pal/image/minigame/image-data.ts
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        function errorCallback () {
+            image.removeEventListener('load', loadCallback);
+            image.removeEventListener('error', errorCallback);
+            if (onComplete) { onComplete(new Error(getError(4930, url))); }
+        }
+
+        image.addEventListener('load', loadCallback);
+        image.addEventListener('error', errorCallback);
```
PPpro: should we use `onload` and `onerror` ? which is public interface delcared in pal/image module

pal/image/native/image-data.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    static downloadImage (url: string,
+        options: Record<string, any>,
+        onComplete: ((err: Error | null, data?: HTMLImageElement | null) => void)): HTMLImageElement {
+        const image = new ImageData();
+
+        // NOTE: on xiaomi platform, we need to force setting img.crossOrigin as 'anonymous'
+        if (ccwindow.location.protocol !== 'file:') {
+            image.crossOrigin = 'anonymous';
```
PPpro:
```suggestion
            image.acceptCORS();
```

change like this as mentioned above

pal/image/native/image-data.ts
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        function errorCallback () {
+            image.removeEventListener('load', loadCallback);
+            image.removeEventListener('error', errorCallback);
+            if (onComplete) { onComplete(new Error(getError(4930, url))); }
+        }
+
+        image.addEventListener('load', loadCallback);
+        image.addEventListener('error', errorCallback);
```
PPpro: we should use `onload` `onerror`

cocos/asset/assets/simple-texture.ts
qiuguohua: The engine internal is to remove the dependencies, the export interface still needs to be retained.

pal/image/minigame/image-data.ts
qiuguohua: Yep, this is an export interface.
image

@types/pal/image.d.ts
qiuguohua: Yep，need to use enum.

cocos/2d/assets/sprite-frame.ts
qiuguohua: This is an error, it should be referencing 'pal/image'.

cocos/3d/skinned-mesh-renderer/skinned-mesh-batch-renderer.ts
qiuguohua: 'SourceImage' is partially compatible with 'TexImageSource'. If you don't add 'as' it will report an error.

@types/pal/image.d.ts
qiuguohua: DownloadImg is implemented for every platform. It is downloading images from the server and DownloadImg for other platforms.

LoadImg can be implemented by defining an imagedata passed to null.

@types/pal/image.d.ts
qiuguohua: Here I was thinking of using pal/image or using relative paths to reference.

@types/pal/image.d.ts
qiuguohua: If we use get, an error will be reported and the type of unknow will be returned.

cocos/tiledmap/tiled-map.ts
qiuguohua: Yep，https://github.com/cocos/cocos-engine/issues/15324

pal/image/base-image-data.ts
qiuguohua: Will report an error because width is a readable property.

@types/pal/image.d.ts
qiuguohua: If this is the case, will it break compatibility, as this interface may be exported for developer

cocos/asset/assets/simple-texture.ts
minggo: Why not export ImageSource? It will keep compatibility and remove the usage of web classes.

pal/image/minigame/image-data.ts
minggo: I can not search it in https://docs.cocos.com/creator/api/zh/result?keyword=downloadImage.

qiuguohua:
@cocos-robot run test cases

cocos/asset/asset-manager/parser.ts
```diff
@@ -80,14 +82,14 @@ export class Parser {
...
      */
-    public parseImage (file: HTMLImageElement | Blob, options: Record<string, any>, onComplete: ((err: Error | null, data?: HTMLImageElement | ImageBitmap | null) => void)): void {
-        if (file instanceof HTMLImageElement) {
-            onComplete(null, file);
+    public parseImage (file: ImageData | Blob, options: Record<string, any>, onComplete: ((err: Error | null, data?: ImageSource | ArrayBufferView | null) => void)): void {
+        if (file instanceof ImageData) {
+            onComplete(null, file.data);
             return;
         }
-        createImageBitmap(file, { premultiplyAlpha: 'none' }).then((result): void => {
```
minggo, line 88: Why remove return type?

cocos/asset/asset-manager/parser.ts
```diff
@@ -80,14 +82,14 @@ export class Parser {
...
+        createImageBitmap(file, { premultiplyAlpha: 'none' }).then((result) => {
             onComplete(null, result);
-        }, (err): void => {
```
minggo, line 90: The same as above.

cocos/asset/assets/image-asset.ts
```diff
@@ -523,43 +485,52 @@ export class ImageAsset extends Asset {
...
-     * @en Image data.
-     * @zh 此图像资源的图像数据。
+     * @en Image data source(include: HTMLCanvasElement | HTMLImageElement | IMemoryImageSource | ImageBitmap | ArrayBufferView).
+     * @zh 图像的来源（包括：HTMLCanvasElement | HTMLImageElement | IMemoryImageSource | ImageBitmap | ArrayBufferView）。
      */
-    get data (): ArrayBufferView | HTMLCanvasElement | HTMLImageElement | ImageBitmap | null {
-        if (isNativeImage(this._nativeData)) {
-            return this._nativeData;
-        }
+    get data (): any {
```
minggo: Why return any? We should avoid using any.

cocos/asset/assets/image-asset.ts
```diff
@@ -523,43 +485,52 @@ export class ImageAsset extends Asset {
...
-            value.format = value.format || this._format;
-        }
         this.reset(value);
     }
 
+    /*
+     * @en Raw image data.
+     * @zh 原始图像数据。
+     */
+    get rawData (): any {
```
minggo: What's the difference between `rawData` and `data`?

cocos/asset/assets/image-asset.ts
minggo: As it returns `ImageData`, and `data` is contained in `ImageData`. So is this API needed?

cocos/asset/assets/simple-texture.ts
```diff
@@ -140,11 +142,11 @@ export class SimpleTexture extends TextureBase {
      * @param level @en Mipmap level to upload the image to. @zh 要上传的 mipmap 层级。
      * @param arrayIndex @en The array index. @zh 要上传的数组索引。
      */
-    public uploadData (source: HTMLCanvasElement | HTMLImageElement | ArrayBufferView | ImageBitmap, level = 0, arrayIndex = 0): void {
+    public uploadData (source: HTMLCanvasElement | HTMLImageElement | ArrayBufferView | ImageBitmap | IMemoryImageSource, level = 0, arrayIndex = 0): void {
```
minggo: Can avoid the usage of `HTMLCanvasElement ` and other web standard API here?

cocos/gfx/empty/empty-device.ts
```diff
@@ -213,6 +214,7 @@ export class EmptyDevice extends Device {
     public copyBuffersToTexture (buffers: Readonly<ArrayBufferView[]>, texture: Texture, regions: Readonly<BufferTextureCopy[]>): void {}
     public copyTextureToBuffers (texture: Readonly<Texture>, buffers: ArrayBufferView[], regions: Readonly<BufferTextureCopy[]>): void {}
     public copyTexImagesToTexture (texImages: Readonly<TexImageSource[]>, texture: Texture, regions: Readonly<BufferTextureCopy[]>): void {}
+    public copyImagesToTexture (imageAssets: Readonly<ImageAsset[]>, texture: Texture, regions: Readonly<BufferTextureCopy[]>) {}
```
minggo: Please add return type.

cocos/gfx/webgl/webgl-device.ts
```diff
@@ -564,4 +565,21 @@ export class WebGLDevice extends Device {
             regions,
         );
     }
+
+    public copyImagesToTexture (
```
minggo: Please add return type.

cocos/asset/asset-manager/parser.ts
qiuguohua: This is a merge error.

cocos/asset/assets/image-asset.ts
qiuguohua: There are three data types:

‘rawData’: data containing the content of the image

'data': the source of the image, including HTMLImageElement, HTMLCanvasElement, etc. This may be better to use dataSource, but because this interface is exported, it may affect compatibility issues

'imageData': image data management object, similar to _nativeAsset, but modifying _nativeAsset may lead to compatibility problems, so the imageData interface is added

cocos/asset/assets/simple-texture.ts
qiuguohua: ImageSource is not exported and is also the internal interface of the engine. The interface for exporting PAL also needs to be considered

cocos/asset/assets/image-asset.ts
qiuguohua: It is needed here because in the web platform, imagesource(HTMLCanvasElement | HTMLImageElement) is used directly. 

In native platform is used raw image data.

cocos/asset/assets/simple-texture.ts
minggo: Why can not export ImageSource?

cocos/asset/assets/image-asset.ts
minggo: Can combine data and imageData? Only use one interface.

qiuguohua: You mean to get data through imageData？But the data interface is already an exported interface. Do I need to deprecate this interface?

cocos/asset/assets/simple-texture.ts
qiuguohua: It seems that the export interface can also rely on internal types. I will reconsider this

cocos/asset/assets/image-asset.ts
minggo: Can just use data?

@types/pal/image.d.ts
```diff
@@ -0,0 +1,87 @@
...
+        destroy (): void;
+
+        /**
+         * Get the url of the image(when source is HTMLImageElement).
+         */
+        get src (): string;
+        /**
+         * Set the url of the image(when source is HTMLImageElement).
+         */
+        set src(url: string);
```
PPpro:
```suggestion
        /**
         * Get and set the url of the image (when source is HTMLImageElement).
         */
        get src (): string;
        set src (url: string);
```

we only need one comment for getter/setter, the same with others

@types/pal/image.d.ts
```diff
@@ -0,0 +1,87 @@
...
+        onload: ((ev: Event) => void) | null;
+        /**
+         * Image load error callback(when the image is HTMLImageElement).
+         */
+        onerror: ((ev: Event) => void) | null;
+
+        /**
+         * Get raw image data, in web platform is image source(like data interface), in native platform is image raw data.
+         */
+        rawData (): unknown;
```
PPpro:
```suggestion
        getRawData (): unknown;
```

if it's not a getter but a method instead, I think it's better to name it as `getRawData() `

it's strange when we call `imageData.rawData()` but access data with `imageData.data`

qiuguohua:
@cocos-robot run test cases

@types/pal/image.d.ts
```diff
@@ -0,0 +1,87 @@
...
+        ): ImageData;
+
+        /**
+         * Image load success callback(when the image is HTMLImageElement).
+         */
+        onload: ((ev: Event) => void) | null;
+        /**
+         * Image load error callback(when the image is HTMLImageElement).
+         */
+        onerror: ((ev: Event) => void) | null;
```
PPpro: can we only use static `loadImage` method, and remove `onload` `onerror` ?

it seems we only need this mothod for engine framework layer

@types/pal/image.d.ts
PPpro: It's better to use pal/image, after engine modulized, we prefer to define type in `.d.ts` files

@types/pal/image.d.ts
PPpro: it's strange, why, what error it reports

pal/image/native/image-data.ts
```diff
@@ -0,0 +1,79 @@
...
+        }
+        super.destroy();
+    }
+    public rawData (): unknown {
+        if (this._imageSource instanceof HTMLCanvasElement) {
+            // @ts-ignore
+            return this._imageSource._data.data;
+        } else if (this._imageSource instanceof HTMLImageElement) {
+            // @ts-ignore
+            return this._imageSource._data;
```
PPpro: we don't recommend using ts-expect-error or ts-ignore, please follow the [codeing-style-guidance](https://github.com/cocos/3d-tasks/blob/master/workflows/coding-style-guidance.md#%E7%A6%81%E7%94%A8-ts-expect-error)

@types/pal/image.d.ts
```diff
@@ -0,0 +1,87 @@
...
+        get compressed(): boolean;
+        get mipmapLevelDataSize(): number[] | undefined;
+        /**
+         * Download images from the server.
+         */
+        static loadImage (
+            url: string,
+            options: Record<string, any>,
+            onComplete: ((err: Error | null, data?: ImageSource | null) => void),
+        ): ImageData;
```
PPpro:
```suggestion
        static loadImage (
            url: string,
            options: Record<string, any>,
        ): Promise<ImageData>;
```
I recommended to design loadImage as Promise based interface, then we could use like this

```ts
ImageData.loadImage('xxx.png', {}).then(imageData => {

}).catch(err => {

})
```

cocos/game/splash-screen.ts
```diff
@@ -137,26 +137,24 @@ export class SplashScreen {
...
-                this.logoImage.imageData.onload = (): void => {
+                ImageData.loadImage(this.settings.base64src).then((imageData) => {
+                    this.logoImage = new ImageAsset();
+                    this.logoImage._nativeAsset = imageData.data;
                     this.initLogo();
                     resolve();
-                };
-                this.logoImage.imageData.onerror = (): void => {
+                }).catch((err) => {
                     reject();
```
PPpro:
```suggestion
                    reject(err);
```

cocos/game/splash-screen.ts
```diff
@@ -137,26 +137,24 @@ export class SplashScreen {
...
-                this.bgImage.imageData.onload = (): void => {
+                ImageData.loadImage(this.settings.bgBase64).then((imageData) => {
+                    this.bgImage = new ImageAsset();
+                    this.bgImage._nativeAsset = imageData.data;
                     this.initBG();
                     resolve();
-                };
-                this.bgImage.imageData.onerror = (): void => {
+                }).catch((err) => {
                     reject();
```
PPpro:
```suggestion
                    reject(err);
```

@types/pal/image.d.ts
```diff
@@ -55,25 +55,12 @@ declare module 'pal/image' {
...
-
-        /**
-         * Image load success callback(when the image is HTMLImageElement).
-         */
-        onload: ((ev: Event) => void) | null;
-        /**
-         * Image load error callback(when the image is HTMLImageElement).
-         */
-        onerror: ((ev: Event) => void) | null;
+        static loadImage (url: string): Promise<ImageData>
```
PPpro:
```suggestion
        static loadImage (urlOrBase64: string): Promise<ImageData>
```

@types/pal/image.d.ts
qiuguohua: ok

pal/image/native/image-data.ts
qiuguohua: Okay, I've removed.

@types/pal/image.d.ts
qiuguohua: Okay, I've removed.

@types/pal/image.d.ts
qiuguohua: OK

cocos/gfx/webgl/webgl-device.ts
qiuguohua: ok

editor/assets/primitives.fbx.meta
```diff
@@ -11,15 +11,16 @@
       "displayName": "",
       "id": "17020",
       "name": "sphere.mesh",
-      "ver": "1.1.0",
```
minggo: Should change meta file?

cocos/asset/assets/image-asset.ts
qiuguohua: yeh，load images with loadImage removes the imagedata dependency.

editor/assets/primitives.fbx.meta
qiuguohua: No need, it should be an error in the merge, I have restored it yesterday. I'm checking why it didn't take effect.

@types/pal/image.d.ts
```diff
@@ -0,0 +1,64 @@
+declare module 'pal/image' {
+    export type ImageSource = import('pal/image/types').ImageSource;
+    export type IMemoryImageSource = import('pal/image/types').IMemoryImageSource;
+    export type PixelFormat = import('cocos/asset/assets/asset-enum').PixelFormat;
+    export class ImageData {
```
dumganhar, line 7: Why not call it `Image` directly? ImageData seems just raw data like https://developer.mozilla.org/en-US/docs/Web/API/ImageData .
`Image` includes `data` property is fine. But `ImageData` also contains `data` property which is confused in my opinion.

qiuguohua: image
They are defined in Dom.

There are currently two types of data, one is imagesource (HtmlImageElement) and the other is Raw image data, both of which currently need to exist.

The data attribute prefers to be compatible, and I understand that it should be named source would be better.

dumganhar: DOM's Image is in global namespace, but our encapsulated `Image` could be in engine's namespace, for example `cc`. Will the class names in different namespace cause `conflict`?

@types/pal/image.d.ts
```diff
@@ -0,0 +1,65 @@
...
+        set width(value: number);
+
+        /**
+         * Get image height.
+         */
+        get height(): number;
+        /**
+         * Set image height(when source is IMemoryImageSource).
+         */
+        set height(value: number);
```
PPpro: one comment for getter/setter

@types/pal/image.d.ts
```diff
@@ -0,0 +1,65 @@
...
+        get format(): PixelFormat;
+        /**
+         * Get if the image is compressed.
+         */
+        get compressed(): boolean;
+        get mipmapLevelDataSize(): number[] | undefined;
+        /**
+         * Load image via local url or web url.
+         */
+        static loadImage (url: string): Promise<ImageData>
```
PPpro:
```suggestion
        static loadImage (url: string, options: Record<string, unknown>): Promise<ImageData>
```
we should support load options, see this use case

https://github.com/cocos/cocos-engine/blob/2d67b2d6c7ed604c4937eefc5f697b762e205398/cocos/asset/asset-manager/download-dom-image.ts#L25-L35

@types/pal/image.d.ts
qiuguohua: It should not be combined here because the behavior of set is not the same

@types/pal/image.d.ts
qiuguohua: The options parameter is not implemented, and it feels unnecessary to pass in the options parameter.

@types/pal/image.d.ts
qiuguohua: OK

@types/pal/image.d.ts
PPpro: It doesn't make sense to have separate getter/setter annotations, because it will only end up being parsed as one annotation by the TS language service when we use `imageData.height`

cocos/asset/assets/simple-texture.ts
```diff
@@ -140,7 +142,7 @@ export class SimpleTexture extends TextureBase {
      * @param level @en Mipmap level to upload the image to. @zh 要上传的 mipmap 层级。
      * @param arrayIndex @en The array index. @zh 要上传的数组索引。
      */
-    public uploadData (source: HTMLCanvasElement | HTMLImageElement | ArrayBufferView | ImageBitmap, level = 0, arrayIndex = 0): void {
+    public uploadData (source: ImageSource | ArrayBufferView | ImageAsset, level = 0, arrayIndex = 0): void {
```
dumganhar: Could we deprecate `uploadData` here?
Add a new function which accesses `(imageAsset: ImageAsset, level = 0, arrayIndex = 0)`

qiuguohua: Yes, I've overloaded it.

cocos/asset/assets/image-asset.ts
```diff
@@ -193,8 +193,12 @@ export class ImageAsset extends Asset {
             let dataOffset = fileHeaderLength;
             for (let i = 0; i < files.length; i++) {
                 const file = files[i];
-                outView.setUint32(COMPRESSED_HEADER_LENGTH + COMPRESSED_MIPMAP_LEVEL_COUNT_LENGTH + i * COMPRESSED_MIPMAP_DATA_SIZE_LENGTH,
-                    file.byteLength, true); //add file data size
+                outView.setUint32(
+                    COMPRESSED_HEADER_LENGTH + COMPRESSED_MIPMAP_LEVEL_COUNT_LENGTH + i * COMPRESSED_MIPMAP_DATA_SIZE_LENGTH,
+                    file.byteLength,
+
```
dumganhar: Empty line ?

cocos/asset/assets/image-asset.ts
```diff
@@ -580,21 +602,46 @@ export class ImageAsset extends Asset {
...
+     * @deprecated please use imagedata structure.
+     */
+    constructor (nativeAsset: HTMLCanvasElement | HTMLImageElement | ImageBitmap);
+    /**
+     * @en Constructing an ImageSource object.
+     * @zh 构造ImageSource对象
+     * @param data @en The image source. @zh 图像数据源。
+     */
+    constructor (nativeAsset?: ImageData | IMemoryImageSource);
+    constructor (nativeAsset?: ImageData | IMemoryImageSource | HTMLCanvasElement | HTMLImageElement | ImageBitmap) {
```
dumganhar: The parameter name `nativeAsset` is wrong.

cocos/asset/assets/simple-texture.ts
```diff
@@ -141,8 +141,28 @@ export class SimpleTexture extends TextureBase {
      * @param source @en The source image or image data. @zh 源图像或图像数据。
      * @param level @en Mipmap level to upload the image to. @zh 要上传的 mipmap 层级。
      * @param arrayIndex @en The array index. @zh 要上传的数组索引。
+     * @deprecated since v3.9, please use `uploadData()` instead.
```
dumganhar: The comment is not clear. Both the function names are `uploadData`.

pal/env/web/env.ts
```diff
@@ -29,7 +29,7 @@ export function findCanvas (): { frame: HTMLDivElement, container: HTMLDivElemen
     const container = document.querySelector('#Cocos3dGameContainer') as HTMLDivElement;
     const canvas = document.querySelector('#GameCanvas') as HTMLCanvasElement;
 
-    return { frame, canvas, container };
+    return { frame, container, canvas };
```
dumganhar: Why change the order?

pal/image/base-image-data.ts
```diff
@@ -90,6 +92,49 @@ export class BaseImageData {
...
+            if ('buffer' in buff) {
+                // es-lint as any
+                data = new Uint8Array((buff as any).buffer, (buff as any).byteOffset, (buff as any).byteLength);
+            } else {
+                rawBuffer = buff;
+                data = new Uint8Array(rawBuffer);
+            }
+        } else {
+            // eslint-disable-next-line no-console
+            console.log('imageBmp copy not impled!');
```
dumganhar: impled -> implemented

```
console.warn(`getRawData doesn't support this source type.`);
```

qiuguohua: Yep, it would be better to use assert here.

cocos/asset/assets/simple-texture.ts
qiuguohua: OK

pal/env/web/env.ts
qiuguohua: This is an unrelated modification because the return values defined for this interface are in a different order.

It's been restored.

SantyWang (approved):

@types/jsb.d.ts
```diff
@@ -339,6 +339,10 @@ declare namespace jsb {
         setVerifyCallback (verifyCallback: (path: string, asset: ManifestAsset) => boolean): void;
         setEventCallback (eventCallback: (event: EventAssetsManager) => void): void;
     }
+
+    export class JSBNativeDataHolder {
+
+    }
```
minggo, line 345: Why export it? Does developers need to use it?

cocos/asset/assets/image-asset.ts
```diff
@@ -609,56 +610,58 @@ export class ImageAsset extends Asset {
...
-            this._format = data.format;
-        } else {
-            this._nativeData = data;
+    public reset (data: HTMLCanvasElement | HTMLImageElement | ImageBitmap): void;
+    /**
+     * @en Reset the source of the image asset.
+     * @zh 重置此图像资源使用的原始图像源。
+     * @param data @en The new source. @zh 新的图片数据源。
+     */
+    public reset (data: IMemoryImageSource): void;
```
minggo, line 658: Why need this declaration as it is part of `public reset (data: IMemoryImageSource | HTMLCanvasElement | HTMLImageElement | ImageBitmap)`.

@types/jsb.d.ts
qiuguohua: In order to declare the type, if the type is not declared, the return would need to be of type any.

cocos/asset/assets/image-asset.ts
qiuguohua: Because the dom related interface needs to be deprecated, but IMemoryImageSource needs to be kept.

cocos/game/splash-screen.ts
```diff
@@ -340,7 +340,8 @@ export class SplashScreen {
         region.texExtent.width = this.bgImage.width;
         region.texExtent.height = this.bgImage.height;
         region.texExtent.depth = 1;
-        device.copyTexImagesToTexture([this.bgImage], this.bgTexture, [region]);
+
+        device.copyImagesToTexture([this.bgImage], this.bgTexture, [region]);
```
minggo: Why change the code?

cocos/game/splash-screen.ts
```diff
@@ -375,7 +376,7 @@ export class SplashScreen {
         region.texExtent.width = this.logoImage.width;
         region.texExtent.height = this.logoImage.height;
         region.texExtent.depth = 1;
-        device.copyTexImagesToTexture([this.logoImage], this.logoTexture, [region]);
+        device.copyImagesToTexture([this.logoImage], this.logoTexture, [region]);
```
minggo: The same as above.

cocos/gfx/base/device.ts
```diff
@@ -335,6 +336,15 @@ export abstract class Device {
...
     public abstract copyTexImagesToTexture (texImages: Readonly<TexImageSource[]>, texture: Texture, regions: Readonly<BufferTextureCopy[]>): void;
 
+    /**
+     * @en Copy image assets to texture.
+     * @zh 拷贝图像资产到纹理。
+     * @param imageAssets The image assets to be copied.
+     * @param texture The texture to copy to.
+     * @param regions The region descriptions.
+     */
+    public abstract copyImagesToTexture (imageAssets: Readonly<ImageAsset[]>, texture: Texture, regions: Readonly<BufferTextureCopy[]>): void;
```
minggo: @bluesky013 please take a look.

@types/jsb.d.ts
minggo: But we don't want to export it to developers. So should find a way to export internally.

cocos/asset/assets/image-asset.ts
minggo: Got it.

cocos/game/splash-screen.ts
qiuguohua: The type of the parameter has changed.
`copyTexImagesToTexture` may need to be deprecated because of the dependency on the dom-related interface.

@types/jsb.d.ts
qiuguohua: The jsb namespace is deprecated and will be used internally as an engine.

cocos/gfx/base/device.ts
bluesky013: `ImageAsset` is a concept in the rendering layer, and gfx is below the rendering layer. It should not depend on `ImageAsset`. It may be necessary to encapsulate the data structure of gfx.

PPpro (approved):

qiuguohua:
@cocos-robot run test cases

minggo (approved):

bluesky013 (approved):

cocos/asset/assets/image-asset.jsb.ts
```diff
@@ -36,82 +38,54 @@ export type ImageAsset = JsbImageAsset;
...
-    }
-
-    return imageSource instanceof jsbWindow.HTMLImageElement || imageSource instanceof jsbWindow.HTMLCanvasElement || isImageBitmap(imageSource);
-}
-
 // TODO: we mark imageAssetProto as type of any, because here we have many dynamic injected property @dumganhar
 const imageAssetProto: any = ImageAsset.prototype;
 
-imageAssetProto._ctor = function (nativeAsset?: ImageSource) {
+imageAssetProto._ctor = function (nativeAsset?: ImageData | IMemoryImageSource | HTMLCanvasElement | HTMLImageElement | ImageBitmap) {
```
dumganhar: Need to rename the parameter name `nativeAsset`.

cocos/asset/assets/image-asset.jsb.ts
```diff
@@ -36,82 +38,54 @@ export type ImageAsset = JsbImageAsset;
...
 Object.defineProperty(imageAssetProto, '_nativeAsset', {
     configurable: true,
     enumerable: true,
     get () {
-        return this._nativeData;
+        return this._imageData.source;
     },
-    set (value: ImageSource) {
-        if (!(value instanceof jsbWindow.HTMLElement) && !isImageBitmap(value)) {
-            (value as IMemoryImageSource).format = (value as IMemoryImageSource).format || this.format;
```
dumganhar, line 97: Was this line removed ? value.format doesn't need to be set to `this.format` now?

cocos/asset/assets/image-asset.ts
```diff
@@ -609,56 +617,58 @@ export class ImageAsset extends Asset {
...
+        } else {
+            this._imageData = new ImageData(imageSource);
         }
     }
 
     /**
      * @en Reset the source of the image asset.
      * @zh 重置此图像资源使用的原始图像源。
      * @param data @en The new source. @zh 新的图片数据源。
+     * @deprecated @en please use `reset(data: IMemoryImageSource)` instead. @zh 请使用`reset(data: IMemoryImageSource)`替换
```
dumganhar, line 650: If we deprecate DOM data type, should we add a new overload method with `ImageData` parameter?

```
public reset(data: ImageData) {...}
```

cocos/asset/assets/simple-texture.ts
```diff
@@ -155,32 +178,24 @@ export class SimpleTexture extends TextureBase {
...
+        gfxDevice.copyImageDatasToTexture([imageData], this._gfxTexture, _regions);
     }
 
     /**
      * @engineInternal
      */
     protected _assignImage (image: ImageAsset, level: number, arrayIndex?: number): void {
-        const data = image.data;
-        if (!data) {
+        if (!image.data) {
```
dumganhar, line 195: Should it be `if (!image.imageData) {` ?

qiuguohua: No need, the data interface maintains compatibility and can be used in this way

cocos/asset/assets/image-asset.ts
qiuguohua: This is not recommended, imagedata doesn't feel perfect yet, and directly recommending users to use it feels unnecessary.

dumganhar (approved):

cocos/asset/assets/image-asset.jsb.ts
qiuguohua: This is implemented in the reset function.

cocos/asset/assets/image-asset.jsb.ts
qiuguohua: OK

PatriceJiang (approved):

minggo (approved):

sushanhong:
@cocos-robot run test cases

PR #15506, opened 2023-06-21
Ignore readability-convert-member-functions-to-static

PatriceJiang:
Resolve: https://github.com/cocos/3d-tasks/issues/7496

https://clang.llvm.org/extra/clang-tidy/checks/readability/convert-
member-functions-to-static.html

## Potential issues of `readability-convert-member-functions-to-static`:

- Reduced code readability: Overuse of static member functions can sometimes reduce the readability of your code, as static member functions are not tied to an instance of a class. This may make your code more difficult to understand, especially in large code bases where context is important.

- Compromised extensibility: If the non-static member functions were ever to require access to non-static members of the class in future iterations of your software, converting these functions to static could introduce the need for significant refactoring down the line.

- Loss of Polymorphism: Static member functions cannot be virtual. Therefore, if your member function is designed to be overridden in derived classes (i.e., it is a virtual function), converting it to static would remove this capability.

### Changelog

* Ignore the clang-tidy option  readability-convert-member-functions-to-static

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/.clang-tidy
+++ b/native/.clang-tidy
@@ -41,6 +41,7 @@ Checks: >
     -readability-implicit-bool-cast,
     -readability-magic-numbers,
     -readability-identifier-length,
+    -readability-convert-member-functions-to-static,
     -bugprone-branch-clone,
     -misc-static-assert,
 


minggo:
I think it depend on the design. If a function can be static, than it is better to use static.

PatriceJiang:
Agree, ant the issue being addressed here is that enabling this option by default can lead to an excess of warning messages.

PatriceJiang:
Given that most similar issues have already been avoided using NOLINT, this behavior will not be modified.

PR #15527, opened 2023-06-25
[AOT] add type for AOT engine

PPpro:

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/core/data/object.ts
+++ b/cocos/core/data/object.ts
@@ -620,7 +620,7 @@ declare namespace CCObject {
  * @return @en Whether it is a CCObject boolean value. @zh 是否为CCObject的布尔值。
  * @engineInternal
  */
-export function isCCObject (object: any): boolean {
+export function isCCObject (object: any): object is CCObject {
     return object instanceof CCObject;
 }
 
--- a/cocos/core/data/object.ts
+++ b/cocos/core/data/object.ts
@@ -68,7 +68,7 @@ const PersistentMask = ~(ToDestroy | Dirty | Destroying | DontDestroy | Deactiva
 // all the hideFlags
 const AllHideMasks = DontSave | EditorOnly | LockedInEditor | HideInHierarchy;
 
-const objectsToDestroy: any = [];
+const objectsToDestroy: CCObject[] = [];
 let deferredDestroyTimer: number | null = null;
 
 function compileDestruct (obj, ctor): Function {
--- a/cocos/particle/renderer/particle-system-renderer-cpu.ts
+++ b/cocos/particle/renderer/particle-system-renderer-cpu.ts
@@ -138,7 +138,7 @@ const _matInsInfo: IMaterialInstanceInfo = {
 };
 
 // TODO: we should not use this type, should use a uniform array type instead.
-// Tracking issue:
+// Tracking issue: https://github.com/cocos/cocos-engine/issues/15553
 export type PVData = [Vec3, Vec3, Vec3, Vec3, number, Vec3 | null, null];
 
 export default class ParticleSystemRendererCPU extends ParticleSystemRendererBase {
--- a/cocos/particle/renderer/particle-system-renderer-cpu.ts
+++ b/cocos/particle/renderer/particle-system-renderer-cpu.ts
@@ -24,7 +24,7 @@
 
 import { EDITOR_NOT_IN_PREVIEW } from 'internal:constants';
 import { builtinResMgr } from '../../asset/asset-manager';
-import { Material } from '../../asset/assets';
+import { Material, Texture2D } from '../../asset/assets';
 import { AttributeName, Format, Attribute, FormatInfos } from '../../gfx';
 import { Mat4, Vec2, Vec3, Vec4, pseudoRandom, Quat, EPSILON, approx, RecyclePool } from '../../core';
 import { MaterialInstance, IMaterialInstanceInfo } from '../../render-scene/core/material-instance';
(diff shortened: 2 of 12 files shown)


cocos/core/data/object.ts
```diff
@@ -620,7 +620,7 @@ declare namespace CCObject {
  * @return @en Whether it is a CCObject boolean value. @zh 是否为CCObject的布尔值。
  * @engineInternal
  */
-export function isCCObject (object: any): boolean {
+export function isCCObject (object: any): object is CCObject {
```
minggo, line 623: I think `boolean` is correct.

cocos/particle/renderer/particle-system-renderer-cpu.ts
```diff
@@ -136,20 +137,24 @@ const _matInsInfo: IMaterialInstanceInfo = {
     subModelIdx: 0,
 };
 
+// TODO: we should not use this type, should use a uniform array type instead.
+// Tracking issue: https://github.com/cocos/cocos-engine/issues/15553
+export type PVData = [Vec3, Vec3, Vec3, Vec3, number, Vec3 | null, null];
```
minggo, line 142: What does it mean?

PPpro: this means that this is an array type, and its element type is uncertain

for example 
```ts
let a: PVData;
a[0] // Vec3 type
a[4] // number type
```

cocos/particle/renderer/particle-system-renderer-gpu.ts
```diff
@@ -162,7 +163,7 @@ export default class ParticleSystemRendererGPU extends ParticleSystemRendererBas
         this._frameTile_velLenScale = new Vec4(1, 1, 0, 0);
         this._unifrom_velLenScale = this._frameTile_velLenScale.clone();
         this._tmp_velLenScale = this._frameTile_velLenScale.clone();
-        this._node_scale = new Vec4();
+        this._node_scale = new Vec3();
```
minggo, line 166: Why change to Vec3 as the type of _node_scale is Vec4.

cocos/scene-graph/node.ts
```diff
@@ -1160,7 +1160,7 @@ export class Node extends CCObject implements ISchedulable, CustomSerializable {
      * ```
      */
     public off (type: string, callback?: AnyFunction, target?: unknown, useCapture: any = false): void {
-        this._eventProcessor.off(type, callback, target, useCapture);
+        this._eventProcessor.off(type as NodeEventType, callback, target, useCapture);
```
minggo, line 1163: Why not just change the type of `type` from `string` to `NodeEventType `?

cocos/particle/renderer/particle-system-renderer-cpu.ts
minggo: > and its element type is uncertain

I think it is defined in the PVData. As `PVData[0]` should be `Vec3`, `PVData[1]` should be `Vec3`, and so on.

cocos/core/data/object.ts
PPpro: we need this type assertion to tell TS language service that if this method returns true, `object` should be an instance of `CCObject`

it's useful when we need to do some **Type Narrowing**, for example

```ts
function test (obj: unknown) {
  if (isCCObject(obj)) {
      obj.isValid   // we can access this property defined in `isCCObject ` class
  } else {
    obj.isValid  // this should emit an error, because it's a type of unknown
  }
}
```

cocos/particle/renderer/particle-system-renderer-gpu.ts
PPpro: _node_scale  should be a type of Vec3, and actually we use it as Vec3 here
https://github.com/cocos/cocos-engine/blob/899a3a989746f729da792d88cb84c6e32710be59/cocos/particle/renderer/particle-system-renderer-gpu.ts#L293-L305

PPpro: it is because `_particleSystem` is a type of `any` before,  we never encounter this type issue

cocos/scene-graph/node.ts
PPpro: it's a public interface, the node currently support registering any custom event name, not only NodeEventType

NodeEventType is actually string value but it's some engine internal enumerate event

cocos/particle/renderer/particle-system-renderer-gpu.ts
minggo: But the declaration of _node_scale is not changed.

cocos/scene-graph/node.ts
minggo: If so, then internally should also be string too.

cocos/particle/renderer/particle-system-renderer-gpu.ts
PPpro: _node_scale is private

cocos/core/data/object.ts
minggo: Got it, it is wired.

cocos/particle/renderer/particle-system-renderer-cpu.ts
PPpro: yes, because it was type of `any` before, it is very wierd to implement like this

we should not introduce different element types in an array, this issue should be solved here https://github.com/cocos/cocos-engine/issues/15553

minggo (approved):

cocos/scene-graph/node.ts
PPpro: I've created a new issue to deprecate `node.on(type: string)`  
https://github.com/cocos/cocos-engine/issues/15557

cocos/tween/actions/action-manager.ts
```diff
@@ -451,14 +452,14 @@ export class ActionManager {
      */
     update (dt: number): void {
         const locTargets = this._arrayTargets;
-        let locCurrTarget;
+        let locCurrTarget: HashElement;
```
SantyWang, line 455: HashElement ?

SantyWang (approved):

PR #15542, opened 2023-06-26
V3.8.0 optimize ui-prop group interface in inspector

arsen2010:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


editor/inspector/utils/prop.js
```diff
@@ -337,93 +331,68 @@ exports.getName = function(dump) {
...
-                child.style.display = 'block';
+    return $group;
+};
+exports.toggleGroup = function($groups) {
+    for (const id in $groups) {
+        const $contents = $groups[id].querySelectorAll('.ui-prop-group-content');
+
+        $contents.forEach($content => {
+            const $props = Array.from($content.querySelectorAll(':scope > ui-prop'));
+            const show = $props.some($prop => getComputedStyle($prop).display !== 'none');
```
doommm, line 349: using `window.getComputedStyle` in a loop can potentially cause performance issues. can we just detect `hidden` attribute directly?

doommm (approved):

editor/inspector/utils/prop.js
doommm: From GPT

Using `Window.getComputedStyle()` in a loop can potentially cause performance issues in JavaScript, depending on how it is used and the number of elements involved. Here's why:

1. Recalculating styles: When you call `Window.getComputedStyle()` on an element, the browser needs to recalculate the styles for that element. This can be an expensive operation, especially if the element is complex or has a large number of child elements. Performing this operation in a loop for multiple elements can significantly impact performance.

2. Reflow and repaint: Each time you access a computed style, the browser may trigger a reflow and repaint, which involves calculating the layout and rendering of the affected elements. This process can be costly, especially if it occurs repeatedly in a loop.

To mitigate these performance concerns, you should avoid calling `Window.getComputedStyle()` in a loop whenever possible. Consider restructuring your code to retrieve computed styles once and store them in variables for later use. This way, you can minimize the number of reflows and repaints.

If you absolutely need to access computed styles within a loop, try to find alternative approaches that reduce the number of iterations or optimize the way styles are handled. For example, you could consider applying CSS classes to the elements beforehand and then accessing the relevant styles directly using class-based selectors or data attributes.

Remember, it's important to profile and measure the performance of your code in different scenarios to determine the actual impact of using `Window.getComputedStyle()` in a loop and identify any potential bottlenecks.

doommm (approved):

VisualSJ (approved):

PR #15590, opened 2023-06-30
fix some eslint error

minggo:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/2d/assembler/label/font-utils.ts
+++ b/cocos/2d/assembler/label/font-utils.ts
@@ -400,7 +400,7 @@ export class LetterAtlas {
     }
 
     public getLetterDefinitionForChar (char: string, labelInfo: ILabelInfo): any {
-        const hash = char.charCodeAt(0) + Number(labelInfo.hash);
+        const hash = char.charCodeAt(0).toString() + labelInfo.hash;
         let letter = this.fontDefDictionary.letterDefinitions[hash];
         if (!letter) {
             const temp = new LetterTexture(char, labelInfo);
--- a/cocos/2d/assembler/label/font-utils.ts
+++ b/cocos/2d/assembler/label/font-utils.ts
@@ -400,7 +400,7 @@ export class LetterAtlas {
     }
 
     public getLetterDefinitionForChar (char: string, labelInfo: ILabelInfo): any {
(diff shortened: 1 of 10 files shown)


cocos/2d/assembler/label/font-utils.ts
```diff
@@ -400,7 +400,7 @@ export class LetterAtlas {
     }
 
     public getLetterDefinitionForChar (char: string, labelInfo: ILabelInfo): any {
-        const hash = char.charCodeAt(0) + labelInfo.hash;
+        const hash = char.charCodeAt(0).toString() + labelInfo.hash;
```
minggo, line 403: As  key of `letterDefinitions ` is string, so use string here.

cocos/core/data/utils/preprocess-class.ts
```diff
@@ -92,7 +92,7 @@ function parseType (val, type, className, propName): void {
         if (type.length > 0) {
             val.type = type = type[0];
         } else {
-            return errorID(5508, className, propName);
+            errorID(5508, className, propName);
```
SantyWang, line 95: Should keep `return` logic;

cocos/audio/audio-manager.ts
```diff
@@ -97,6 +97,7 @@ export class AudioManager {
             });
         }
         if (audioInfoToDiscard) {
+            // eslint-disable-next-line @typescript-eslint/no-floating-promises
             audioInfoToDiscard.audio.stop();
```
PPpro:
```suggestion
            audioInfoToDiscard.audio.stop().catch((err) => {});
```

it‘s better to catch the error of promise

cocos/core/data/utils/preprocess-class.ts
minggo: The return type is void.

SantyWang: I know, but we should not execute the code below.

SantyWang (approved):

cocos/audio/audio-manager.ts
minggo: fixed

PPpro (approved):

PR #15607, opened 2023-07-03
Add execution timeout detection

PatriceJiang:
Re: https://github.com/cocos/3d-tasks/issues/10975

[Test Case](https://github.com/cocos/cocos-test-projects/pull/815)

By setting the `native.monitor.blocking.timeout` parameter, developers can specify the maximum execution time for `gameTick`. If `gameTick` does not complete successfully within this set time frame, the system will trigger the `native.monitor.blocking.callback` callback function. The first argument of this callback function will return a string representation of the JavaScript call stack. Notably, if `timeout` is set to 0, then this monitoring mechanism will be disabled.

This monitoring mechanism facilitates developers to swiftly obtain and collect the JavaScript call stack information when game logic stutters, thereby enabling appropriate debugging and optimization.

Known Issues:
 - The mechanism for blocking checks is based on signals. This could lead to some potential issues, including conflicts with signals from third-party libraries or signals defined by the user themselves.

### Changelog

* Add `gameTick` blocking detection mechanism 

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/native-binding/impl.ts
+++ b/cocos/native-binding/impl.ts
@@ -156,4 +156,5 @@ export const native = {
     Manifest: globalJsb.Manifest,
     saveImageData: globalJsb.saveImageData,
     process: globalJsb.process,
+    blkDct: globalJsb.blkDct,
 };
--- a/cocos/native-binding/impl.ts
+++ b/cocos/native-binding/impl.ts
@@ -156,5 +156,5 @@ export const native = {
     Manifest: globalJsb.Manifest,
     saveImageData: globalJsb.saveImageData,
     process: globalJsb.process,
-    blkDct: globalJsb.blkDct,
+    monitor: globalJsb.monitor,
 };
--- a/cocos/native-binding/index.ts
+++ b/cocos/native-binding/index.ts
@@ -1403,8 +1403,8 @@ export declare namespace native {
     }
 
     /**
-     * @en A module to detect blocking execution of the gameTick procedure.
-     * @zh 用于检测 gameTick 执行阻塞的模块
+     * @en A module to detect blocking execution of the gameTick procedure. 'blkDct' is an abbreviation for 'blocking detection'.
+     * @zh 用于检测 gameTick 执行阻塞的模块. 'blkDct' 是 'blocking detection' 的简写.
      */
     namespace blkDct {
         /**
--- a/cocos/native-binding/index.ts
+++ b/cocos/native-binding/index.ts
@@ -1403,49 +1403,56 @@ export declare namespace native {
     }
 
     /**
-     * @en A module to detect blocking execution of the gameTick procedure. 'blkDct' is an abbreviation for 'blocking detection'.
-     * @zh 用于检测 gameTick 执行阻塞的模块. 'blkDct' 是 'blocking detection' 的简写.
+     * @en A module for monitoring the engine's operational status.
+     * @zh 用于检测引擎的运行状态.
      */
-    namespace blkDct {
+    namespace monitor {
+
         /**
-         * @en
-         * The timeout field allows developers to define the maximum execution time, in milliseconds, for
-         * the gameTick function. If the gameTick function exceeds this specified time limit,
-         * the system will invoke the native.blockingDetection.callback function.
-         * The first argument of this callback function will provide a string representation
-         * of the JavaScript call stack. It is important to note that setting timeout
-         * to 0 will disable this monitoring mechanism.
-         *
-         * Please be aware that the mechanism for blocking checks is based on signals.
-         * This could lead to some potential issues, including conflicts with signals from
-         * third-party libraries or signals defined by the user themselves.
-         *
-         * @zh
-         * timeout 字段允许开发者定义 gameTick 函数的最大执行时间, 单位为毫秒(ms)。
-         * 如果 gameTick 函数超出这个指定的时间限制,系统将调用 native.blockingDetection.callback
-         * 回调函数。这个回调函数的第一个参数将提供 JavaScript 调用栈的字符串表示。
-         * 将 timeout 设置为 0 将禁用这个监控机制。
-         *
-         * 请注意，阻塞检查的机制是基于信号（signal）的。这可能会引发一些潜在问题，
-         * 包括与第三方库的信号或用户自定义的信号监听发生冲突。
+         * @en A module for detecting blocking execution in the gameTick procedure.
+         * @zh 用于检测 gameTick 执行阻塞的模块.
          */
-        // eslint-disable-next-line import/no-mutable-exports
-        export let timeout: number;
-        /**
-         * @en
-         * Callback function triggered when a blocking operation is detected during the execution of gameTick.
-         * The first argument provides a string representation of the JavaScript call stack.
-         * This callback can be used to handle or log the occurrence of blocking operations in the gameTick function.
-         * It is recommended to investigate and optimize any blocking operations to ensure smooth gameplay and responsiveness.
-         * If the timeout parameter is set to 0, this callback will not be triggered as the monitoring mechanism is disabled.
-         * @zh
-         * 当在 gameTick 执行过程中检测到阻塞操作时 (超过 timeout) 触发的回调函数。
-         * 第一个参数提供了 JavaScript 调用栈的字符串表示形式。
-         * 此回调函数可用于处理或记录 gameTick 中阻塞操作的发生。
-         * 建议调查和优化任何阻塞操作，以确保游戏的流畅性和响应性。
-         * 如果将 timeout 参数设置为 0，则禁用此监控机制，回调函数将不会被触发。
-         */
-        // eslint-disable-next-line import/no-mutable-exports
-        export let callback: (stackStr?: string) => void;
+        namespace blocking {
+            /**
+             * @en
+             * The timeout field allows developers to define the maximum execution time, in milliseconds, for
+             * the gameTick function. If the gameTick function exceeds this specified time limit,
+             * the system will invoke the native.blockingDetection.callback function.
+             * The first argument of this callback function will provide a string representation
+             * of the JavaScript call stack. It is important to note that setting timeout
+             * to 0 will disable this monitoring mechanism.
+             *
+             * Please be aware that the mechanism for blocking checks is based on signals.
+             * This could lead to some potential issues, including conflicts with signals from
+             * third-party libraries or signals defined by the user themselves.
+             *
+             * @zh
+             * timeout 字段允许开发者定义 gameTick 函数的最大执行时间, 单位为毫秒(ms)。
+             * 如果 gameTick 函数超出这个指定的时间限制,系统将调用 native.blockingDetection.callback
+             * 回调函数。这个回调函数的第一个参数将提供 JavaScript 调用栈的字符串表示。
+             * 将 timeout 设置为 0 将禁用这个监控机制。
+             *
+             * 请注意，阻塞检查的机制是基于信号（signal）的。这可能会引发一些潜在问题，
+             * 包括与第三方库的信号或用户自定义的信号监听发生冲突。
+             */
+            // eslint-disable-next-line import/no-mutable-exports
+            export let timeout: number;
+            /**
+             * @en
+             * Callback function triggered when a blocking operation is detected during the execution of gameTick.
+             * The first argument provides a string representation of the JavaScript call stack.
+             * This callback can be used to handle or log the occurrence of blocking operations in the gameTick function.
+             * It is recommended to investigate and optimize any blocking operations to ensure smooth gameplay and responsiveness.
+             * If the timeout parameter is set to 0, this callback will not be triggered as the monitoring mechanism is disabled.
+             * @zh
+             * 当在 gameTick 执行过程中检测到阻塞操作时 (超过 timeout) 触发的回调函数。
+             * 第一个参数提供了 JavaScript 调用栈的字符串表示形式。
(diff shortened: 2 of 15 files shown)


native/cocos/bindings/jswrapper/v8/ScriptEngine.h
```diff
@@ -422,6 +432,7 @@ class ScriptEngine final {
     bool _isWaitForConnect;
 
     uint32_t _vmId;
+    uint32_t _blockingTimeout{1000};
```
minggo: What's the unit? Please add comment for it.

native/cocos/bindings/manual/jsb_global.h
```diff
@@ -76,3 +77,8 @@ bool jsb_run_script(const ccstd::string &filePath, se::Value *rval = nullptr);
 bool jsb_run_script_module(const ccstd::string &filePath, se::Value *rval = nullptr); // NOLINT(readability-identifier-naming)
 
 bool jsb_global_load_image(const ccstd::string &path, const se::Value &callbackVal); // NOLINT(readability-identifier-naming)
+
+/**
+ * get the blocking timeout in milliseconds
+ */
+int32_t cc_get_blocking_timeout();     // NOLINT(readability-identifier-naming)
```
minggo: Other functions starts as `jsb`, why not obey the rule?

native/cocos/bindings/jswrapper/v8/ScriptEngine.h
PatriceJiang: This is a unsued field, it will be removed

native/cocos/bindings/manual/jsb_global.h
PatriceJiang: The blockingTimeout field, in reality, has no connection with JSB, yet I haven't identified a more suitable location for its storage. Any recommendations?

cocos/native-binding/impl.ts
```diff
@@ -157,3 +157,19 @@ export const native = {
...
+        if (desc) {
+            Object.defineProperty(native, attr, desc);
+        }
+    });
+
+    Object.defineProperty(native, 'onBlockingDetected', {
+        enumerable: true,
+        configurable: true,
+        get() { return globalJsb.onBlockingDetected; },
+        set(fn) { globalJsb.onBlockingDetected = fn; }
```
dumganhar: Could we just use `value: globalJsb.onBlockingDetected` here?

cocos/native-binding/index.ts
```diff
@@ -1401,4 +1401,45 @@ export declare namespace native {
...
+     * The first argument of this callback function will provide a string representation
+     * of the JavaScript call stack. It is important to note that setting blockingTimeout
+     * to 0 will disable this monitoring mechanism.
+     *
+     * Please be aware that the mechanism for blocking checks is based on signals.
+     * This could lead to some potential issues, including conflicts with signals from
+     * third-party libraries or signals defined by the user themselves.
+     *
+     * @zh
+     * blockingTimeout字段允许开发者定义gameTick函数的最大执行时间。
```
dumganhar: Add spaces between `gameTick`.

native/cocos/bindings/jswrapper/v8/ScriptEngine.cpp
```diff
@@ -115,11 +116,12 @@ ccstd::string stackTraceToString(v8::Local<v8::StackTrace> stack) {
         stackStr += tmp;
         stackStr += "]";
         stackStr += (funcName.empty() ? "anonymous" : funcName.c_str());
-        stackStr += "@";
```
dumganhar, line 118: Why remove `@` here?

native/cocos/bindings/jswrapper/v8/ScriptEngine.cpp
```diff
@@ -1204,6 +1206,16 @@ bool ScriptEngine::callFunction(Object *targetObj, const char *funcName, uint32_
...
 }
 
+
+void ScriptEngine::requestInterrupt(void (*callback)(void*), void *data) {
+    using context_t = std::tuple<decltype(callback), void*>;
+    _isolate->RequestInterrupt(+[](v8::Isolate * /*isolate*/, void *data){
+        auto *ctx = reinterpret_cast<context_t*>(data);
+        std::get<0>(*ctx)(std::get<1>(*ctx));
+        delete ctx;
+    }, new context_t(callback, data));
```
dumganhar: Use `ccnew` instead?

cocos/native-binding/impl.ts
PatriceJiang: Developer may assign `native.onBlockingDetected` which should be synchorize to the `globalJsb` object

native/cocos/bindings/jswrapper/v8/ScriptEngine.cpp
PatriceJiang: The format is adjusted according the Devtool output format

native/cocos/engine/Engine.cpp
```diff
@@ -285,6 +301,13 @@ void Engine::tick() {
 
         prevTime = std::chrono::steady_clock::now();
 
+#if WATCHDOG_ENABLED
+        WatchDog watchdog(
```
minggo: Please don't create it if not needed.

native/cocos/engine/Engine.cpp
```diff
@@ -70,6 +71,12 @@
 #endif
 #include "profiler/Profiler.h"
 
+#if !CC_EDITOR && (CC_PLATFORM == CC_PLATFORM_ANDROID || CC_PLATFORM == CC_PLATFORM_WINDOWS || CC_PLATFORM == CC_PLATFORM_MAC_OSX || CC_PLATFORM == CC_PLATFORM_MAC_IOS)
```
dumganhar: Using `CC_PLATFORM_MACOS` instead of `CC_PLATFORM_MAC_OSX`
`CC_PLATFORM_IOS` instead of `CC_PLATFORM_MAC_IOS`.

CC_PLATFORM_MAC_OSX and CC_PLATFORM_MAC_IOS were deprecated.

native/cocos/engine/Engine.cpp
PatriceJiang: The blockingTimeout object is created at runtime, and it cannot be controlled through macros. When it is disabled, the overhead of this object is already small.

PatriceJiang: Otherwise, we need consider timeout to call new/delete. That can be more expensive.

native/cocos/bindings/jswrapper/v8/ScriptEngine.cpp
PatriceJiang: fixed

native/cocos/engine/Engine.cpp
PatriceJiang: Fixed

cocos/native-binding/index.ts
PatriceJiang: fixed

native/cocos/bindings/jswrapper/v8/ScriptEngine.h
PatriceJiang: Removed

native/cocos/bindings/manual/jsb_global.h
minggo: The value is only used in WatchDog, so it is better to do everything in global. May be can use in ts like:
```
let watchDog = new native.WatchDog(timeout);
watchDog.callback = xxx;
...
```

dumganhar:
It will be great if we add a method in ScriptEngine to listen on `watch dog timeout event`. For instance, 
```c++
void ScriptEngine::setScriptANRCallback(const ScriptANRCallback& cb);
```
If ANR callback is triggered, developer could send the ANR information to server.

PatriceJiang:
> It will be great if we add a method in ScriptEngine to listen on `watch dog timeout event`. For instance,
> 
> ```c++
> void ScriptEngine::setScriptANRCallback(const ScriptANRCallback& cb);
> ```
> 
> If ANR callback is triggered, developer could send the ANR information to server.

ScriptEngine itself does not have the concept of ANR, and ANR evaluation is performed within the engine's tick function. And the ScriptEngine has no way to provide an ANR starting point.

dumganhar:
So it probably should be a API in `Engine` class ?

native/cocos/bindings/jswrapper/v8/ScriptEngine.cpp
dumganhar: Got it.

PatriceJiang:
What kind of  API were you referring to? The only relevant API I can provide is something like setBlockingTimeout , but callbacks are set within the JS code. The JS side does not have access to an engine class.

native/cocos/bindings/manual/jsb_global.h
PatriceJiang: When developers create the native.WatchDog object, they're also required to invoke the corresponding destroy method, which could be a significant imposition. Moreover, with respect to JS code, it's not possible to monitor all methods within a single tick in the same manner as in our current implementation. This discrepancy is tied to how setTimeout/requestAnimationFrame functions are implemented.

PatriceJiang:
```js
native.blockingDetection.timeout = 100;
native.blockingDetection.callback = (jsStack)=>{...}
```
Would these modifications potentially negate the need to make changes to the global namespace?

And the `cc_get_blocking_timeout` is removed. `Engine::setBlockingTimeout` is used instead.

native/cocos/platform/android/AndroidWatchDog.cpp
```diff
@@ -0,0 +1,133 @@
...
+#include <sys/types.h>
+#include <unistd.h>
+#include <algorithm>
+#include <chrono>
+#include <csignal>
+#include <cstdlib>
+#include <ctime>
+#include <functional>
+#include <memory>
+#include "../WatchDog.h"
```
minggo: Please don't use relative path.

cocos/native-binding/index.ts
```diff
@@ -1401,4 +1401,51 @@ export declare namespace native {
          */
         export const argv: Readonly<string[]>;
     }
+
+    /**
+     * @en A module used to detect blocking execution of the gameTick procedure.
+     * @zh 用于检测 gameTick 执行阻塞的模块
+     */
+    namespace blockingDetection {
```
minggo: `blockingDetection ` is too long for a namespace.

PatriceJiang: What about `blkDct` ?

native/cocos/platform/android/AndroidWatchDog.cpp
PatriceJiang: fixed

cocos/native-binding/index.ts
```diff
@@ -1401,4 +1401,51 @@ export declare namespace native {
          */
         export const argv: Readonly<string[]>;
     }
+
+    /**
+     * @en A module to detect blocking execution of the gameTick procedure.
+     * @zh 用于检测 gameTick 执行阻塞的模块
+     */
+    namespace blkDct {
```
dumganhar: Is `blkDct` an abbreviation of `Blocking Detector`?

native/cocos/bindings/jswrapper/v8/ScriptEngine.h
```diff
@@ -317,6 +317,16 @@ class ScriptEngine final {
      */
     uint32_t getVMId() const { return _vmId; }
 
+    /**
+     * @brief
+     * This function asks ScriptEngine to interrupt lengthy JavaScript operations and run the provided callback, passing the supplied data to it.
+     * Once the callback completes, control returns to the JavaScript code. Multiple interrupt requests may be ongoing.
+     * This function can be called from a different thread without needing a Locker. The registered callback should not reenter the interrupted Isolate.
+     * @param callback
+     * @param data
```
dumganhar: Missing comments for callback and data arguments.

cocos/native-binding/index.ts
PatriceJiang: comment updated.
its short for `blocking detection`

native/cocos/bindings/jswrapper/v8/ScriptEngine.h
PatriceJiang: Updated

native/cocos/bindings/jswrapper/v8/ScriptEngine.h
```diff
@@ -322,8 +322,8 @@ class ScriptEngine final {
      * This function asks ScriptEngine to interrupt lengthy JavaScript operations and run the provided callback, passing the supplied data to it.
      * Once the callback completes, control returns to the JavaScript code. Multiple interrupt requests may be ongoing.
      * This function can be called from a different thread without needing a Locker. The registered callback should not reenter the interrupted Isolate.
-     * @param callback
-     * @param data
+     * @param callback the callback to be called when interrupt happens.
```
dumganhar: when `interruption` happens.

`interrupt` is a verb.

PatriceJiang: fixed

minggo (approved):

dumganhar (approved):

bofeng-song:
Android platform self-testing found that SIGUSR1 should be changed to SIGUSR2, otherwise the signals sent by SIGUSR1 will be intercepted by the system.

PR #15628, opened 2023-07-05
Optimize unit test; support overriding constants in per-test-suite

shrinktofit:
Re: #

### Changelog

* Optimize unit test.

  We have the requirement to await for an initialization function before some tests. Before, we implement this by expose a [waitThis](https://github.com/cocos/cocos-engine/blob/08d9a344571b6e2995242f3a8f31cf2a06c30a6b/tests/test-environment.ts#L8) in global within test environment. However in jest 28.x, they make a change so we can directly export an async function in [**setupFiles**](https://jestjs.io/docs/configuration#setupfiles-array). So here I adjusted to use this mechanism.

* Support overriding constants in per test suite.

  Sometimes we need to test a feature under different "constants". For example, we need to test the deserialization behaviour under both `SUPPORT_JIT` is true or false.

  If a test suite(ie. a test file) needs to override the default constant value. One may create a JSON config file at `<dir-of-test-file>/<name-of-the-test-file-including-extension>.config.json`, called as "test suite config". Within that config, one can specify the constant overrides. The config's content can be configured according to [`TestSuiteConfig`](https://github.com/cocos/cocos-engine/pull/15628/files#diff-d2c64ef8176f957584f6f863c240883afe45fe2695556e756c4e0aa62d34119dR2-R4).

  Note, this PR also includes a [test](https://github.com/cocos/cocos-engine/pull/15628/files#diff-86c3f0dcf4501d40cf0107971dfce64298753972cba07e3bca6fdbfa3a686932R5) for the constant overriding. It can as well be referenced as example.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/tests/init.ts
+++ b/tests/init.ts
@@ -2,14 +2,15 @@ jest.mock(
     'internal:constants',
     () => {
         const actual = jest.requireActual('./constants-for-test');
-        const { getCurrentTestSuiteConfig } = jest.requireActual('./utils/test-suite-config');
+        const { getCurrentTestSuiteConfig } = jest.requireActual('./utils/test-suite-config') as
+            typeof import('./utils/test-suite-config');
         const config = getCurrentTestSuiteConfig();
-        if (!config.constantsOverride) {
+        if (!config.constantOverrides) {
             return actual;
         }
         return {
             ...actual,
-            ...config.constantsOverride,
+            ...config.constantOverrides,
         };
     },
     { virtual: true, },
--- a/tests/init.ts
+++ b/tests/init.ts
@@ -1,6 +1,17 @@
 jest.mock(
     'internal:constants',
-    () => jest.requireActual('./constants-for-test'),
+    () => {
+        const actual = jest.requireActual('./constants-for-test');
+        const { getCurrentTestSuiteConfig } = jest.requireActual('./utils/test-suite-config');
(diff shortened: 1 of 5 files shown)


tests/init.ts
```diff
@@ -190,4 +202,6 @@ globalThis.waitThis((async () => {
     initBuiltinPhysicsMaterial();
     await game.init(config);
     await game.run();
-})());
+}
+
+module.exports = bootstrap;
```
shrinktofit, line 208: See https://jestjs.io/docs/configuration#setupfiles-array

tests/test-environment.ts
```diff
@@ -1,25 +1,36 @@
...
-        const promises = this._promises;
-        Object.defineProperty(this.global, 'waitThis', {
-            get: () => {
-                return (promise: Promise<unknown>) => {
-                    promises.push(promise);
-                };
-            },
-        });
+        const [_, context] = args;
+        this._testPath = context.testPath;
```
PPpro, line 10: where this `testPath` comes from ?

shrinktofit: https://github.com/jestjs/jest/blob/main/packages/jest-environment/src/index.ts#L16

PPpro (approved):

minggo (approved):

PR #15630, opened 2023-07-05
Tweak deserialization code

shrinktofit:
Re: #

### Changelog

* Abstract repeat code in deserialzation.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/serialization/deserialize-dynamic.ts
+++ b/cocos/serialization/deserialize-dynamic.ts
@@ -100,7 +100,7 @@ function compileDeserializeJIT (self: _Deserializer, klass: CCClassConstructor<u
     const sources = [
         'var prop;',
     ];
-    const fastMode = misc.BUILTIN_CLASSID_RE.test(js.getClassId(klass));
+    const fastMode = canBeDeserializedInFastMode(klass);
     // sources.push('var vb,vn,vs,vo,vu,vf;');    // boolean, number, string, object, undefined, function
 
     for (let p = 0; p < props.length; p++) {
@@ -172,7 +172,7 @@ function compileDeserializeJIT (self: _Deserializer, klass: CCClassConstructor<u
 }
 
 function compileDeserializeNative (_self: _Deserializer, klass: CCClassConstructor<unknown>): CompiledDeserializeFn {
-    const fastMode = misc.BUILTIN_CLASSID_RE.test(js.getClassId(klass));
+    const fastMode = canBeDeserializedInFastMode(klass);
     const shouldCopyId = js.isChildClassOf(klass, cclegacy.Node) || js.isChildClassOf(klass, cclegacy.Component);
     let shouldCopyRawData = false;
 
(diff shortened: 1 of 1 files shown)


cocos/serialization/deserialize-dynamic.ts
```diff
@@ -208,16 +199,9 @@ function compileDeserializeNative (_self: _Deserializer, klass: CCClassConstruct
...
-                } else {
-                    const defaultType = typeof defaultValue;
-                    isPrimitiveType = defaultType === 'string'
-                                      || defaultType === 'number'
-                                      || defaultType === 'boolean';
-                }
+                isPrimitiveType = isPrimitiveTypeInFastMode(defaultValue, userType);
             }
-            if (fastMode && isPrimitiveType) {
+            if (isPrimitiveType) {
```
shrinktofit: Note, **according to original code**, `isPrimitive === true` has already meant `fastMode === true`.

jareguo: Not necessarily, user-defined fields could also be primitive?

jareguo (approved):

cocos/serialization/deserialize-dynamic.ts
```diff
@@ -136,17 +136,8 @@ function compileDeserializeJIT (self: _Deserializer, klass: CCClassConstructor<u
...
-                isPrimitiveType = userType instanceof CCClass.Attr.PrimitiveType || userType === ENUM_TAG || userType === BITMASK_TAG;
-            } else {
-                const defaultType = typeof defaultValue;
-                isPrimitiveType = defaultType === 'string'
-                                  || defaultType === 'number'
-                                  || defaultType === 'boolean';
-            }
-
-            if (isPrimitiveType) {
+            const isPrimitiveTypeInFastMode = checkIsPrimitiveTypeInFastMode(defaultValue, userType);
```
minggo: Can i know why name it as fast mode?

shrinktofit: You can ask @jareguo . I also documented the definition here: https://github.com/cocos/cocos-engine/pull/15630/commits/a5a2ccd3c2e2b378705478bba900cd3a6a542cc5 .

jareguo: Yes, just as shrinktofit's remarks. FastMode does not verify every instance data that is loaded, but simply processes it uniformly. This is an optimization strategy that can only be used when the types of instance data are certain to match, and is currently only effective for types built into the engine. When the type of the engine really changes and needs to be compatible, the editor can complete the data migration, so all types of the engine are fast. However, the instance data in the user project may change continuously as the code is adjusted, so the loaded data needs to be checked twice.

cocos/serialization/deserialize-dynamic.ts
```diff
@@ -281,6 +265,28 @@ function compileDeserializeNative (_self: _Deserializer, klass: CCClassConstruct
...
+ * In fast mode, deserialization of the class will go into an optimized way:
+ * each class property will be examined whether to be primitive according to their default value
+ * and type. Finally, all primitive properties would be together deserialized using simple assignment,
+ * without performing in-loop check.
+ */
+function canBeDeserializedInFastMode (klass: any): boolean {
+    return misc.BUILTIN_CLASSID_RE.test(js.getClassId(klass));
+}
+
+function checkIsPrimitiveTypeInFastMode (defaultValue: any, userType: any): boolean {
```
minggo: As these two functions are check whether can treat them as fast mode, and TypeScript can not override function, so i suggest to rename them to
- canBeDeserializedInFastModeForClass
- canBeDeserializedInFastModeForType

cocos/serialization/deserialize-dynamic.ts
minggo: Good to know. Thanks.

cocos/serialization/deserialize-dynamic.ts
shrinktofit: > As these two functions are check whether can treat them as fast mode, and TypeScript can not override function, so i suggest to rename them to
> 
> * canBeDeserializedInFastModeForClass
> * canBeDeserializedInFastModeForType

No, the function you outlined is not what you expressed. It expresses "in fast mode, is a property with `defaultValue` and `userType`" is primitive?

minggo: Then what does fast mode mean for the two functions?

minggo: fast mode in two functions have different meaning?

minggo (approved):

PR #15638, opened 2023-07-06
Fix deserialization of engine class in certain circumstance

shrinktofit:
This closes https://github.com/cocos/3d-tasks/issues/7909

### Changelog

* Previously, if a class have cc name "cc.", and it has a property with primitive type but has neither default value nor `@type`, the property will be incorrectly deserialized. For example:

  ```ts
  @ccclass('cc.Whatever')
  class Cls { @serializable x: number; }
  
  const o = new Cls();
  o.x = 1;
  ```

 Serialize then deserialize `o` results the deserialized `o.x ` being empty object literal `{}` instead of `1`. Note, this error only appears in "non-JIT" branch of deserialization.

  This PR fixes it.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/serialization/deserialize-dynamic.ts
+++ b/cocos/serialization/deserialize-dynamic.ts
@@ -235,7 +235,7 @@ function compileDeserializeNative (_self: _Deserializer, klass: CCClassConstruct
             if (prop === undefined) {
                 continue;
             }
-            if (!fastMode && typeof prop !== 'object') {
+            if (typeof prop !== 'object') {
                 o[propName] = prop;
             } else {
                 // fastMode (so will not simpleProp) or object
--- a/tests/core/deserialize.test.ts
+++ b/tests/core/deserialize.test.ts
@@ -3,6 +3,7 @@ import { property } from '../../cocos/core/data/class-decorator';
 import { ccclass, type } from '../../cocos/core/data/decorators';
 import { deserialize } from '../../cocos/serialization/deserialize';
 import { BitMask } from '../../cocos/core/value-types/bitmask';
+import './serialization/deserialize-common-tests';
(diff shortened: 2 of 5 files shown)


cocos/serialization/deserialize-dynamic.ts
```diff
@@ -235,7 +235,7 @@ function compileDeserializeNative (_self: _Deserializer, klass: CCClassConstruct
             if (prop === undefined) {
                 continue;
             }
-            if (!fastMode && typeof prop !== 'object') {
+            if (typeof prop !== 'object') {
                 o[propName] = prop;
```
jareguo, line 239: There is no problem with this change. fastMode was judged only because it was assumed that no type serialized fields would be officially declared. In fact, this assumption should not affect the actual code logic.

cocos/serialization/deserialize-dynamic.ts
```diff
@@ -235,7 +235,7 @@ function compileDeserializeNative (_self: _Deserializer, klass: CCClassConstruct
             if (prop === undefined) {
```
jareguo, line 235: Related question: The unassigned field is undefined by default, why does it go below, it should continue here.

jareguo (approved):

cocos/serialization/deserialize-dynamic.ts
shrinktofit: Solved after conversation.

cocos/serialization/deserialize-dynamic.ts
shrinktofit: Solved after conversation.

PR #15782, opened 2023-07-19
Fix videoplayer status after pause&finish

novain1102:
Re: # related PR: https://github.com/cocos/cocos-test-projects/pull/824

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/video/video-player-impl.ts
+++ b/cocos/video/video-player-impl.ts
@@ -178,6 +178,7 @@ export abstract class VideoPlayerImpl {
     }
 
     public onEnded (e: Event): void {
+        this._playing = false;
         this.dispatchEvent(EventType.COMPLETED);
     }
 
@@ -193,7 +194,6 @@ export abstract class VideoPlayerImpl {
         }
     }
 
-    //
     public play (): void {
         if (this._loadedMeta || this._loaded) {
             this.canPlay();
--- a/cocos/video/video-player-impl.ts
+++ b/cocos/video/video-player-impl.ts
@@ -163,11 +163,11 @@ export abstract class VideoPlayerImpl {
     }
 
     public onPause (e: Event): void {
+        this._playing = false;
         if (this._ignorePause) {
             this._ignorePause = false;
             return;
         }
-        this._playing = false;
         this.dispatchEvent(EventType.PAUSED);
     }
 


cocos/video/video-player-impl.ts
```diff
@@ -163,11 +163,11 @@ export abstract class VideoPlayerImpl {
     }
 
     public onPause (e: Event): void {
+        this._playing = false;
         if (this._ignorePause) {
             this._ignorePause = false;
             return;
         }
-        this._playing = false;
```
minggo, line 170: There is `_ignorePause`. I don't quite understand its usage. But from its name, it seems to ignore stop. And if want to ignore stop, then should not change the status?

novain1102: image
As you can see, when call stop(), it will set `_ignorePause ` to ture, which is reasonable.
But then it call `video.pause()` then trigger `onPause ` in web.
Q: Why? It should be call `video.stop()` then trigger `onStop` , like on other platform:
image
image
A: Because on web, we don't have such `video.stop()` interface. So, we use `pause()` instead of `stop()`.

Based on the above, I think we should set `_playing` to false to make sure, in this condition, the video status is not playing. (obviously, when trigger `onPause`, `_playing` status shouldn't be true).
Otherwise, we got wrong `isplaying` value when click `stop` in web.

> But from its name, it seems to ignore stop. And if want to ignore stop, then should not change the status?

By the way, as far as i'm concerned, `_ignorePause ` is just to prevent to `pause` repeatly when video already `pause` or `stop` and emit extra `Event.PAUSE`. So it is safe to set `_playing = false` in advance.

minggo: How about add `stopped` status on web? Then can just remove `_ignorePause`. The `stop` implementation may be like this
```typescript
public stop (): void {
    if (this._status == STOPPED) {
        return;
    }

    this.pause();
    this._status = STOPPED;
    ...
}
```

minggo: And i think only _status variable is needed.

novain1102: > How about add `stopped` status on web? Then can just remove `_ignorePause`. The `stop` implementation may be like this
> 
> ```ts
> public stop (): void {
>     if (this._status == STOPPED) {
>         return;
>     }
> 
>     this.pause();
>     this._status = STOPPED;
>     ...
> }
> ```

Right now , `_ignorePause ` is not only use on web but also on other platform. So I do not recommand change implementation in small version like 3.8.1; In fact, there are other implementations also not reasonable enough. Maybe we can do some refactory work all together later on.

minggo: Sounds reasonable.

minggo (approved):

PR #15884, opened 2023-08-07
refine: erase unneed audio operation & fix potential bug

bofeng-song:
Re: # https://github.com/cocos/3d-tasks/issues/7318

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/audio/audio-source.ts
+++ b/cocos/audio/audio-source.ts
@@ -38,9 +38,9 @@ enum AudioSourceEventType {
     ENDED = 'ended',
 }
 
-class OperationInfo {
+class DelayedOperation {
     op: string = '';
-    params: any;
+    params: any[] | null = null;
 }
 
 /**
@@ -76,7 +76,7 @@ export class AudioSource extends Component {
     private _cachedCurrentTime = -1;
 
     // An operation queue to store the operations before loading the AudioPlayer.
-    private _operationsBeforeLoading: OperationInfo[] = [];
+    private _operationsBeforeLoading: DelayedOperation[] = [];
     private _isLoaded = false;
 
     private _lastSetClip: AudioClip | null = null;
@@ -143,6 +143,7 @@ export class AudioSource extends Component {
             this._player = player;
             this._syncStates();
             this.node?.emit(_LOADED_EVENT);
+        // eslint-disable-next-line @typescript-eslint/no-empty-function
         }).catch((e) => {});
     }
 
@@ -183,7 +184,9 @@ export class AudioSource extends Component {
     @tooltip('i18n:audio.loop')
     set loop (val) {
         this._loop = val;
-        this._player && (this._player.loop = val);
+        if (this._player) {
+            this._player.loop = val;
+        }
     }
     get loop (): boolean {
         return this._loop;
@@ -219,6 +222,7 @@ export class AudioSource extends Component {
     @range([0.0, 1.0])
     @tooltip('i18n:audio.volume')
     set volume (val) {
+        // eslint-disable-next-line no-console
         if (Number.isNaN(val)) { console.warn('illegal audio volume!'); return; }
         val = clamp(val, 0, 1);
         if (this._player) {
@@ -280,6 +284,7 @@ export class AudioSource extends Component {
     public getPCMData (channelIndex: number): Promise<AudioPCMDataView | undefined> {
         return new Promise((resolve) => {
             if (channelIndex !== 0 && channelIndex !== 1) {
+                // eslint-disable-next-line no-console
                 console.warn('Only support channel index 0 or 1 to get buffer');
                 resolve(undefined);
                 return;
@@ -357,6 +362,7 @@ export class AudioSource extends Component {
         audioManager.discardOnePlayingIfNeeded();
         // Replay if the audio is playing
         if (this.state === AudioState.PLAYING) {
+            // eslint-disable-next-line @typescript-eslint/no-empty-function
             this._player?.stop().catch((e) => {});
         }
         const player = this._player;
@@ -381,6 +387,7 @@ export class AudioSource extends Component {
             this._operationsBeforeLoading.push({ op: 'pause', params: null });
             return;
         }
+        // eslint-disable-next-line @typescript-eslint/no-empty-function
         this._player?.pause().catch((e) => {});
     }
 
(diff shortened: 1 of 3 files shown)


cocos/audio/audio-source.ts
```diff
@@ -38,6 +38,11 @@ enum AudioSourceEventType {
     ENDED = 'ended',
 }
 
+class OperationInfo {
+    op: string = '';
+    params: any;
```
PPpro:
```suggestion
    params: any[] | null;
```

cocos/audio/audio-source.ts
```diff
@@ -38,6 +38,11 @@ enum AudioSourceEventType {
     ENDED = 'ended',
 }
 
+class OperationInfo {
```
PPpro: where we used this class, can we use the `OperationInfo ` exported from `operation-queue.ts` ? or should it be an interface ?

pal/audio/operation-queue.ts
```diff
@@ -37,6 +38,38 @@ export interface OperationQueueable {
...
+        } else if (seekSearched) {
+            reserveOps.push(opInfo);
+            break;
+        } else if (reserveOps.length === 0) {
+            reserveOps.push(opInfo);
+        }
+    }
+    reserveOps.forEach((opInfo): void => {
+        const index = instance._operationQueue.indexOf(opInfo);
+        instance._operationQueue.splice(index, 1);
```
PPpro: it's better to ensure index is not -1

pal/audio/operation-queue.ts
```diff
@@ -37,6 +38,38 @@ export interface OperationQueueable {
...
+    });
+    instance._operationQueue.forEach((opInfo): void => {
+        instance._eventTarget.emit(opInfo.id.toString());
```
PPpro: this emit message means the operation is finished, are you sure ?

cocos/audio/audio-source.ts
bofeng-song: Used by _operationsBeforeLoading

cocos/audio/audio-source.ts
bofeng-song: Currenttly, we don't need an array

pal/audio/operation-queue.ts
bofeng-song: I known, what potential problems will it cause?

bofeng-song: Currently, only the play interface has a then processing. 
this.node?.emit(AudioSourceEventType.STARTED, this);

cocos/audio/audio-source.ts
bofeng-song: AudioSource only needs two variables, a function name and a parameter; it does not require as many parameter requirements as operation-queue.

cocos/audio/audio-source.ts
```diff
@@ -38,6 +38,11 @@ enum AudioSourceEventType {
...
+    op: string = '';
```
minggo: It is better to create a new type for operation type as only some operations are valid.

cocos/audio/audio-source.ts
```diff
@@ -428,15 +433,19 @@ export class AudioSource extends Component {
...
-                this._operationsBeforeLoading.length = 0;
-            }
-        }).catch((e): void => {});
+        if (this._player) {
+            this._player.loop = this._loop;
+            this._player.volume = this._volume;
+            this._operationsBeforeLoading.forEach((opInfo): void => {
+                if (opInfo.op === 'seek') {
+                    this._cachedCurrentTime = opInfo.params as number;
+                    this._player?.seek(this._cachedCurrentTime).catch((e): void => {});
```
minggo: _player? -> _player as it is checked in if condition.

minggo: And please fix the ESLint error.

minggo:
It is better to use state machine for these operations in future.

cocos/audio/audio-source.ts
PPpro: https://github.com/cocos/cocos-engine/blob/6adec7f7f6699ce30fc63c201236511eb1cde952/cocos/audio/audio-source.ts#L450

I saw here you use it as an array

and somewhere it could be null, so the more strict type should be `any[] | null`

cocos/audio/audio-source.ts
PPpro: could it be an interface instead of a class， I see you only use it as an interface, which is only a type
but class will exist in runtime

cocos/audio/audio-source.ts
bofeng-song: If params is set to any instead of any[], it will result in op being included in params and there will be a warning.

pal/audio/operation-queue.ts
PPpro: we only emit the finished message when the operation is finished, see the line no 49 https://github.com/cocos/cocos-engine/blob/f024d3f6a4c86f02d0f9343dcb4a27f30f1dd41d/pal/audio/operation-queue.ts#L46-L52


Otherwise, the call order of the queue will be messed up.

PPpro (approved):

cocos/audio/audio-source.ts
```diff
@@ -345,13 +356,14 @@ export class AudioSource extends Component {
      */
     public play (): void {
         if (!this._isLoaded && this.clip) {
-            this._operationsBeforeLoading.push('play');
+            this._operationsBeforeLoading.push({ op: 'play', params: null });
```
minggo: It is better not use magic string, such as `play`. It is better to define it as a const value, then other codes can just refer to the type. For example, may define it as 
```typescript
class AudioOperationType {
    const PLAY: string = 'play';
    ...
}
```

cocos/audio/audio-source.ts
```diff
@@ -38,6 +38,11 @@ enum AudioSourceEventType {
     ENDED = 'ended',
 }
 
+interface DelayedOperation {
```
minggo: DelayedOperation -> AudioOperationInfo

cocos/audio/audio-source.ts
```diff
@@ -38,7 +38,14 @@ enum AudioSourceEventType {
...
-interface DelayedOperation {
+enum AudioOperationType {
+     PLAY = 'play',
+     STOP = 'stop',
+     PAUSE = 'pause',
+     SEEK = 'seek'
+}
+
+interface AudioOperationInfo {
     op: string;
```
minggo: The type is `AudioOperationType `.

minggo (approved):

PR #16118, opened 2023-08-28
A better eslint rules config

finscn:
A better eslint rules config    (maybe).

需要在 package.json里加入下列依赖项: 
```
  "devDependencies": {
    "@eslint-stylistic/metadata": "^1.4.1",
    "@stylistic/eslint-plugin": "^1.4.1",
    "@stylistic/eslint-plugin-migrate": "^1.4.1",
    "@typescript-eslint/eslint-plugin": "^6.12.0",
    "@typescript-eslint/parser": "^6.12.0",
    "eslint": "^8.54.0",
    "eslint-plugin-import": "^2.29.0"
  },
```
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -235,7 +235,10 @@ rules:
         objects: always-multiline,
         imports: never,
         exports: never,
-        functions: never
+        functions: never,
+        enums: always-multiline,
+        generics: never,
+        tuples: never,
     }]
     '@stylistic/quote-props': [error, consistent-as-needed]
     '@stylistic/function-paren-newline': off
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -201,7 +201,6 @@ rules:
 
     ##### eslint-stylistic rules #####
 
-    '@stylistic/eol-last': error
     '@stylistic/no-trailing-spaces': error
     '@stylistic/space-before-blocks': error
     '@stylistic/space-before-function-paren': [warn, always] # we require this explicitl
@@ -303,3 +302,4 @@ rules:
         enforceForFunctionPrototypeMethods: false,
         ignoreJSX: all
     }]
+    '@stylistic/eol-last': error
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -85,35 +85,43 @@ rules:
     prefer-destructuring: off # auto-fix is not smart enough to merge different instances
     default-case-last: off # Place default case clause to first make it more clear that this switch statement has handled all cases
 
-    prefer-template: off
-
     func-names: off
     default-param-last: off
     one-var: [error, never]
     arrow-body-style: [error, as-needed]
-    id-denylist: [error, any, unknown, Number, number, Boolean, boolean, String, Undefined]
+    id-denylist: [error,
+        any,
+        unknown,
+        Number, number,
+        Boolean, boolean,
+        String, # string,
+        Undefined, undefined,
+    ]
     id-match: off
     no-cond-assign: off
     no-constant-condition: off
     no-prototype-builtins: off
 
-    # constructor-super: error
-    # curly: [error, all]
-    # no-debugger: error
-    # no-duplicate-case: error
-    # no-eval: error
-    # no-sparse-arrays: error
-    # no-trailing-spaces: error
-    # no-var: error
-    # prefer-const: error
-    # prefer-numeric-literals: off
+    constructor-super: error
+    curly: [error, all]
+    no-debugger: error
+    no-duplicate-case: error
+    no-eval: error
+    no-new-wrappers: error
+    no-sparse-arrays: error
+    no-trailing-spaces: error
+    no-var: error
+    prefer-const: error
+    prefer-numeric-literals: off
+    prefer-template: off
+    eol-last: error
 
     ##### TYPESCRIPT-SPECIFIC RULE OVERRIDES #####
 
     '@typescript-eslint/no-unused-expressions': [error, {
         allowShortCircuit: true,
         allowTernary: true,
-        allowTaggedTemplates: false
+        allowTaggedTemplates: false,
     }]
     '@typescript-eslint/no-empty-function': [off, {
       allow: [
@@ -125,7 +133,7 @@ rules:
     }]
 
     '@typescript-eslint/ban-ts-comment': [error, {
-        'ts-expect-error': true,
+        'ts-expect-error': false,
         'ts-ignore': true,
         'ts-nocheck': true,
         'ts-check': false,
@@ -155,11 +163,11 @@ rules:
     '@typescript-eslint/no-non-null-assertion': off # sometimes we just know better than the compiler
     '@typescript-eslint/no-namespace': [warn, { # we need to declare static properties
         allowDeclarations: true,
-        allowDefinitionFiles: true
+        allowDefinitionFiles: true,
     }]
     '@typescript-eslint/restrict-template-expressions': [warn, { # concatenations of different types are common, e.g. hash calculations
         allowNumber: true,
-        allowBoolean: true
+        allowBoolean: true,
     }]
 
 
@@ -168,7 +176,7 @@ rules:
     # For example, https://babeljs.io/docs/en/babel-plugin-transform-typescript#istsx
     # forbids angle bracket style if `isTSX` is enabled.
     '@typescript-eslint/consistent-type-assertions': [error, {
-        assertionStyle: 'as'
+        assertionStyle: 'as',
     }]
 
     # Prefer the interface style.
@@ -177,8 +185,14 @@ rules:
     '@typescript-eslint/no-this-alias': off
     '@typescript-eslint/no-duplicate-type-constituents': [error, {
         ignoreIntersections: false,
-        ignoreUnions: true
+        ignoreUnions: true,
     }]
+    '@typescript-eslint/explicit-member-accessibility': [error, {
+         accessibility: 'no-public',
+    }]
+    '@typescript-eslint/member-ordering': off
+    '@typescript-eslint/naming-convention': off
+    '@typescript-eslint/no-unnecessary-boolean-literal-compare': off
 
 
     # '@stylistic/key-spacing': [error, {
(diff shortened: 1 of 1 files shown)


.eslintrc.yaml
```diff
@@ -163,3 +149,78 @@ rules:
...
+
+
+    # '@typescript-eslint/no-unsafe-argument': off
+
+    max-len: [warn, 250] # more lenient on max length per line
+    no-console: warn # prefer the uniform logging methods
+    '@typescript-eslint/no-empty-function': off
+    no-multi-spaces: [warn, {'ignoreEOLComments': true, 'exceptions': {'Property': true, 'VariableDeclarator': false } }]
+
+    comma-dangle: [error, {
```
shrinktofit: Why don't you choose to disallow trailing commas? I think it'd better to allow. For example, if you had the following array:

```ts
const array = [
  a,
  b
];
```

In this time, you want to append a new item, the diff might be look like:

```diff
- b
+b,
+c
```

In compare with the allowing case, the diff would be:

```diff
  b,
+c,
```

finscn: I've changed the config. I think you are right.

shrinktofit:
This PR needs to be refined for following reasons:

1. The PR refers too many lint rule changes, which is hard to be reviewed.

2. To convince our developers, it's required that "why should it be this" reason is documented for each lint rule. For example, we've had:
  
  image


3. To not bother later PRs, PRs issuing lint changes should also issues the changes applied to related files.

finscn:
> This PR needs to be refined for following reasons:
> 
> 1. The PR refers too many lint rule changes, which is hard to be reviewed.
> 2. To convince our developers, it's required that "why should it be this" reason is documented for each lint rule. For example, we've had:
> 
> image
> 
> 3. To not bother later PRs, PRs issuing lint changes should also issues the changes applied to related files.

大部分lint的修改 难以通过一两句注释说明, 而且我的英文不好.
  我非常愿意通过远程会议等方式 和你们做进一步沟通.

.eslintrc.yaml
```diff
@@ -163,3 +160,100 @@ rules:
...
+    # no-debugger: error
+    # no-duplicate-case: error
+    # no-eval: error
+    # no-sparse-arrays: error
+    # no-trailing-spaces: error
+    # no-var: error
+    # prefer-const: error
+    # prefer-numeric-literals: off
+
+    prefer-template: off
```
dumganhar: Why to disable `prefer-template`?

.eslintrc.yaml
```diff
@@ -163,3 +160,100 @@ rules:
...
+    '@typescript-eslint/no-unsafe-argument': warn
```
dumganhar: This change the default `error` to `warn`. Any reasons for this?

dumganhar: Refer to https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-unsafe-argument.md

I think marking it as `error` is better.

.eslintrc.yaml
```diff
@@ -163,3 +160,100 @@ rules:
...
+
+    max-len: [warn, 250] # more lenient on max length per line
+    no-console: warn # prefer the uniform logging methods
```
dumganhar: Code in engine should not use `console.log/warn/error/info`  directly.  `error` is better.

.eslintrc.yaml
```diff
@@ -163,3 +160,100 @@ rules:
...
+    max-statements-per-line: [error, { max: 1 }]
+
+    comma-dangle: [error, {
+        'arrays': always-multiline,
+        'objects': always-multiline,
+        'imports': never,
+        'exports': never,
+        'functions': never
+    }]
+    quote-props: [error, consistent-as-needed]
```
dumganhar: This config will make 
```ts
const object1 = {
    foo: 'bar',
    baz: 42,
    'qux-lorem': true, // string with `-` could not be a key without quote.
};
```

to be changed to

```ts
const object1 = {
    'foo': 'bar', // quote ' was added for `foo` property
    'baz': 42,  // quote ' was added for `baz` property
    'qux-lorem': true,
};
```

So I think keep the default behavior is better.

.eslintrc.yaml
```diff
@@ -163,3 +160,100 @@ rules:
...
+    function-paren-newline: off
```
dumganhar: Its default value is `multiline` which i think it's better. 

If we disable this config, the following code
```ts
function foo (
    bar, baz
) {}

const foo1 = function (
    bar, baz
) {};

const foo2 = (
    bar, baz
) => {};
```

will `not` be formated to 

```ts
function foo (bar, baz) {}

const foo1 = function (bar, baz) {};

const foo2 = (bar, baz) => {};
```

.eslintrc.yaml
```diff
@@ -163,3 +160,100 @@ rules:
...
+    space-unary-ops: error
```
dumganhar: This rule `space-unary-ops: error` is enabled by default. Why do we have to set it explicitly？

.eslintrc.yaml
```diff
@@ -163,3 +160,100 @@ rules:
...
+        'singleLine': {
+            'beforeColon': false,
+            'afterColon': true
+        },
+        'multiLine': {
+            'beforeColon': false,
+            'afterColon': true,
+            'align': colon
+        }
+    }]
```
dumganhar, line 228: This rule is not good for our code. If we add a longer property in object, all previous properties will change.
For instance, if we get
```ts
const obj = {
    hello: 'world',
    a    : 'bbb',
};
```

If we want to add a new property called `foooooooo` like:

```ts
const obj = {
    hello    : 'world',  // code changed
    a        : 'bbb',      // code changed
    foooooooo: 'ahaha',
};
```
We just added a new property, but the whole object definition code were changed.

.eslintrc.yaml
finscn: 字符串模板性能更低,  开启该功能会强制使用字符串模板, 即使简单的一次拼接(如 `"Hello " + user.name`) 也会被转换成 `'Hello ${user.name}'`. 得不偿失.

.eslintrc.yaml
finscn: 因为按照目前的cocos代码, 如果是 error, 那么会大量标红.如果cocos愿意大规模重构代码, error当然是更好的

.eslintrc.yaml
finscn: 理论上是这样, 但是 `console.log` 无法避免. 无论是开发者的代码 还是引擎代码. 总是会在某些时候需要在控制台打印一些信息. 即使生产环境也是如此.

.eslintrc.yaml
finscn: 在实际代码中, 一致性也是很重要的. 在一个json/object的代码块中, key的形式应该尽量保证一致.
 而且在现代编辑器中 有没有引号 会导致 key的颜色不同.  如下图:

image

如果实际代码中出现这种 key的颜色不一致的情况, 其实是对阅读代码产生干扰的.

.eslintrc.yaml
finscn: 嗯 这个我可以理解.按你说的问题不大.

.eslintrc.yaml
finscn: 实际测试下来 ,  space-unary-ops 并不是error,  可能是被继承的配置篡改了.
```
extends:
    - eslint:recommended
    - airbnb-base
```
其实我建议不要继承这些所谓的 "已有的最佳实践", 这样的话 eslint的配置就是黑盒的了.
可以去看看     eslint:recommended 和  airbnb-base里是怎么设置的 适合cocos的就copy过来. 不适合的就删除或者改掉.

.eslintrc.yaml
finscn: 单行 function 就和 单行 if  单行 for 一样,  除了 "节省行数"  并没有任何其他好处. 反而破坏代码的一致性. 而且会造成一定的阅读困难.
要不要把函数写成一行 应该是开发者根据函数复杂度和代码长度自己决定.
打开这个选项, 会出现很多 "虽然变为了单行, 但是单行长度很长" 的情况. 并不便于代码的阅读.

我个人完全无法接受   单行函数,  单行if  单行for .

.eslintrc.yaml
```diff
@@ -163,3 +160,100 @@ rules:
...
+
+    func-names: off
```
dumganhar: The default value of `func-names` is `warn`. Why do we need to disable the warning?
Anonymous functions will hard for debugging.

.eslintrc.yaml
```diff
@@ -163,3 +160,100 @@ rules:
...
+    default-param-last: off
```
dumganhar: https://eslint.org/docs/latest/rules/default-param-last

It's better to put default argument to the last. Why should we disable this check?

.eslintrc.yaml
dumganhar: I agree with this.

.eslintrc.yaml
finscn: 打开这个开关, 对于  `arr.forEach( (item, index) => {})`  还有很多 callback/ hook 函数 也要求写名字 太讨厌了.

.eslintrc.yaml
finscn: 在实际开发中我经常有类似: 
function   foo(a:number , b:number =1,  c?:string) {  ... }  这样的需求.
就是  可选参数在最后一个.  强制给可选参数赋默认值, 会破坏原有逻辑.   `不传参`和`传默认值` 是完全不同的概念

dumganhar: Got it.

minggo:
一下子改动太多的规则了，会导致引擎出现很多的错误。关于规则的改动得一个个来。

finscn:
> 一下子改动太多的规则了，会导致引擎出现很多的错误。关于规则的改动得一个个来。

但是实际情况是 这么久了, 是一个都没改过啊....

minggo:
我的意思是你 PR 的时候不要一下子改太多。最好一次一个。这样每个配置可以单独讨论，不会耦合在一起。

Issue #15912, opened 2023-08-09
support a mininum pal module to varify engine build tool


PPpro:
https://github.com/cocos/cocos-engine/issues/12647

the core module is too big to make it modularized https://github.com/cocos/cocos-engine/issues/14293
I'll try to modularize the `pal/env` `pal/pacer` or `pal/wasm` module first

and then handle the other build tool issues: https://github.com/cocos/cocos-engine/issues/14294 https://github.com/cocos/cocos-engine/issues/14295

the core module task will be latter

PPpro:
wait for task https://github.com/cocos/cocos-engine/issues/15895

PR #16196, opened 2023-09-06
support @pal/env and @pal/utils modules

PPpro:
Re: https://github.com/cocos/cocos-engine/issues/15912

### Changelog

* support `@pal/env` and `@pal/utils` modules with __node modules__
* this should be a practice for engine modularize workflow
* fix compile-native-ts workflow with `TsImportType` expression https://github.com/cocos/cocos-ccbuild/pull/40

### Rely on PR
- build tools support https://github.com/cocos/cocos-editor/pull/2511
- `TsImportType` support for ts engine build https://github.com/cocos/cocos-ccbuild/pull/40

### NOTE

should run `npm i` to link the modules

### Ts Module vs. Node Modules
- our legacy implementation for pal is __Ts Modules__, eg.
```ts
declare module 'pal/env' {}
```
- now we implement pal as __Node Modules__, eg.
```json5
// pal/env/package.json
{
  name: "@pal/env",
  // ...
}
```
- The problem with the __Ts Modules__ is that it cannot reference enumerated types, and the type definition and implementation cannot be put together.
- The __Node Modules__ can solve the above problems and provide good module encapsulation through conditional export
- now we can set the type definition of module by `types` filed
https://github.com/cocos/cocos-engine/blob/ed2ed4a35087a6d8df01b6036fe3250d1d13795e/modules/pal/env/package.json#L12
- and set the platform specific export by these conditional export 
https://github.com/cocos/cocos-engine/blob/ed2ed4a35087a6d8df01b6036fe3250d1d13795e/modules/pal/env/package.json#L13-L20

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


modules/pal/env/package.json
```diff
@@ -0,0 +1,23 @@
...
+      "native": "./src/native/env.ts",
+      "web": "./src/web/env.ts",
+      "minigame": {
+        "default": "./src/minigame/env.ts",
+        "oppo": "./src/runtime/env.ts",
+        "vivo": "./src/runtime/env.ts",
+        "huawei": "./src/runtime/env.ts"
+      }
+    }
+  }
```
PPpro: here are the module conditional exports, which should replace the platform related moduleOverride machnism in `cc.config.json`, see this schema design https://github.com/cocos/cocos-ccbuild/blob/f4d039f0bed5a5c753ce67d0f04c4fc8bafd130f/modules/modularize/src/module-config.ts#L53-L87

modules/pal/env/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "../../../schemas/module-config.schema.json",
```
PPpro: each module config should reference this schema

modules/pal/env/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "../../../tsconfig.json",
```
PPpro: each module `tsconfig.json` should extends the root `tsconfig.json`

tsconfig.json
```diff
@@ -3,6 +3,7 @@
     "target": "es6",
     "lib": ["es2015", "es2017", "dom"],
     "module": "commonjs",
+    "moduleResolution": "Node16",
```
PPpro, line 6: we use node module resolution policy to support the engine modularize

tsconfig.json
```diff
@@ -14,6 +15,7 @@
     "experimentalDecorators": true,
     "isolatedModules": true,
     "forceConsistentCasingInFileNames": true,
+    "preserveSymlinks": false,
```
PPpro, line 18: use the real file path instead of symlink path

cocos/game/game.ts
```diff
@@ -25,7 +25,7 @@
 
 import { DEBUG, EDITOR, NATIVE, PREVIEW, TEST, EDITOR_NOT_IN_PREVIEW } from 'internal:constants';
 import { systemInfo } from 'pal/system-info';
-import { findCanvas, loadJsFile } from 'pal/env';
+import { findCanvas, loadJsFile } from '@pal/env';
```
minggo, line 28: I thought pal is a module, why can import from its subclass? Or `pal/env` is a module, and pal contains more modules?

PPpro: all the pal modules can be merged together, but it need times to organize module structure

`@pal/env` `@pal/utils` is the smallest module I could thought.

cocos/game/game.ts
```diff
@@ -41,6 +41,7 @@ import { bindingMappingInfo } from '../rendering/define';
 import { ICustomJointTextureLayout } from '../3d/skeletal-animation/skeletal-animation-utils';
 import { IPhysicsConfig } from '../physics/framework/physics-config';
 import { effectSettings } from '../core/effect-settings';
+import { log, error } from '../core/platform/debug';
```
minggo: Can import from `core`?

modules/pal/utils/package.json
```diff
@@ -0,0 +1,15 @@
+{
```
minggo: Why need to put into `modules/`?

PPpro: I want to be isolated from code that is not modularized

All scripts that have not been modularized are placed outside the modules folder

pal/minigame/baidu.ts
```diff
@@ -23,9 +23,9 @@
 */
 
 import { IMiniGame } from 'pal/minigame';
-import { checkPalIntegrity, withImpl } from '../integrity-check';
+import { checkPalIntegrity, withImpl, cloneObject, createInnerAudioContextPolyfill } from '@pal/utils';
 import { Orientation } from '../screen-adapter/enum-type';
-import { cloneObject, createInnerAudioContextPolyfill } from '../utils';
+import { warn } from '../../cocos/core/platform/debug';
```
minggo, line 28: Can import from core instead of `cocos/core/platform/debug`?

cocos/game/game.ts
PPpro: yes, we can in `game.ts`

pal/minigame/baidu.ts
PPpro: better not, it is easy to cause circular references

minggo: Currently can use it like this. We should import from core only at the end.

modules/pal/utils/package.json
minggo: You can record it in other way, but not use folder to do it.

cocos/game/game.ts
minggo: Ok.

modules/pal/utils/package.json
PPpro: we also need to config the module workspaces here, the module declared in these directory could be linked while we execute `npm i` command
https://github.com/cocos/cocos-engine/blob/6b3fa742382acfe3b2e8dd5e5712a5594cd9c1d5/package.json#L15-L18

PPpro: I think they still needs to be put in a separate folder, because each module is an independent ts project, they have their own `tsconfig.json`, which is no longer suitable for putting together with the other scripts

minggo: Can just use pal folder. Maybe there is pal/config folder to record configuration files. It is strange to see
```
modules/a
modules/b
```
in engine. Because a and b should be a module.

PPpro: the`/modules/pal` can be merged into `/pal` after we modularize all pal modules

PPpro: It seems it's OK to merge into pal directory
https://github.com/cocos/cocos-engine/pull/16196/commits/9dc9cf5feba2745028cef6def66597b5e631669a

pal/env/src/index.d.ts
```diff
@@ -0,0 +1,2 @@
+export function findCanvas (): { frame: HTMLDivElement, container: HTMLDivElement, canvas: HTMLCanvasElement };
+export function loadJsFile (path: string): Promise<void>;
```
PPpro, line 2: this comes from the removed env.d.ts https://github.com/cocos/cocos-engine/pull/16196/files#diff-9cbea41d2702dc9457f92742092de2d9d06374d0aa38684cdcb94f80debf17e7

minggo (approved):

pal/minigame/baidu.ts
PPpro: yes for sure

shrinktofit (approved):

qiuguohua (approved):

Issue #15910, opened 2023-08-09
support @cocos/ambient-type internal module


PPpro:
this module provide with ambient type, which will include all types defined in https://github.com/cocos/cocos-engine/tree/develop/%40types

also support its own build task, for now this build task is to generate `'internal:constants'` module

PPpro:
wait for task https://github.com/cocos/cocos-engine/issues/15895

PPpro:
I didn't see what benefits this module could bring to the engine's workflow. Instead, it would increase maintenance costs, so I decided not to deal with this task anymore and closed the PR： https://github.com/cocos/cocos-engine/pull/16186

PR #16215, opened 2023-09-07
support @types/cc-ambient-types 

PPpro:
Re: https://github.com/cocos/cocos-engine/issues/15910

### Changelog

- support `@types/cc-ambient-types` module to provide global types definition for all modules

### Why we need this module ?

Before, we reference all the types in one root `tsconfig.json`, now we have more and more separate modules and each module has their own `tsconfig.json`, but the `types` config cannot be extended from the root `tsconfig.json` because the reference path is relative path.
If we don't provide this `@types/cc-ambient-types` module, we need to config `types` in each module's `tsconfig.json`, which would be much more cumbersome.

### Breaking Change
- remove `npm run build-declaration` and `npm run build-const` workflow and merge them into `npm run build` workflow
- `npm run build` doesn't build engine anymore which is meaningless
- `npm run build` will automatically run after `npm i`
- `npm run build` will run all the modules build task with `turbo build`, and run the root build task with `npm run build-root`
- in the future, we also want to remove `gulp build-debug-infos && npm run bundle-adapter && npm run fetch-ral` workflow, we only need one `npm run build` command to finish all the build tasks.

### CI Failed Analysis

The interface check failed because the diff content is out of the maximum of string 65536.

Here are the interface compare data, most of the diff comes from the private type which is named as path. We can ignore these diff.
[before.zip](https://github.com/cocos/cocos-engine/files/12549394/before.zip)
[after.zip](https://github.com/cocos/cocos-engine/files/12549405/after.zip)
image

the __\<Native\>Generate Code__ failure is fix here https://github.com/cocos/cocos-engine/pull/16214 https://github.com/cocos/cocos-engine/pull/16214/commits/d3231a12b892ea82f149825ffced6f35b8b6acc7

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


PPpro:
image

packages/cc-ambient-types/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "build": "node ./scripts/build-consts.js"
+  },
+  "license": "MIT",
+  "types": "./src/index.d.ts",
+  "exports": {
+    ".": {
+      "node": "./src/index.d.ts",
+      "types": "./src/index.d.ts"
+    },
+    "./query": "./scripts/query.js"
```
PPpro, line 16: we add this `query` export to provide some interface to query info of the module, such as `getDtsFiles()`

https://github.com/cocos/cocos-engine/blob/bb83a8653fa50185e9ad4f62e028c57d45f699fe/scripts/compile-native-ts.js#L88-L89

packages/cc-ambient-types/src/index.d.ts
```diff
@@ -0,0 +1,30 @@
...
+/// <reference path="./pal/image.d.ts"/>
+
+// external types
+/// <reference path="../../../native/external/emscripten/external-wasm.d.ts"/>
+/// <reference path="../../../native/external/emscripten/bullet/bullet.d.ts"/>
+/// <reference path="../../../native/external/emscripten/physx/physx.d.ts"/>
+/// <reference path="../../../native/external/emscripten/webgpu/webgpu.d.ts"/>
+/// <reference path="../../../native/external/emscripten/spine/spine.d.ts"/>
+/// <reference path="../../../native/external/emscripten/meshopt/meshopt_decoder.d.ts"/>
+/// <reference path="../../../native/external/emscripten/box2d/box2d.d.ts"/>
```
PPpro, line 30: we add types reference here

packages/cc-ambient-types/src/pal/audio.d.ts
```diff
@@ -50,7 +50,7 @@ declare module 'pal/audio' {
          * @param opts Load options.
          * @returns The audio player.
          */
-        static load (url: string, opts?: import('pal/audio/type').AudioLoadOptions): Promise<AudioPlayer>;
+        static load (url: string, opts?: import('../../../../pal/audio/type').AudioLoadOptions): Promise<AudioPlayer>;
```
PPpro, line 53: these relative paths could be fixed after they are modularized

tsconfig.json
```diff
@@ -23,33 +23,8 @@
...
-      // pal
-      "./@types/pal/system-info",
-      "./@types/pal/screen-adapter",
-      "./@types/pal/minigame",
-      "./@types/pal/audio",
-      "./@types/pal/input",
-      "./@types/pal/pacer",
-      "./@types/pal/wasm",
-      "./@types/pal/image",
+      "cc-ambient-types",
```
PPpro, line 26: we solve the relative path issue in this way, see https://www.typescriptlang.org/tsconfig#types

minggo (approved):

shrinktofit (approved):

PR #16235, opened 2023-09-11
feat: support touch info query in input module

PPpro:
Re: https://github.com/cocos/cocos-engine/issues/15983

### Changelog

- support touch info query in input module, thanks @finscn for the suppor PR https://github.com/cocos/cocos-engine/pull/15985
- this PR gives an optimized version on this commit https://github.com/cocos/cocos-engine/pull/16235/commits/38fee133096da655c33a4525765d77abe70e7224 https://github.com/cocos/cocos-engine/pull/16235/commits/0955f777634b81159a42cc08531370e4feab6752, and cherry pick to the branch v3.8.2

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/input/input.ts
+++ b/cocos/input/input.ts
@@ -248,9 +248,8 @@ export class Input {
      * @zh
      * 通过 touch ID 获取 touch对象。
      */
-    public getTouch (touchID: number): Touch | undefined {
-        const touch = touchManager._touchMap.get(touchID);
-        return touch;
+    public getTouch (touchID: number): Readonly<Touch> | undefined {
+        return touchManager.getTouch(touchID);
     }
 
     /**
@@ -270,7 +269,7 @@ export class Input {
      * 获取当前 touch 对象的数量。
      */
     public getTouchCount (): number {
-        return touchManager._touchMap.size;
+        return touchManager.getTouchCount();
     }
 
     /**
@@ -308,7 +307,7 @@ export class Input {
     private _simulateEventTouch (eventMouse: EventMouse): void {
         const eventType = pointerEventTypeMap[eventMouse.type];
         const touchID = 0;
-        const touch = touchManager.getTouch(touchID, eventMouse.getLocationX(), eventMouse.getLocationY());
+        const touch = touchManager.getOrCreateTouch(touchID, eventMouse.getLocationX(), eventMouse.getLocationY());
         if (!touch) {
             return;
         }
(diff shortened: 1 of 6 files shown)


cocos/input/input.ts
```diff
@@ -229,6 +241,38 @@ export class Input {
...
         this._eventTarget.off(eventType, callback, target);
     }
+
+    /**
+     * @en
+     * Get touch object by touch ID.
+     * @zh
+     * 通过 touch ID 获取 touch对象。
+     */
+    public getTouch (touchID: number): Touch | undefined {
```
minggo: I think it is better to return a readonly object.

cocos/input/input.ts
```diff
@@ -229,6 +241,38 @@ export class Input {
...
+        return touch;
+    }
+
+    /**
+     * @en
+     * Get all the current touches objects as array.
+     * @zh
+     * 获取当前 所有touch对象 的数组。
+     */
+    public getAllTouches (): Touch[] {
```
minggo, line 261: The same as above.

pal/input/touch-manager.ts
```diff
@@ -22,39 +22,25 @@
...
 
 const tempVec2 = new Vec2();
 
 class TouchManager {
     /**
      * A map from touch ID to touch object.
+     * @engineInternal
      */
-    private _touchMap: Map<number, Touch>;
+    public _touchMap: Map<number, Touch>;
```
minggo: Why change to public?

cocos/input/input.ts
minggo: And it is not a good idea to add Input.getTouch() as Input is used for all types input event, such as touch, keyboard and so on.

PPpro: > And it is not a good idea to add Input.getTouch() as Input is used for all types input event, such as touch, keyboard and so on

I think it's OK to add this query interface, latter we could also add `getKeyDown() getKeyUp()` interface to support querying keyboard key state

PPpro: > I think it is better to return a readonly object.

Readonly is useless for a Touch object, we provide setting method on Touch such as `setPoint()` `setPrevPoint()`, Readonly won't work on this.

so that's why I support a `clone()` method on Touch class to guide developers to protect the Touch data

pal/input/touch-manager.ts
PPpro: I've changed to private 
https://github.com/cocos/cocos-engine/pull/16235/commits/f3810f074d42f657cc39828f63bc0a551c2e0336

cocos/input/input.ts
PPpro: the same answer above

minggo (approved):

Issue #12647, opened 2022-09-01
Adjust engine folder structure


minggo:
As discussed moths ago, we will adjust engine folder structure to make it more clear. The first level folder structure looks like

## TS
- [x] core
  - Serialization/DeSerialization
  - memory management
  - math
  - core data structure
  - pal
  - file system
  - event system
- [x] audio
- [x] physics
- [x] gfx
- [x] rendering
- [x] render-scene
- [x] scene-graph
- [x] renderer-2d
- [x] renderer-3d
- [x] ui
- [x] input
- [x] misc
- [x] particle
- [x] animation
- [x] terrian
- [x] asset
- [x] extension
   - [x] spine
   - [x] dragonbone
- [x] profiler
## CPP

- [ ] core
  - Serialization/DeSerialization
  - memory management
  - math
  - core data structure
  - pal
  - file system
  - event system
- [ ] audio
- [ ] physics
- [ ] gfx
- [ ] rendering
- [ ] render-scene
- [ ] scene-graph
- [ ] renderer-2d
- [ ] renderer-3d
- [ ] ui
- [ ] input
- [ ] network
- [ ] misc
- [ ] particle
- [ ] animation
- [ ] terrian
- [ ] asset
- [ ] extension
   - [ ] spine
   - [ ] dragonbone
- [ ] scripting
- [ ] threading
- [ ] profiler

minggo:
TS folder structure is almost done. May adjust a little if find circular dependency in future. As modular mechanism is not ready, and @PPpro can not finish it in v3.7, so will not adjust c++ folder, as may modify the folder structure again when merging the same module of TS/c++ files into a folder.

So will delay this issue to next version.

PR #16247, opened 2023-09-12
support @base/global module

PPpro:
Re: 
https://github.com/cocos/cocos-engine/issues/14293
https://github.com/cocos/cocos-engine/issues/12647

## Changelog

* remove the internal `legacyCC`
* the only public interface is `cclegacy`
* support interactive command utils `node ./scripts/recast-import.js`, this tools gives an interactive way to help fix the import statement, like this
image
```ts
import { js } from './core';
// should change to 
import { js } from '@cocos/core' 
```
* setup a `@base/global` module to encapsure the `global-export.ts` in core module

## What is `@base` ? What is `@pal` ?

the engine layer structure should be like this

</img>

I think we need another layer called `base`, which should includes the basic value type like `Vec2` `Vec3` `Size` `Color` `Touch`, and basic infrastructure like the `cclegacy` `ccwindow` global variables.

### before we provide this module

our implementation is wierd for now, because our `core` layer includes the `base` module, 
so here we generated a circular reference like `core -> pal -> base -> core`
image

### after that

so here I wanna create a new layer called `base`, after that the dependency graph should be like this
image

### and that's where the module names come from

`@base/global` `@pal/env` `@pal/utils` means which layer we should put the modules in. 
and __maybe__ we could have the layers in the future below:
- `@base/xxx`: includes some basic value types and infrastructure, provided to the upper layer
- `@pal/xxx` 
- `@core/xxx` 
- `@resources/xxx`
- `@cocos/xxx`: the function layer of cocos engine, or we call it the framework layer
- `@editor/xxx`: to provide some editor related features (I think this should be our final target to develop editor with cocos engine)

## How to merge modules

Later if we need to merge all the `@pal/xxx` into one modules named `pal`, all we need to do is 
- put all the modules together
- setup an index file for this module
- replace and merge all the import statement in engine repo, like
```ts
import { touchInputSource } from '@pal/input';
import { findCanvas } from '@pal/env';
// should change to 
import { touchInputSource, findCanvas } from 'pal';

// maybe we could support a tool to do this
```

We also need to take care whether the module is public, if it's a public module, we need to do some migration after we change the module dependency graph, and here is another task to support the migration mechanism https://github.com/cocos/3d-tasks/issues/17452

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


minggo:
Yep, i like the idea to have base module. But from the codes above, it seems there are more modules in base space. Can i know what modules will be included in base space? Currently, i saw `@base/global`, any other modules?

PPpro:
> Yep, i like the idea to have base module. But from the codes above, it seems there are more modules in base space. Can i know what modules will be included in base space? Currently, i saw @base/global, any other modules?

for now I can say, it should also includes 
- `@base/debug`: provide some log utils and debug info
- `@base/js`: our js utils
- `@base/event`: `EventTarget` is all the layers need to emit event
- `@base/value-types`:  includes 
  - some math types like `Vec2` `Vec3` `Mat3` ...
  - `Size` `Color` `Rect`
  - `Touch`, and maybe `Event`?

@minggo

cocos/asset/asset-manager/deprecated.ts
```diff
@@ -24,14 +24,18 @@
...
+import { AssetManager } from './asset-manager';
 import { resources } from './bundle';
 import dependUtil from './depend-util';
-import downloader, { Downloader } from './downloader';
+import downloader from './downloader';
+import { Downloader } from './downloader';
 import { getUuidFromURL, transform } from './helper';
-import parser, { Parser } from './parser';
+import parser from './parser';
+import { Parser } from './parser';
```
minggo, line 38: Can just write like
```
import { parser, Parser } from './parser';
```
?

cocos/asset/asset-manager/downloader.ts
```diff
@@ -23,9 +23,11 @@
 */
 
 import { BUILD, EDITOR, EDITOR_NOT_IN_PREVIEW } from 'internal:constants';
-import { sys, js, misc, path, cclegacy } from '../../core';
+import { sys, js, misc, path } from '../../core';
+import { cclegacy } from '@base/global';
 import Cache from './cache';
-import downloadFile, { FileProgressCallback } from './download-file';
+import downloadFile from './download-file';
+import { FileProgressCallback } from './download-file';
```
minggo, line 30: The same as above.

cocos/asset/asset-manager/deprecated.ts
PPpro: yes, the recast-import tools auto separates the default import and named import, I'll fix this mannually

cocos/asset/asset-manager/downloader.ts
PPpro: https://github.com/cocos/cocos-engine/pull/16247/commits/747bea8e1163cddc6efb57de109318dff61ee256

cocos/asset/asset-manager/deprecated.ts
PPpro: https://github.com/cocos/cocos-engine/pull/16247/commits/747bea8e1163cddc6efb57de109318dff61ee256

cocos/asset/asset-manager/downloader.ts
minggo: There are many codes is modified like this. I don't mark all these codes.

PPpro: I am gonna keep this two separate import statement and revert the fix here https://github.com/cocos/cocos-engine/pull/16247/commits/701ccf5bda2190454a28bb71281966211921e6c5

because:
1. I think it's not a good implementation to merge `default import` and `named import` together.
```ts
import downloadFile from './download-file';  // this is default import
import { FileProgressCallback } from './download-file';  // this is named import
```
TS also support `type import`, try to consider this situation
```ts
import type downloadFile from './download-file';
import type { FileProgressCallback } from './download-file';

// there is no way to merge this two import statement
// TS syntax doesn't support to merge like this:
import type downloadFile, type { FileProgressCallback } from './download-file';
// also TS syntax doesn't support:
import type downloadFile, { FileProgressCallback } from './download-file';
import downloadFile, type { FileProgressCallback } from './download-file';
```

so I think we should take a unified approach to separate the `default import` and `named import`:
```ts
// value import 
import downloadFile from './download-file';
import { FileProgressCallback } from './download-file';
// type import 
import type downloadFile from './download-file';
import type { FileProgressCallback } from './download-file';
```

2. another reason is that for the implementation of `recast-import` tool, we've tried to merge the same import statement:
```ts
import { Vec2 } from './core/math';
import { cclegacy } from './core/index';
// should be merged to
import { Vec2, cclegacy  } from './core/index';
```
but for `default import` statement, we cannot merge them, try to consider this situation:
```ts
// we don't know how to merge this three import statement
import defaultName1 from './core/index';
import defaultName2 from './core/index';
import { cclegacy }from './core/index';

// TS syntax doesn't support:
import defaultName1, defaultName2, { cclegacy  } from './core/index';
```

So I think it's better that we don't try merging `default import` and `named import`, and keep them separated instead.

minggo: So should we use default export?

PPpro: yes we can use, it's actually a syntax sugar 
```ts
import defaultName from './core/index';
// equals to 
import { default as defaultName } from './core/index';
```

it's better not to use default export

minggo (approved):

PR #16298, opened 2023-09-21
optimize: move assert* interface into @base/debug module

PPpro:
Re: 
https://github.com/cocos/cocos-engine/issues/14293
https://github.com/cocos/cocos-engine/issues/12647

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


cocos/2d/renderer/batcher-2d.ts
```diff
@@ -25,11 +25,12 @@
 import { DEBUG, JSB } from 'internal:constants';
 import { cclegacy } from '@base/global';
 import { assert } from '@base/debug';
+import { assertIsTrue } from '@base/debug/internal';
```
minggo, line 28: Is this function used internally?

cocos/2d/renderer/render-data.ts
```diff
@@ -23,11 +23,12 @@
 */
 
 import { DEBUG, JSB } from 'internal:constants';
+import { assert } from '@base/debug';
+import { assertIsTrue } from '@base/debug/internal';
```
minggo, line 27: Why are `assert` and `assetIsTrue` in different modules?

PPpro: `assetIsTrue` is not public, I think it should be in internal export, they are not different modules, but different exprot condition

cocos/2d/renderer/batcher-2d.ts
PPpro: yes

cocos/2d/renderer/render-data.ts
minggo: Why can't it be public?

PPpro: I just wanna keep it internal, because we haven't export this interface ever before, 
And it's weird to provide this interface originally, because we have `assert` already, why we still provide `assertIsTrue`, I think it's an interface desigh issue.

minggo: If so, we should not use it internally too. Though i know it is nothing about this issue.

minggo (approved):

PR #16168, opened 2023-09-01
fix tween.sequence

zxx43:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

* [ ] needs automatic test cases check.
  > Manual trigger with `@cocos-robot run test cases` afterward.
* [ ] does not change any runtime related code or build configuration
  > If any reviewer thinks the CI checks are needed, please uncheck this option, then close and reopen the issue.

-------

### Compatibility Check

This pull request:

* [ ] changes public API, and have ensured backward compatibility with [deprecated features](https://github.com/cocos/cocos-engine/blob/v3.5.0/docs/contribution/deprecated-features.md).
* [ ] affects platform compatibility, e.g. system version, browser version, platform sdk version, platform toolchain, language version, hardware compatibility etc.
* [ ] affects file structure of the build package or build configuration which requires user project upgrade.
* [ ] **introduces breaking changes**, please list all changes, affected features and the scope of violation.

fix issue: https://github.com/cocos/3d-tasks/issues/17561

PR #16443, opened 2023-10-27
Fix tween.sequence not supporting single parameter

qiuguohua:
Re: #

### Changelog

https://github.com/cocos/3d-tasks/issues/17561
https://github.com/cocos/cocos-engine/pull/16168

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/tween/actions/action-interval.ts
+++ b/cocos/tween/actions/action-interval.ts
@@ -291,7 +291,7 @@ export class Sequence extends ActionInterval {
         const sequence = new Sequence();
         sequence.initWithTwoActions(actionOne, actionTwo);
         return sequence;
-    }
+    };
 
     private _actions: ActionInterval[] = [];
     private _split = 0;
--- a/cocos/tween/actions/action-interval.ts
+++ b/cocos/tween/actions/action-interval.ts
@@ -454,7 +454,7 @@ export class Sequence extends ActionInterval {
  * const seq = sequence(actArray);
  */
 // todo: It should be use new
-export function sequence (/* Multiple Arguments */tempArray: any): ActionInterval {
+export function sequence (/* Multiple Arguments */tempArray: any): ActionInterval | null {
     const paramArray = (tempArray instanceof Array) ? tempArray : arguments;
     if (paramArray.length === 1) {
         return paramArray[0] as ActionInterval;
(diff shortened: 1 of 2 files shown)


cocos/tween/actions/action-interval.ts
```diff
@@ -465,12 +467,12 @@ export function sequence (/* Multiple Arguments */tempArray: any): ActionInterva
...
         for (let i = 1; i <= last; i++) {
             if (paramArray[i]) {
-                result = Sequence._actionOneTwo(result, paramArray[i]);
+                result = Sequence._actionOneTwo(result as ActionInterval, paramArray[i] as ActionInterval);
             }
         }
     }
 
-    return result;
+    return result as ActionInterval;
```
minggo, line 475: Why not just  change the type of  `result` to `ActionInterval `?

qiuguohua: Because the input type is any, we need to assign the value.

minggo: Can define it as ActionInterval instead of any.

qiuguohua: The input parameter is of type array, and the elements in the array are any.
This has to do with the type of the input array, and this needs more modification.

minggo: What I mean is change 
```typescript
let result: any = null;
```
to 
```
let result: ActionInterval = null;
```

`result` is defined in this function, I think we can change it.

qiuguohua: ok

cocos/tween/actions/action-interval.ts
```diff
@@ -454,25 +454,27 @@ export class Sequence extends ActionInterval {
  * const seq = sequence(actArray);
  */
 // todo: It should be use new
-export function sequence (/* Multiple Arguments */tempArray: any): ActionInterval {
+export function sequence (/* Multiple Arguments */tempArray: any): ActionInterval | null {
```
minggo: It is better no change the return type. It may break compatibility.

qiuguohua: It may not work, because the result may be null. eslint will report an error.

The original meaning of this function was to not allow error conditions, and if an error condition occurs, it will be printed out using the logID

minggo: How about just return a new ActionInterval without any initialization instead of null?

qiuguohua: It feels like this doesn't identify if it's reasonable or illegal. If it's illegal, it's also used as a legal value

minggo: The original implementation also return an uninitialized object.

qiuguohua: Yeah, it feels reasonable to return null.

qiuguohua: If there is an error, I think it makes sense to return null to let the user know directly. This doesn't break compatibility if there is no error.

minggo: You can see the CI, it break compatibility. The game project may meet the same problem.

minggo: Or maybe we should revert the modification. Just use any as it is. Or it will create an useless object, which will add gc overhead.

qiuguohua: ok

minggo (approved):

PR #16612, opened 2023-12-15
Fix building error for simulator.

Y-way:


minggo:
It looks good to me. But i am curious how to reproduce it as we also built it too, but have't met any problem.

Y-way:
> It looks good to me. But i am curious how to reproduce it as we also built it too, but have't met any problem.

step 1. Create SimulatorApp solution by 'glup gen-simulator' at the 'cocos-engine/native' subdirectory.
step 2. Open the simulator with vs, then build the SimulatorApp.

qiuguohua (approved):

minggo (approved):

Issue #16514, opened 2023-11-17
UIOpacity does not work


qiuguohua:
### Cocos Creator version

> 3.6.3

### System information

native

### Issue description

UIOpacity does not work

### Steps to reproduce

https://forum.cocos.org/t/topic/153865

PR #16642, opened 2023-12-27
Fix uiopacity change when parent node changes.

qiuguohua:
Re: #

### Changelog

https://github.com/cocos/cocos-engine/issues/16514

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/2d/components/ui-opacity.ts
+++ b/cocos/2d/components/ui-opacity.ts
@@ -28,6 +28,7 @@ import { Component } from '../../scene-graph/component';
 import { misc } from '../../core';
 import { UIRenderer } from '../framework/ui-renderer';
 import { Node } from '../../scene-graph';
+import { NodeEventType } from '../../scene-graph/node-event';
 
 /**
  * @en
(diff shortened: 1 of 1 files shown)


cocos/2d/components/ui-opacity.ts
```diff
@@ -133,8 +134,32 @@ export class UIOpacity extends Component {
...
+        const uiOp = node.getComponent<UIOpacity>(UIOpacity);
+        if (uiOp) {
+            UIOpacity.setEntityLocalOpacityDirtyRecursively(node, true, 1, false);
+        }
+    }
+
     public onEnable (): void {
-        // If the ancestor node has a uiopacity component, it will be initialized when initializing 
+        this.node.on(NodeEventType.CHILD_ADDED, this._parentAdded, this);
+        this.node.on(NodeEventType.CHILD_REMOVED, this._parentRemoved, this);
```
zhefengzhang, line 161: Need to turn off the listener in the onDisable callback

qiuguohua: OK

cocos/2d/components/ui-opacity.ts
```diff
@@ -133,8 +134,32 @@ export class UIOpacity extends Component {
     @serializable
     protected _opacity = 255;
 
+    protected _parentAdded (node: Node): void {
+        const parent: Node |null = node.getParent();
+        if (parent == null) {
+            return;
+        }
```
minggo, line 141: Why should check it as it is a callback of CHILD_ADDED.

cocos/2d/components/ui-opacity.ts
```diff
@@ -133,8 +134,32 @@ export class UIOpacity extends Component {
...
+        const render = parent._uiProps.uiComp as UIRenderer;
+        if (render && render.color) {
+            return;
```
minggo, line 144: Why return here?

qiuguohua: If it is a renderable element, the child nodes do not need to respond to changes.

cocos/2d/components/ui-opacity.ts
qiuguohua: Because getParent may be null

minggo: It is called when a child is added, then why parent will be null?

qiuguohua:
Fixed in another pr ：https://github.com/cocos/cocos-engine/pull/16828

Issue #16661, opened 2024-01-04
Spine animtion can not paused from 3.8.2


yanjifa:
### Cocos Creator version

3.8.2

### System information

Mac OS Edge Browser

### Issue description

maybe cause by this
https://github.com/cocos/cocos-engine/blob/77de3b09ed8a61a63e20565978be9509bfc80e2b/cocos/spine/skeleton-system.ts#L81

https://github.com/cocos/cocos-engine/blob/348b7ed803c0f167857baf73226085ab40b48efe/cocos/spine/skeleton.ts#L1030

@bofeng-song 

### Steps to reproduce

empty

PR #16672, opened 2024-01-09
Fix spine can not be paused while working at realTime mode

bofeng-song:
Re: #
https://github.com/cocos/cocos-engine/issues/16661
https://github.com/cocos/cocos-engine-external/pull/450

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


cocos/spine/skeleton-cache.ts
```diff
@@ -302,7 +302,10 @@ export class AnimationCache {
     }
 
     public destroy (): void {
-        spine.SkeletonSystem.destroySpineInstance(this._instance);
+        if (this._instance) {
+            this._instance.destroy();
+            this._instance = null as any;
```
dumganhar: If it could be null, should we define its type with ` | null`?

native/cocos/editor-support/spine-wasm/spine-skeleton-instance.cpp
```diff
@@ -34,7 +33,6 @@ static void trackEntryCallback(AnimationState *state, EventType type, TrackEntry
 
 SpineSkeletonInstance::SpineSkeletonInstance() {
     _model = new SpineModel();
-    SpineSkeletonSystem::addSpineInstance(this);
```
dumganhar, line 37: Why delete SpineSkeletonSystem ?

bofeng-song: Its no need an additional helper class to release the spineSkeletonInstance object.

cocos/spine/skeleton-cache.ts
bofeng-song: If null is added to the declaration, all the places where it is used need to be checked for null, and there will be a bit more changes to be made.

dumganhar: Yes，but that‘s the correct way since it could be `null`.

dumganhar (approved):

minggo (approved):

PR #16686, opened 2024-01-12
Fix chunk as null causing editor exception.

qiuguohua:
Re: #

### Changelog

https://forum.cocos.org/t/topic/155516

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/2d/renderer/static-vb-accessor.ts
+++ b/cocos/2d/renderer/static-vb-accessor.ts
@@ -159,7 +159,7 @@ export class StaticVBAccessor extends BufferAccessor {
 
     public allocateChunk (vertexCount: number, indexCount: number): StaticVBChunk | null {
         const byteLength = vertexCount * this.vertexFormatBytes;
-        if (vertexCount <= this._vCount) {
+        if (vertexCount <= this._vCount || indexCount < this._iCount) {
             errorID(9004, byteLength);
             return null;
         }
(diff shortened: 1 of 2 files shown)


cocos/2d/renderer/static-vb-accessor.ts
```diff
@@ -178,7 +178,7 @@ export class StaticVBAccessor extends BufferAccessor {
             if (entry) break;
         }
         // Allocation fail
-        if (!entry) {
+        if (!entry && vertexCount <= this._vCount) {
```
minggo: I think should check it at the beginning of this function, for example
```typescript
public allocateChunk (vertexCount: number, indexCount: number): StaticVBChunk | null {
    if (vertexCount > this._vCount) {
        return null;
    }
    ...
}
```

qiuguohua: I think it's appropriate to put it here as well. Because most of the cases will not exceed the size, and only a very small part of the cases will have errors. Putting it at the top doesn't improve performance.

minggo: Add at top of the function will not affect performance for other situations but can improve performance in special cases, why not do it?

qiuguohua: It's a error that needs to be fixed. It currently needs to be fixed at the user, so performance is not really a concern in this case.

minggo: Is there any concern to add it at the begin of the function?

qiuguohua: ok

minggo (approved):

Issue #16487, opened 2023-11-08
[bug] Framerate will not be stable after switch to 60fps from 30fps


dumganhar:
It's reported here: https://forum.cocos.org/t/topic/154035

PR #16716, opened 2024-02-02
Reduce the amplitude of the frequency increase

qiuguohua:
Re: #
https://github.com/cocos/cocos-engine/issues/16487

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/pal/pacer/pacer-web.ts
+++ b/pal/pacer/pacer-web.ts
@@ -116,8 +116,8 @@ export class Pacer {
         if (!this._isPlaying) {
             return;
         }
-        const nowTime = performance.now();
-        const elapseTime = nowTime - this._startTime;
+        const currTime = performance.now();
+        const elapseTime = currTime - this._startTime;
         const elapseFrame = Math.floor(elapseTime / this._frameTime);
         if (elapseFrame < this._frameCount) {
             const sleepMs = this._frameCount * this._frameTime - elapseTime;
@@ -139,8 +139,8 @@ export class Pacer {
         if (this._callback) {
             this._callback();
         }
-        if (nowTime - this._startTime > FRAME_RESET_TIME) {
-            this._startTime = nowTime;
+        if (currTime - this._startTime > FRAME_RESET_TIME) {
+            this._startTime = currTime;
             this._frameCount = 1;
         }
         this._timeId = setTimeout(this._handleRAF, 0);
--- a/pal/pacer/pacer-web.ts
+++ b/pal/pacer/pacer-web.ts
@@ -30,6 +30,7 @@ const FRAME_RESET_TIME = 2000;
 
 export class Pacer {
     private _stHandle = 0;
+    private _timeId = 0;
     private _onTick: (() => void) | null = null;
     private _targetFrameRate = 60;
     private _frameTime = 0;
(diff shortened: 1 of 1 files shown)


pal/pacer/pacer-web.ts
```diff
@@ -106,25 +104,46 @@ export class Pacer {
...
-            this._frameCount = 0;
+        this._frameCount++;
+        if (this._callback) {
+            this._callback();
+        }
+        if (currTime - this._startTime > FRAME_RESET_TIME) {
+            this._startTime = currTime;
+            this._frameCount = 1;
         }
+        this._timeId = setTimeout(this._handleRAF, 0);
```
dumganhar: Why do we need another timer to trigger _handleRAF?
`RAF` means `requestAnimationFrame`, its callback is invoked every frame.

pal/pacer/pacer-web.ts
```diff
@@ -106,25 +104,46 @@ export class Pacer {
...
         }
-        if (performance.now() - this._startTime > FRAME_RESET_TIME) {
-            this._startTime = performance.now();
-            this._frameCount = 0;
+        this._frameCount++;
+        if (this._callback) {
+            this._callback();
+        }
+        if (currTime - this._startTime > FRAME_RESET_TIME) {
+            this._startTime = currTime;
```
dumganhar: Why do we need to reset `_startTime` and `_frameCount` in 2 seconds?

pal/pacer/pacer-web.ts
qiuguohua: If 'elapseFrame < 16ms' calling requestAnimationFrame may add an extra frame of time. This results in occasional 30ms frames

pal/pacer/pacer-web.ts
qiuguohua: My personal opinion may be to avoid adding frames all the time. And it also avoids error amplification because there is frame-filling logic.

pal/pacer/pacer-web.ts
```diff
@@ -106,25 +104,46 @@ export class Pacer {
...
         const elapseFrame = Math.floor(elapseTime / this._frameTime);
         if (elapseFrame < this._frameCount) {
-            this._rAF.call(window, this._handleRAF);
-        } else {
-            this._frameCount++;
-            if (this._callback) {
-                this._callback();
+            const sleepMs = this._frameCount * this._frameTime - elapseTime;
+            if (sleepMs > 16) {
+                this._stHandle = this._rAF.call(window, this._handleRAF);
```
dumganhar: If running on high frame rate devices, for example, 120FPS, the `requireAnimationFrame` interval is `8ms` instead of `16ms`. So  will `this._rAF.call` here be invoked twice ?

qiuguohua: I've tried searching for a way to get the screen refresh rate, but haven't found one yet. Therefore, at high resolution, use settimeout for delay

pal/pacer/pacer-web.ts
```diff
@@ -106,25 +104,46 @@ export class Pacer {
...
-        if (performance.now() - this._startTime > FRAME_RESET_TIME) {
-            this._startTime = performance.now();
-            this._frameCount = 0;
+        this._frameCount++;
+        if (this._callback) {
+            this._callback();
+        }
+        if (currTime - this._startTime > FRAME_RESET_TIME) {
+            this._startTime = currTime;
+            this._frameCount = 1;
```
dumganhar: Why reset the `_frameCount` to 1 rather than 0?

dumganhar (approved):

minggo (approved):

PR #16731, opened 2024-02-18
fix a few errors in VKCommandBuffer.cpp

jcyuan:
1, should be Access mask, not Image layout
2, the ASSERT is useless because it's already tested above in the `missed` value
3, same with item 1
4, `uint32_t memoryBarrierCount` should not be 0 when `pMemoryBarrier` is not null

### DIFF
--- a/native/cocos/renderer/gfx-vulkan/VKCommandBuffer.cpp
+++ b/native/cocos/renderer/gfx-vulkan/VKCommandBuffer.cpp
@@ -847,7 +847,7 @@ void CCVKCommandBuffer::pipelineBarrier(const GeneralBarrier *barrier, const Buf
                 bool fullBarrier = ccBarrier->getInfo().type == BarrierType::FULL;
                 bool missed = _barrierEvents.find(ccTexture) == _barrierEvents.end();
                 if (!fullBarrier && !missed) {
-                    //CC_ASSERT(_barrierEvents.find(ccTexture) != _barrierEvents.end());
+                    CC_ASSERT(_barrierEvents.find(ccTexture) != _barrierEvents.end());
                     VkEvent event = _barrierEvents.at(ccTexture);
                     scheduledEvents.push_back(event);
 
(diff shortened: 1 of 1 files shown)


native/cocos/renderer/gfx-vulkan/VKCommandBuffer.cpp
```diff
@@ -892,7 +892,7 @@ void CCVKCommandBuffer::pipelineBarrier(const GeneralBarrier *barrier, const Buf
                 bool fullBarrier = ccBarrier->getInfo().type == BarrierType::FULL;
                 bool missed = _barrierEvents.find(ccBuffer) != _barrierEvents.end();
                 if (!fullBarrier && !missed) {
-                    CC_ASSERT(_barrierEvents.find(ccBuffer) != _barrierEvents.end());
+                    // CC_ASSERT(_barrierEvents.find(ccBuffer) != _barrierEvents.end());
```
star-e, line 895: It looks like the `CC_ASSERT` is needed. If `missed` is `false`, it means `fullBarrier` must be true to avoid the failing CC_ASSERT.

native/cocos/renderer/gfx-vulkan/VKCommandBuffer.cpp
```diff
@@ -965,8 +965,8 @@ void CCVKCommandBuffer::pipelineBarrier(const GeneralBarrier *barrier, const Buf
         }
 
         if (!fullBufferBarriers.empty() || !fullImageBarriers.empty()) {
-            vkCmdPipelineBarrier(_gpuCommandBuffer->vkCommandBuffer, fullSrcStageMask, fullDstStageMask, 0, 0, pMemoryBarrier,
-                                 fullBufferBarriers.size(), fullBufferBarriers.data(), fullImageBarriers.size(), fullImageBarriers.data());
+            vkCmdPipelineBarrier(_gpuCommandBuffer->vkCommandBuffer, fullSrcStageMask, fullDstStageMask, 0, pMemoryBarrier ? 1 : 0,
+                                 pMemoryBarrier, fullBufferBarriers.size(), fullBufferBarriers.data(), fullImageBarriers.size(), fullImageBarriers.data());
```
star-e, line 969: Shall we change the `if (!fullBufferBarriers.empty() || !fullImageBarriers.empty())` to `if (!fullBufferBarriers.empty() || !fullImageBarriers.empty() || pMemoryBarrier)` ?

`pMemoryBarrier` might not be executed.

native/cocos/renderer/gfx-vulkan/VKCommandBuffer.cpp
jcyuan: ah, i got it, sorry about that. reverted

native/cocos/renderer/gfx-vulkan/VKCommandBuffer.cpp
jcyuan: personally i think it's fine, because `memoryBarrierCount` arg for `vkCmdPipelineBarrier` decides it will be used or not.

PR #16732, opened 2024-02-18
fix a few errors in VKCommands.cpp

jcyuan:
1, removed the wrong `return`, because we still need to retry with the default way to allocate
2, just typo: `depthSetncil` -> `depthStencil`
3, clear: useless line


minggo (approved):

native/cocos/renderer/gfx-vulkan/VKCommands.cpp
```diff
@@ -159,7 +159,6 @@ void cmdFuncCCVKCreateTexture(CCVKDevice *device, CCVKGPUTexture *gpuTexture) {
                                              pVkImage, pVmaAllocation, &res);
             if (!result) {
                 gpuTexture->memoryAllocated = false;
-                return;
```
star-e: Here we use TBR memoryless texture (memory on chip, or subpass attachment without load/store) and do not allocate memory. I'm not sure the change is correct or not.

jcyuan: i got it~ yes this is a point. maybe we can have some strategy pattern or macro or whatever to make a improve on this? like, on WIN, it's just fine with the 'retry' way.
anyway i have just reverted this change, now about other 2 changes i think it's safe to be merged.

star-e (approved):

star-e:
@cocos-robot run test cases

PR #16733, opened 2024-02-18
fix the wrong condition in VKDevice.cpp

jcyuan:
the wrong conditions:
there are 2 lines above the last check line `if (swapchain->gpuSwapchain()) {`, so there will probably be null pointer access exception caused.

### DIFF
--- a/native/cocos/renderer/gfx-vulkan/VKDevice.cpp
+++ b/native/cocos/renderer/gfx-vulkan/VKDevice.cpp
@@ -620,6 +620,9 @@ void CCVKDevice::acquire(Swapchain *const *swapchains, uint32_t count) {
     vkPresentBarriers.resize(count, presentBarrier);
     for (uint32_t i = 0U; i < count; ++i) {
         auto *swapchain = static_cast<CCVKSwapchain *>(swapchains[i]);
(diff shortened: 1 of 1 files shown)


minggo (approved):

native/cocos/renderer/gfx-vulkan/VKDevice.cpp
```diff
@@ -620,6 +620,10 @@ void CCVKDevice::acquire(Swapchain *const *swapchains, uint32_t count) {
     vkPresentBarriers.resize(count, presentBarrier);
     for (uint32_t i = 0U; i < count; ++i) {
         auto *swapchain = static_cast<CCVKSwapchain *>(swapchains[i]);
+        if (!swapchain->gpuSwapchain()) {
+            continue;
+        }
```
star-e: We might change this guard to `CC_EXPECTS(swapchain->gpuSwapchain());`. I believe it is always valid.

jcyuan: yes~ agree with it!
change pulled

star-e:
@cocos-robot run test cases

PR #16753, opened 2024-02-23
Fix label inactivation without calling destroy to release resources

qiuguohua:
Re: #

### Changelog

https://github.com/cocos/cocos-engine/issues/16721

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/2d/components/label.ts
+++ b/cocos/2d/components/label.ts
@@ -39,10 +39,8 @@ import { BlendFactor } from '../../gfx';
 import { TextStyle } from '../assembler/label/text-style';
 import { TextLayout } from '../assembler/label/text-layout';
 import { TextOutputLayoutData, TextOutputRenderData } from '../assembler/label/text-output-data';
-import { CCObject } from '../../core/data/object';
 
 const tempColor = Color.WHITE.clone();
-const IsOnLoadCalled = CCObject.Flags.IsOnLoadCalled;
 /**
  * @en Enum for horizontal text alignment.
  *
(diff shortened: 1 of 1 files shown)


minggo:
I don't understand why it can fix the issue.

qiuguohua:
> I don't understand why it can fix the issue.

super._onpredestroy checks to see if _isOnLoadCalled is included, and calls ondestroy if it is.

Here we rewrite _onpredestroy to determine if it contains _isOnLoadCalled, if so, we go directly to the original logic, if not, we release the resources that have already been applied.

minggo:
Got it, i think it is better to add `override` for the function. And please fix the CI error.

minggo (approved):

huhudage:
After apply this merge request, if the Label component was destroyed before node's first activation, the Label component will not be removed from node, because super._onPreDestroy() will not be called.

Or is the following a better way?

```
    public _onPreDestroy (): void {
        super._onPreDestroy();

        if (this._isOnLoadCalled) {
            return;
        }
        // If _objFlags does not contain IsOnLoadCalled, it is possible to destroy the ttfSpriteFrame.
        this.destroyTtfSpriteFrame();
    }
```

qiuguohua:
> After apply this merge request, if the Label component was destroyed before node's first activation, the Label component will not be removed from node, because super._onPreDestroy() will not be called.
> 
> Or is the following a better way?
> 
> ```
>     public _onPreDestroy (): void {
>         super._onPreDestroy();
> 
>         if (this._isOnLoadCalled) {
>             return;
>         }
>         // If _objFlags does not contain IsOnLoadCalled, it is possible to destroy the ttfSpriteFrame.
>         this.destroyTtfSpriteFrame();
>     }
> ```

Yes, you're right. I'll fix it later

PR #16766, opened 2024-03-04
[Add] raytracing data preparation

GengineJS:


native/cocos/scene/raytracing/Def.h
```diff
@@ -0,0 +1,56 @@
...
+
+#include <cstdint>
+namespace cc {
+namespace scene {
+namespace raytracing {
+struct InstanceInfo {
+    uint32_t mMeshPrimID = 0;
+    uint32_t mTransformID = 0;
+    uint32_t mMeshID = 0;
+    uint32_t mMeshPrimitiveLocalID = 0;
```
star-e: Please use cocos naming convention.

star-e (approved):

native/cocos/scene/raytracing/Description.cpp
```diff
@@ -0,0 +1,27 @@
+#include "Description.h"
```
star-e, line 21: add license

native/cocos/scene/raytracing/RayTracing.h
```diff
@@ -0,0 +1,72 @@
...
+namespace scene {
+    class RenderScene;
+namespace raytracing {
+
+class RayTracing : public RefCounted {
+public:
+    explicit RayTracing(RenderScene* scene);
+    ~RayTracing() override = default;
+    // getter
+    static inline const ccstd::vector<Mat4>& getTransforms()  { return description->transforms; }
```
star-e: Here `inline` is unnecessary in class definition and can be removed. Function definition in class body is `inline` by default. `inline` in c++ means external linkage and can be used to avoid ODR (one definition rule). https://stackoverflow.com/questions/4193639/inline-function-linkage

PR #16771, opened 2024-03-06
Supplementing spriteframe's clone function

qiuguohua:
https://github.com/cocos/cocos-engine/issues/16744

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/2d/assets/sprite-frame.ts
+++ b/cocos/2d/assets/sprite-frame.ts
@@ -1439,14 +1439,6 @@ export class SpriteFrame extends Asset {
         sp._texture = this._texture;
         sp._isFlipUVX = this._isFlipUVX;
         sp._isFlipUVY = this._isFlipUVY;
-        sp._pixelsToUnit = this._pixelsToUnit;
-        sp._pivot.set(this._pivot);
-        sp._meshType = this._meshType;
-        sp._packable = this._packable;
-        sp._pixelsToUnit = this._pixelsToUnit;
-        sp._meshType = this._meshType;
-        sp._extrude = this._extrude;
-        sp._customOutLine.splice(0, sp._customOutLine.length, ...this._customOutLine);
         if (this._original) {
             sp._original = {
                 _texture: this._original._texture,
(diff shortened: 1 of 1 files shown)


cocos/2d/assets/sprite-frame.ts
```diff
@@ -1438,9 +1439,24 @@ export class SpriteFrame extends Asset {
...
+        } else {
+            sp._original = null;
+        }
+        sp._packable = this._packable;
         sp._pixelsToUnit = this._pixelsToUnit;
         sp._pivot.set(this._pivot);
         sp._meshType = this._meshType;
+        sp._extrude = this._extrude;
+        sp._customOutLine.splice(0, sp._customOutLine.length, ...this._customOutLine);
+        sp._mesh = this._mesh;
```
qiuguohua: I'm not sure it's needed here. mesh currently doesn't implement the clone function.

minggo: It is not a clone because they share the same mesh. What i think clone is a deep copy.

cocos/2d/assets/sprite-frame.ts
```diff
@@ -1438,9 +1439,28 @@ export class SpriteFrame extends Asset {
...
         sp._pivot.set(this._pivot);
         sp._meshType = this._meshType;
+        sp._extrude = this._extrude;
+        sp._customOutLine.splice(0, sp._customOutLine.length, ...this._customOutLine);
+        sp._minPos = this._minPos;
+        sp._maxPos = this._maxPos;
+        if (this._mesh) {
+            // Creates a new mesh, and 'this' creates the mesh in the same way. So we can make a copy like this.
+            // It must be placed last because the mesh will depend on some of its members when it is created.
+            sp.ensureMeshData();
```
minggo: The logic is not correct, `ensureMeshData ` will invoke `_initVertices` which will reset vertices to default values.

qiuguohua: It doesn't matter, the mesh and vertices are a combination. If this's mesh was generated by calling ensureMeshData, then sp should be the same in this way. The vertices and mesh will end up being the same, and it won't happen that the mesh is generated and then the vertices are modified.

qiuguohua: However, createmesh can be called to speed up the creation process

minggo (approved):

PR #16783, opened 2024-03-09
check max mipmap level for texture creation

jcyuan:
check max level count
actually there is already a function `getLevelCount` in the `GFXTexture.cpp` to calculate the max count, but i didn't move this function to somewhere for common use because not sure where should i put it.

### DIFF
--- a/native/cocos/renderer/gfx-validator/TextureValidator.cpp
+++ b/native/cocos/renderer/gfx-validator/TextureValidator.cpp
@@ -104,7 +104,8 @@ void TextureValidator::doInit(const TextureInfo &info) {
     }
 
     if (hasFlag(info.flags, TextureFlagBit::GEN_MIPMAP)) {
-        CC_ASSERT(info.levelCount > 1);
+        auto maxLevelCount = std::floor(std::log2(std::max({ info.width, info.height, info.depth }))) + 1;
+        CC_ASSERT(info.levelCount > 1 && info.levelCount <= maxLevelCount);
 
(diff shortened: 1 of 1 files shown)


native/cocos/renderer/gfx-validator/TextureValidator.cpp
```diff
@@ -104,7 +104,8 @@ void TextureValidator::doInit(const TextureInfo &info) {
     }
 
     if (hasFlag(info.flags, TextureFlagBit::GEN_MIPMAP)) {
-        CC_ASSERT(info.levelCount > 1);
+        auto maxLevelCount = std::floor(std::log2(std::max({ info.width, info.height, info.depth }))) + 1;
```
minggo, line 107: I think don't have to `+1` for some situations, for example `std::max({ info.width, info.height, info.depth }) == 8`.

jcyuan: check here: https://github.com/cocos/cocos-engine/blob/cde48d1750c7f7579fa40d506f963ea6ed8f9130/native/cocos/renderer/gfx-base/GFXTexture.cpp#L84-L86
and this method actually could be moved to somewhere for common use.

minggo: Got it, I think it is better to use _actor->getLevelCount() to get the value. And we may fix the logic of  _actor->getLevelCount() in future.

PR #16802, opened 2024-03-14
Added unit tests

qiuguohua:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/tests/ui/sprite-frame.test.ts
+++ b/tests/ui/sprite-frame.test.ts
@@ -1,0 +1,88 @@
+import { SpriteFrame } from "../../cocos/2d";
+import { Texture2D } from "../../cocos/asset/assets";
+import { Rect, Size, Vec2 } from "../../exports/base";
+
+test('sprite-frame.clone', () => {
+    let sp = new SpriteFrame;
+
+    sp.insetTop = 100;
+    expect(sp.insetTop).toStrictEqual(100);
+    sp.insetBottom = 200;
+    expect(sp.insetBottom).toStrictEqual(200);
+    sp.insetLeft = 300;
+    expect(sp.insetLeft).toStrictEqual(300);
+    sp.insetRight = 400;
+    expect(sp.insetRight).toStrictEqual(400);
+
+    let rc = new Rect(0, 0, 100, 100);
+    sp.rect = rc;
+    expect(sp.rect).toEqual(rc);
+
+    let sz = new Size(100, 100);
+    sp.originalSize = sz
+    expect(sp.originalSize).toEqual(sz);
+
+    let offset = new Vec2(100, 200);
+    sp.offset = offset;
+    expect(sp.offset).toEqual(offset);
+
+    sp.rotated = true;
+    expect(sp.rotated).toEqual(true);
+
+    let tex = new Texture2D();
+    sp.texture = tex;
+    expect(sp.texture).toEqual(tex);
+
+    let atlasUuid = '123456787abcd';
+    sp.atlasUuid = atlasUuid;
+    expect(sp.atlasUuid).toStrictEqual(atlasUuid);
+
+    expect(sp.width).toEqual(tex.width);
+    expect(sp.height).toEqual(tex.height);
+
+    sp.flipUVX = true;
+    sp.flipUVY = false;
+    expect(sp.flipUVX).toEqual(true);
+    expect(sp.flipUVY).toEqual(false);
+
+    sp.packable = true;
+    expect(sp.packable).toEqual(true);
+
+    // default value
+    expect(sp.pixelsToUnit).toEqual(100);
+    expect(sp.pivot).toEqual(new Vec2(0.5, 0.5));
+    
+    sp.setRotated(false);
+    expect(sp.rotated).toEqual(false);
+    // Make sure the mesh is created
+    sp.ensureMeshData();
+    
+    let sp_clone = sp.clone();
+    expect(sp_clone.insetTop).toStrictEqual(sp.insetTop);
+    expect(sp_clone.insetBottom).toStrictEqual(sp.insetBottom);
+    expect(sp_clone.insetLeft).toStrictEqual(sp.insetLeft);
+    expect(sp_clone.insetRight).toStrictEqual(sp.insetRight);
+    expect(sp_clone.rect).toEqual(sp.rect);
+    expect(sp_clone.originalSize).toEqual(sp.originalSize);
+    expect(sp_clone.offset).toEqual(sp.offset);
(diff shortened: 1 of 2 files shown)


tests/ui/sprite-frame.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    let sp = new SpriteFrame;
+
+    sp.insetTop = 100;
+    expect(sp.insetTop).toStrictEqual(100);
+    sp.insetBottom = 200;
+    expect(sp.insetBottom).toStrictEqual(200);
+    sp.insetLeft = 300;
+    expect(sp.insetLeft).toStrictEqual(300);
+    sp.insetRight = 400;
+    expect(sp.insetRight).toStrictEqual(400);
```
minggo, line 14: Why need to check these functions?

tests/ui/sprite-frame.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    expect(sp_clone.pixelsToUnit).toEqual(sp.pixelsToUnit);
+    expect(sp_clone.pivot).toEqual(sp.pivot);
+    expect(sp_clone.mesh).toEqual(sp.mesh);
+    expect(sp_clone.uv).toEqual(sp.uv);
+    expect(sp_clone.unbiasUV).toEqual(sp.unbiasUV);
+    expect(sp_clone.uvSliced).toEqual(sp.uvSliced);
+    expect(sp_clone.vertices).toEqual(sp.vertices);
+    expect(sp_clone.original).toEqual(sp.original);
+    expect(sp_clone.trimmedBorder).toEqual(sp.trimmedBorder);
```
minggo, line 136: I think should also make sure the object variables should refer to different object. For example
```typescript
sp_clone.offset.x = sp_clone.offset.x + 1;
// sp_clone.offset should not be equal to sp.offset. Then we can make sure sp_clone.offset and sp.offset are different objects.
```

tests/ui/sprite-frame.test.ts
qiuguohua: '_capInsets' is a protected member and does not have an interface to get. So these functions can be tested indirectly.

tests/ui/sprite-frame.test.ts
qiuguohua: Ok

tests/ui/sprite-frame.test.ts
minggo: If you want to test other things, i think you should do in another function. This function just test clone.

qiuguohua: This test needs to be set. If it is not set it will be the default value, and the default value is no guarantee that it is a true copy. 
Checking that it is set correctly and testing the get function can be merged together. Otherwise there will be duplicate work

minggo: Yep, they need to be set, but don't have to be test here. Maybe you should write other test cases to test the value is set correctly.

qiuguohua: Ok

tests/ui/sprite-frame.test.ts
```diff
@@ -78,11 +97,55 @@ test('spriteframe.clone', () => {
...
+    sp_clone.vertices?.indexes.splice(0,  sp_clone.vertices?.indexes.length);
+    expect(sp_clone.vertices?.indexes !== sp.vertices?.indexes).toEqual(true);
+    sp_clone.vertices?.uv.splice(0,  sp_clone.vertices?.uv.length);
+    expect(sp_clone.vertices?.uv !== sp.vertices?.uv).toEqual(true);
+    sp_clone.vertices?.nuv.splice(0,  sp_clone.vertices?.nuv.length);
+    expect(sp_clone.vertices?.nuv !== sp.vertices?.nuv).toEqual(true);
+    sp_clone.vertices?.minPos.add3f(100,200,300);
+    expect(sp_clone.vertices?.minPos !== sp.vertices?.minPos).toEqual(true);
+    sp_clone.vertices?.maxPos.add3f(100,200,300);
+    expect(sp_clone.vertices?.maxPos !== sp.vertices?.maxPos).toEqual(true);;
```
minggo, line 133: I think can treat vertices as a whole. For example, just modify `vertices.rawPosition`, then check vertices again. Don't have to modify every element of vertices.

tests/ui/sprite-frame.test.ts
```diff
@@ -78,11 +97,55 @@ test('spriteframe.clone', () => {
...
     expect(sp_clone.original).toEqual(sp.original);
+
     expect(sp_clone.trimmedBorder).toEqual(sp.trimmedBorder);
+    sp_clone.trimmedBorder.add4f(100,200,300,400);
+    expect(sp_clone.trimmedBorder !== sp.trimmedBorder).toEqual(true);
+    
+    expect(sp_clone.insetTop).toStrictEqual(sp.insetTop);
+    // The modifications here will modify the other test data, so it needs to be placed backward.
+    sp_clone.insetTop += 1;
```
minggo: It is a number, so i don't think need to change its value and check again.

tests/ui/sprite-frame.test.ts
```diff
@@ -78,11 +97,55 @@ test('spriteframe.clone', () => {
...
+    expect(sp_clone.insetTop !== sp.insetTop).toEqual(true);
+    expect(sp_clone.insetBottom).toStrictEqual(sp.insetBottom);
+    expect(sp_clone.insetLeft).toStrictEqual(sp.insetLeft);
+    expect(sp_clone.insetRight).toStrictEqual(sp.insetRight);
+
+    expect(sp_clone.texture).toEqual(sp.texture);
+    sp_clone.texture = new Texture2D();
```
minggo: It creates a new texture, i must be different, so i don't think need to do it like this.

tests/ui/sprite-frame.test.ts
qiuguohua: The vertexes are copied separately, and not all copies are necessarily normal.

tests/ui/sprite-frame.test.ts
qiuguohua: OK

tests/ui/sprite-frame.test.ts
qiuguohua: It seems like the textures are shared, this was the previous design

minggo: Yep, then don't have to test it.

qiuguohua: OK

minggo (approved):

PR #16828, opened 2024-03-27
Optimize uiopacity component

qiuguohua:
Re: #

### Changelog

https://github.com/cocos/cocos-engine/issues/16514

document:
https://github.com/cocos/3d-tasks/discussions/18155

test-case:
https://github.com/cocos/cocos-test-projects/pull/857

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/2d/components/ui-opacity.ts
+++ b/cocos/2d/components/ui-opacity.ts
@@ -108,15 +108,15 @@ export class UIOpacity extends Component {
      *             @zh 递归的节点。
      * @param node @en Is the color dirty.
      *             @zh color是否dirty。
-     * @param interruptParentOpacity @en The parent node's opacity.
-     *                               @zh 父节点的opacity。
-     * @param recursiveWithoutUiOp @en Components that just recursively do not have uiopacity and RenderColor.
-     *                             @zh 仅仅递归没有uiopacity和RenderEntity的组件。
+     * @param parentOpacity @en The parent node's opacity.
+     *                      @zh 父节点的opacity。
+     * @param recursiveWithoutUiOp @en Recursing only nodes without uiopacity.
+     *                             @zh 仅仅递归没有uiopacity的组件。
      */
     public static setEntityLocalOpacityDirtyRecursively (
         node: Node,
         dirty: boolean,
-        interruptParentOpacity: number,
+        parentOpacity: number,
         recursiveWithoutUiOp: boolean,
     ): void {
         if (!node.isValid) {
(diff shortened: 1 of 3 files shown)


cocos/2d/components/ui-opacity.ts
```diff
@@ -106,50 +127,93 @@ export class UIOpacity extends Component {
 
         const render = node._uiProps.uiComp as UIRenderer;
         const uiOp = node.getComponent<UIOpacity>(UIOpacity);
-        let interruptOpacity = interruptParentOpacity;// if there is no UIOpacity component, it should always equal to 1.
 
+        if (recursiveWithoutUiOp && uiOp) {
```
minggo: `recursiveWithoutUiOp ` means recursing only nodes without `uiopacity` component,  and `uiOp` means it has a `uiopacity` component. It seems conflict here.

cocos/2d/components/ui-opacity.ts
```diff
@@ -96,8 +97,28 @@ export class UIOpacity extends Component {
...
+     * @zh
+     * 递归设置localopacity。
+     *
+     * @param node @en recursive node.
+     *             @zh 递归的节点。
+     * @param node @en Is the color dirty.
+     *             @zh color是否dirty。
+     * @param parentOpacity @en The parent node's opacity.
+     *                      @zh 父节点的opacity。
+     * @param recursiveWithoutUiOp @en Recursing only nodes without uiopacity.
```
minggo: without uiopacity -> without `UIOpacity` component.

cocos/2d/components/ui-opacity.ts
```diff
@@ -96,8 +97,28 @@ export class UIOpacity extends Component {
...
+     *                             @zh 仅仅递归没有uiopacity的组件。
```
minggo: uiopacity -> UIOpacity

cocos/2d/components/ui-opacity.ts
```diff
@@ -106,50 +127,93 @@ export class UIOpacity extends Component {
...
+            this._parentOpacity = this._getParentOpacity(parent);
+            opacity = this._parentOpacity;
+        }
+        UIOpacity.setEntityLocalOpacityDirtyRecursively(this.node, true, opacity, false);
+    }
+
+    protected _setEntityLocalOpacityRecursively (opacity: number): void {
+        if (!JSB) {
+            return;
+        }
```
minggo, line 195: Why return if not JSB?

cocos/2d/components/ui-opacity.ts
qiuguohua: 'uiOp ! = null' is indicating that the UiOpacity component exists. ‘recursiveWithoutUiOp‘ indicates that it doesn't need to exist, so both conditions are met when handling the situation.

cocos/2d/components/ui-opacity.ts
qiuguohua: ok

cocos/2d/components/ui-opacity.ts
qiuguohua: This is only necessary for native

minggo: Could you please add some comment why only need in native?

qiuguohua: Because JSB's localOpacity value is present in the renderEntity, but non-JSB's are not.

kaxifakl:
I think change 'walk' function in Batcher2D.cpp to the same logic in Batcher2D.ts is a good solution. 
Get opacity value from parent and just bind value with jsb on native.

qiuguohua:
> I think change 'walk' function in Batcher2D.cpp to the same logic in Batcher2D.ts is a good solution.
Get opacity value from parent and just bind value with jsb on native.

Yes. This is a very good idea that can be moved on to the next stage.

minggo (approved):

PR #16842, opened 2024-03-28
multi-fingers touch bug fix

zhaoyang57:


cocos2d/core/platform/CCInputManager.js
```diff
@@ -217,13 +217,19 @@ let inputManager = {
                 ccTouch._setPrevPoint(selTouch._prevPoint);
                 ccTouch._lastModified = now;
                 handleTouches.push(ccTouch);
+
+                // event will be distributed multiple times on opharmony platform, requires filtering by windowId
```
qiuguohua: opharmony to openharmony?

cocos2d/core/platform/CCInputManager.js
```diff
@@ -217,13 +217,19 @@ let inputManager = {
...
+                if(globalThis.oh && selTouch.windowId == selTouch._id) {
```
qiuguohua: Why can't we do this in eventmanager?

cocos2d/core/event-manager/CCTouch.js
```diff
@@ -35,6 +35,7 @@
  */
 cc.Touch = function (x, y, id) {
     this._lastModified = 0;
+    this.windowId = undefined;
```
qiuguohua: Is it possible to make this variable private?

minggo:
Could you please describe:
- what's the issue
- how did you fix it

cocos2d/core/platform/CCInputManager.js
zhaoyang57: done

cocos2d/core/platform/CCInputManager.js
minggo: Is it a real window id or just a tag to distinguish touches? If it is a tag, then please use a more meaningful name.

zhaoyang57: > Why can't we do this in eventmanager?

Because windowId cannot be obtained in eventmanager.

cocos2d/core/event-manager/CCTouch.js
zhaoyang57: Why need to add private tag in js code?

qiuguohua: Just like _lastModified, it is private and obtains the value through a function. We agree that variables starting with an underscore are private

cocos2d/core/platform/CCInputManager.js
zhaoyang57: > Is it a real window id or just a tag to distinguish touches? If it is a tag, then please use a more meaningful name.

 it is a tag. I have modified the name already.

cocos2d/core/event-manager/CCTouch.js
zhaoyang57: > Just like _lastModified, it is private and obtains the value through a function. We agree that variables starting with an underscore are private

ok, I have modified the name to _currentTouchId

qiuguohua: If it is private, should the value be obtained through member methods?

zhaoyang57: > If it is private, should the value be obtained through member methods?

fixed.

qiuguohua (approved):

minggo (approved):

Issue #16266, opened 2023-09-15
Transparent renderers stop rendering when resolution changes


hulcypdiscord:
### Cocos Creator version

3.8

### System information

WebGL 2.0

### Issue description

When Post Processing is enabled for the Main Camera, if the Canvas size changes transparent renderers stop rendering.

### Steps to reproduce

1. Create New Project
2. Add default particle system in view of the camera
3. Enable the Custom Render Pipeline
4. Set the CUSTOM_PIPELINE_NAME to Custom
5. Attach FXAA to the Main Camera and enable Post Processing on the Main Camera
6. Play in browser
7. Switch between different resolutions using the picker in the top left
8. Notice after swapping between a couple resolutions the particles stop rendering

### Minimal reproduction project

[transparency-broken-repro.zip](https://github.com/cocos/cocos-engine/files/12621995/transparency-broken-repro.zip)

pandamicro:
Should be fixed in https://github.com/cocos/cocos-engine/pull/16157

star-e:
> Should be fixed in #16157

@GengineJS is backporting it to 3.8.1

GengineJS:
https://github.com/cocos/cocos-engine/pull/16321

hulcypdiscord:
@pandamicro Transparency no longer breaks using the repro steps I outlined but now the Frame Buffer is just stuck with the contents of the last frame before the window is resized sometimes. So end user experience is the game looks "frozen" until you resize the window again.

hulcypdiscord:
Bump. My project cannot upgrade to 3.8.1 because the new issue is worse than the previous one.

GengineJS:
@hulcypdiscord  What does the new question refer to? Is it still a problem that transparent objects need to be restored by resizing the window?

GengineJS:
Is it convenient to post a recorded video or a simple demo for troubleshooting?

hulcypdiscord:
https://github.com/cocos/cocos-engine/assets/75751645/1dfbe74d-df85-4cf3-8811-a21881d89dfa

Here is an attached video using the same sample project provided in this issue reproducing the new issue.

GengineJS:
OK, received, we will provide a solution as soon as possible

GengineJS:
Related questions in the video have been merged into PR   https://github.com/cocos/cocos-engine/pull/16524

hulcypdiscord:
[...]

PR #16911, opened 2024-04-18
Resolve Transparent Rendering Breaking when changing screen Resolution

hulcypdiscord:
Resolves this issue: https://github.com/cocos/cocos-engine/issues/16266

The above issue was partially fixed by detecting that the screen size changed indicating the frame buffer resources needed to be rebuilt. However the specific issue of Transparency breaking is due to the fact that the Transparent Render Pass for the Post Process pipeline references the same Texture resources used by the Opaque Render Pass. Each Render Pass owns its own Framebuffer but the underlying Texture resources are shared. When the above issue occurs it is a result of the Transparent Render Pass frame buffer no longer referencing the same Texture resources as the Opaque Render Pass.

1. Initial screen resolution set and Render Pass objects initialized
2. Transparent Frame Buffer is initialized and points to the same Texture resource as the Opaque Render Pass
3. Change screen resolution
4. New Render Pass objects are created keyed off the render params used to generate the hashId
5. Everything works fine
6. Change screen resolution back to the original size
7. The Render Pass objects already exist so call resetResrouces
8. Opaque Render Pass `resetResources` called and detects the screen resolution has changed so it rebuilds it's frame buffer object
9. This triggers calls to `Texture.resize` on the color and depth textures associated with the Render Pass
10. Calling `resize` destroys the old webgl textures and creates new ones that the Opaque frame buffer binds to when it is recreated
11. Transparent Render Pass `resetResources` called but because it shares the same Textures with the Opaque Render Pass it does not detect that any resolution changes have been made so it does not rebuild it's frame buffer, resulting in the Transparent Render Pass frame buffer still binding to old webgl texture handles that are different than the ones bound to the Opaque Render Pass.

I am not certain that my solution here is the best way to solve this problem but it has resolved the issue for my project and does properly showcase what the issue is.

### Changelog

* 

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/gfx/base/framebuffer.ts
+++ b/cocos/gfx/base/framebuffer.ts
@@ -73,6 +73,10 @@ export abstract class Framebuffer extends GFXObject {
         return this._height;
     }
 
+    public get needsRebuild(): boolean {
+        return false;
+    }
+
     protected _renderPass: RenderPass | null = null;
     protected _colorTextures: (Texture | null)[] = [];
     protected _depthStencilTexture: Texture | null = null;
--- a/cocos/gfx/webgl/webgl-framebuffer.ts
+++ b/cocos/gfx/webgl/webgl-framebuffer.ts
@@ -29,7 +29,7 @@ import { WebGLDeviceManager } from './webgl-define';
 import { IWebGLGPUFramebuffer, IWebGLGPUTexture } from './webgl-gpu-objects';
 import { WebGLRenderPass } from './webgl-render-pass';
 
-import { WebGLTexture } from './webgl-texture';
+import { WebGLTexture as ccWebGLTexture } from './webgl-texture';
 
 export class WebGLFramebuffer extends Framebuffer {
     get gpuFramebuffer (): IWebGLGPUFramebuffer {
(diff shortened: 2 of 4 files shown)


minggo:
@hulcypdiscord please fix warnings.

star-e:
Thanks for your pr! It helps a lot for finding the bug.
Currently we are figuring out how to solve this problem, and it might affect more places than framebuffer.

hulcypdiscord:
Yeah, I honestly did not expect y'all to necessarily merge this PR but it felt like the simplest way to surface this issue as this was the solution I came up with that has solved our use case in the short term.

GengineJS:
@hulcypdiscord  We made some minor adjustments based on the pull request (PR) you submitted. Thank you for your contribution, and we would also appreciate it if you could review the changes.

https://github.com/cocos/cocos-engine/pull/16927

hulcypdiscord:
Looks good to me! Thanks!

minggo:
@GengineJS please fix the conflict.

hulcypdiscord:
Closing this PR as the issue has been resolved with the above mentioned PR by GengineJS

PR #16912, opened 2024-04-18
Fix for particle system render bounds center getting out of sync

hulcypdiscord:
I discovered this issue when trying to enable render culling for Particle Systems in our game to improve performance. I noticed that Particle Systems were being culled while they were clearly visible on screen. It turns out the issue is that the Particle System render culling bounding box center is not recomputed every frame, it merely has a movement offset applied to the current bounding box center. This can result in the bounding box center being drastically offset from the actually location of the particle system if the particle system is moved while the Node is disabled. For our situation this was due to us  making use of a Pooling system to re-use one shot particle effects.

This change clears the _oldPos value used when computing the offset delta to be applied to the bounding box so that it is reset to the current position of the particle system when it is re-enabled and the bounding box is re-computed.

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/particle/particle-system.ts
+++ b/cocos/particle/particle-system.ts
@@ -1108,9 +1108,9 @@ export class ParticleSystem extends ModelRenderer {
         this.processor.onDisable();
         if (this._trailModule) this._trailModule.onDisable();
         if (this._boundingBox) {
-            this._boundingBox = null;
-            this._oldPos = null;
+            this._boundingBox = null;            
         }
+        this._oldPos = null;
         if (this._culler) {
             this._culler.clear();
             this._culler.destroy();
--- a/cocos/particle/particle-system.ts
+++ b/cocos/particle/particle-system.ts
@@ -1108,7 +1108,7 @@ export class ParticleSystem extends ModelRenderer {
         this.processor.onDisable();
         if (this._trailModule) this._trailModule.onDisable();
         if (this._boundingBox) {
-            this._boundingBox = null;            
+            this._boundingBox = null;
         }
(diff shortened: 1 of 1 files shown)


cocos/particle/particle-system.ts
```diff
@@ -1109,6 +1109,7 @@ export class ParticleSystem extends ModelRenderer {
         if (this._trailModule) this._trailModule.onDisable();
         if (this._boundingBox) {
             this._boundingBox = null;
+            this._oldPos = null;
```
dumganhar: Good catch.

But why put the logic inside `if (this._boundingBox)`? 

Should it be outside `if (this._boundingBox)` ?

hulcypdiscord: Mostly consistency, but looking over the code again, I can see cases where the boundBox has already been set to null before onDisable is called so probably safest to just always set it to null. Good call out.

cocos/particle/particle-system.ts
```diff
@@ -1108,8 +1108,9 @@ export class ParticleSystem extends ModelRenderer {
         this.processor.onDisable();
         if (this._trailModule) this._trailModule.onDisable();
         if (this._boundingBox) {
-            this._boundingBox = null;
+            this._boundingBox = null;
```
dumganhar: Please fix the eslint error: `Trailing spaces not allowed`.

dumganhar (approved):

dumganhar:
It was merged. Thanks.

Issue #16929, opened 2024-04-25
A Spine demo crashes on native platform.


dumganhar:
Reported at https://forum.cocos.org/t/topic/157119 

[归档(1).zip](https://github.com/cocos/cocos-engine/files/15106143/1.zip)

PR #16935, opened 2024-04-26
Fix divide zero issue in mat4.toSRT which causes node._lrot with NaN values.

dumganhar:
Re: #16929

If node._lrot.x is NaN, it will be serialized to scene data with `null` value.

企业微信截图_b5056fb5-4838-440b-9b90-774731779425

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/core/math/mat4.ts
+++ b/cocos/core/math/mat4.ts
@@ -897,44 +897,48 @@ export class Mat4 extends ValueType {
      * @param s The corresponding scaling vector
      */
     public static toSRT<InType extends IMat4Like, VecLike extends IVec3Like> (m: InType, q: Quat | null, v: VecLike | null, s: VecLike | null): void {
+        // Translation
+        if (v) {
+            Vec3.set(v, m.m12, m.m13, m.m14);
+        }
+
+        // Scale
         const sx = Vec3.set(v3_1, m.m00, m.m01, m.m02).length();
         const sy = Vec3.set(v3_1, m.m04, m.m05, m.m06).length();
         const sz = Vec3.set(v3_1, m.m08, m.m09, m.m10).length();
-
-        if (sx !== 0) {
-            m3_1.m00 = m.m00 / sx;
-            m3_1.m01 = m.m01 / sx;
-            m3_1.m02 = m.m02 / sx;
-        } else {
-            m3_1.m00 = m3_1.m01 = m3_1.m02 = 0;
-        }
-        if (sy !== 0) {
-            m3_1.m03 = m.m04 / sy;
-            m3_1.m04 = m.m05 / sy;
-            m3_1.m05 = m.m06 / sy;
-        } else {
-            m3_1.m03 = m3_1.m04 = m3_1.m05 = 0;
-        }
-        if (sz !== 0) {
-            m3_1.m06 = m.m08 / sz;
-            m3_1.m07 = m.m09 / sz;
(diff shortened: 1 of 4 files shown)


cocos/core/math/mat4.ts
```diff
@@ -930,27 +900,43 @@ export class Mat4 extends ValueType {
...
-            Vec3.set(v, m.m12, m.m13, m.m14);
-        }
-        if (q) {
+
+        if (sx !== 0) {
             m3_1.m00 = m.m00 / sx;
             m3_1.m01 = m.m01 / sx;
             m3_1.m02 = m.m02 / sx;
+        } else {
+            m3_1.m00 = m3_1.m01 = m3_1.m02 = 0;
```
minggo: Is else needed? If length is 0, then all values should be 0.

cocos/core/math/mat4.ts
```diff
@@ -930,27 +900,43 @@ export class Mat4 extends ValueType {
...
+
+        const det = Mat3.determinant(m3_1);
+        if (s) {
+            s.x = sx;
+            s.y = sy;
+            s.z = sz;
+            if (det < 0) {
+                s.x *= -1;
+            }
+        }
```
star-e, line 940: Here I don't understand, why inverse s.x?

dumganhar: https://github.com/cocos/cocos-engine/blob/v3.8.3/cocos/core/math/mat4.ts#L904

It's from the original logic。

### There is an explain from AI:


When the determinant is less than 0, it signifies two important physical meanings:

1. Reversal of Orientation: The sign of the determinant indicates whether a transformation preserves the orientation of space. When the determinant is negative, it means that the transformation reverses the orientation of space, flipping from the positive direction to the negative direction. Geometrically, this implies that objects are oriented oppositely after the transformation compared to the original orientation.
2. Volume Reversal: The absolute value of the determinant represents the scaling factor of the space volume defined by the transformation. When the determinant is negative, it implies that the transformation results in a reversal of the space volume, meaning that after the transformation, the volume of objects in space is reversed, becoming opposite.

dumganhar: @star-e I found some reference in Ogre engine:

https://github.com/OGRECave/ogre/blob/26c02c910dcc3c90580111403e6e40285ee785ad/OgreMain/src/OgreMatrix3.cpp#L626

minggo (approved):

cocos/core/math/mat4.ts
```diff
@@ -935,22 +911,36 @@ export class Mat4 extends ValueType {
...
+
+        // Scale too close to zero, can't decompose rotation.
+        if (sx === 0 || sy === 0 || sz === 0) {
+            if (q) {
+                Quat.identity(q);
+            }
+            return;
         }
+
+        m3_1.m00 = m.m00 / sx;
```
dumganhar, line 923: The original code did not check whether the sx value is 0, and divide 0 will make m3_1.m00 be a NaN value.

GengineJS (approved):

cocos/core/math/mat4.ts
star-e: I think it mirrors the x-axis. As long as the conversion is consistent, it will be fine.

star-e (approved):

PR #16965, opened 2024-05-10
[V3.8.5 pipeline] Add window resize handler.

star-e:
Refactored custom pipeline resize logic.
Previously users need to detect render window resize by themselves. It is tedious and hard to do.
New approach will call user defined resize handler.
New approach will only be enabled, if user defined `gameWindowResize` handler.

### Changelog

* Add resize related members to the `RenderWindow` class.
* Add `colorName`, `depthStencilName` to the `RenderWindow` class. Those are used by custom pipeline.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


star-e:
@cocos-robot run test cases

cocos/render-scene/core/render-window.ts
```diff
@@ -103,6 +150,14 @@ export class RenderWindow {
     protected _hasOffScreenAttachments = false;
     protected _framebuffer: Framebuffer | null = null;
     protected _device: Device | null = null;
+    // _renderWindowId is used to identify the render window in the render pipeline
+    protected _renderWindowId = _windowCount++;
+    // _isResized is used to indicate whether the render window is resized
+    protected _isResized = true;
```
minggo, line 156: Why default value is true?

star-e: When render window is initialized, we consider it is resized. So the resize handler can be triggered and user can register resources in the handler.

minggo: Got it.

minggo (approved):

PR #16966, opened 2024-05-10
simplify logic

minggo:
- just check y can determine use which type of function
- don't add extra parameters and reduce operation

### DIFF
--- a/cocos/scene-graph/node.ts
+++ b/cocos/scene-graph/node.ts
@@ -41,7 +41,6 @@ import { PrefabInfo, PrefabInstance } from './prefab/prefab-info';
 import { NodeEventType } from './node-event';
 import { Event } from '../input/types';
 import type { NodeEventProcessor } from './node-event-processor';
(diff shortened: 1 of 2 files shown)


cocos/scene-graph/node.ts
```diff
@@ -2106,27 +2106,30 @@ export class Node extends CCObject implements ISchedulable, CustomSerializable {
...
-            this.invalidateChildren(TransformBit.POSITION);
+            Vec3.copy(this._lpos, val as Vec3);
+        } else {
+            if (z === undefined) {
+                z = localPosition.z;
+            }
 
-            if (this._eventMask & TRANSFORM_ON) {
-                this.emit(NodeEventType.TRANSFORM_CHANGED, TransformBit.POSITION);
+            if (val as number === localPosition.x  && y === localPosition.y && z === localPosition.z) {
```
dumganhar: there is a redundant `space` after `localPosition.x`

dumganhar (approved):

Issue #13229, opened 2022-11-02
Can't tween color of cc.Sprite


BenoitFreslon:
### Cocos Creator version

3.6.1

### System information

OSX

### Issue description

Hello,

I want to tween the color of a cc.Sprite with this line:

    `    tween(this.sprite1).to(3, { color:Color.TRANSPARENT }).start();`

But the **color** property is not present.
![image](https://user-images.githubusercontent.com/758680/199543100-cd53446c-6c2a-4627-af51-ba02887c6687.png)

Regarding this post, this bug should be fixed in CocosCretor 3.6
https://discuss.cocos2d-x.org/t/fade-in-and-out/55902/2

But I got this problem with CocosCreator 3.6.1.

I need to add UIOpacity component to change the opacity property.

`   tween(this.sprite1.getComponent(UIOpacity)).to(1, { opacity:0 }).start();`

### Steps to reproduce

Add a Sprite component then try to tween the color.
`tween(this.sprite1).to(3, { color:Color.TRANSPARENT }).start();`

minggo:
@bofeng-song please discuss with @qiuguohua , who will take charge of it.

minggo:
@bofeng-song please discuss with @qiuguohua , who will take charge of it.

dumganhar:
@BenoitFreslon Cocos Creator uses `Entity Component` architecture. So `Node` is an entity without any rendering relative functionality. So you can't tween `color` property on Node since it doesn't exist on Node.

dumganhar:
I misunderstand the issue. 
The issue is to create a tween color for sprite component, it is not relative to color property on `Node`. I will fix this issue soon.

BenoitFreslon:
Yes it is.
If I remeber the old documentation shows an example with a tween to change the color sprite. That's why I posted this.
Thanks.

dumganhar:
@BenoitFreslon , thanks for your report. I submitted a pull request here https://github.com/cocos/cocos-engine/pull/16977
If you get time, you could apply this patch to verify whether it works for you.

PR #16948, opened 2024-05-06
Fix lots of type issues in Tween system

dumganhar:
Re: #16937

If developers enable `strict` mode for their projects and write some code by using tween system like the following:

```ts
class MyCustomData {
    constructor (a: number) {
        this._aaa = a;
    }
    _aaa = 0;
}

class TestOne {
    _customData = new MyCustomData(0);

    foo (): void {
        const tweenDuration: number = 1.0;
        const t = tween(this._customData).to(
            tweenDuration,
            new MyCustomData(100),
            this,
        ).start();
    }

    onUpdate (target?: MyCustomData, ratio?: number): void {
        if (target && ratio) {
            // Write the code
        }
    }
}
```

Before this PR, the above code will trigger a ts compiler error:

```
Argument of type 'this' is not assignable to parameter of type 'ITweenOption | undefined'.
  Type 'TestOne' is not assignable to type 'ITweenOption'.
    Type 'this' is not assignable to type 'ITweenOption'.
      Type 'TestOne' is not assignable to type 'ITweenOption'.
        Types of property 'onUpdate' are incompatible.
          Type '(target?: MyCustomData | undefined, ratio?: number | undefined) => void' is not assignable to type '(target?: object | undefined, ratio?: number | undefined) => void'.
            Types of parameters 'target' and 'target' are incompatible.
              Type 'object | undefined' is not assignable to type 'MyCustomData | undefined'.
                Property '_aaa' is missing in type '{}' but required in type 'MyCustomData'.ts(2345)
MoveSprite.ts(20, 5): '_aaa' is declared here.
this: this
```

After apply this PR, the error will disappear, and also we get strong type restriction for `target`. That means it will trigger an error of type mismatch if the `onUpdate` target parameter's type is not matched with `_this._customData`.
For example, we write a wrong code to make the onUpdate target parameter type to `Node`:

```ts
class TestOne {
    _customData = new MyCustomData(0);

    foo (): void {
        const tweenDuration: number = 1.0;
        const t = tween(this._customData).to(
            tweenDuration,
            new MyCustomData(100),
            this,
        ).start();
    }

    onUpdate (target?: Node, ratio?: number): void { // Wrote wrong type of target
        if (target && ratio) {
            // Write the code
        }
    }
}
```

TS compiler will make an error: 

```
Argument of type 'this' is not assignable to parameter of type 'ITweenOption<MyCustomData> | undefined'.
  Type 'TestOne' is not assignable to type 'ITweenOption<MyCustomData>'.
[...]

PR #16977, opened 2024-05-12
[v3.8.5] Fix the bug of tween color property of Sprite.

dumganhar:
Re: #13229

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/core/math/color.ts
+++ b/cocos/core/math/color.ts
@@ -31,8 +31,10 @@ import { legacyCC } from '../global-exports';
 import { assertIsTrue } from '../data/utils/asserts';
 import { Vec4 } from './vec4';
 import { Vec3 } from './vec3';
+import { extractPropsSymbol } from '../utils/misc';
 
 const toFloat = 1 / 255;
+const extractColorProperties: string[] = ['r', 'g', 'b', 'a'];
 
 /**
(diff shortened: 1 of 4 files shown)


dumganhar:
ESLint issues are fixed in this PR: https://github.com/cocos/cocos-engine/pull/16948

So I don't fix them here.

dumganhar:
ESLint errors were fixed at https://github.com/cocos/cocos-engine/pull/16948 , so don't fix them here.

minggo:
I was thinking if we need to change Color implementation. Use r, g, b, a instead of _val.

dumganhar:
Yes, it cause some issues that we need to use bit operation in many places.
And Color._var is public and it uses a number instead of 4 numbers which does save some memory. 
So I think it's not the right time to make the change currently.

minggo:
> Yes, but Color._var is public and it uses a number instead of 4 numbers which does save some memory. I think, currently, it's not the right time to make the change.

I see. So i want to create a ticket in forum about it. I want to do it like this because it affects tween system now.

dumganhar:
Fixed at https://github.com/cocos/cocos-engine/pull/17021
Close this one.

PR #16979, opened 2024-05-12
[v3.8.5] Support difference targets in sequence and parallel chain

dumganhar:
Re: #16978

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


dumganhar:
TODO: 
- [x] handle `.then`
- [x] nest tween in deep depth
- [x] nest sequence and parallel
- [x] re-targetting
- [x] clone with target

minggo:
What's the usage with different targets?

dumganhar:
Could see the new code I added in the test case. Or refer to https://forum.cocos.org/t/topic/157814/7?u=dumganhar

minggo:
I was thinking if we need to change Color implementation. Use r, g, b, a instead of _val.

cocos/tween/actions/action.ts
```diff
@@ -45,6 +45,7 @@ export class Action {
 
     protected originalTarget: Node | null = null;
     protected target: Node | null = null;
+    public workerTarget: Node | null = null;
```
minggo: What does `workerTarget ` mean?

minggo: And what's the difference between `workerTarget ` and `target`?

dumganhar: `workerTarget` is the actual target tween action runs on when the target initialized in `sequence`, `parallel` or `then` is differernt from the root target. For example:

```ts
tween(this.node).parallel( // this.node is the root target, start method will assign this.node as `target`
  tween(this.node.getComponent(UITransform))).to(XXXXX), // UITransform target is different from `this.node`, so it's the actual target this tween wants to run on. I call it `workerTarget`
  tween(this.node.getComponent(UIOpacity))).to(XXXXX)
)
.start()
```

When `start` is inovked, only the root target `this.node` will be set to every internal actions in the old code, so any other targets in `sequence`, `parallel` and `then` will be ignored. 

I added a new property `workerTarget` to support this feature. But yes, there is probably a better name for it. Maybe `proxyTarget` or just `subTarget`.

dumganhar:
@cocos-robot run test cases

dumganhar:
@cocos-robot run test cases

cocos/tween/actions/action-interval.ts
```diff
@@ -29,6 +29,25 @@ import { FiniteTimeAction, Action } from './action';
 import { macro, logID, errorID } from '../../core';
 import { ActionInstant } from './action-instant';
 
+// Extra action for making a Sequence or Spawn when only adding one action to it.
+class ExtraAction extends FiniteTimeAction {
```
minggo: How about rename to DummyAction?

cocos/tween/actions/action.ts
```diff
@@ -207,112 +208,3 @@ export abstract class FiniteTimeAction extends Action {
 
     abstract reverse (): FiniteTimeAction;
 }
-
-/*
- * Changes the speed of an action, making it take longer (speed > 1)
- * or less (speed < 1) time. <br/>
- * Useful to simulate 'slow motion' or 'fast forward' effect.
- */
-export class Speed extends Action {
```
minggo, line 216: Why delete it?

dumganhar: It's unused code.

dumganhar:
@cocos-robot run test cases

cocos/tween/tween.ts
```diff
@@ -127,7 +146,7 @@ export class Tween<T extends object = any> {
         if (this._finalAction) {
             TweenSystem.instance.ActionManager.removeAction(this._finalAction);
         }
-        this._finalAction = this._union();
+        this._finalAction = this._union(false);
```
dumganhar, line 149: The final action doesn't need to update workerTarget since it will replace workerTarget (subTarget) with the `root target`.

cocos/tween/actions/action-interval.ts
```diff
@@ -389,6 +394,19 @@ export class Sequence extends ActionInterval {
         action._reversed = true;
         return action;
     }
+
+    updateWorkerTarget<T> (workerTarget: T): void {
```
minggo, line 398: Why need this function?

cocos/tween/actions/action-interval.ts
dumganhar: It's ok for me. Actually, the name of `ExtraAction` is come from the code in `cocos2d-x`.

cocos/tween/actions/action.ts
minggo: Is it exported?

cocos/tween/actions/action-interval.ts
minggo: It seems DummyAction is more meaningful as it does nothing, just make the tween system happy?

cocos/tween/actions/action.ts
dumganhar: No, all actions tween module use are internal.

cocos/tween/tween.ts
minggo: What does `workerTarget` mean?

cocos/tween/actions/action-interval.ts
dumganhar: It's used for `update` the workerTarget (subTarget or call it runningTarget) in Sequence/Spawn actions.
Sequence and Spawn actions are containers for storing some other actions ( child actions) .  Child actions may have different target to apply to, so we need a method to `update` the target.

cocos/tween/actions/action-interval.ts
dumganhar: Renamed.

cocos/tween/tween.ts
dumganhar: I answered it above, maybe you could refer it by scroll up this page.

cocos/tween/actions/action.ts
dumganhar: Or to make it clearer, `targetToApply`? 
If developer apply the same root target for different sub tweens, then `targetToApply` is the root target.
```ts
const targetToApply =  this.targetToApply ?? this.target; // targetToApply may be null if all sub tweens have the same targets.
```

minggo: What will happen if workerTarget is not a component, but a node, for example
```typescript
tween(this.node).parallel( // this.node is the root target, start method will assign this.node as `target`
  tween(this.node).to(XXXXX),
  tween(this.node.getComponent(UIOpacity))).to(XXXXX)
)
.start()
```

dumganhar: In this example, 
```ts
tween(this.node).parallel(   // this.node is the target for root tween, target = this.node, workerTarget = null
  tween(this.node).to(XXXXX), // target = this.node, workerTarget = this.node
  tween(this.node.getComponent(UIOpacity))).to(XXXXX)  // target = this.node, workerTarget = this.node
)
.start()
```
In fact, each tween owns a target, the targets in sub-tween of `parallel/sequence` will be lost since the parallel/sequence only pick up the internal actions of tween stores.

dumganhar: And since the target type is a generic `T`, it could be any object types, not only `Node`/`Component`.

minggo: But the declaration type is `Node | null`.

dumganhar: I think it's `unknown` type now.  T for Tween<T> will contain generic type, while passing to Action, it's an unknown type.

Construct Tween<T> with T or Tween.clone/target with <U>.
```ts
class Tween<T extends object = any> {
    constructor(target?: T | null) {...}
    clone<U extends object = any> (target: U): Tween<U> { ... }
    target<U extends object = any> (target: U): Tween<U> { ... }
}
```
When we start a tween, target with T will be passed to ActionManager:

```ts
addAction<T> (action: Action | null, target: T, paused: boolean): void {
    ...
    action.startWithTarget(target); // It will call action's startWithTarget<T>(target)
}

startWithTarget<T> (target: T | null): void {
        this.originalTarget = target;
        this.target = target;  // We use unknown type to store target
    }

    protected originalTarget: unknown = null;
    protected target: unknown = null;
    public workerTarget: unknown = null;

```

In tween-actions.ts, since we need to recover the `T` for the callback ( onStart/onUpdate/onComplete) , we convert unknown to T.
```ts
    startWithTarget<U> (target: U | null): void {
        const isEqual: TypeEquality<T, U> = true;
        if (!isEqual) return;
        super.startWithTarget(target);

        const workerTarget = (this.workerTarget ?? this.target) as T;
        if (!workerTarget) return;
        ......
        if (this._opts!.onStart) { this._opts!.onStart(workerTarget); }
```

```ts
update (t: number): void {
        const workerTarget = this.workerTarget ?? this.target;
        if (!workerTarget) return;
        ......
        if (opts.onUpdate) { opts.onUpdate(workerTarget as T, t); }
        if (t === 1 && opts.onComplete) { opts.onComplete(workerTarget as T); }
```

minggo (approved):

dumganhar:
Fix a bug about union here: https://github.com/cocos/cocos-engine/pull/17020

Issue #17022, opened 2024-05-21
When using bitmapfont in taobao-minigame, the game becomes unresponsive.


zhefengzhang:
### Cocos Creator version

2.4.12

### System information

windows10

### Issue description

![image](https://github.com/cocos/cocos-engine/assets/35944775/9c2af519-b203-439a-8d71-1745863c6121)

### Steps to reproduce

![image](https://github.com/cocos/cocos-engine/assets/35944775/e32999ed-4363-4d85-8abd-da83f0c1e590)
Continuous rapid clicking of the button

### Minimal reproduction project

[demo.zip](https://github.com/cocos/cocos-engine/files/15388379/demo.zip)

zhefengzhang:
因为淘宝真机运行的 JS 虚拟机是 quickjs 的定制版，模拟器是 v8，所以在对数据类型对象进行处理时会有一些差异。此问题是因为淘宝真机上执行下面这种代码时会报错，而在淘宝模拟器上不会。
```
var a = new Unit16Array(2);
a[2] = 0;
```

目前可以修改 assembler-2d.js 中的代码暂时解决此问题。
![image](https://github.com/cocos/cocos-engine/assets/35944775/3b93915b-6df2-4598-83a5-fe3e09ae65eb)

liuhy-liu:
2.4.15-0328，使用上述demo，真机运行黑屏报错，如下，模拟器白屏
![Image](https://github.com/user-attachments/assets/178b282a-23d5-490f-a3a5-e3b30554e4f2)

PR #17023, opened 2024-05-21
fixed taobao-minigame use bitmapfont sometimes error

zhefengzhang:
Re: https://github.com/cocos/cocos-engine/issues/17022

### DIFF
--- a/cocos2d/core/renderer/assembler-2d.js
+++ b/cocos2d/core/renderer/assembler-2d.js
@@ -127,7 +127,7 @@ export default class Assembler2D extends Assembler {
             vertexId = offsetInfo.vertexOffset;
         for (let i = 0, l = iData.length; i < l; i++) {
             //Because taobao-minigame does not allow out-of-bounds access to Uint16Array, otherwise it will report an error and cause the game to be unresponsive
-            if (indiceOffset + 1 >= ibuf.length) break;
+            if (cc.sys.platform === cc.sys.TAOBAO_MINIGAME && indiceOffset + 1 >= ibuf.length) break;
             ibuf[indiceOffset++] = vertexId + iData[i];
         }
     }
--- a/cocos2d/core/renderer/assembler-2d.js
+++ b/cocos2d/core/renderer/assembler-2d.js
@@ -127,7 +127,7 @@ export default class Assembler2D extends Assembler {
             vertexId = offsetInfo.vertexOffset;
         for (let i = 0, l = iData.length; i < l; i++) {
             //Because taobao-minigame does not allow out-of-bounds access to Uint16Array, otherwise it will report an error and cause the game to be unresponsive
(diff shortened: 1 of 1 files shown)


cocos2d/core/renderer/assembler-2d.js
```diff
@@ -126,6 +126,8 @@ export default class Assembler2D extends Assembler {
             indiceOffset = offsetInfo.indiceOffset,
             vertexId = offsetInfo.vertexOffset;
         for (let i = 0, l = iData.length; i < l; i++) {
+            //Because taobao-minigame does not allow out-of-bounds access to Uint16Array, otherwise it will report an error and cause the game to be unresponsive
+            if (cc.sys.platform === cc.sys.TAOBAO_MINIGAME && indiceOffset + 1 >= ibuf.length) break;
```
bofeng-song, line 130: indiceOffset range from 0?

bofeng-song: The last member can not be set normally.

bofeng-song (approved):

cocos2d/core/renderer/assembler-2d.js
```diff
@@ -126,6 +126,8 @@ export default class Assembler2D extends Assembler {
             indiceOffset = offsetInfo.indiceOffset,
             vertexId = offsetInfo.vertexOffset;
         for (let i = 0, l = iData.length; i < l; i++) {
+            //Because taobao-minigame does not allow out-of-bounds access to Uint16Array, otherwise it will report an error and cause the game to be unresponsive
+            if (cc.sys.platform === cc.sys.TAOBAO_MINIGAME && indiceOffset >= ibuf.length) break;
```
minggo: - why do it in for loop, can determine how much data to copy before for loop
- this PR means that on tabao platform, some data will not be copied into ibuf, can it work correctly on tabao platform?

zhefengzhang: Yes, after project testing, this PR will not cause any abnormalities in the game, and it has solved the problem of the game not responding. I will try to handle this issue in a more performant way.

zhefengzhang:
When increasing the value of indiceOffset, it should not exceed the array length of buffer._iData. 

image

image

cocos2d/core/renderer/assembler-2d.js
```diff
@@ -125,7 +125,7 @@ export default class Assembler2D extends Assembler {
         let ibuf = buffer._iData,
             indiceOffset = offsetInfo.indiceOffset,
             vertexId = offsetInfo.vertexOffset;
-        for (let i = 0, l = iData.length; i < l; i++) {
+        for (let i = 0, l = this.indicesCount; i < l; i++) {
             ibuf[indiceOffset++] = vertexId + iData[i];
```
minggo, line 132: It may exceed iData too. So i think should use the min value of them, for example
```typescript
const indicesToFill = min(this.indicesCount, iData.length);
```

minggo (approved):

PR #17033, opened 2024-05-23
add spine safety protection

zhefengzhang:
在大型项目中，由于资源的加载和释放的情况更加复杂，在某些情况下切换游戏场景会出现 spine 错误。
In the large project，because the resource load and release is more complicated，the switching game scene will appear spine error in certain cases.
a30ae047-5875-4c2c-9f68-4e0fa2612a3e

### DIFF
--- a/cocos/spine/assembler/simple.ts
+++ b/cocos/spine/assembler/simple.ts
@@ -161,7 +161,7 @@ function realTimeTraverse (comp: Skeleton): void {
     const vPtr = model.vPtr;
     const iPtr = model.iPtr;
     const ibuf = rd.indices!;
-    const HEAPU8 = spine.wasmUtil.wasm.HEAPU8;
+    const HEAPU8: Uint8Array = spine.wasmUtil.wasm.HEAPU8;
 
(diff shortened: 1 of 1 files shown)


bofeng-song (approved):

bofeng-song (approved):

minggo:
@zhefengzhang 
Did you found the real reason or just protect to avoid the issue?

zhefengzhang:
I understand the cause, but I can only temporarily use this approach to avoid the issue. The reason for this problem is that while the resources of the old scene are being automatically destroyed, the new scene is simultaneously loading resources and initializing the spine components. If a user's project has a large number of resources, and the destruction of resources from the old scene and the loading of resources for the new scene cannot be executed in an ordered queue, there is a chance that the spine components might be initialized using spine resources that have already been destroyed.
我知道原因，但是只能暂时使用这种方式规避。出现此问题的原因是旧场景资源正在自动销毁的同时，新场景也同时正在加载资源和初始化 spine 组件。如果用户的项目资源较多，旧场景的资源销毁与新场景的资源加载目前不能在有序队列执行，就会可能出现 spine 组件初始化时使用的是已经被销毁过的 spine 资源。

minggo:
Got it, thanks.

Issue #17078, opened 2024-05-31
[feature] Tween supports to start at a specific timepoint


dumganhar:
This feature was requested from https://forum.cocos.org/t/topic/157814/78?u=dumganhar

PR #17105, opened 2024-06-05
Add Tween.startAt(time) API and add more unit test.

dumganhar:
Re: #17078

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/tween/actions/action-interval.ts
+++ b/cocos/tween/actions/action-interval.ts
@@ -126,7 +126,7 @@ export abstract class ActionInterval extends FiniteTimeAction {
             this._firstTick = false;
             let elapsed = 0;
             if (this._startTime > 0) {
-                // _startTime only take effect in the first tick after tween starts. So reset it to 0 after the first tick.
+                // _startTime only takes effect in the first tick after tween starts. So reset it to 0 after the first tick.
                 elapsed = this._startTime;
                 this._startTime = 0;
             }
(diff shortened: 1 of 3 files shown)


cocos/tween/actions/action-interval.ts
```diff
@@ -80,6 +82,11 @@ export abstract class ActionInterval extends FiniteTimeAction {
         }
     }
 
+    setStartTime (time: number): void {
```
minggo, line 84: Does it support negative values?

dumganhar: I think it will not, I will add a valid time checking in this function.

minggo: If so, then i think don't have to add a variable to check if it is set or not, just check the value. The default value can be set to a negative value.

dumganhar: Updated and added two unit tests for this.

dumganhar: > If so, then i think don't have to add a variable to check if it is set or not, just check the value. The default value can be set to a negative value.

Good point. I think the default value could be 0, I will update the code.

dumganhar: Updated

cocos/tween/actions/action-interval.ts
```diff
@@ -118,7 +124,13 @@ export abstract class ActionInterval extends FiniteTimeAction {
         dt *= this._speed;
         if (this._firstTick) {
             this._firstTick = false;
-            this._elapsed = 0;
+            let elapsed = 0;
```
minggo: It seems this variable is not needed.

dumganhar: Fixed.

minggo (approved):

PR #17124, opened 2024-06-11
Fix the problem that the RT is upside down in the GL backend

GengineJS:


native/cocos/core/assets/Material.cpp
```diff
@@ -228,6 +228,30 @@ const MaterialPropertyVariant *Material::getProperty(const ccstd::string &name,
     return nullptr;
 }
 
+void Material::prepareInfo(const cc::IMaterialInfo::DefinesType &patch, ccstd::vector<MacroRecord> &curr) {
+    size_t len = _effectAsset != nullptr ? _effectAsset->_techniques[_techIdx].passes.size() : 1;
+    const auto *pOneElement = ccstd::get_if<MacroRecord>(&patch);
```
minggo: - don't use Hungarian notation
- it is better to use a more meaningful name, i don't know what oneElement mean

dumganhar:
So the issue isn't caused by wrong js binding for Material::copy?

Is this PR ( https://github.com/cocos/cocos-engine/pull/17122 ) needed?

native/cocos/core/assets/Material.cpp
```diff
@@ -228,6 +228,30 @@ const MaterialPropertyVariant *Material::getProperty(const ccstd::string &name,
...
+            }
+        }
+    } else {
+        const auto *pPatchArray = ccstd::get_if<ccstd::vector<MacroRecord>>(&patch);
+        if (pPatchArray != nullptr) {
+            const auto &patchArray = *pPatchArray;
+            size_t len = patchArray.size();
+            curr.resize(len);
+            for (size_t i = 0; i < len; ++i) {
+                curr[i] = patchArray[i];
```
dumganhar: Shall we use `curr.emplace_back(...)` here since `curr` is an in-out parameter and it probably contains elements yet?

native/cocos/core/assets/Material.h
```diff
@@ -317,6 +317,7 @@ class Material : public Asset {
     bool uploadProperty(scene::Pass *pass, const ccstd::string &name, const MaterialPropertyVariant &val);
     void bindTexture(scene::Pass *pass, uint32_t handle, const MaterialProperty &val, uint32_t index = 0);
 
+    void prepareInfo(const cc::IMaterialInfo::DefinesType &patch, ccstd::vector<MacroRecord> &curr);
     template <typename T1, typename T2>
     void prepareInfo(const T1 &patch, ccstd::vector<T2> &cur) {
```
dumganhar, line 373: There is already a `template` version of `prepareInfo` here . Why do we need an overload version with the same logic here?

GengineJS:
> So the issue isn't caused by wrong js binding for Material::copy?
> 
> Is this PR ( #17122 ) needed?

Yes, it is also necessary.

native/cocos/core/assets/Material.h
GengineJS: Because MacroRecord is of the type unordered_map<ccstd::string, MacroValue>, what we need to do is to add to the unordered_map, not to add to a vector.

native/cocos/core/assets/Material.cpp
GengineJS: I think it directly replaces the content, rather than adding through the method of addition.

native/cocos/core/assets/Material.cpp
```diff
@@ -228,6 +228,30 @@ const MaterialPropertyVariant *Material::getProperty(const ccstd::string &name,
     return nullptr;
 }
 
+void Material::prepareInfo(const cc::IMaterialInfo::DefinesType &patch, ccstd::vector<MacroRecord> &curr) {
```
star-e: Shall we use
```
template <typename T1>
void prepareInfo(const T1 &patch, ccstd::vector<MacroRecord> &cur) {
    // xxx
}
```
It overloads on `ccstd::vector<MacroRecord>` and is more general than `const cc::IMaterialInfo::DefinesType &patch`.

minggo (approved):

star-e (approved):

native/cocos/core/assets/Material.h
```diff
@@ -334,9 +338,33 @@ class Material : public Asset {
             if (macroRecordArray != nullptr) {
                 const auto &patchArray = *macroRecordArray;
                 size_t len = patchArray.size();
-                curr.resize(len);
+                if (curr.size() < len) curr.resize(len);
```
dumganhar, line 341: The checking of `if (curr.size() < len)` is redundant.
https://en.cppreference.com/w/cpp/container/vector/resize

- Resizes the container to contain count elements, does nothing if count == size().

- If the current size is greater than count, the container is reduced to its first count elements.

- If the current size is less than count,

  - 1) additional [default-inserted](https://en.cppreference.com/w/cpp/named_req/DefaultInsertable) elements are appended.
  - 2) additional copies of value are appended.
```

dumganhar:
@cocos-robot run test cases

dumganhar (approved):

dumganhar:
@cocos-robot run test cases

dumganhar:
@cocos-robot run test cases

dumganhar:
@cocos-robot run test cases

PR #17157, opened 2024-06-14
feat(security): prevent RCE in various workflows

avivkeller:
CC @minggo

### DIFF
--- a/.github/workflows/web-interface-check.yml
+++ b/.github/workflows/web-interface-check.yml
@@ -13,6 +13,7 @@ jobs:
     if: 
       (! contains(github.event.pull_request.body, '[X] does not change any runtime related code or build configuration'))
     runs-on: ubuntu-latest
+    environment: external
     steps:
       - uses: actions/setup-node@v2
         with:
@@ -47,8 +48,7 @@ jobs:
       - uses: actions/checkout@v3
         name: Checkout Head Ref
         with:
-          ref: 'refs/pull/${{ github.event.pull_request.number }}/merge' # Don't check out the head commit, checkout the "merge commit" instead
-          path: './engine-HEAD'
+          ref: ${{ github.event.pull_request.merge_commit_sha }}
       - run: |
           "EXT_VERSION_HEAD=$(node ./engine-HEAD/.github/workflows/get-native-external-version.js)" >> $env:GITHUB_ENV
         shell: pwsh
--- a/.github/workflows/automation_labels.yml
+++ b/.github/workflows/automation_labels.yml
@@ -13,13 +13,11 @@ jobs:
     steps:
       - name: Check organization
         id: get_check_code
+        env:
+          user: ${{ github.event.pull_request.user.login }}
         run: |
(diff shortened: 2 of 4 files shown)


minggo:
@RedYetiDev thanks for the PR.

shaoqiangcai (approved):
his modification makes our actions more standardized

.github/workflows/web-interface-check.yml
```diff
@@ -13,6 +13,7 @@ jobs:
     if: 
       (! contains(github.event.pull_request.body, '[X] does not change any runtime related code or build configuration'))
     runs-on: ubuntu-latest
+    environment: external
```
dumganhar, line 16: Hi @RedYetiDev , thanks for your contribution.
May I ask why do we need to set an environment variable `external` here ?

avivkeller: Setting this value to `external` means that the workflow won't proceed without approval from a repository admin

minggo: Does it mean the admin should manually to approve the action every time?

dumganhar: > Setting this value to external means that the workflow won't proceed without approval from a repository admin

@RedYetiDev , we need the `web-interface-check.yml` run every time when Pull Request is updated even when it isn't approved.
So I don't know why should it have to be run after `approval`. Any specific security reasons?

dumganhar: There will be an issue if we run this workflow after approval.

The old strategy:

1. Developer send a pull request
2. web-interface-check.yml will be run automatically
3. Repo maintainer review the `interface change` comment sent by `web-interface-check` task
4. Repo maintainer approved the Pull Request
5. Pull Request was merged

If we mark the it as `external`, the flow will be 

1. Developer send a pull request
2. web-interface-check.yml will not run.
3. Repo maintainer approved the Pull Request
4. Repo may forget to wait the `web-interface-check.yml` to finish and merge the Pull Request, the interface changes will not be reviewed.

So make the `environment` to `external` will let each maintainer notice that there is a task being run after approved, and it will be easily forgotten.

avivkeller: Sorry that this change wouldn't be the best anti-RCE method.

This file runs without approval, and uses local scripts, leading to RCE.

I'd suggest reworking the workflow to run on `pull_request` rather than `pull_request_target`.

I'm out-of-town right now, so I'll be unable to implement such a change.

dumganhar: @RedYetiDev ok,thanks for the advice, I will check that.

dumganhar: Reference: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

dumganhar:
I found that this Pull Request was closed since the source repository was deleted.
I cherry-picked the the commit in this PR : https://github.com/cocos/cocos-engine/pull/17258 and reverted the changes in web-interface-check.yml.

I will send another PR to resolve the RCE issue in web-interface-check.yml.

@RedYetiDev thanks for this PR.

.github/workflows/web-interface-check.yml
dumganhar: In the end, fix it here : https://github.com/cocos/cocos-engine/pull/17264

@RedYetiDev  Could you please help to review that PR? Thank you.

PR #17160, opened 2024-06-15
modify the userdata type of IExposedAttributes

whaqzhzd:
Replace the IExposedAttributes userdata types from Record < string, any > to interface IExposedAttributesUserData extends Record < string, any> {}. The advantage of this modification is that we can take advantage of the type namesakes interface merging feature of ts to achieve better type hints.

For example, we have a type file in which we define the following:

```ts
declare module "cc" {
  export namespace __private {
    export interface _cocos_core_data_utils_attribute_defines__IExposedAttributesUserData extends  my interface type {
       // other prototype
    }
  }
}
```

So when we use
```typescript
@property({ type: property type, userData: { /*There will be our own defined type information. Very convenient/*  } }) _property: property type= null!;
```

### DIFF
--- a/cocos/core/data/class.ts
+++ b/cocos/core/data/class.ts
@@ -521,7 +521,6 @@ function parseAttributes (constructor: Function, attributes: PropertyStash, clas
         }
         parseSimpleAttribute('slide', 'boolean');
         parseSimpleAttribute('unit', 'string');
-        parseSimpleAttribute('userData', 'object');
         parseSimpleAttribute('radioGroup', 'boolean');
     }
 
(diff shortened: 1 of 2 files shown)


cocos/core/data/class.ts
```diff
@@ -594,6 +593,7 @@ function parseAttributes (constructor: Function, attributes: PropertyStash, clas
     parseSimpleAttribute('min', 'number');
     parseSimpleAttribute('max', 'number');
     parseSimpleAttribute('step', 'number');
+    parseSimpleAttribute('userData', 'object');
```
dumganhar, line 596: In the old logic, `userData` was exposed only in DEV mode. 
Why does it need to be exposed in Release mode too?

whaqzhzd: Well, in that case, I don't necessarily need him, I just happened to see such a field in the Property decorator, and from the comments, it doesn't seem to be limited to use under editor or dev. I'm just adding some class tolerance for my properties to do something later. Both so I can develop a new decorator to do this.

dumganhar: So will you reverse the `userData` changes?

dumganhar (approved):

PR #17172, opened 2024-06-17
fixed memory leak on x86 android devices.

dumganhar:
Re: #

https://forum.cocos.org/t/topic/158470

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/native/cocos/base/Macros.h
+++ b/native/cocos/base/Macros.h
@@ -409,3 +409,7 @@ It should work same as apples CFSwapInt32LittleToHost(..)
         #define CC_FORCE_INLINE inline
     #endif
 #endif
+
+// Macro definition to check if a type has padding
+#define CC_CHECK_STRUCT_NO_PADDING(T) \
+    static_assert(std::has_unique_object_representations_v<T>, "ERROR: Type has padding bytes!")
--- a/native/cocos/renderer/gfx-base/GFXDef-common.h
+++ b/native/cocos/renderer/gfx-base/GFXDef-common.h
@@ -1291,7 +1291,9 @@ struct ALIGNAS(8) ColorAttachment {
     LoadOp loadOp{LoadOp::CLEAR};
     StoreOp storeOp{StoreOp::STORE};
     GeneralBarrier *barrier{nullptr};
-
+#if CC_CPU_ARCH == CC_CPU_ARCH_32
+    uint32_t _padding{0};
+#endif
     EXPOSE_COPY_FN(ColorAttachment)
 };
 
@@ -1305,7 +1307,9 @@ struct ALIGNAS(8) DepthStencilAttachment {
     LoadOp stencilLoadOp{LoadOp::CLEAR};
     StoreOp stencilStoreOp{StoreOp::STORE};
     GeneralBarrier *barrier{nullptr};
-
+#if CC_CPU_ARCH == CC_CPU_ARCH_32
+    uint32_t _padding{0};
+#endif
     EXPOSE_COPY_FN(DepthStencilAttachment)
 };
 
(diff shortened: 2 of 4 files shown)


star-e (approved):

native/cocos/renderer/gfx-base/GFXDef-common.h
```diff
@@ -1401,7 +1405,9 @@ struct ALIGNAS(8) BufferBarrierInfo {
 
     Queue *srcQueue{nullptr}; // @ts-nullable
     Queue *dstQueue{nullptr}; // @ts-nullable
-
+#if CC_CPU_ARCH == CC_CPU_ARCH_32
+    uint32_t _padding{0};
```
star-e: `uint32_t _padding` might be added between `size` and `discardContents` without macro.

dumganhar: I will check that

dumganhar:
@cocos-robot run test cases

native/cocos/renderer/gfx-base/GFXDef-common.h
```diff
@@ -1396,7 +1400,7 @@ struct ALIGNAS(8) BufferBarrierInfo {
 
     uint32_t offset{0};
     uint32_t size{0};
-
+    uint32_t _padding{0};
```
star-e, line 1403: I forgot the padding changed member initialization order. It can be problematic.

dumganhar: Do we depend on the order？

star-e: I have searched the codebase and it seems aggregate initialization is not used.
So it might be fine.

star-e (approved):

native/cocos/base/Macros.h
```diff
@@ -409,3 +409,7 @@ It should work same as apples CFSwapInt32LittleToHost(..)
         #define CC_FORCE_INLINE inline
     #endif
 #endif
+
+// Macro definition to check if a type has padding
+#define CC_CHECK_STRUCT_NO_PADDING(T) \
+    static_assert(std::has_unique_object_representations_v<T>, "ERROR: Type has padding bytes!")
```
minggo, line 415: Can not use `std::has_unique_object_representations_v<T>` to check padding, as in the following codes, it also returns false
```c++
struct D {
    int* ptr;
};

struct E {
    int id;
    std::string name;
    double* data;
};
```

Will return false.

dumganhar: It could work for primitve types, and all gfx structs that need the added padding manually could work by `std:: has_unique_object_representations_v `

So do we need to consider stl containers, like `string`, `vector`, `list`, `map` in struct? 
And if so , any suggestion for how to check those types?

dumganhar: The example you wrote:
```cpp
struct D {
    int* ptr;
};
```
returns true.

You could paste code 
```cpp
#include <type_traits>
#include <stdint.h>
#include <iostream>
#include <cstddef> 

#define CHECK_NO_PADDING(T) \
    static_assert(std::has_unique_object_representations_v<T>, "Type " #T " has padding bytes!")


struct D {
    int* ptr;
};

struct E {
    int id;
    std::string name;
    double* data;
};

int main() {

  
  CHECK_NO_PADDING(D); // return true, since int* is 4 bytes,  the default padding is 4 bytes too.
  CHECK_NO_PADDING(E);  // return false, 
   

    return 0;
}

```

in [cpp.sh](https://cpp.sh)  to test. The environment on `cpp.sh` is x86 .

minggo: > So do we need to consider stl containers, like string, vector, list, map in struct?
And if so , any suggestion for how to check those types?

It depends on the usage. But as this function is defined in Macros.h, so should add comments to say that it only works for primitive types.

minggo: > struct D {
    int* ptr;
};
returns true.

Yep, i was deceived by ChatGPT.

dumganhar: I will add a comment later.

PR #17179, opened 2024-06-19
[Fixed] Fix the issue where Fxaa reports an error and cannot run under custom pipeline in WebGPU

GengineJS:


editor/assets/chunks/common/texture/texture-lod.chunk
```diff
@@ -22,7 +22,11 @@ vec4 fragTextureLod (samplerCube tex, vec3 coord, float lod) {
       return textureCube(tex, coord, lod); // fallback to bias
     #endif
   #else
-    return textureLod(tex, coord, lod);
+    #if defined(CC_USE_WGPU)
+      return textureLod(tex, coord, lod);
+    #else
+      return textureLod(tex, coord, lod);
+    #endif
```
star-e: The code is same in both `if` branch.

editor/assets/chunks/common/texture/texture-lod.chunk
```diff
@@ -46,6 +54,10 @@ vec4 fragTextureGrad (samplerCube tex, vec3 P, vec3 dPdx, vec3 dPdy) {
       return textureCube(tex, P);
     #endif
   #else
-    return textureGrad(tex, P, dPdx, dPdy);
+    #if defined(CC_USE_WGPU)
+      return textureLod(tex, P, 0.0);
```
star-e, line 54: wgsl does not have `textureGrad`?

editor/assets/default_renderpipeline/builtin-pipeline.ts
```diff
@@ -438,7 +438,7 @@ if (rendering) {
                 : nativeHeight;
 
             // Render Window (UI)
-            ppl.addRenderWindow(window.colorName, Format.BGRA8, nativeWidth, nativeHeight, window);
+            ppl.addRenderWindow(window.colorName, Format.RGBA8, nativeWidth, nativeHeight, window);
```
star-e, line 441: Here I have made a design mistake. We shouldn't specify texture format in `addRenderWindow`. I suggest we don't use the 'format' parameter and get the format from the render window directly.
The reason is that the swapchain format might be 'rgba8' or 'bgra8', which is platform dependent. So we need to get the format from `RenderWindow` instead of hardcode.

editor/assets/chunks/common/texture/texture-lod.chunk
GengineJS: > wgsl does not have `textureGrad`?

Yes, the problem occurred during the compilation with glslang.js.

star-e: image
wgsl has `textureSampleGrad`. It seems strange.

star-e (approved):

star-e:
@cocos-robot run test cases

sushanhong:
@cocos-robot run test cases

sushanhong:
@cocos-robot run test cases

GengineJS:
https://github.com/cocos/cocos-test-projects/pull/874

star-e (approved):

star-e:
@cocos-robot run test cases

PR #17213, opened 2024-06-24
GeometryRenderer.addSpline also needs to support 'useTransform' parameter like other API in this class.

dumganhar:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/rendering/geometry-renderer.ts
+++ b/cocos/rendering/geometry-renderer.ts
@@ -512,7 +512,7 @@ export class GeometryRenderer {
         depthTest = true,
         unlit = false,
         useTransform = false,
-        transform = new Mat4(),
+        transform: Readonly<Mat4> = Mat4.IDENTITY,
     ): void {
         const deltaPhi   = Math.PI * 2.0 / segmentsU;
         const deltaTheta = Math.PI / 2.0 / hemiSegmentsV;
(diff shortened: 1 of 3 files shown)


cocos/rendering/geometry-renderer.ts
```diff
@@ -1017,10 +1017,16 @@ export class GeometryRenderer {
         }
     }
 
-    public addSpline (spline: geometry.Spline, color: Color, index = 0xffffffff, knotSize = 0.5, segments = 32, depthTest = true): void {
+    public addSpline (spline: geometry.Spline, color: Color, index = 0xffffffff, knotSize = 0.5, segments = 32, depthTest = true, useTransform = false, transform = new Mat4()): void {
```
minggo: Will it create a Mat4 instance if it is not passed?

minggo: It is better use Mat4.IDENTITY.

minggo: Or maybe can just change the API to
```typescript
 public addSpline (spline: geometry.Spline, color: Color, index = 0xffffffff, knotSize = 0.5, segments = 32, depthTest = true, transform?: Mat4)
```

dumganhar: Just because to keep the same API style in this class. You could refer to the other member functions.
The old style was using `default parameter`, I don't want to change that since it will not be unified with other member functions.

dumganhar: > Will it create a Mat4 instance if it is not passed?

Yep, it will create the Mat4 instance here if transform is not passed in. But the reason is that I don't want make the API inconsistent.

dumganhar: > It is better use Mat4.IDENTITY.

I agree this.

minggo (approved):

PR #17223, opened 2024-06-26
Optimize ubo logic and fix the problem that the screen is upside down when the floating point format is turned on and the uniform member setting does not work.

GengineJS:


GengineJS:
@cocos-robot run test cases

star-e (approved):

PR #17254, opened 2024-06-29
fixed #17207: Add Tween.updateUntil low-level tween API 

### LINKED ISSUE #17207: [Tween System] Unlimited time update api
### Cocos Creator version

3.8.2

### System information

MacOS

### Issue description

Right now we have a finite time 'update' api. good job!

As discussed in the previous issue, this api gives users more freedom and is the basis for some speed-driven animations, hopefully soon.

```
tween()
    .update(
        (target: T, dt: number, ...args: R) => {
            const [speed] = args;
            target.string = String(Number(target.string) + speed);

            if (target.string === "100") {
                return false;
            }
        },
        ...args,
    )
    .start();
```

Details: https://github.com/cocos/cocos-engine/issues/17076

### Relevant error log output

_No response_

### Steps to reproduce

.

_No response_

dumganhar:
Re: #17207

Suppose there is an object (A) who wants to trace another object (B), when A gets close to B, A plays an animation to scale itself to 2 in 0.25s and reverse to scale 1 later. We could add an 'updateUntil' action to make A trace B, if A gets close to B, the callback function returns true, which means that 'updateUnitl' should be finished now, otherwise, the callback will be invoked every frame and the actions after it will not be executed 'until' its callback returns true.

I wrote the following code to demostrate.
```ts
import { _decorator, Component, Node, tween, v3, Vec3 } from 'cc';
const { ccclass, property } = _decorator;

const positionTmp = new Vec3();
const positionTmp2 = new Vec3();

@ccclass('TweenTest')
export class TweenTest extends Component {

    @property(Node)
    targetNode: Node | null = null;

    start() {
        tween(this.node)
            .updateUntil((curNode: Node, dt: number)=>{
                const d = Vec3.copy(positionTmp2, this.targetNode!.position).subtract(curNode.position);
                const length = d.length();
                if (length < 10) {
                    return true;
                }
                
                const newPos = Vec3.copy(positionTmp, curNode.position).add(d.normalize().multiplyScalar(length / 10 * dt * 10));
                curNode.setPosition(newPos);
                return false;
            })
            .by(0.25, { scale: v3(1, 1, 0) }, { easing: 'cubicInOut' }).id(1)
            .reverse(1)
            .call((curNode?: Node)=>{
                const newPos = v3((Math.random() - 0.5) * 400, (Math.random() - 0.5) * 400, 0);
                if (newPos.y < 100 && newPos.y > 0) newPos.y = 100;
                if (newPos.y > -100 && newPos.y < 0) newPos.y = -100;

                if (newPos.x < 100 && newPos.x > 0) newPos.x = 100;
                if (newPos.x > -100 && newPos.x < 0) newPos.x = -100;
                curNode?.setPosition(newPos);
            })
            .union()
            .repeatForever()
            .start();

        tween(this.node)
            .by(1, { angle: 360 })
            .repeatForever()
            .start();

    }
}
```
And I recorded a video for that:

https://github.com/cocos/cocos-engine/assets/493372/4c7df413-7eff-4137-aafa-bc37df732102

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


dumganhar:
@smallmain could review this now.

dumganhar:
@cocos-robot run test cases

dumganhar:
TODO: Update the `updateUntil` API documentation.

cocos/tween/actions/action-interval.ts
```diff
@@ -759,9 +801,20 @@ export class Spawn extends ActionInterval {
...
+            }
+        }
+
+        if (this._two) {
+            if (!this._finished || this._two.isUnknownTime()) {
+                this._two.update(t);
+            }
+        }
+
+        this._finished = t === 1;
```
dumganhar, line 824: This assignment will cause issues if `timeScale` is a negative value.

smallmain:
amazing!

dumganhar:
@cocos-robot run test cases

cocos/tween/actions/action.ts
```diff
@@ -279,4 +279,6 @@ export abstract class FiniteTimeAction extends Action {
     abstract clone (): FiniteTimeAction;
 
     abstract reverse (): FiniteTimeAction;
+
+    abstract isUnknownTime (): boolean;
```
minggo: It is better to rename to `isUnknownDuration`

minggo: Or rename to `isDynamicDuration`

dumganhar: Yep, it's a better name. I will update it.

cocos/tween/actions/action-interval.ts
```diff
@@ -809,6 +862,22 @@ export class Spawn extends ActionInterval {
...
+        const two = this._two;
+
+        if (one == null || two == null) return false;
+
+        const isOneUnknownTime = one.isUnknownTime();
+        const isTwoUnknownTime = two.isUnknownTime();
+
+        if (isOneUnknownTime || isTwoUnknownTime) {
+            if (isOneUnknownTime && isTwoUnknownTime) return true;
+            else if (this._finished) return true;
```
minggo, line 884: Don't understand the logic, why return true if this is finished.?

dumganhar: Suppose that there are two action A and B in Spawn, A is an action with unknown duration (0s) while B is an action with constant duration (e.g. 1 second), when 1 second passes, B will be done, so the whole Spawn is done, but A is not resovled, so the Spawn is still in `unknown duration` state.

cocos/tween/actions/action.ts
dumganhar: Updated

dumganhar:
@cocos-robot run test cases

cocos/tween/actions/action-interval.ts
```diff
@@ -360,6 +376,19 @@ export class Sequence extends ActionInterval {
...
         return null;
     }
+
+    isUnknownTime (): boolean {
+        if (this._actions.length === 0) return false;
+
+        const one = this._actions[0];
+        const two = this._actions[1];
+
+        if (this._last < 1) {
```
minggo, line 390: Can change to `this_last === 0`

dumganhar:
@cocos-robot run test cases

cocos/tween/actions/action-interval.ts
dumganhar: We can't. Because there is another state, this._last == -1, which means all actions in Sequence were not updated.
So `< 1` here means `0` or `-1`.

cocos/tween/actions/action-interval.ts
minggo: In this situation, i think it should return true event this._finished is false. May be can change the code like this
```typescript
return (isOneUnknownTime || isTwoUnknownTime);
```

minggo (approved):

cocos/tween/actions/action-interval.ts
dumganhar: `(isOneUnknownTime || isTwoUnknownTime);` means one of the actions in Spawn is unknown duration, another one is not. Still take the above example:
- Spawn starts
- A, B update at the same time, A is `unknown duration`, B is fixed duration (1 second) , the `Spawn`'s duration is 1s since `unknown duration` will always return 0.
- Suppose 0.5 second was passed, A keeps its `unknown duration` state, B `is not` done. B needs to be continued get `update` with a `t` increased. So the `Spawn` is `not` in `unknown duration` state.
- While 1 seconds was passed, A still keeps its `unknown duration` state, B `is` done. B doesn't needs to be continued get `update` now. But A is still waiting, so the whole `Spawn` needs to be waiting too. And the `Spawn` is in `unknown duration` state to avoid `elasped time` being added `dt`.
- A is resolved, A itself is not in `unknown duration` state, the `Spawn` is not in `unknown duration` state now.
- The spawn is done.

dumganhar:
@cocos-robot run test cases

sushanhong:
@cocos-robot run test cases

sushanhong:
@cocos-robot run test cases

sushanhong:
@cocos-robot run test cases

sushanhong:
@cocos-robot run test cases

sushanhong:
@cocos-robot run test cases

dumganhar:
@cocos-robot run test cases

dumganhar:
@cocos-robot run test cases

shaoqiangcai:
@cocos-robot run test cases

PR #17291, opened 2024-07-03
[Optimize] the logic of the UBO code and fix the issue where the profiler does not display in special circumstances.

GengineJS:


star-e (approved):

star-e:
Will merge after run test cases succeeded.

PR #17335, opened 2024-07-09
Optimize the depthattachment of renderpass and fix the webgpu sampler type error and fix the problem that webgpu post-processing does not work

GengineJS:


cocos/gfx/webgpu/webgpu-sampler.ts
```diff
@@ -64,13 +70,34 @@ export class WebGPUSampler extends Sampler {
         };
     }
 
-    public createGPUSampler (mipLevel: number = 1): void {
-        if (this._gpuSampler && !this.gpuSampler.gpuSampler) {
-            this._gpuSampler.mipLevel = mipLevel;
-            const device = WebGPUDeviceManager.instance;
-            this._hasChange = true;
-            WebGPUCmdFuncCreateSampler(device, this._gpuSampler);
+    private _computeHash (info: IWebGPUGPUSampler): number {
```
star-e: I think `hash` is not a hash value. `_computeHash` and `hash` might be renamed to `_computeSamplerKey` and `key`.

star-e (approved):

PR #17374, opened 2024-07-16
The Web end has added offscreen Multisample Anti-Aliasing (MSAA), modified the blitTexture, and optimized the renderTexture.

GengineJS:


cocos/gfx/webgl/webgl-swapchain.ts
```diff
@@ -245,12 +245,12 @@ export class WebGLSwapchain extends Swapchain {
             WebGLDeviceManager.instance.capabilities.maxTextureUnits,
             WebGLDeviceManager.instance.capabilities.maxVertexAttributes,
         );
-
+        this._format = premultipliedAlpha ? Format.RGBA8 : Format.RGB8;
```
star-e, line 248: Why not use RGBA8 for both cases?
`premultiplied alpha` should not have affected pixel format.

cocos/asset/assets/render-texture.ts
```diff
@@ -53,6 +53,10 @@ const _windowInfo: IRenderWindowInfo = {
     renderPassInfo: passInfo,
 };
 
+function isWebGPU (): boolean {
```
minggo, line 56: Why need to do special check for webgpu? How about metal and vulkan?

cocos/gfx/webgl/webgl-swapchain.ts
```diff
@@ -191,7 +191,7 @@ export function getExtensions (gl: WebGLRenderingContext): IWebGLExtensions {
 
     return res;
 }
-
+const premultipliedAlpha = false;
```
minggo, line 194: Why need this variable?

cocos/gfx/webgl/webgl-swapchain.ts
GengineJS: When premultipliedAlpha is set to false, the format of the swapchain is RGB8, not RGBA8. You can obtain this through gl.drawingBufferFormat.

star-e: I see.

cocos/asset/assets/render-texture.ts
GengineJS: In WebGPU, it is possible to force the format to RGBA8, but the official recommendation is to call the function navigator.gpu.getPreferredCanvasFormat() to use the most suitable format for the swapchain. This function may return BGRA8 on some devices, so it is necessary to make a distinction.

cocos/gfx/webgl/webgl-swapchain.ts
GengineJS: This attribute is used to obtain the format of the current swapchain. If it is false, the actual drawingBufferFormat is RGB8, which can lead to format mismatch errors when offscreen MSAA is blitted to the screen.

GengineJS: In fact, the RGB8 format for the swapchain can bring some performance improvements to the rendering process of readPixels or renderTexture, but to avoid breaking compatibility, I still use RGBA8 when assigning the on-screen format.

minggo: But it is const, it seems meaningless.

cocos/asset/assets/render-texture.ts
minggo: But why not just get format from swapchain for all platforms?

cocos/gfx/webgl2/webgl2-commands.ts
```diff
@@ -3406,23 +3406,30 @@ export function WebGL2CmdFuncBlitTexture (
         cache.glReadFramebuffer = srcFramebuffer;
     }
 
-    if (cache.glFramebuffer !== dstFramebuffer) {
-        gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER, dstFramebuffer);
-        cache.glFramebuffer = dstFramebuffer;
-    }
-
```
minggo, line 3413: Why delay binding it?

cocos/gfx/webgl2/webgl2-commands.ts
```diff
@@ -3406,23 +3406,30 @@ export function WebGL2CmdFuncBlitTexture (
...
+        gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER, null);
+        cache.glFramebuffer = null;
+    }
+    if (!isSwapchain) {
+        if (dst.glTexture) {
+            gl.framebufferTexture2D(gl.DRAW_FRAMEBUFFER, dstAttachment, dst.glTarget, dst.glTexture, dstMip);
+        } else {
+            gl.framebufferRenderbuffer(gl.DRAW_FRAMEBUFFER, dstAttachment, gl.RENDERBUFFER, dst.glRenderbuffer);
+        }
     }
```
minggo, line 3432: It is better to make the logic more clearly like this:
```javascript
if (isSwapchain) {
    if (cache.glFramebuffer !== null) {
        gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER, null);
        cache.glFramebuffer = null;
    }
} else {
    if (cache.glFramebuffer !== dstFramebuffer) {
        gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER, dstFramebuffer);
        cache.glFramebuffer = dstFramebuffer;
    }

    if (dst.glTexture) {
        gl.framebufferTexture2D(gl.DRAW_FRAMEBUFFER, dstAttachment, dst.glTarget, dst.glTexture, dstMip);
    } else {
        gl.framebufferRenderbuffer(gl.DRAW_FRAMEBUFFER, dstAttachment, gl.RENDERBUFFER, dst.glRenderbuffer);
    }
}
```

cocos/gfx/base/device.ts
```diff
@@ -77,14 +77,6 @@ export abstract class Device {
         return this._cmdBuff as CommandBuffer;
     }
 
-    /**
-     * @en The current format of the swapchain being used.
-     * @zh 当前使用的swapchain的格式。
-     */
-    get swapchainFormat (): Format {
-        return this._swapchainFormat;
```
dumganhar, line 85: gfx interfaces are public.
Will it  cause compatiblity issue if removing the getter?

cocos/gfx/webgl2/webgl2-commands.ts
```diff
@@ -3406,23 +3406,30 @@ export function WebGL2CmdFuncBlitTexture (
...
         gl.framebufferRenderbuffer(gl.READ_FRAMEBUFFER, srcAttachment, gl.RENDERBUFFER, src.glRenderbuffer);
     }
-
-    if (dst.glTexture) {
-        gl.framebufferTexture2D(gl.DRAW_FRAMEBUFFER, dstAttachment, dst.glTarget, dst.glTexture, dstMip);
-    } else {
-        gl.framebufferRenderbuffer(gl.DRAW_FRAMEBUFFER, dstAttachment, gl.RENDERBUFFER, dst.glRenderbuffer);
+    const status = gl.checkFramebufferStatus(gl.READ_FRAMEBUFFER);
+    if (status !== gl.FRAMEBUFFER_COMPLETE) {
+        warn('Framebuffer is not complete', status);
```
dumganhar, line 3416: It's better to use `warnID` and add the warning message in EngineErrorMap.md

Issue #17012, opened 2024-05-20
Loading assets synchronously causes rendering to freeze


dumganhar:
Reported at https://forum.cocos.org/t/topic/154102

Using ThreadPool to load assets as which we do for image loading may be an option.

jordiwang:
这里3.8.5会解决吗

dumganhar:
@jordiwang 

来了：https://github.com/cocos/cocos-engine/pull/17464

PR #17479, opened 2024-08-01
[optimization] free string memory after JSON is parsed.  

dumganhar:
V8's garbage collector only take care of JS Heap and ignore the external string size.  So  ExternalStringResource::Dispose may be delayed which will cause memory in high value.

Re: #17012

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

## Greptile Summary

Optimized memory management for external strings in JSON parsing by ensuring prompt memory release.

- Modified `native/cocos/bindings/jswrapper/v8/Utils.cpp` to call `delete this` in `ExternalStringResource::Dispose`.
- Removed atomic counter `gExternalCount` for tracking external strings.
- Explicitly cleared and shrunk string memory post-parsing to reduce memory usage.
- Adjusted `native/cocos/bindings/jswrapper/v8/Object.cpp` to align with new memory management logic.

### DIFF
--- a/native/cocos/bindings/jswrapper/v8/Object.cpp
+++ b/native/cocos/bindings/jswrapper/v8/Object.cpp
@@ -401,13 +401,23 @@ Object *Object::createJSONObject(const ccstd::string &jsonStr) {
 }
 
 Object *Object::createJSONObject(std::u16string &&jsonStr) {
-    auto v8Str = v8::String::NewExternalTwoByte(__isolate, ccnew internal::ExternalStringResource(std::move(jsonStr)));
+    auto *external = ccnew internal::ExternalStringResource(std::move(jsonStr));
+    external->addRef();
+
+    auto v8Str = v8::String::NewExternalTwoByte(__isolate, external);
     if (v8Str.IsEmpty()) {
+        external->release();
         return nullptr;
     }
 
     v8::Local<v8::Context> context = __isolate->GetCurrentContext();
     v8::MaybeLocal<v8::Value> ret = v8::JSON::Parse(context, v8Str.ToLocalChecked());
+    
+    // After v8::JSON::Parse, the memory of u16string could be freed.
+    external->freeMemory();
+    external->release();
+    external = nullptr;
+    
     if (ret.IsEmpty()) {
         return nullptr;
     }
--- a/native/cocos/bindings/jswrapper/v8/Object.cpp
+++ b/native/cocos/bindings/jswrapper/v8/Object.cpp
@@ -402,11 +402,8 @@ Object *Object::createJSONObject(const ccstd::string &jsonStr) {
 
 Object *Object::createJSONObject(std::u16string &&jsonStr) {
     auto *external = ccnew internal::ExternalStringResource(std::move(jsonStr));
-    external->addRef();
-
     auto v8Str = v8::String::NewExternalTwoByte(__isolate, external);
     if (v8Str.IsEmpty()) {
-        external->release();
         return nullptr;
     }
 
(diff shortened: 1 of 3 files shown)


native/cocos/bindings/jswrapper/v8/Utils.cpp
```diff
@@ -243,6 +243,39 @@ void clearPrivate(v8::Isolate *isolate, ObjectWrap &wrap) {
     }
 }
 
+// ExternalStringResource
+ExternalStringResource::ExternalStringResource(std::u16string &&s)
+: _s(std::move(s)) {
```
minggo, line 248: s is right reference, should use std::move?

native/cocos/bindings/jswrapper/v8/Object.cpp
```diff
@@ -401,13 +401,23 @@ Object *Object::createJSONObject(const ccstd::string &jsonStr) {
 }
 
 Object *Object::createJSONObject(std::u16string &&jsonStr) {
-    auto v8Str = v8::String::NewExternalTwoByte(__isolate, ccnew internal::ExternalStringResource(std::move(jsonStr)));
+    auto *external = ccnew internal::ExternalStringResource(std::move(jsonStr));
+    external->addRef();
```
minggo: The default reference count is 1, should add ref here?

dumganhar:
@cocos-robot run test cases

native/cocos/bindings/jswrapper/v8/Object.cpp
dumganhar: I think about again. It seems doesn't need the reference count here.
Because `auto v8Str = v8::String::NewExternalTwoByte(__isolate, external);` will keep the external buffer until GC comes, so it's safe to invoke `freeMemory` after `v8::JSON::Parse`.

I will update it.

dumganhar: UPDATED, remove reference count.

native/cocos/bindings/jswrapper/v8/Utils.cpp
dumganhar: Yes, need it.
Test code:

```cpp
class Foo {
public:
    Foo(int a)
    : a(a)
    {
        printf("constructor\n");
    }
    
    ~Foo() {
        printf("~Foo: %d\n", a);
    }
    
    Foo(const Foo& o) {
        a = o.a;
        printf("copy constructor ...\n");
    }
    
    Foo(Foo && o) {
        a = o.a;
        o.a = 0;
        printf("move constructor ...\n");
    }
    
    Foo& operator=(const Foo &o) {
        a = o.a;
        printf("copy assignment ...\n");
        return *this;
    }
    
    Foo& operator=(Foo &&o) {
        a = o.a;
        o.a = 0;
        printf("move assignment ...\n");
        return *this;
    }
    
    int a = 0;
};
```

```cpp
    class Bar {
    public:
        Bar(Foo && foo)
        : _foo(std::move(foo))
        {}
        
    private:
        Foo _foo;
    };
    
    {
        Foo hello(123);
        Bar b(std::move(hello));
        printf("after bar\n");
    }
/*
constructor
move constructor ...
after bar
~Foo: 123
~Foo: 0
*/
```

```cpp
    class Bar {
    public:
        Bar(Foo && foo)
        : _foo(foo)
        {}
        
    private:
        Foo _foo;
    };
    
    {
        Foo hello(123);
        Bar b(std::move(hello));
        printf("after bar\n");
    }

/*
constructor
copy constructor ...
after bar
~Foo: 123
~Foo: 123
*/
```

minggo: good to know

minggo (approved):

PR #17496, opened 2024-08-06
Reduce graph api code

star-e:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

## Greptile Summary

The pull request renames several graph-related methods across multiple files in the Cocos Engine, focusing on brevity and consistency in naming conventions.

- Renamed 'outEdges' to 'oe', 'outDegree' to 'od', 'inEdges' to 'ie', and 'inDegree' to 'id' in `cocos/rendering/custom/graph.ts`.
- Updated `LayoutGraph` and `LayoutGraphData` classes in `cocos/rendering/custom/layout-graph.ts` with similar method name changes.
- Modified `SubpassGraph`, `ResourceGraph`, and `RenderGraph` classes in `cocos/rendering/custom/render-graph.ts` to use the new shortened method names.
- These changes may impact code readability and require updates in other parts of the codebase that rely on the old method names.

### DIFF
--- a/cocos/rendering/custom/compiler.ts
+++ b/cocos/rendering/custom/compiler.ts
@@ -344,7 +344,7 @@ class PassManagerVisitor extends DefaultVisitor {
 
     set resId (value: number) {
         this._resId = value;
-        this._colorMap.colors.length = context.renderGraph.numVertices();
+        this._colorMap.colors.length = context.renderGraph.nv();
     }
     get resId (): number {
         return this._resId;
@@ -354,7 +354,7 @@ class PassManagerVisitor extends DefaultVisitor {
         this._resId = resId;
         this._passVisitor = new PassVisitor(context);
         this._graphView = new ReferenceGraphView<RenderGraph>(context.renderGraph);
-        this._colorMap = new VectorGraphColorMap(context.renderGraph.numVertices());
+        this._colorMap = new VectorGraphColorMap(context.renderGraph.nv());
     }
     get graphView (): ReferenceGraphView<RenderGraph> { return this._graphView; }
     get colorMap (): VectorGraphColorMap { return this._colorMap; }
@@ -460,7 +460,7 @@ export class Compiler {
     compile (rg: RenderGraph): void {
         context.set(this._pipeline, this._resourceGraph, rg, this._layoutGraph);
         context.pipeline.resourceUses.length = 0;
-        this._visitor.colorMap.colors.length = context.resourceGraph.numVertices();
+        this._visitor.colorMap.colors.length = context.resourceGraph.nv();
         depthFirstSearch(this._resourceGraph, this._visitor, this._visitor.colorMap);
 
         if (DEBUG) {
@@ -516,7 +516,7 @@ export class ResourceManagerVisitor extends DefaultVisitor {
     private _resVisitor: ResourceVisitor;
     constructor (context: CompilerContext) {
         super();
-        this._colorMap = new VectorGraphColorMap(context.resourceGraph.numVertices());
+        this._colorMap = new VectorGraphColorMap(context.resourceGraph.nv());
         this._resourceGraph = context.resourceGraph;
         this._resVisitor = new ResourceVisitor(context);
     }
--- a/cocos/rendering/custom/define.ts
+++ b/cocos/rendering/custom/define.ts
@@ -426,7 +426,7 @@ export function getDescBinding (descId, descData: DescriptorSetData): number {
 export function getDescBindingFromName (bindingName: string): number {
     const pipeline = cclegacy.director.root.pipeline as WebPipeline;
     const layoutGraph = pipeline.layoutGraph;
-    const vertIds = layoutGraph.vertices();
+    const vertIds = layoutGraph.v();
     const descId = layoutGraph.attributeIndex.get(bindingName);
     let currDesData: DescriptorSetData;
     for (const i of vertIds) {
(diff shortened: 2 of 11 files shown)


smallmain:
It's scary...

star-e:
> It's scary...

Yes, it seems scary.

These methods are only used in engine implementation. To reduce code size, methods and members will be shortened aggressively.
Readability will be poor, so we are adding TS Doc comments to improve reasoning.

smallmain:
> > It's scary...
> 
> Yes, it seems scary.
> 
> These methods are only used in engine implementation. To reduce code size, methods and members will be shortened aggressively. Readability will be poor, so we are adding TS Doc comments to improve reasoning.



> > It's scary...
> 
> Yes, it seems scary.
> 
> These methods are only used in engine implementation. To reduce code size, methods and members will be shortened aggressively. Readability will be poor, so we are adding TS Doc comments to improve reasoning.

I looked at the Terser.

​https://terser.org/docs/cli-usage/#cli-mangling-property-names---mangle-props

It seems like there are plenty of options to restrict Mangle to certain properties on certain files, so why go to the trouble of doing it manually? 

Even if you create a 'mangle.json' list file, maintain which files and which properties can be reduced, at least it won't affect development, and is it easier to maintain?

star-e:
> I looked at the Terser.
> 
> ​https://terser.org/docs/cli-usage/#cli-mangling-property-names---mangle-props
> 
> It seems like there are plenty of options to restrict Mangle to certain properties on certain files, so why go to the trouble of doing it manually?
> 
> Even if you create a 'mangle.json' list file, maintain which files and which properties can be reduced, at least it won't affect development, and is it easier to maintain?

We are currently looking at tools that can reduce code. Before we finding a suitable tool, the code size still needs to be reduced.
There are also several other reasons.
1. These codes are generated and the generator is generally easier to maintain. (It also has a list to maintain which properties are reduced.)
2. Introducing another tool might make the building process more error prone. And bugs will be harder to catch.
3. We need to handle TS/C++ inter-operation. In C++ these names are not shortened. The new tool might not able to handle this situation. Currently we are using the generator to solve this problem.

smallmain:
> > I looked at the Terser.
> > ​https://terser.org/docs/cli-usage/#cli-mangling-property-names---mangle-props
> > It seems like there are plenty of options to restrict Mangle to certain properties on certain files, so why go to the trouble of doing it manually?
> > Even if you create a 'mangle.json' list file, maintain which files and which properties can be reduced, at least it won't affect development, and is it easier to maintain?
> 
> We are currently looking at tools that can reduce code. Before we finding a suitable tool, the code size still needs to be reduced. There are also several other reasons.
> 
> 1. These codes are generated and the generator is generally easier to maintain. (It also has a list to maintain which properties are reduced.)
> 2. Introducing another tool might make the building process more error prone. And bugs will be harder to catch.
> 3. We need to handle TS/C++ inter-operation. In C++ these names are not shortened. The new tool might not able to handle this situation. Currently we are using the generator to solve this problem.

That's fine, as long as we don't make the changes manually.

Terser is a popular library and most likely already in the engine's build chain.

I prefer to make improvements in existing tools (including avoiding dealing with TS/C++ glue code) that can be applied to user/engine code more quickly.

smallmain:
> > I looked at the Terser.
> > ​https://terser.org/docs/cli-usage/#cli-mangling-property-names---mangle-props
> > It seems like there are plenty of options to restrict Mangle to certain properties on certain files, so why go to the trouble of doing it manually?
> > Even if you create a 'mangle.json' list file, maintain which files and which properties can be reduced, at least it won't affect development, and is it easier to maintain?
> 
> We are currently looking at tools that can reduce code. Before we finding a suitable tool, the code size still needs to be reduced. There are also several other reasons.
> 
> 1. These codes are generated and the generator is generally easier to maintain. (It also has a list to maintain which properties are reduced.)
> 2. Introducing another tool might make the building process more error prone. And bugs will be harder to catch.
> 3. We need to handle TS/C++ inter-operation. In C++ these names are not shortened. The new tool might not able to handle this situation. Currently we are using the generator to solve this problem.

If it's automatic, I'm even less clear, why commit to the codebase instead of processing it as a build artifact?

star-e:
> If it's automatic, I'm even less clear, why commit to the codebase instead of processing it as a build artifact?

We will add it to build process in the future. Currently we are focusing on other parts of the engine.

smallmain:
> > If it's automatic, I'm even less clear, why commit to the codebase instead of processing it as a build artifact?
> 
> We will add it to build process in the future. Currently we are focusing on other parts of the engine.

Okay, thank you for your answer.

minggo (approved):

dumganhar:
> Introducing another tool might make the building process more error prone. And bugs will be harder to catch.

I agree this. At this point, we don't have time to test `terser mangle configuration`.
Last week, I played `rollup's terser plugin` , I wrote a `mangle regex`: `properties: \w+\$$` means every property ends with `$` should be reduced.

And I found a potential issue that it generate wrong code.

Something like:


```ts
// __virtual-cc-xxx.js
export("a$") {... }
```

```ts
// cc.js
Size = module.ne; // but here should be `a$`.
```

It's probably a bug in rollup or terser, which generates wrong code in module's chunk file.

In the end, I have to modify the regex to `regex: /^[a-zA-Z_][a-zA-Z0-9_]{3,}\$$/,` to avoid that bug.

PR #17529, opened 2024-08-12
Remove unused code in gfx/base, gfx/webgl module and mangle properties to reduce js size.

dumganhar:
Add WebGLConstants.ts for inlining enums. Depends on: https://github.com/cocos/cocos-ccbuild/pull/65

https://github.com/cocos/cocos-engine/pull/17664

After this changes, the package size could reduce 100KB.

gfx WebGL2 backend：187KB -> 103KB

Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

## Greptile Summary

The pull request focuses on reducing JavaScript size by mangling property names and removing unused code in the gfx/base and gfx/webgl modules.

- **Property Renaming**: Private properties across multiple files have been renamed by appending a `$` symbol to their names (e.g., `_layout` to `_layout$`). This affects files like `cocos/gfx/base/buffer.ts`, `cocos/gfx/base/command-buffer.ts`, and others.
- **Import Path Updates**: Import paths for functions like `murmurhash2_32_gc` have been updated to be more specific in files such as `cocos/gfx/base/render-pass.ts`.
- **New Utility Functions**: Introduced a new `ceil` function in `cocos/gfx/base/define.ts` to replace multiple instances of `Math.ceil`.
- **Consistency Across Codebase**: Ensured that all references to renamed properties are updated accordingly to prevent runtime errors.
- **No Functional Changes**: The changes are syntactical and should not affect the runtime behavior or introduce new functionality.


cocos/gfx/webgl/webgl-commands.ts
```diff
@@ -22,114 +22,119 @@
...
 } from '../base/define';
-import { WebGLStateCache } from './webgl-state-cache';
+
+import { WebGLConstants } from '../gl-constants';
+import { assertID, debugID, error, errorID } from '../../core/platform/debug';
+import { cclegacy } from '../../core/global-exports';
+
+function max (a: number, b: number): number {
+    return Math.max(a, b);
+}
```
minggo: It adds one function invoking, i don't think it is not a good idea to do it like this.

cocos/gfx/base/define.ts
```diff
@@ -2163,6 +2161,10 @@ export function IsPowerOf2 (x: number): boolean {
     return x > 0 && (x & (x - 1)) === 0;
 }
 
+function ceil (x: number): number {
+    return Math.ceil(x);
+}
```
minggo: It introduces a function call. So i don't think it is a good idea to do it like this to reduce pacakge size.

cocos/gfx/webgl/webgl-commands.ts
dumganhar: UPDATED:

Use function reference instead of defining a function for Math.ceil and Math.max
```ts
const max = Math.max;
```

cocos/gfx/base/define.ts
dumganhar: UPDATED:

Use function reference instead of defining a function for Math.ceil and Math.max
```ts
const ceil = Math.ceil;
```

dumganhar: https://github.com/cocos/cocos-engine/pull/17529/commits/5035f0b59c007c7385865e589d0a13b59f6fc3fb

cocos/gfx/webgl/webgl-commands.ts
dumganhar: https://github.com/cocos/cocos-engine/pull/17529/commits/5035f0b59c007c7385865e589d0a13b59f6fc3fb

dumganhar:
TODO: 
- [x] gfx-webgl2 is also needed to optimized.
- [ ] Need to check whether source map works.

dumganhar:
@cocos-robot run test cases

dumganhar:
@cocos-robot run test cases

minggo (approved):

cocos/gfx/gl-constants.ts
```diff
@@ -0,0 +1,562 @@
+/* eslint-disable @typescript-eslint/no-duplicate-enum-values */
+export enum WebGLConstants {
```
star-e, line 2: What is this `gl-constants.ts` file for?

star-e: Shall we use `const enum`, since these values will never be changed.

dumganhar: Currently, both enum and `const enum` will be `inlined` since I wrote a custom babel plugin in ccbuild.
I remember that I got same errors when using `const enum`. 
It seems that we could use const enum here, but could not use `const enum` for the exported type since that will break compatibility.


> What is this gl-constants.ts file for?
The default WebGL constants is not enum which will not be inlined. So I add this file to inline the enums.
For example:
```ts
//libdom.d.ts
interface WEBGL_color_buffer_float {
    readonly RGBA32F_EXT: 0x8814;
    readonly FRAMEBUFFER_ATTACHMENT_COMPONENT_TYPE_EXT: 0x8211;
    readonly UNSIGNED_NORMALIZED_EXT: 0x8C17;
}
```

dumganhar:
@cocos-robot run test cases

star-e (approved):

PR #17539, opened 2024-08-15
export Bullet

minggo:
Bullet is exported until this PR: https://github.com/cocos/cocos-engine/pull/16350. I don't know why that PR removed the codes as the modification out of the range of its title.

And there is forum ticket discussion it: https://forum.cocos.org/t/topic/158556.

## Greptile Summary

This PR reintroduces the global export of the Bullet physics library in the Cocos Engine, addressing a previous removal and community discussion.

- Added `globalThis.Bullet = bt as any;` in `cocos/physics/bullet/instantiated.ts` to make Bullet globally accessible after initialization
- The change ensures Bullet is only exposed globally after it has been fully initialized
- While addressing the need for global access, this approach may introduce potential security risks and namespace pollution
- Consider exploring alternative methods to provide necessary access while maintaining better encapsulation

### DIFF
--- a/cocos/physics/bullet/instantiated.ts
+++ b/cocos/physics/bullet/instantiated.ts
@@ -87,7 +87,6 @@ interface BtCache {
 
 // eslint-disable-next-line import/no-mutable-exports
 export let bt = {} as Bullet.instance;
-globalThis.Bullet = bt as any;
 export const btCache = {} as BtCache;
 btCache.BODY_CACHE_NAME = 'body';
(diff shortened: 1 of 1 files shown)


dumganhar (approved):

PPpro:
I think we can add a comment here, or it will  be removed next time because the global Bullet is used nowhere in engine

@minggo 

it's also not a good practice to export Bullet globally

minggo:
Developers can invoke Bullet features as they need. Cocos can not wrap all features.

PR #17592, opened 2024-08-31
Update node.ts setSiblingIndex support negative index

kaokei:
eg.
-1 equals siblings.length -1.
-2 equals siblings.length -2.
-3 equals siblings.length -3.

Re: #

### Changelog

* Support negative index in setSiblingIndex method

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

## Greptile Summary

This PR updates the setSiblingIndex method in the Node class to support negative indices for more convenient node positioning within the scene graph.

- Modifies `setSiblingIndex` in `cocos/scene-graph/node.ts` to handle negative indices
- Allows positioning nodes relative to the end of the sibling list (e.g. -1 for last position)
- Affects core functionality of node positioning and hierarchy management
- Requires careful review to ensure consistency with existing behavior
- May impact other engine components that rely on sibling indexing

### DIFF
--- a/cocos/scene-graph/node.ts
+++ b/cocos/scene-graph/node.ts
@@ -630,7 +630,7 @@ export class Node extends CCObject implements ISchedulable, CustomSerializable {
             return;
         }
         const siblings = this._parent._children;
-        index = index !== -1 ? index : siblings.length - 1;
+        index = index >= 0 ? index : siblings.length + index;
         const oldIndex = siblings.indexOf(this);
         if (index !== oldIndex) {
             siblings.splice(oldIndex, 1);
--- a/native/cocos/core/scene-graph/Node.cpp
+++ b/native/cocos/core/scene-graph/Node.cpp
@@ -353,7 +353,8 @@ void Node::setSiblingIndex(index_t index) {
         return;
     }
     ccstd::vector<IntrusivePtr<Node>> &siblings = _parent->_children;
-    index = index != -1 ? index : static_cast<index_t>(siblings.size()) - 1;
+    index = index >= 0 ? index : static_cast<index_t>(siblings.size()) + index;
+    index = index >= 0 ? index : 0;
     index_t oldIdx = getIdxOfChild(siblings, this);
     if (index != oldIdx) {
         if (oldIdx != CC_INVALID_INDEX) {


minggo:
It will break compatibility.

kaokei:
> It will break compatibility.

@minggo 
Is there something I haven't considered?
From my understanding, the original code `index = index !== -1 ? index : siblings.length - 1;` only accounts for scenarios where the index is a positive number or -1.
However, the new code `index = index >= 0 ? index : siblings.length + index;` not only supports positive numbers but also supports a wider range of negative numbers.
The new code should encompass the behavior of the old one. Could you provide some cases to illustrate where there are breaking changes?

minggo:
My fault. I doesn't break compatibility.

And this PR just make code more clear, it doesn't support more index range. As the current implementation also supports -2 and -3.

cocos/scene-graph/node.ts
```diff
@@ -630,7 +630,7 @@ export class Node extends CCObject implements ISchedulable, CustomSerializable {
             return;
         }
         const siblings = this._parent._children;
-        index = index !== -1 ? index : siblings.length - 1;
+        index = index >= 0 ? index : siblings.length + index;
         const oldIndex = siblings.indexOf(this);
```
dumganhar, line 634: Would it run well if passing a negative value that smaller than `-siblings.length`?

dumganhar:
https://github.com/cocos/cocos-engine/blob/v3.8.5/native/cocos/core/scene-graph/Node.cpp#L347

Also need to modify the cpp code for native platforms.

kaokei:
> https://github.com/cocos/cocos-engine/blob/v3.8.5/native/cocos/core/scene-graph/Node.cpp#L347
> 
> Also need to modify the cpp code for native platforms.

@dumganhar 
Actually, I'm not familiar with C++ code. Could you help me review if this implementation is feasible? The main change is adding logic to force the index to be greater than or equal to 0.

cocos/scene-graph/node.ts
dumganhar: @kaokei 
Could you take a look at this question?
If `index` < `-siblings.length`, for example, it's `-siblings.length - 10`,
After this logic `index = index >= 0 ? index : siblings.length + index;`
The index will be `siblings.length + (-siblings.length - 10)` = `-10`, -10 will be a invalid index.

dumganhar:
> @dumganhar
> Actually, I'm not familiar with C++ code. Could you help me review if this implementation is feasible? The main change is adding logic to force the index to be greater than or equal to 0.

OK, I will be glad to update the cpp code later.

dumganhar:
@kaokei I found you have already updated the cpp code. It looks good.

cocos/scene-graph/node.ts
dumganhar: It seems `splice` will check whether the index is in the range and do a clamp operation.

dumganhar (approved):

native/cocos/core/scene-graph/Node.cpp
```diff
@@ -353,7 +353,8 @@ void Node::setSiblingIndex(index_t index) {
         return;
     }
     ccstd::vector<IntrusivePtr<Node>> &siblings = _parent->_children;
-    index = index != -1 ? index : static_cast<index_t>(siblings.size()) - 1;
+    index = index >= 0 ? index : static_cast<index_t>(siblings.size()) + index;
+    index = index >= 0 ? index : 0;
```
minggo, line 357: if index < 0, then `siblings.insert(siblings.begin() + index, this);` will crash.

minggo (approved):

minggo:
@kaokei could you please send it to v3.8.5? Thanks.

kaokei:
> @kaokei could you please send it to v3.8.5? Thanks.

done. @dumganhar

minggo:
@cocos-robot run test cases

PR #17595, opened 2024-09-03
[Optimized] upload buffer

GengineJS:
## Greptile Summary

This pull request optimizes buffer management in the custom rendering pipeline, focusing on improving performance and reducing unnecessary updates. Key changes include:

- Introduced `currUniform` property in `DescBuffManager` class for more efficient uniform handling
- Implemented single buffer approach in `DescBuffManager`, simplifying buffer management
- Added change detection before buffer updates to minimize unnecessary operations
- Introduced `idxRD` parameter in several functions for improved indexing in buffer-related operations
- Enhanced error handling and boundary checking in buffer-related functions for increased stability

These changes should lead to improved rendering performance by reducing buffer operations and optimizing memory usage.

### DIFF
--- a/cocos/rendering/custom/define.ts
+++ b/cocos/rendering/custom/define.ts
@@ -448,15 +448,16 @@ export function getDescBindingFromName (bindingName: string): number {
     return -1;
 }
 
-const uniformMap: Map<Buffer, Float32Array> = new Map();
 class DescBuffManager {
     private buffers: Buffer[] = [];
     private currBuffIdx: number = 0;
     private device: Device;
+    public currUniform: Float32Array;
     constructor (bufferSize: number, numBuffers: number = 2) {
         const root = cclegacy.director.root;
         const device = root.device;
         this.device = device;
+        this.currUniform = new Float32Array(bufferSize / 4);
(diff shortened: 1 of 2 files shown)


cocos/rendering/custom/define.ts
```diff
@@ -468,36 +469,37 @@ class DescBuffManager {
         }
     }
     getCurrentBuffer (): Buffer {
+        if (this.buffers.length === 1) {
+            return this.buffers[0];
+        }
```
star-e: Why is this check needed? I think `this.buffers` is correctly resized in `constructor`.

GengineJS: This is a determination of the array length, not the internal length of the buffer.

minggo (approved):

cocos/rendering/custom/define.ts
star-e: I mean we can remove this guard, the rest of the code still works. I think this is an optimization, because when `this.buffers.length === 1`, we don't need to do the modulo?

star-e (approved):

PR #17624, opened 2024-09-10
Bugfix: preloadScene would not return the actual data on the onComplete callback

MrcSnm:
### Changelog
preloadScene would not return the actual data on the onComplete callback, this caused issues on the following code.

```ts
director.preloadScene("scene", (err, scene) =>
{ 
      //Always true since no scene is being sent
      if(!scene) throw err;
});
```
*

-------

### Continuous Integration

This pull request:

* [ ] needs automatic test cases check.
  > Manual trigger with `@cocos-robot run test cases` afterward.
* [ ] does not change any runtime related code or build configuration
  > If any reviewer thinks the CI checks are needed, please uncheck this option, then close and reopen the issue.

-------

### Compatibility Check

This pull request:

* [ ] changes public API, and have ensured backward compatibility with [deprecated features](https://github.com/cocos/cocos-engine/blob/v3.5.0/docs/contribution/deprecated-features.md).
* [ ] affects platform compatibility, e.g. system version, browser version, platform sdk version, platform toolchain, language version, hardware compatibility etc.
* [ ] affects file structure of the build package or build configuration which requires user project upgrade.
* [ ] **introduces breaking changes**, please list all changes, affected features and the scope of violation.

PR #17625, opened 2024-09-10
Show correct type for preloadScene

MrcSnm:
Re: #17624 

### Changelog
The old type was misleading
*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/game/director.ts
+++ b/cocos/game/director.ts
@@ -447,7 +447,7 @@ export class Director extends EventTarget {
      * @param sceneName @en The name of the scene to load @zh 场景名称。
      * @param onLoaded @en Callback to execute once the scene is loaded @zh 加载回调。
      */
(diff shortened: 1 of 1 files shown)


cocos/game/director.ts
```diff
@@ -819,6 +819,7 @@ export declare namespace Director {
 
     export type OnUnload = () => void;
 
+    export type OnScenePreloaded = (error?: Error) => void;
```
minggo, line 822: Yep, rename it to OnScenePreloaded is more clear. But it will break compatibility.

MrcSnm: Yeah the main problem is that the current interface is really misleading since the extra argument. I wanted instead to at least make it valid

minggo: Sorry, as keep compatible is important. So can not merge this PR. And as v3.8.5 is released. So this branch will be deleted, only add a tag for 3.8.5.

PR #17671, opened 2024-09-27
 uses physx shipped with engine on bytedance

minggo:
@cocos-robot run test cases


dumganhar:
@cocos-robot run test cases

dumganhar (approved):

dumganhar:
@cocos-robot run test cases

PR #17697, opened 2024-10-10
do not terminate loop if mask mismatch

minggo:
- If ReportFixture returns 0, then it will terminate the loop for raycast(Can refer to box2d source code for detail information). So just return -1 here.
- Fix eslint issue: add type for variables and return correct type

### DIFF
--- a/cocos/physics-2d/box2d-wasm/platform/physics-ray-cast-callback.ts
+++ b/cocos/physics-2d/box2d-wasm/platform/physics-ray-cast-callback.ts
@@ -44,9 +44,9 @@ export class PhysicsRayCastCallback {// extends B2.RayCastCallback {
         PhysicsRayCastCallback._fractions.length = 0;
     }
 
-    static ReportFixture (fixture: number, point: B2.Vec2, normal: B2.Vec2, fraction: number): any {
+    static ReportFixture (fixture: number, point: B2.Vec2, normal: B2.Vec2, fraction: number): number {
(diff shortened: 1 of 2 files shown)


cocos/physics-2d/box2d-wasm/platform/physics-ray-cast-callback.ts
```diff
@@ -44,9 +44,9 @@ export class PhysicsRayCastCallback {// extends B2.RayCastCallback {
         PhysicsRayCastCallback._fractions.length = 0;
     }
 
-    static ReportFixture (fixture: number, point: B2.Vec2, normal: B2.Vec2, fraction: number): any {
+    static ReportFixture (fixture: number, point: B2.Vec2, normal: B2.Vec2, fraction: number): number {
         if ((B2.FixtureGetFilterData(fixture).categoryBits & PhysicsRayCastCallback._mask) === 0) {
-            return 0;
+            return -1;
```
dumganhar, line 49: What does `-1` mean?

minggo: Can refer to https://github.com/cocos/box2d_wasm/blob/wasm/include/box2d/b2_world_callbacks.h#L154 for more information.

```
-1 to filter, 0 to terminate, fraction to clip the ray for closest hit, 1 to continue
```

dumganhar (approved):

PR #17708, opened 2024-10-14
Fix issue with registering multiple event listeners for a single event

onerain88:
The use of `arr.find(listener)` can cause previously added listeners to be executed during the search process when multiple listeners are added. Therefore, a modification has been made.

Currently, there is an ESLint error in the `<Web> Quality / Run ESLint`, and it seems that there might be an issue with the ESLint configuration. I noticed that in [eslint.ps1#L36](https://github.com/cocos/cocos-engine/blob/74595e62c3375a50403ff6f4f882779ed280e7d8/.github/workflows/eslint.ps1#L36), it filters files under the cocos directory, but the tsconfig.json does not include files under the cocos directory.
However, since I am not very familiar with the project, I have not made any hasty modifications.

### Changelog

* Fix issue with registering multiple event listeners for a single event

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/native-binding/impl.ts
+++ b/cocos/native-binding/impl.ts
@@ -69,7 +69,7 @@ if (NATIVE) {
                 this.eventMap.set(eventName, []);
             }
             const arr = this.eventMap.get(eventName);
-            if (!arr.find(listener)) {
+            if (!arr.includes(listener)) {
                 arr.push(listener);
             }
         },


minggo:
> The use of arr.find(listener) can cause previously added listeners to be executed during the search process when multiple listeners are added.

I don't quite understand why it has problem. Could you please describe in more detail? Thanks.

onerain88:
> > The use of arr.find(listener) can cause previously added listeners to be executed during the search process when multiple listeners are added.
> 
> I don't quite understand why it has problem. Could you please describe in more detail? Thanks.

You can try running the following code, which is extracted from `impl.ts`, and you will find that during the process of `if (!arr.find(listener))`, the body of the listener function is executed.

```js
const eventMap = new Map();

function addNativeEventListener(eventName, listener) {
  if (!eventMap.get(eventName)) {
    eventMap.set(eventName, []);
  }
  const arr = eventMap.get(eventName);
  if (!arr.find(listener)) {
    arr.push(listener);
  }
}

function listener() {
  console.log('listen');
}

addNativeEventListener('event', listener);
addNativeEventListener('event', listener);
```

output:

```sh
node .\index.js
listen
```

minggo:
Thanks. I think it is better to change to use `arr.includes(listener)`

minggo (approved):

PR #17710, opened 2024-10-14
raycast result includes closest hit fraction

minggo:
https://github.com/cocos/cocos-engine/issues/17672


minggo:
Have to change other physics engines implementation.

dumganhar (approved):

PR #17742, opened 2024-10-19
support friction float

longchuan:
support friction float

### DIFF
--- a/cocos/rendering/custom/executor.ts
+++ b/cocos/rendering/custom/executor.ts
@@ -1318,7 +1318,7 @@ class DeviceRenderScene implements RecordingInterface {
     public record (): void {
         const devicePass = this._currentQueue.devicePass;
         const sceneCulling = context.culling;
-        if (!devicePass.indexOfRD) this._updateRenderData();
+        this._updateRenderData();
         this._applyViewport();
         // Currently processing blit and camera first
         if (this.blit) {


cocos/physics/framework/assets/physics-material.ts
```diff
@@ -119,11 +117,11 @@ export class PhysicsMaterial extends Asset {
...
     @type(CCFloat)
     @range([0, 1, 0.01])
-    get restitution (): number {
+    get restitution(): number {
         return this._restitution;
     }
 
-    set restitution (value) {
+    set restitution(value) {
         if (!math.equals(this._restitution, value)) {
```
dumganhar: Do we also need to fix `_restitution` here since it's also a float?

cocos/physics/framework/assets/physics-material.ts
```diff
@@ -61,15 +61,13 @@ export class PhysicsMaterial extends Asset {
      */
     @editable
     @type(CCFloat)
-    get friction (): number {
+    get friction(): number {
         return this._friction;
     }
 
-    set friction (value) {
-        if (!math.equals(this._friction, value)) {
```
dumganhar: Perhaps, we need to check whether the value is changed by `math.approx` rather than `math.equals`.
Because if value is not changed, the `PhysicsMaterial.EVENT_UPDATE` event should not be emitted.

PR #17761, opened 2024-10-22
support friction CCFLOAT

longchuan:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ native/external/
 
 !templates/**/*
 templates/**/.DS_Store
+templates/**/.gradle
 
 platforms/runtime/local-commit.json
 platforms/runtime/common/web-adapter.js
--- a/cocos/physics/framework/assets/physics-material.ts
+++ b/cocos/physics/framework/assets/physics-material.ts
@@ -66,10 +66,8 @@ export class PhysicsMaterial extends Asset {
     }
 
     set friction (value) {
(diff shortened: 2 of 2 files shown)


minggo:
Current implementation also supports float though it is strict equal. What's wrong with current implementation?

longchuan:
> Current implementation also supports float though it is strict equal. What's wrong with current implementation?

js' float support many dots,and bullet's float support many dots too.if the friction was bigger than one,the friction could be max,So,the dots should more than six.Limit the float's dot was meaningless setting.The same problem as position's edit.Editor limits the value,but the scene file has no limits.That is conflictive.

minggo:
I don't quite understand the issue. Could you give an example to explain it?

longchuan:
> I don't quite understand the issue. Could you give an example to explain it?

image
look at this,if I set the friction to 3e-15 or 4e-15,there will have diffrent results.But if I set the friction to 1 or 2,the result will not be different.

minggo:
1 and 2 should be different.

longchuan:
> 1 and 2 should be different.

image
reffer to the requirement, the ball could not output.and blocked by the friction
for the output,the rollingFriction should be a large one.
one friction has one result,I got the different results by editing friction.

PR #17782, opened 2024-10-29
reuse _tempVec2 and avoid creating new temple object

minggo:
_calculateMovePercentDelta() just uses anchor.x and anchor.y, and will not pass it to other functions, so it is safe to reuse _moveDeltaOptions and _tempVec2.

The best solution is change the parameter from `_calculateMovePercentDelta(options: MoveDeltaOptions) ` to `_calculateMovePercentDelta(anchor: Vec2, applyToHorizontal: boolean, applyToVertical: boolean)`. Then even don't have to:
- define MoveDeltaOptions
- create _moveDeltaOptions
- and don't have assign _moveDeltaOptions before invoking _calculateMovePercentDelta()

But it will break the compatibility, as _calculateMovePercentDelta() is a protected function.

### DIFF
--- a/cocos/ui/scroll-view.ts
+++ b/cocos/ui/scroll-view.ts
@@ -78,12 +78,7 @@ const eventMap = {
     'scroll-began': 12,
 };
 
-interface MoveDeltaOptions {
-    anchor: Vec2,
-    applyToHorizontal: boolean,
-    applyToVertical: boolean,
-}
-const _moveDeltaOptions: MoveDeltaOptions = {
+const _moveDeltaOptions = {
     anchor: _tempVec2,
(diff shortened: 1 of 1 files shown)


minggo:
@cocos-robot run test cases

minggo:
@cocos-robot run test cases

minggo:
@cocos-robot run test cases

sushanhong:
@cocos-robot run test cases

qiuguohua (approved):

minggo:
@cocos-robot run test cases

sushanhong:
@cocos-robot run test cases

Issue #17715, opened 2024-10-15
Reduce package size in 3.8.5


minggo:

dumganhar:
[[v3.8.5] Support inline enum and mangle private variables to reduce engine size.](https://github.com/cocos/cocos-engine/pull/17703) 
https://github.com/cocos/cocos-engine/pull/17724

PR #17783, opened 2024-10-29
[v3.8.5] Reduce package size for spine module

dumganhar:
Re: #17715

■3.8.4
-rw-r--r--  1 james  staff  653581 Oct 11 14:08 spine.asm.js
-rw-r--r--  1 james  staff   57346 Oct 11 14:08 spine.js.mem
-rwxr-xr-x  1 james  staff  351821 Oct 11 14:08 spine.wasm
-rw-r--r--  1 james  staff   56455 Oct 11 14:08 spine.wasm.js
■3.8.5
-rw-r--r--@ 1 james  staff  377837 Nov  1 06:36 spine.asm.js ( save 42%  )
-rw-r--r--@ 1 james  staff   19563 Nov  1 06:36 spine.js.mem （ save 65%）
-rwxr-xr-x@ 1 james  staff  198881 Nov  1 06:36 spine.wasm ( save 43%）
-rw-r--r--@ 1 james  staff   21944 Nov  1 06:36 spine.wasm.js （save 61%）

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:


native/cocos/editor-support/spine/SpineString.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+String::String(const char *chars, size_t len, bool own) {
+    if (!chars) {
+        _length = 0;
+        _buffer = NULL;
+    } else {
+        _length = len;
+        if (!own) {
+            _buffer = SpineExtension::calloc<char>(_length + 1, __SPINE_FILE__, __SPINE_LINE__);
+            _buffer[_length] = '\0';
```
dumganhar, line 50: Ensure the last element is `\0`.

dumganhar:
@cocos-robot run test cases

dumganhar:
@cocos-robot run test cases

.github/workflows/emscripten-patches/embind/bind.cpp
```diff
@@ -0,0 +1,205 @@
...
+        return strdup("<unknown error>");
+    }
+#else
+    return strdup(ti->name());
+#endif
+  } else {
+    //cjh char str[80];
+    // sprintf(str, "%p", reinterpret_cast<const void*>(ti));
+    // return strdup(str);
+    return pointerToHexString(reinterpret_cast<const void*>(ti));
```
dumganhar, line 69: Don't use `sprintf` since it will take in lots of code in libc.

.github/workflows/emscripten-patches/embind/bind.cpp
```diff
@@ -0,0 +1,205 @@
...
+  register_float<float>("float");
+  register_float<double>("double");
+
+/*cjh
+  _embind_register_std_string(TypeID<std::string>::get(), "std::string");
+  _embind_register_std_string(
+    TypeID<std::basic_string<unsigned char>>::get(), "std::basic_string<unsigned char>");
+  _embind_register_std_wstring(TypeID<std::wstring>::get(), sizeof(wchar_t), "std::wstring");
+  _embind_register_std_wstring(TypeID<std::u16string>::get(), sizeof(char16_t), "std::u16string");
+  _embind_register_std_wstring(TypeID<std::u32string>::get(), sizeof(char32_t), "std::u32string");
```
dumganhar, line 172: Don't register `std::string` since all code is changed to use `spine::String` which is smaller.

.github/workflows/emscripten-patches/embind/bind.cpp
```diff
@@ -0,0 +1,205 @@
...
+  register_memory_view<char>("emscripten::memory_view<char>");
+  register_memory_view<signed char>("emscripten::memory_view<signed char>");
+  register_memory_view<unsigned char>("emscripten::memory_view<unsigned char>");
+
+  register_memory_view<short>("emscripten::memory_view<short>");
+  register_memory_view<unsigned short>("emscripten::memory_view<unsigned short>");
+  register_memory_view<int>("emscripten::memory_view<int>");
+  register_memory_view<unsigned int>("emscripten::memory_view<unsigned int>");
+  register_memory_view<long>("emscripten::memory_view<long>");
+  register_memory_view<unsigned long>("emscripten::memory_view<unsigned long>");
```
dumganhar, line 191: We don't use `memory_view` feature.

cocos/spine/lib/spine-define.ts
```diff
@@ -372,173 +176,38 @@ function overrideProperty_RegionAttachment (): void {
...
-        ['bendDirection', prototype.getBendDirection, prototype.setBendDirection],
-        ['compress', prototype.getCompress, prototype.setCompress],
-        ['stretch', prototype.getStretch, prototype.setStretch],
-        ['mix', prototype.getMix, prototype.setMix],
-        ['softness', prototype.getSoftness, prototype.setSoftness],
-        ['active', prototype.getActive, prototype.setActive],
-    ];
-    propertyPolyfills.forEach((prop): void => {
-        js.getset(prototype, prop[0], prop[1], prop[2]);
-    });
```
dumganhar, line 423: Remove lots of `property` polyfill code in spine-define.ts since we could use `.property(...)` to bind them.

native/cocos/editor-support/spine-creator-support/SkeletonRenderer.cpp
```diff
@@ -1074,7 +1074,7 @@ void SkeletonRenderer::setSlotTexture(const std::string &slotName, cc::Texture2D
             region->setRendererObject(attachmentVertices);
         }
         V3F_T2F_C4B *vertices = attachmentVertices->_triangles->verts;
-        auto UVs = region->getUVs();
+        const auto &UVs = region->getUVs();
```
dumganhar, line 1077: getUVs return `reference` so use `const auto &` to avoid Vector copy.

native/cocos/editor-support/spine-creator-support/Vector2.cpp
```diff
@@ -17,15 +17,15 @@ void Vector2::setX(float x) {
...
+float Vector2::getX() const {
     return x;
 }
 
 void Vector2::setY(float y) {
     this->y = y;
 }
 
-float Vector2::getY() {
+float Vector2::getY() const {
```
dumganhar, line 28: Add const suffix to indicate that getter will not modify the data of this class.

native/cocos/editor-support/spine-wasm/AtlasAttachmentLoaderExtension.cpp
```diff
@@ -40,9 +40,10 @@ void AtlasAttachmentLoaderExtension::configureAttachment(Attachment *attachment)
         auto *region = static_cast<AtlasRegion *>(regionAttachment->getRendererObject());
         auto *attachmentVertices = new AttachmentVertices(4, quadTriangles, 6, pages.indexOf(region->page));
         V3F_T2F_C4B *vertices = attachmentVertices->_triangles->verts;
+        const auto &uvs = regionAttachment->getUVs();
         for (int i = 0, ii = 0; i < 4; ++i, ii += 2) {
-            vertices[i].texCoord.u = regionAttachment->getUVs()[ii];
-            vertices[i].texCoord.v = regionAttachment->getUVs()[ii + 1];
+            vertices[i].texCoord.u = uvs[ii];
+            vertices[i].texCoord.v = uvs[ii + 1];
```
dumganhar, line 46: Avoid invoking `getUVs` functions many times which will save some code size and improve performance a little bit.

native/cocos/editor-support/spine-wasm/CMakeLists.txt
```diff
@@ -1,32 +1,76 @@
...
+set(CMAKE_BUILD_TYPE "MinSizeRel")
+# set(CMAKE_BUILD_TYPE "RelWithDebInfo")
+# set(CMAKE_BUILD_TYPE "Debug")
+
+if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+    set(ENABLE_CLOSURE_COMPILER 0)
+    set(SPINE_EXTRA_FLAGS "")
+else()
+    set(ENABLE_CLOSURE_COMPILER 1)
+    set(SPINE_EXTRA_FLAGS "-Oz")
```
dumganhar, line 26: -Oz will make binary size even smaller

native/cocos/editor-support/spine-wasm/spine-mesh-data.cpp
```diff
@@ -12,22 +12,19 @@ void SpineMeshData::initMeshMemory() {
     if (vBuf) return;
     const auto vCount = 65535;
     const auto byteStride = 7 * sizeof(float);
-    vBuf = new uint8_t[2 * vCount * byteStride];
-    iBuf = new uint16_t[8 * 65535];
+    vBuf = static_cast<uint8_t*>(::malloc(sizeof(uint8_t) * (2 * vCount * byteStride))); // 2 * 64KB * 28 = 3.5MB
+    iBuf = static_cast<uint16_t*>(::malloc(sizeof(uint16_t) * (8 * 65535))); // 512KB
```
dumganhar, line 16: Use c malloc to allocate memory for raw data.

native/cocos/editor-support/spine-wasm/spine-model.cpp
```diff
@@ -1,58 +1,56 @@
 #include "spine-model.h"
 
 SpineModel::SpineModel() {
-    SpineModel::data = new std::vector<uint32_t>(6, 0);
+    _data.setSize(6, 0);
```
dumganhar, line 4: Avoid to use `std::vector`, use `spine::Vector` instead which will make binary smaller.

native/cocos/editor-support/spine-wasm/spine-skeleton-instance.cpp
```diff
@@ -80,9 +78,9 @@ TrackEntry *SpineSkeletonInstance::setAnimation(float trackIndex, const std::str
     return trackEntry;
 }
 
-void SpineSkeletonInstance::setSkin(const std::string &name) {
+void SpineSkeletonInstance::setSkin(const spine::String &name) {
     if (!_skeleton) return;
-    _skeleton->setSkin(name.c_str());
```
dumganhar, line 85: Use spine::String could also avoid temporary spine::String constructed here.

native/cocos/editor-support/spine-wasm/spine-skeleton-instance.cpp
```diff
@@ -355,20 +353,20 @@ void SpineSkeletonInstance::collectMeshData() {
         SpineMeshData::moveIB(currMesh.iCount);
         // record debug shape info
         if (_userData.debugMode) {
-            SpineDebugShape debugShape;
+            size_t currentShapesLen = _debugShapes.size();
+            _debugShapes.setSize(currentShapesLen + 1, {});
+            SpineDebugShape& debugShape = _debugShapes[currentShapesLen];
```
dumganhar, line 358: Avoid creating temporary `SpineDebugShape` object and copy to spine::Vector<>.
Create it in spine::Vector<> and get the reference to modify its content will make the binary smaller since less wasm directives be generated.

native/cocos/editor-support/spine-wasm/spine-skeleton-instance.cpp
```diff
@@ -355,20 +353,20 @@ void SpineSkeletonInstance::collectMeshData() {
...
             debugShape.iCount = currMesh.iCount;
-            _debugShapes.push_back(debugShape);
         }
 
         currMesh.blendMode = static_cast<uint32_t>(slot->getData().getBlendMode());
         if (_userData.useSlotTexture) {
-            auto iter = slotTextureSet.find(slot);
-            if (iter != slotTextureSet.end()) {
-                currMesh.textureID = iter->second;
+            if (_slotTextureSet.containsKey(slot)) {
```
dumganhar, line 368: Use spine::HashMap instead of `std::map` will may the binary smaller.

native/cocos/editor-support/spine-wasm/spine-skeleton-instance.cpp
```diff
@@ -406,8 +404,8 @@ AnimationState *SpineSkeletonInstance::getAnimationState() {
     return _animState;
 }
 
-void SpineSkeletonInstance::setMix(const std::string &from, const std::string &to, float duration) {
-    _animStateData->setMix(from.c_str(), to.c_str(), duration);
+void SpineSkeletonInstance::setMix(const spine::String &from, const spine::String &to, float duration) {
+    _animStateData->setMix(from, to, duration);
```
dumganhar, line 408: Avoid many string copy here. Cool.

native/cocos/editor-support/spine-wasm/spine-skeleton-instance.cpp
```diff
@@ -469,19 +463,19 @@ void SpineSkeletonInstance::resizeSlotRegion(const std::string &slotName, uint32
...
         region->updateOffset();
-        auto attachmentVertices = static_cast<AttachmentVertices *>(region->getRendererObject());
+        auto *attachmentVertices = static_cast<AttachmentVertices *>(region->getRendererObject());
         if (createNew) {
             attachmentVertices = attachmentVertices->copy();
             region->setRendererObject(attachmentVertices);
         }
         V3F_T2F_C4B *vertices = attachmentVertices->_triangles->verts;
-        auto UVs = region->getUVs();
+        const auto &UVs = region->getUVs();
```
dumganhar, line 472: Use reference to avoid copying uvs.

native/cocos/editor-support/spine-wasm/spine-skeleton-instance.h
```diff
@@ -1,13 +1,11 @@
...
+#pragma once
+
 #include <spine/spine.h>
 
-#include <map>
+#include <cstdint>
 #include <string>
 #include "mesh-type-define.h"
 #include "spine-model.h"
-using namespace spine;
```
dumganhar, line 9: Don't `using namespace` in header file since it will spread the namespace to all files those include this header.

native/cocos/editor-support/spine-wasm/spine-type-export.cpp
```diff
@@ -1,70 +1,17 @@
...
 #include <type_traits>
-#include <vector>
 #include "spine-skeleton-instance.h"
 #include "spine-wasm.h"
 #include "Vector2.h"
 
-using namespace emscripten;
 using namespace spine;
 
-namespace {
```
dumganhar, line 11: The ` wasm export` file was optimized a lot, need to be patient to review it.

native/cocos/editor-support/spine-wasm/spine-type-export.cpp
```diff
@@ -1,70 +1,17 @@
...
-std::string STRING_SP2STD(const spine::String &str) {
-    std::string stdStr(str.buffer(), str.length());
-    return stdStr;
-}
-
-const spine::String STRING_STD2SP(const std::string &str) {
```
dumganhar, line 17: NO NEED to do a `std::string <-> spine::String` conversion any more.

native/cocos/editor-support/spine-wasm/spine-type-export.cpp
```diff
@@ -85,20 +34,467 @@ using SPVectorPathConstraintPtr = Vector<PathConstraint*>;
...
+        } \
+    }; \
+}}
+
+
+#define DEFINE_SPINE_CLASS_TYPEID(cls) \
+namespace emscripten { namespace internal { \
+    template<> \
+    constexpr TYPEID getLightTypeID<cls>(const cls& value) { \
+        return value.getRTTI().getClassName(); \
```
dumganhar, line 62: Spine runtime has already supported `RTTI` functionality, so we disable which in standard c++ by `-fno-exception` in CMakeLists.txt. This could save about 20 KB wasm file size (uncompressed).

dumganhar: To use our own version of TypeID system, emscripten provide some internal APIs whose we could override here.

native/cocos/editor-support/spine-wasm/spine-type-export.cpp
```diff
@@ -85,20 +34,467 @@ using SPVectorPathConstraintPtr = Vector<PathConstraint*>;
...
+    template<> \
+    struct LightTypeID<const cls&> { \
+        static constexpr TYPEID get() { \
+            return #cls; \
+        } \
+    }; \
+}}
+
+#define GETTER_RVAL_TO_PTR(ClassType, Method, ReturnType) \
+    optional_override([](ClassType &obj) { return const_cast<ReturnType>(&obj.Method()); })
```
dumganhar, line 139: A easy way to make the emscripten API `.property(...)` return a pointer rather than a reference.
Returning a pointer which will not new an object and do a copy operation.

native/cocos/editor-support/spine-wasm/spine-type-export.cpp
```diff
@@ -85,20 +34,467 @@ using SPVectorPathConstraintPtr = Vector<PathConstraint*>;
...
+DEFINE_SPINE_CLASS_TYPEID(DeformTimeline)
+DEFINE_SPINE_CLASS_TYPEID(EventTimeline)
+DEFINE_SPINE_CLASS_TYPEID(DrawOrderTimeline)
+DEFINE_SPINE_CLASS_TYPEID(IkConstraintTimeline)
+DEFINE_SPINE_CLASS_TYPEID(TransformConstraintTimeline)
+DEFINE_SPINE_CLASS_TYPEID(PathConstraintPositionTimeline)
+DEFINE_SPINE_CLASS_TYPEID(PathConstraintMixTimeline)
+DEFINE_SPINE_CLASS_TYPEID(VertexEffect)
+DEFINE_SPINE_CLASS_TYPEID(JitterVertexEffect)
+DEFINE_SPINE_CLASS_TYPEID(SwirlVertexEffect)
```
dumganhar, line 182: For classes that have inheritance relationship, we need to use `DEFINE_SPINE_CLASS_TYPID` to make them could be `uppercast` and `downercast` in wasm glue code.

native/cocos/editor-support/spine-wasm/spine-type-export.cpp
```diff
@@ -85,20 +34,467 @@ using SPVectorPathConstraintPtr = Vector<PathConstraint*>;
...
+DEFINE_ALLOW_RAW_POINTER(SkeletonData)
+DEFINE_ALLOW_RAW_POINTER(Skin)
+DEFINE_ALLOW_RAW_POINTER(Animation)
+DEFINE_ALLOW_RAW_POINTER(AnimationStateData)
+DEFINE_ALLOW_RAW_POINTER(TrackEntry)
+DEFINE_ALLOW_RAW_POINTER(IkConstraintData)
+DEFINE_ALLOW_RAW_POINTER(PathConstraintData)
+DEFINE_ALLOW_RAW_POINTER(TransformConstraintData)
+DEFINE_ALLOW_RAW_POINTER(SPVectorUnsignedShort)
+DEFINE_ALLOW_RAW_POINTER(SPVectorFloat)
```
dumganhar, line 202: GETTER in `.property(...)` doesn't support a function that returns `raw pointer`, so we use `DEFINE_ALLOW_RAW_POINTER` to make embind core code know that we need to return `raw pointer` for the getter.

native/cocos/editor-support/spine-wasm/spine-type-export.cpp
```diff
@@ -85,20 +34,467 @@ using SPVectorPathConstraintPtr = Vector<PathConstraint*>;
...
+        size_t (VecType::*size)() const = &VecType::size;
+        T& (VecType::*get)(size_t) = &VecType::operator[];
+        return emscripten::class_<spine::Vector<T>>(name)
+            .template constructor<>()
+            .function("resize", setSize)
+            .function("size", size)
+            .function("get", get)
+            .function("set", emscripten::optional_override([](VecType& obj, int index, const T& value){
+                obj[index] = value;
+            }), emscripten::allow_raw_pointers());
```
dumganhar, line 252: Add support for `spine::Vector` binding.

native/cocos/editor-support/spine-wasm/spine-type-export.cpp
```diff
@@ -112,232 +508,53 @@ EMSCRIPTEN_BINDINGS(spine) {
...
-    register_vector<EventData *>("VectorEventData");
-    register_vector<Event *>("VectorEvent");
-    register_vector<ConstraintData *>("VectorConstraintData");
-    register_vector<IkConstraint *>("VectorIkConstraint");
-    register_vector<PathConstraint *>("VectorPathConstraint");
-    register_vector<TransformConstraint *>("VectorTransformConstraint");
-    register_vector<IkConstraintData *>("VectorIkConstraintData");
-    register_vector<TransformConstraintData *>("VectorTransformConstraintData");
-    register_vector<PathConstraintData *>("VectorPathConstraintData");
-    register_vector<TrackEntry *>("VectorTrackEntry");
```
dumganhar, line 137: Remove unused std::vector binding registeration.

native/cocos/editor-support/spine-wasm/spine-type-export.cpp
```diff
@@ -112,232 +508,53 @@ EMSCRIPTEN_BINDINGS(spine) {
...
-        }), allow_raw_pointer<TrackEntry*>());
-
-    class_<SPVectorUpdatablePtr>("SPVectorUpdatablePtr")
-        .constructor<>()
-        .function("resize", &SPVectorUpdatablePtr::setSize)
-        .function("size", &SPVectorUpdatablePtr::size)
-        .function("get", &SPVectorUpdatablePtr::operator[], allow_raw_pointers())
-        .function("set",optional_override([](SPVectorUpdatablePtr &obj, int index, Updatable *value) {
-            obj[index] = value;
-        }), allow_raw_pointer<Updatable*>());
```
dumganhar, line 304: Use REGISTER_SPINE_VECTOR instead to simply code.

native/cocos/editor-support/spine-wasm/spine-type-export.cpp
```diff
@@ -112,232 +508,53 @@ EMSCRIPTEN_BINDINGS(spine) {
...
 
     class_<Vector2>("Vector2")
         .constructor<>()
         .constructor<float, float>()
-        .function("setX", &Vector2::setX)
-        .function("getX", &Vector2::getX)
-        .function("setY", &Vector2::setY)
-        .function("getY", &Vector2::getY)
+        .property("x", &Vector2::x)
+        .property("y", &Vector2::y)
```
dumganhar, line 548: Binding field directly makes smaller binary and better performance.

native/cocos/editor-support/spine-wasm/spine-type-export.cpp
```diff
@@ -112,232 +508,53 @@ EMSCRIPTEN_BINDINGS(spine) {
...
-        .function("equals", select_overload<String&(const String&)>(&String::operator=))
-        .function("buffer", &String::buffer, allow_raw_pointer<const char*>())
-        //.function("estr", optional_override([](String &obj) {
-        //    auto str = emscripten::val(obj.buffer());
-        //    return str; }), allow_raw_pointers())
-        .function("strPtr", optional_override([](String &obj) {
-            return reinterpret_cast<uintptr_t>(obj.buffer());}), allow_raw_pointers())
-        .function("str", optional_override([](String &obj) {
-            std::string stdStr(obj.buffer(), obj.length());
-            return stdStr; }), allow_raw_pointers());
```
dumganhar, line 334: Doesn't need to bind `spine::String` anymore, it's treated like `std::string`.

native/cocos/editor-support/spine-wasm/spine-type-export.cpp
```diff
@@ -347,60 +564,45 @@ EMSCRIPTEN_BINDINGS(spine) {
...
-        .constructor<>()
-        .function("triangulate", &Triangulator::triangulate)
-        .function("decompose", &Triangulator::decompose, allow_raw_pointers());
+    class_<HasRendererObject>("HasRendererObject")
+        .constructor<>();
+
+    // class_<Triangulator>("Triangulator")
+    //     .constructor<>()
+    //     .function("triangulate", &Triangulator::triangulate)
+    //     .function("decompose", &Triangulator::decompose, allow_raw_pointers());
```
dumganhar, line 573: Doesn't need to bind `Triangulator` since it's only used in `SkeletonClipping` class and it's a private property of SkeletonClipping.

.github/workflows/emscripten-patches/embind/bind.cpp
minggo: Is `bind.cpp` only used for spine?

cocos/spine/skeleton.ts
```diff
@@ -924,7 +924,7 @@ export class Skeleton extends UIRenderer {
             let cache = this._skeletonCache.getAnimationCache(this._skeletonData!.uuid, name);
             if (!cache) {
                 cache = this._skeletonCache.initAnimationCache(this.skeletonData!.uuid, this._skeletonData!, name);
-                cache?.setSkin(this._skinName);
+                if (cache && this._skinName) cache.setSkin(this._skinName);
```
minggo, line 929: Should check `this._skinName`?

native/cocos/editor-support/spine-wasm/spine-type-export.cpp
```diff
@@ -347,60 +564,45 @@ EMSCRIPTEN_BINDINGS(spine) {
...
     class_<ConstraintData>("ConstraintData")
         .constructor<const String &>()
-        .function("getName", optional_override([](ConstraintData &obj) { return STRING_SP2STD(obj.getName()); }))
-        .function("getOrder", &ConstraintData::getOrder)
-        .function("setOrder", &ConstraintData::setOrder)
-        .function("getSkinRequired", &ConstraintData::isSkinRequired)
-        .function("setSkinRequired", &ConstraintData::setSkinRequired);
+        .property("name", &ConstraintData::getName)
+        .property("order", &ConstraintData::_order)
+        .property("skinRequired", &ConstraintData::_skinRequired);
```
dumganhar, line 579: Use `.property` instead of `.function` will make binary size smaller and make spine-define.ts smaller.

native/cocos/editor-support/spine-wasm/spine-mesh-data.cpp
minggo: 65535 -> vCount

native/cocos/editor-support/spine-wasm/wasmSpineExtension.cpp
```diff
@@ -29,44 +30,23 @@ char *WasmSpineExtension::_readFile(const String &path, int *length) {
 void *WasmSpineExtension::_alloc(size_t size, const char *file, int line) {
     SP_UNUSED(file);
     SP_UNUSED(line);
-
-    if (size == 0)
-        return 0;
-    void *ptr = new uint8_t[size];
-    return (void *)ptr;
+    return ::malloc(sizeof(uint8_t) * size);
```
minggo, line 36: Why not check size before malloc?

native/cocos/editor-support/spine-wasm/spine-type-export.cpp
```diff
@@ -410,109 +612,76 @@ EMSCRIPTEN_BINDINGS(spine) {
         .function("aabbIntersectsSkeleton", &SkeletonBounds::aabbIntersectsSkeleton)
         .function("containsPoint", optional_override([](SkeletonBounds &obj, float x, float y) {
             return obj.containsPoint(x, y); }),allow_raw_pointers())
-        .function("containsPointPolygon", optional_override([](SkeletonBounds &obj,Polygon* polygon, float x, float y) {
-            return obj.containsPoint(polygon, x, y); }),allow_raw_pointers())
+        // .function("containsPointPolygon", optional_override([](SkeletonBounds &obj,Polygon* polygon, float x, float y) {
+            // return obj.containsPoint(polygon, x, y); }),allow_raw_pointers())
```
dumganhar, line 616: `Polygon` is not bound, the definition in `spine-core.d.ts` is
```ts
containsPointPolygon(polygon: ArrayLike<number>, x: number, y: number): boolean;
```
So polygon should be an array rather than a `Polygon` object. I will support that later.

native/cocos/editor-support/spine-wasm/wasmSpineExtension.cpp
```diff
@@ -29,44 +30,23 @@ char *WasmSpineExtension::_readFile(const String &path, int *length) {
...
 void *WasmSpineExtension::_calloc(size_t size, const char *file, int line) {
     SP_UNUSED(file);
     SP_UNUSED(line);
-
-    if (size == 0)
-        return 0;
-    uint8_t *ptr = new uint8_t[size];
-    if (ptr) memset(ptr, 0, size);
-    return (void *)ptr;
+    return ::calloc(1, size);
```
minggo, line 45: Why not check size before calloc?

native/cocos/editor-support/spine-wasm/spine-type-export.cpp
```diff
@@ -610,194 +761,129 @@ EMSCRIPTEN_BINDINGS(spine) {
...
+
     class_<Atlas>("TextureAtlas")
         .constructor<const String &, TextureLoader *, bool>()
-        .function("findRegion", optional_override([](Atlas &obj, const std::string &name) {
-            return obj.findRegion(STRING_STD2SP(name)); }), allow_raw_pointers());
+        .function("findRegion", &Atlas::findRegion, allow_raw_pointers());
 
     class_<PowInterpolation, base<Interpolation>>("Pow")
-        .constructor<int>()
-        .function("apply", &PowInterpolation::apply);
```
dumganhar, line 629: apply is a virtual function in base class `Interpolation`, no need to bind it in sub class.

native/cocos/editor-support/spine-wasm/wasmSpineExtension.cpp
```diff
@@ -29,44 +30,23 @@ char *WasmSpineExtension::_readFile(const String &path, int *length) {
...
     SP_UNUSED(line);
-
-    if (size == 0)
-        return 0;
-    uint8_t *mem = new uint8_t[size];
-    memcpy(mem, ptr, size);
-    delete[](char *) ptr;
-    ptr = mem;
-    return mem;
+    return ::realloc(ptr, sizeof(uint8_t) * size);
```
minggo, line 56: Why not check size before realloc?

native/cocos/editor-support/spine/AnimationState.h
```diff
@@ -230,8 +230,9 @@ class SP_API TrackEntry : public SpineObject, public HasRendererObject {
     void setListener(AnimationStateListener listener);
 
     void setListener(AnimationStateListenerObject* listener);
-
+#ifndef __EMSCRIPTEN__
 private:
+#endif
```
minggo, line 235: Why need it?

native/cocos/editor-support/spine/Extension.cpp
```diff
@@ -47,7 +47,9 @@ void SpineExtension::setInstance(SpineExtension *inValue) {
 }
 
 SpineExtension *SpineExtension::getInstance() {
+#ifndef __EMSCRIPTEN__
     if (!_instance) _instance = spine::getDefaultExtension();
+#endif
```
minggo, line 52: Why need to change it?

native/cocos/editor-support/spine/Extension.cpp
```diff
@@ -108,6 +112,9 @@ void DefaultSpineExtension::_free(void *mem, const char *file, int line) {
 }
 
 char *DefaultSpineExtension::_readFile(const String &path, int *length) {
+#ifdef __EMSCRIPTEN__
+    return nullptr;
+#else
```
minggo: The code is included in `#ifndef __EMSCRIPTEN__`.

dumganhar: Fixed.

native/cocos/editor-support/spine/Extension.cpp
dumganhar: Since doesn't need getDefaultExtension. It's a singleton and always lives in the whole lifetime.
It's really redutant. You could refer to the code:
```ts
void SpineWasmUtil::spineWasmInit() {
    // LogUtil::Initialize();
    SpineExtension* extension = new WasmSpineExtension();
    SpineExtension::setInstance(extension); // Set the singleton here.

    SpineMeshData::initMeshMemory();

    //LogUtil::PrintToJs("spineWasmInit");
}

void SpineWasmUtil::spineWasmDestroy() {
    auto* extension = SpineExtension::getInstance();
    delete extension;
    freeStoreMemory();
    SpineMeshData::releaseMeshMemory();
    // LogUtil::ReleaseBuffer();
}
```

Remove `getDefaultExtension()` invocation for emscripten will save some code size.

native/cocos/editor-support/spine/Extension.cpp
```diff
@@ -116,12 +123,14 @@ char *DefaultSpineExtension::_readFile(const String &path, int *length) {
...
     fread(data, 1, *length, file);
     fclose(file);
 
     return data;
+#endif
 }
 
 DefaultSpineExtension::DefaultSpineExtension() : SpineExtension() {
 }
+#endif
```
minggo, line 132: It seems it doesn't match.

dumganhar: Fixed.

native/cocos/editor-support/spine/IkConstraintData.h
```diff
@@ -39,6 +39,7 @@ namespace spine {
 class BoneData;
 
 class SP_API IkConstraintData : public ConstraintData {
+    RTTI_DECL
```
minggo, line 42: Why need it?

native/cocos/editor-support/spine/MathUtil.cpp
```diff
@@ -42,14 +42,9 @@
...
-const float MathUtil::Pi_2 = 3.1415926535897932385f * 2;
-const float MathUtil::Deg_Rad = (3.1415926535897932385f / 180.0f);
-const float MathUtil::Rad_Deg = (180.0f / 3.1415926535897932385f);
-
-float MathUtil::abs(float v) {
-    return ((v) < 0 ? -(v) : (v));
-}
+RTTI_IMPL_NOPARENT(Interpolation)
+RTTI_IMPL(PowInterpolation, Interpolation)
+RTTI_IMPL(PowOutInterpolation, Interpolation)
```
minggo, line 47: Why need it?

native/cocos/editor-support/spine/TransformConstraint.h
```diff
@@ -73,11 +73,13 @@ class SP_API TransformConstraint : public Updatable {
     float getShearMix();
     void setShearMix(float inValue);
 
-    bool isActive();
+    virtual bool isActive();
```
minggo, line 76: Why need virtual?

native/cocos/editor-support/spine/Extension.cpp
minggo: Got it.

native/cocos/editor-support/spine-wasm/wasmSpineExtension.cpp
dumganhar: In fact, malloc/calloc support to access `zero` value to size parameter. But it's an undefined behavior in c standard.
https://en.cppreference.com/w/c/memory/malloc
> If size is zero, the behavior of malloc is implementation-defined. For example, a null pointer may be returned. Alternatively, a non-null pointer may be returned; but such a pointer should not be [dereferenced](https://en.cppreference.com/w/c/language/operator_member_access), and should be passed to [free](https://en.cppreference.com/w/c/memory/free) to avoid memory leaks.

Emscripten uses `dlmalloc` to implement c standard malloc/calloc/free  by default. dlmalloc will return a small chunk of memory if passing 0 to it.

Yes, to avoid the undefined behaviior, I think adding back the `zero check` is ok.

dumganhar: Reverted.

native/cocos/editor-support/spine-wasm/wasmSpineExtension.cpp
dumganhar: Reverted the `checking size zero` logic.

native/cocos/editor-support/spine/IkConstraintData.h
dumganhar: Because we don't want the default c++ rtti support which will make the package size 20KB bigger.
Spine has already provoided RTTI strategy, so we use `RTTI_DECL`/`RTTI_IMPL` to mark the classes that need the RTTI information. RTTI information is used in getting a unique type id for a `instance` which like c++ `typeid(*pointer)`.

With the RTTI information, a c++ function return a `Base` pointer which points a `Sub` instance will be downcast to `Sub` JS object in emscripten layer. Otherwise, without the RTTI info, in JS layer, we only could see `Base` class.

// c++
```c++
class Base {};

class Sub {};

class Foo {
    Base* getBase() { return new Sub(); }
};
```

// js
```js
const instance = foo.getBase(); // it's a Base instance without RTTI info, but in fact, it should be seen as a `Sub` instance.
```

native/cocos/editor-support/spine/MathUtil.cpp
dumganhar: Explained above.

native/cocos/editor-support/spine/TransformConstraint.h
dumganhar: Just the ensure that `isActive/setActive` is also virtual functions.
In fact, using the `override` suffix may be better, but I think spine runtime may support older compiler that only support c++98 standard which doesn't support `override` keyword.

In C++11 or later, you would get a warning like:
企业微信截图_29906fd4-2212-4aef-a7b3-7203d47f738e

I think it's a best practice of declaring a function with `virtual prefix`  or `override suffix` that overrides super's virtual function. It will be more explicit.

minggo: Got it.

minggo (approved):

bofeng-song (approved):

native/cocos/editor-support/spine-wasm/spine-type-export.cpp
```diff
@@ -85,20 +34,467 @@ using SPVectorPathConstraintPtr = Vector<PathConstraint*>;
...
+        return String(v->data, v->length, false);
+    }
+};
+
+}  // namespace internal
+}  // namespace emscripten
+
+#define ENABLE_EMBIND_TEST 0
+
+#if ENABLE_EMBIND_TEST
```
dumganhar, line 327: Add some embind test and disable it by default.

native/cocos/editor-support/spine-wasm/spine-type-export.cpp
dumganhar: Created another issue: https://github.com/cocos/cocos-engine/issues/17811

native/cocos/editor-support/spine/AnimationState.h
dumganhar: For wasm/asmjs binding, we bind the private/protected properties directly which could reduce package size,
since the `setter/getter` function only does `property assignment` and `returning property`.

native/cocos/editor-support/spine/Atlas.cpp
```diff
@@ -51,7 +51,7 @@ Atlas::Atlas(const String &path, TextureLoader *textureLoader, bool createTextur
     const char *lastSlash = lastForwardSlash > lastBackwardSlash ? lastForwardSlash : lastBackwardSlash;
     if (lastSlash == path) lastSlash++; /* Never drop starting slash. */
     dirLength = (int)(lastSlash ? lastSlash - path.buffer() : 0);
-    dir = SpineExtension::calloc<char>(dirLength + 1, __FILE__, __LINE__);
+    dir = SpineExtension::calloc<char>(dirLength + 1, __SPINE_FILE__, __SPINE_LINE__);
```
dumganhar, line 54: __SPINE_FILE__ is defined to empty string for emscripten build. That will save some bytes.

native/cocos/editor-support/spine/Bone.cpp
```diff
@@ -198,10 +200,10 @@ void Bone::updateWorldTransform(float x, float y, float rotation, float scaleX,
             break;
         }
     }
-    _a *= _skeleton.getScaleX();
-    _b *= _skeleton.getScaleX();
-    _c *= _skeleton.getScaleY();
-    _d *= _skeleton.getScaleY();
```
dumganhar, line 204: Refactor `Bone::updateWorldTransform`, save some directives by invoking the getter at the beginning of the function.

native/cocos/editor-support/spine/Bone.cpp
```diff
@@ -310,193 +312,22 @@ Bone *Bone::getParent() {
...
-void Bone::setWorldX(float inValue) {
-    _worldX = inValue;
-}
-
-float Bone::getWorldY() {
-    return _worldY;
-}
-
-void Bone::setWorldY(float inValue) {
-    _worldY = inValue;
```
dumganhar, line 474: Move the implementation of getter/setter to header and inline them.
Inlining them could make wasm/asmjs package size smaller and improve performance a little bit.

native/cocos/editor-support/spine/Constraint.h
```diff
@@ -43,8 +43,6 @@ class SP_API Constraint : public Updatable {
 
     virtual ~Constraint();
 
-    virtual void update() = 0;
```
dumganhar, line 46: update function was already defined in its parent class `Updatable`. No need to define it here again.

native/cocos/editor-support/spine/CurveTimeline.h
```diff
@@ -45,8 +45,6 @@ class SP_API CurveTimeline : public Timeline {
 
     virtual void apply(Skeleton& skeleton, float lastTime, float time, Vector<Event*>* pEvents, float alpha, MixBlend blend, MixDirection direction) = 0;
 
-    virtual int getPropertyId() = 0;
```
dumganhar, line 48: getPropertyId is defined in Timeline, doesn't need to define the pure virtual function here again.

native/cocos/editor-support/spine/Json.cpp
```diff
@@ -41,9 +41,38 @@ THE SOFTWARE.
...
+        result <<= 4;
+        if (c >= '0' && c <= '9') {
+            result |= (c - '0');
+        } else if (c >= 'A' && c <= 'F') {
+            result |= (c - 'A' + 10);
+        } else if (c >= 'a' && c <= 'f') {
+            result |= (c - 'a' + 10);
+        }
+    }
+    return true;
```
dumganhar, line 72: A simpler version of `sscanf(ptr + 1, "%4x", &uc);` which could make the package size smaller.

native/cocos/editor-support/spine/Json.cpp
```diff
@@ -378,15 +415,15 @@ const char *Json::parseNumber(Json *item, const char *num) {
         }
 
         while (*ptr >= '0' && *ptr <= '9') {
-            exponent = (exponent * 10.0) + (*ptr - '0');
+            exponent = (exponent * 10) + (*ptr - '0');
```
dumganhar, line 418: *ptr is a value between `0` and `9`, exponent could only be an integer. So define it as uint32_t and use integer version of pow funciton (`MathUtil::ipow`) could make the package size smaller again.

native/cocos/editor-support/spine/MathUtil.cpp
```diff
@@ -108,21 +65,23 @@ bool MathUtil::isNan(float v) {
...
     float u = random();
     float d = max - min;
     if (u <= (mode - min) / d) return min + sqrt(u * d * (mode - min));
     return max - sqrt((1 - u) * d * (max - mode));
 }
 
-float MathUtil::pow(float a, float b) {
-    return (float)::pow(a, b);
-}
+uint64_t MathUtil::ipow(uint64_t base, uint32_t exp) {
```
dumganhar, line 75: Add an integer version of pow function. It's simpler and higher performance than the builtin float version.

native/cocos/editor-support/spine/MathUtil.h
```diff
@@ -31,20 +31,21 @@
...
-
-#include <string.h>
+#include <spine/RTTI.h>
+#include <math.h>
+#include <cstdint>
 
 namespace spine {
 
-class SP_API MathUtil : public SpineObject {
+class SP_API MathUtil {
```
dumganhar, line 40: MathUtil doesn't need to be implemented from 'SpineObject'.

dumganhar:
@cocos-robot run test cases

native/cocos/editor-support/spine/SpineString.h
```diff
@@ -32,176 +32,47 @@
...
-    }
-
-    String &append(const String &other) {
-        size_t len = other.length();
-        size_t thisLen = _length;
-        _length = _length + len;
-        bool same = other._buffer == _buffer;
-        _buffer = SpineExtension::realloc(_buffer, _length + 1, __FILE__, __LINE__);
-        memcpy((void *)(_buffer + thisLen), (void *)(same ? _buffer : other._buffer), len + 1);
-        return *this;
```
dumganhar, line 168: Moving most methods of spine::String to .cpp to avoid inline too many code.

native/cocos/editor-support/spine/TwoColorTimeline.cpp
```diff
@@ -78,24 +78,26 @@ void TwoColorTimeline::apply(Skeleton &skeleton, float lastTime, float time, Vec
...
-                color.r += (color.r - slot._data.getColor().r) * alpha;
-                color.g += (color.g - slot._data.getColor().g) * alpha;
-                color.b += (color.b - slot._data.getColor().b) * alpha;
-                color.a += (color.a - slot._data.getColor().a) * alpha;
-
-                Color &darkColor = slot.getDarkColor();
-                darkColor.r += (darkColor.r - slot._data.getDarkColor().r) * alpha;
-                darkColor.g += (darkColor.g - slot._data.getDarkColor().g) * alpha;
-                darkColor.b += (darkColor.b - slot._data.getDarkColor().b) * alpha;
+                color.r += (color.r - dataColor.r) * alpha;
```
dumganhar, line 93: Don't invoke `getColor`, `getDarkColor` many times.

native/cocos/editor-support/spine/Vector.h
```diff
@@ -72,14 +75,14 @@ class SP_API Vector : public SpineObject {
         return _size;
     }
 
-    inline void setSize(size_t newSize, const T &defaultValue) {
+    void setSize(size_t newSize, const T &defaultValue) {
```
dumganhar, line 78: Remove `inline` keyword of function that contains many lines. Let compiler deside whether it should be inlined.

native/cocos/editor-support/spine/Vector.h
```diff
@@ -164,6 +167,12 @@ class SP_API Vector : public SpineObject {
         return _buffer[inIndex];
     }
 
+    inline const T &operator[](size_t inIndex) const {
```
dumganhar, line 170: Add const version of `[]` accessor.

bofeng-song (approved):

PR #17801, opened 2024-11-04
simplify codes

minggo:
Reduce duplicated codes can reduce package size https://github.com/cocos/cocos-engine/issues/17715

### DIFF
--- a/cocos/ui/scroll-view.ts
+++ b/cocos/ui/scroll-view.ts
@@ -500,14 +500,7 @@ export class ScrollView extends ViewGroup {
      * ```
      */
     public scrollToBottom (timeInSecond?: number, attenuated = true): void {
-        assignMoveDeltaOption(0, 0, false, true);
-        const moveDelta = this._calculateMovePercentDelta(_moveDeltaOptions);
-
-        if (timeInSecond) {
-            this._startAutoScroll(moveDelta, timeInSecond, attenuated);
-        } else {
-            this._moveContent(moveDelta, true);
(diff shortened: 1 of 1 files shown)


cocos/ui/scroll-view.ts
```diff
@@ -857,7 +787,11 @@ export class ScrollView extends ViewGroup {
      * ```
      */
     public scrollToPercentVertical (percent: number, timeInSecond?: number, attenuated?: boolean): void {
-        assignMoveDeltaOption(0, percent, false, true);
+        this._doScroll(0, percent, false, true, timeInSecond, attenuated);
+    }
+
+    private _doScroll (x: number, y: number, applyToHorizontal: boolean, applyToVertical: boolean, timeInSecond?: number, attenuated = true): void {
+        assignMoveDeltaOption(x, y, applyToHorizontal, applyToVertical);
```
qiuguohua, line 794: The function assignMoveDeltaOption can be removed?

minggo: No, it will update `_moveDeltaOptions`

qiuguohua: I mean to put the implementation of assignMoveDeltaOption directly here?

qiuguohua (approved):

cocos/ui/scroll-view.ts
minggo: It is better not. Current implementation make it more clearly.

PR #17809, opened 2024-11-05
do not new object as possible

minggo:


minggo:
@cocos-robot run test cases

qiuguohua (approved):

minggo:
@cocos-robot run test cases

minggo:
@cocos-robot run test cases

PR #17837, opened 2024-11-11
Fix surface-shader

GengineJS:


cocos/rendering/custom/web-pipeline-types.ts
```diff
@@ -901,37 +901,48 @@ export class RenderDrawQueue {
...
-        while (!res.done) {
-            res.value.clear();
-            res = it.next();
-        }
-        this.queue.clear();
+        this.passInstances.clear();
+        const instanceBuffers = this.instanceBuffers;
+        instanceBuffers.forEach((instance) => {
+            instance.clear();
+        });
```
minggo, line 935: Should also clear instanceBuffers?

star-e (approved):

cocos/rendering/custom/web-pipeline-types.ts
star-e: I think gfx instance buffers are reused. They are not recreated every frame.

GengineJS: The instanceBuffers here will be reused, so there's no need to clear them

minggo: And the size of it is not changed?

GengineJS: The reuse logic is here： image

minggo (approved):

PR #17864, opened 2024-11-15
Add node judgment

wanghui187:
Add node judgment


minggo (approved):

qiuguohua:
https://github.com/cocos/2d-tasks/issues/3884#issuecomment-2469838444

PR #17980, opened 2024-11-28
Fix the error caused by the lag in the WebGPU API

GengineJS:


GengineJS:
https://github.com/cocos/3d-tasks/issues/18624

star-e (approved):

PR #18004, opened 2024-12-05
To fix the issue of image stretching after resizing a reflection probe.

GengineJS:


star-e (approved):

cocos/3d/reflection-probe/reflection-probe-component.ts
```diff
@@ -105,10 +105,6 @@ export class ReflectionProbe extends Component {
 
     private _position = new Vec3(0, 0, 0);
 
-    constructor () {
-        super();
-    }
```
dumganhar, line 110: Removing the default constructor will make the package size bigger. So don't remove it.

cocos/3d/reflection-probe/reflection-probe-component.ts
```diff
@@ -327,6 +323,12 @@ export class ReflectionProbe extends Component {
         }
     }
 
+    private _handleResize$ (): void {
```
dumganhar, line 326: Don't add `$` now since we will write a babel plugin to mangle the private variables.

PR #18049, opened 2024-12-17
Adjust the timing for building the rendergraph in the pipeline.

GengineJS:


star-e (approved):

cocos/game/director.ts
```diff
@@ -809,7 +809,7 @@ export class Director extends EventTarget {
      * @en Build custom render pipeline
      * @zh 构建自定义渲染管线
      */
-    private buildRenderPipeline (): void {
+    public buildRenderPipeline (): void {
```
star-e, line 813: Shall we add `@engineInternal`?

native/cocos/renderer/pipeline/custom/NativePipeline.cpp
```diff
@@ -1310,6 +1311,19 @@ void buildLayoutGraphNodeBuffer(
     }
 }
 
+se::Value buildRPVal;
+void buildRenderPipeline() {
+    if (buildRPVal.isUndefined()) {
```
dumganhar, line 1316: buildRPValue needs to be reset while `ScriptEngine` shutdowns, otherwise the after `game.restart()`, buildRPVal will keep the old value that previous `ScriptEngine` holds.
There is an interface called `addBeforeCleanupHook` that could make a hook callback which will be triggered when cleanup ScriptEngine. E.g.
```cpp
se::ScriptEngine::getInstance()->addBeforeCleanupHook([]() {
        // Doing something to release or reset your `se` resources.
});
```

GengineJS: get

star-e (approved):

dumganhar (approved):

PR #18052, opened 2024-12-18
Fix the UI error issues caused by adjustments to the batching and splitting of UI elements.

GengineJS:


star-e:
Are all the `!`s need to be changed?

GengineJS:
> Are all the `!`s need to be changed?

It seems unnecessary at the moment; the impact of resetState is not significant, and testing has shown no issues.

star-e:
Shall we only change the `!` causing the problem? Although `!` is error prone, I think it's better than checking nullity everywhere.
@GengineJS @dumganhar

dumganhar:
I think we need to check whether renderData is null in all places since we modified the behavior in UIRenderer which is a base class.

star-e:
> I think we need to check whether renderData is null in all places since we modified the behavior in UIRenderer which is a base cass.

I see. So we should change all the `renderData`.

star-e (approved):

qiuguohua (approved):

dumganhar:
@cocos-robot run test cases

dumganhar (approved):

cocos/2d/framework/ui-renderer.ts
```diff
@@ -320,6 +320,10 @@ export class UIRenderer extends Renderer {
         this.node.on(NodeEventType.ANCHOR_CHANGED, this._nodeStateChange, this);
         this.node.on(NodeEventType.SIZE_CHANGED, this._nodeStateChange, this);
         this.node.on(NodeEventType.PARENT_CHANGED, this._colorDirty, this);
+        // If the renderData is invalid, it needs to be rebuilt to recalculate the batch processing.
+        if (!this._renderData && this._flushAssembler) {
+            this._flushAssembler();
```
dumganhar, line 325: This will cause the issue: https://forum.cocos.org/t/topic/165212/66
In ParticleSystem2D component, this._renderData is always `null`, so each time `onEnable` is invoked, `ParticleSystem2D._flushAssembler` method will be invoked which will create new RenderData in it.
```ts
    protected _flushAssembler (): void {
        const assembler = ParticleSystem2D.Assembler.getAssembler(this);

        if (this._assembler !== assembler) {
            this._assembler = assembler;
        }
        if (this._assembler && this._assembler.createData) {
// ---> this._simulator.renderData will be overridden.
            this._simulator.renderData = this._assembler.createData(this);
            this._simulator.renderData.particleInitRenderDrawInfo(this.renderEntity); 
            this._simulator.initDrawInfo();
        }
    }
```

jucktwen20-arch (commented):
?

PR #18144, opened 2025-01-09
fixed #18139: Bug: [Scene] frame should not be invalid, it may have been released.

### LINKED ISSUE #18139: Bug: [Scene] frame should not be invalid, it may have been released.
Reported at https://forum.cocos.org/t/topic/164672

This issue is fired in v3.8.5.
## How to Reproduce?

- Create a prefab
- Add a label node to prefab
- Set the label's cache mode to `CHAR`
- Save the prefab
- Close the prefab
- Open the prefab again
- Close the prefab again

Get the error log:

企业微信截图_ffa63752-2083-4b40-93f9-0b5bea53e26d

## Reason

The behavior of opening or closing a prefab will trigger scene switch, there is a global variable called `_sharedAtlas` in `letter-font.ts` whose texture will be cleared and re-constructed.

```ts
// letter-font.ts

let _shareAtlas: LetterAtlas | null  = null;

export const letterFont = js.mixin(bmfontUtils, {
    getAssemblerData () {
        if (!_shareAtlas) {
            _shareAtlas = new LetterAtlas(_atlasWidth, _atlasHeight);
        }

        return _shareAtlas.getTexture() as LetterRenderTexture | null;
    },
```

```ts
// font-utils.ts

export class LetterAtlas {
        ......

    constructor (width: number, height: number) {
            ......
        // Subscribe the event of before loading scene.
        director.on(DirectorEvent.BEFORE_SCENE_LAUNCH, this.beforeSceneLoad, this);
    }

    public beforeSceneLoad (): void {
        this.clearAllCache();
    }

    public clearAllCache (): void {
        this.destroy(); // Will destroy the internal texture instance

        const texture = new LetterRenderTexture(); // Re-create the internal texture instance
        texture.initWithSize(this._width, this._height);

        this.fontDefDictionary.texture = texture;
    }

    public destroy (): void {
        this.reset();
        const dict = this.fontDefDictionary;
        if (dict && dict.texture) {
            dict.texture.destroy();
            dict.texture = null;
        }
    }
```

There is a new functionality that could preview a prefab in inspector window in v3.8.5.
企业微信截图_11f7a0b5-7424-4b32-b207-dfc5a8fa2cb9

It uses an individual `scene` which will never be destroyed and never be run.
Some code snippet:

```ts
// preview/Interactive-preview.ts

class InteractivePreview extends PreviewBase {
    protected scene!: Scene;
    protected cameraComp!: Camera;
    protected camera: renderer.scene.Camera | null = null;
    protected worldAxisNode!: Node;
    protected axisCameraOffset = new Vec3(0, 0, 40);

    protected isMouseLeft = false;
    protected isMouseMiddle = false;

    protected enableAxis = true;
    protected worldAxis: PreviewWorldAxis | null = null;

    protected enableGrid = true;
    protected grid: Grid |null = null;

    protected enableSkybox = true;
    protected skybox: SkyboxInfo | null = null;

    public createNodes(scene: Scene) {

    }

    public init(registerName: string, queryName: string) {
        this.scene = new Scene(registerName);

        if (this.enableSkybox) {
            this.skybox = this.scene.globals.skybox;
            this.scene.globals.skybox.enabled = true;
        }

        this.cameraComp = new Node(registerName + 'camera').addComponent(Camera);

        this.cameraComp.node.setParent(this.scene);
        this.cameraComp.node.setPosition(0, 1, 2.5);
        this.cameraComp.node.lookAt(Vec3.ZERO);
        this.cameraComp.near = 0.01;
        this.cameraComp.enabled = false;

        this.createNodes(this.scene);

       // Load the scene manually instead of running it.
        this.scene._load();
        this.scene._activate();

        this.cameraComp.clearColor = new Color(71, 71, 71, 255);
        this.camera = this.cameraComp.camera;
        ......
        // Create the Preview buffer which will create an offscreen RenderWindow. See the code bellow.
        this.previewBuffer = new PreviewBuffer(registerName, queryName, this.scene);
        // Change the render window for camera which created in preview buffer.
        this.camera.changeTargetWindow(this.previewBuffer.window);
      ........
    }
```

```ts
// preview/buffer.ts

class PreviewBuffer extends EventEmitter {
    ......
    constructor(registerName: string, name: string, scene: any = null) {
        super();
        ......
        this.createWindow(); // Create render window
        this.queue = [];
    }

    createWindow(uuid: string | null = null) {
        if (uuid && this.windows[uuid]) {
            this.window = this.windows[uuid];
            return;
        }
        const root = cc.director.root;
        const renderPassInfo = new gfx.RenderPassInfo(
            [new gfx.ColorAttachment(root.mainWindow.swapchain.colorTexture.format)],
            new gfx.DepthStencilAttachment(root.mainWindow.swapchain.depthStencilTexture.format),
        );
        renderPassInfo.colorAttachments[0].barrier = root.device.getGeneralBarrier(new gfx.GeneralBarrierInfo(0, gfx.AccessFlagBit.FRAGMENT_SHADER_READ_TEXTURE));
        const window = root.createWindow({
            title: this._name,
            width: this.width,
            height: this.height,
            renderPassInfo,
            isOffscreen: true, // Mark it as an offscreen render window.
        });
        this.window = window;
        uuid && (this.windows[uuid] = window);
    }

```

So we get an independent `preview-prefab-scene` who contains an offscreen render window and all its children will be walked and rendered to offscreen texture even the `preview-prefab-scene` is not the current active scene.

When the global variable `_sharedAtlas` re-creates the altas texture, the label with char mode will not know that the texture is destroyed and changed. It will use the destroyed texture to render the label in `preview-prefab-scene`. Then, we get the error message in Label._render.

```ts
// label.ts
protected _render (render: IBatcher): void {
       // BOMB,  this._texture was destroyed.
        render.commitComp(this, this.renderData, this._texture, this._assembler!, null);
    }
```

### Solution 1

The global variable `_sharedAtlas` should be associated with scene, which means each scene should have an individual `_sharedAtlas` instance. So when editor switchs scenes, the label cached in `preview-prefab-scene` will hold a valid texture.

And add a public interface to re-create the texture in `_sharedAtlas`, this is for editor to release the memory manually.

I don't find a good place the store all `_sharedAtlas` instances ( one per scene).

### Solution 2

Fire an event when the texture re-created, Label component subscribes that event and update the texture.  This solution will make the label cached in `preview-prefab-scene` always change the texture and fill new letter textures to altas texture which will cause some redundant operations.

### Solution 3

Don't re-create the letter altas texture while scene switch in editor mode, add a `reset` method for editor to decide when to re-create it.

### Solution 4

Use Asset.refCount to decide whether could destroy the altas texture.

dumganhar:
Re: #18139

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/2d/assets/bitmap-font.ts
+++ b/cocos/2d/assets/bitmap-font.ts
@@ -52,10 +52,29 @@ export interface ILetterDefinition {
 
 export class FontAtlas {
     public letterDefinitions: ILetterDefinition = {};
-    public declare texture: TextureBase | null;
+    public _texture: TextureBase | null = null;
 
     constructor (texture: TextureBase | null) {
         this.texture = texture;
+    }
+
+    public set texture (texture: TextureBase | null) {
+        const oldTexture = this._texture;
+        if (oldTexture === texture) return;
+        if (oldTexture) {
+            oldTexture.decRef(false);
+            if (oldTexture.refCount <= 0) {
+                oldTexture.destroy();
+            }
+        }
+        if (texture) {
+            texture.addRef();
+        }
+        this._texture = texture;
+    }
+
+    public get texture (): TextureBase | null {
+        return this._texture;
     }
 
     public addLetterDefinitions (letter: string, letterDefinition: FontLetterDefinition): void {
@@ -73,7 +92,7 @@ export class FontAtlas {
     }
 
     public getTexture (): TextureBase | null {
-        return this.texture;
+        return this._texture;
     }
 
     public getLetter (key: string): FontLetterDefinition {
--- a/cocos/2d/assembler/label/font-utils.ts
+++ b/cocos/2d/assembler/label/font-utils.ts
@@ -30,6 +30,7 @@ import { BufferTextureCopy } from '../../../gfx';
 import { safeMeasureText, BASELINE_RATIO, MIDDLE_RATIO, getBaselineOffset, getSymbolCodeAt } from '../../utils/text-utils';
 import { director, DirectorEvent } from '../../../game/director';
 import { ccwindow } from '../../../core/global-exports';
+import { TextureBase } from '../../../asset/assets/texture-base';
 
 export interface ISharedLabelData {
     canvas: HTMLCanvasElement;
@@ -366,12 +367,11 @@ export class LetterAtlas {
         this.reset();
         const dict = this.fontDefDictionary;
         if (dict && dict.texture) {
-            dict.texture.destroy();
             dict.texture = null;
         }
     }
 
-    getTexture (): any {
+    getTexture (): TextureBase | null {
         return this.fontDefDictionary.getTexture();
     }
 
--- a/cocos/2d/components/label.ts
+++ b/cocos/2d/components/label.ts
@@ -505,6 +505,7 @@ export class Label extends UIRenderer {
         }
         if (this._cacheMode === CacheMode.CHAR) {
             this._ttfSpriteFrame = null;
+            this.destroyLetterTexture();
         }
 
         this._cacheMode = value;
@@ -932,10 +933,20 @@ export class Label extends UIRenderer {
         // this._textRenderData = null;
         // this._textLayoutData = null;
 
-        // texture cannot be destroyed in here, lettertexture image source is public.
-        this._letterTexture = null;
+        this.destroyLetterTexture();
 
         super.onDestroy();
+    }
+
+    private destroyLetterTexture (): void {
+        const letterTexture = this._letterTexture;
+        if (letterTexture) {
+            letterTexture.decRef(false);
+            if (letterTexture.refCount <= 0) {
+                letterTexture.destroy();
+            }
+        }
+        this._letterTexture = null;
     }
 
     /**
@@ -1035,8 +1046,13 @@ export class Label extends UIRenderer {
             }
         } else {
             if (this.cacheMode === CacheMode.CHAR) {
-                this._letterTexture = this._assembler!.getAssemblerData();
-                this._texture = this._letterTexture;
+                const oldLetterTexture = this._letterTexture;
+                const letterTexture = this._assembler!.getAssemblerData();
+                if (letterTexture !== oldLetterTexture) {
+                    this.destroyLetterTexture();
+                    letterTexture.addRef();
+                }
+                this._texture = this._letterTexture = letterTexture;
             } else if (!this._ttfSpriteFrame) {
                 this._ttfSpriteFrame = new SpriteFrame();
                 this._assemblerData = this._assembler!.getAssemblerData();


cocos/2d/assets/bitmap-font.ts
```diff
@@ -52,12 +52,31 @@ export interface ILetterDefinition {
...
 
     constructor (texture: TextureBase | null) {
         this.texture = texture;
     }
 
+    public set texture (texture: TextureBase | null) {
+        const oldTexture = this._texture;
+        if (oldTexture === texture) return;
+        if (oldTexture) {
+            oldTexture.decRef(false);
```
minggo, line 65: Why should pass `false`?

dumganhar: The texture asset is maintained by Label module internally, not by the assetmanager. So it should not be controlled by asset-manager. See its definition:
```ts
/**
     * @en
     * Decrease the reference count and it will be auto released when refCount equals 0.
     *
     * @zh
     * 减少资源的引用，如果引用数量为 0，则将自动释放该资源。
     *
     * @return @en The asset itself. @zh 此资源本身。
     *
     */
    public decRef (autoRelease = true): Asset {
        if (this._ref > 0) {
            this._ref--;
        }
        if (autoRelease) {
            (cclegacy.assetManager as AssetManager).getReleaseManager().tryRelease(this);
        }
        return this;
    }
```

minggo: What happen if not passing false and let assetmanager to manage it?

dumganhar: First, assetManager should not mangle the assets that is not loaded by it.
Second, if pass true, it will go to the following logic in release-manager.ts . 
See the `// --> balabala` comments.
```ts
public tryRelease (asset: Asset, force = false): void {
        if (!(asset instanceof Asset)) { return; }
        if (force) { // ---> force is false, will not go inside
            this._free(asset, force);
            return;
        }

        this._toDelete.add(asset._uuid, asset); // ---> add to _toDelete array. asset._uuid is '' (empty)
        if (TEST) return;
        if (!this._eventListener) {
            this._eventListener = true;
            misc.callInNextTick(this._freeAssets.bind(this)); // --> Go here to invoke _freeAssets in next tick
        }
    }

    private _freeAssets (): void {
        this._eventListener = false;
        this._toDelete.forEach((asset): void => {
            this._free(asset); // --> Invoke _free
        });
        this._toDelete.clear();
    }

    private _free (asset: Asset, force = false): void {
        const uuid = asset._uuid;  // --> uuid is empty
        this._toDelete.remove(uuid);

        if (!isValid(asset, true) || this._dontDestroyAssets.indexOf(uuid) !== -1) { return; }

        if (!force) { // --> force is false
            if (asset.refCount > 0) { // --> If refCount > 0, the next line will return.
                if (checkCircularReference(asset) > 0) { return; } // --> return directly
            }
        }

        // remove from cache
        assets.remove(uuid);  // --> it's not in the cache since it's not loaded by assetManager
        const depends = dependUtil.getDeps(uuid); // --> depends is empty.
        for (let i = 0, l = depends.length; i < l; i++) {
            const dependAsset = assets.get(depends[i]);
            if (dependAsset) {
                dependAsset.decRef(false);
                // no need to release dependencies recursively in editor
                if (!EDITOR) {
                    this._free(dependAsset, false);
                }
            }
        }
        // only release non-gc asset in editor
        if (!EDITOR) {
            asset.destroy();  // --> If refCount is 0, go here to destroy it.
        }
        dependUtil.remove(uuid);
       ...
    }
```

dumganhar: It will go inside lots of logic in release-manager.ts and hide the destroy strategy and delay the `destroy` operation to next frame.
I think Assets not loaded by AssetManager should not be mantained by AssetManager/ReleaseManager.

minggo (approved):

PR #18160, opened 2025-01-11
[v3.8.6] Mark some properties as `@dontmangle`.

dumganhar:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/core/curves/keyframe-curve.ts
+++ b/cocos/core/curves/keyframe-curve.ts
@@ -281,8 +281,9 @@ export class KeyframeCurve<TKeyframeValue> implements CurveBase, Iterable<KeyFra
     }
 
     // Times are always sorted and 1-1 correspond to values.
+    /** @dontmangle */
     protected _times: number[] = [];
-
+    /** @dontmangle */
     protected _values: TKeyframeValue[] = [];
 }
 
--- a/cocos/core/curves/keyframe-curve.ts
+++ b/cocos/core/curves/keyframe-curve.ts
@@ -280,10 +280,18 @@ export class KeyframeCurve<TKeyframeValue> implements CurveBase, Iterable<KeyFra
         return iNext;
     }
 
-    // Times are always sorted and 1-1 correspond to values.
-    /** @dontmangle */
+    /**
+     * Times are always sorted and 1-1 correspond to values.
+     * @dontmangle
+     * NOTE: _times is a serializable property set by `CCClass.fastDefine`,
+     * so it should not be mangled while `mangleProtected` is true in `<<ProjectRoot>>/engine-mangle-config.json`.
+     */
     protected _times: number[] = [];
-    /** @dontmangle */
+    /**
+     * @dontmangle
+     * NOTE: _values is a serializable property set by `CCClass.fastDefine`,
+     * so it should not be mangled while `mangleProtected` is true in `<<ProjectRoot>>/engine-mangle-config.json`.
+     */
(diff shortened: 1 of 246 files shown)


cocos/core/curves/keyframe-curve.ts
```diff
@@ -281,8 +281,9 @@ export class KeyframeCurve<TKeyframeValue> implements CurveBase, Iterable<KeyFra
     }
 
     // Times are always sorted and 1-1 correspond to values.
+    /** @dontmangle */
     protected _times: number[] = [];
-
+    /** @dontmangle */
     protected _values: TKeyframeValue[] = [];
```
minggo, line 295: protected member variable will be mangled?

dumganhar: The file `engine-mangle-config.json` supports to control whether to mangle `protected` properties globally, its default value is `false` which means `protected` properties will not be mangled. But if developers set it to `true`, then all protected properties will be mangled too. The file format is :

```json
{
	"common": {
	    "mangleProtected": false, // <------- here
	    "mangleList": [
	    	"MyClass2",
	    	"MyClass.foo",
	    	"MyClass.getBar"
	    ],
	    "dontMangleList": [
	        "BaseFactory",
	        "Slot",
	        "MyClass.hello"
	    ]
	},
	"web-mobile": {
		"extends": "common",
		"mangleList": [
			"MyClass3",
			"MyClass4.haha",
			"MyClass4.w111uwu"
		],
		"dontMangleList": [
			"MyClass5.skldjfl"
		]
	}
}
```

So if developer set `mangleProtected` to true, some properties are set by `cc.fastDefine` to mark them as serializable, so they should not be mangled. In this case, we should make sure some properties in engine should not be mangled by jsdoc tag `@dontmangle`.

dumganhar: For example:
```ts
CCClass.fastDefine('cc.AnimationCurve', AnimationCurve, {
    _curve: null, // <---- the protected property _curve should not be mangled.
});
```

minggo: Got it. If so, then how to determine if a protected variable can be mangled or not? What i mean is how to know if need to add `@@dontmangle` for protected variables?

dumganhar: Need to review the code to decide. So the `mangleProtected` is disabled (false) by default.

dumganhar:
> Package size ⤴ 76 bytes

Because the private property `AnimationCurve._curve` should not be mangled since it's a serializable property used by `cc.fastDefine`.

cocos/core/curves/keyframe-curve.ts
minggo: > Need to review the code to decide.

Is there a simple rule to determine it? If not, then it will be hard to maintain the codes. And how to make sure it work or not, is there a CI to check it?

dumganhar:
@minggo I added some comment for the API being marked as `@dontmangle` and reverted changes in component.ts

cocos/core/curves/keyframe-curve.ts
dumganhar: Currently, we haven't enable CI to run the feature of `mangleProperties`. I will discuss with QA team after editor is updated.

minggo (approved):

PR #18162, opened 2025-01-11
[v3.8.6] Add freezeColor internal method to reduce code size.

dumganhar:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/core/math/color.ts
+++ b/cocos/core/math/color.ts
@@ -42,6 +42,10 @@ const A_INDEX = 3;
 const mathAbs = Math.abs;
 const mathMax = Math.max;
 
+function freezeColor (r: number, g: number, b: number, a: number): Readonly<Color> {
+    return Object.freeze(new Color(r, g, b, a));
+}
+
 /**
  * @en Representation of RGBA colors.<br/>
  * Each color component is an integer value with a range from 0 to 255.<br/>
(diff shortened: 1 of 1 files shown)


cocos/core/math/color.ts
```diff
@@ -42,23 +42,27 @@ const A_INDEX = 3;
...
+    public static WHITE = freezeColor(255, 255, 255, 255);
+    public static GRAY = freezeColor(127, 127, 127, 255);
+    public static BLACK = freezeColor(0, 0, 0, 255);
+    public static TRANSPARENT = freezeColor(0, 0, 0, 0);
+    public static RED = freezeColor(255, 0, 0, 255);
+    public static GREEN = freezeColor(0, 255, 0, 255);
+    public static BLUE = freezeColor(0, 0, 255, 255);
+    public static CYAN = freezeColor(0, 255, 255, 255);
+    public static MAGENTA = freezeColor(255, 0, 255, 255);
+    public static YELLOW = freezeColor(255, 255, 0, 255);
```
minggo, line 65: Can just assign Object.freeze to a variable? Such as
```
const freezeColor = Object.freeze;
```

dumganhar: Probably we could, I didn't test that.
This depends on whether the implementation of `Object.freeze` uses `this` object since we change this object by `const freezeColor = Object.freeze`. 

For safety, it's better to use `const freezeColor = Object.freeze.bind(Object);`.

dumganhar: After thinking a while, I think we should not use `const freezeColor = Object.freeze` since that will generate more code.

```ts
public static BLUE = freezeColor(new Color(0, 0, 255, 255)); 
```

will be mangled to 

Color.BLUE = aa(new Color(0, 0, 255, 255));

while using `function`
```ts
public static BLUE = freezeColor(0, 0, 255, 255);
```

will be mangled to 

Color.BLUE = aa(0, 0, 255, 255);  // shorter and without `new Color(...)`

minggo: ok

minggo (approved):

Issue #18056, opened 2024-12-18
Mangle private properties


dumganhar:
1. Write a plugin to mangle private properties.
- [x] Mangle private properties automatically
- [x] Add `@mangle` jsDoc tag
- [x] Add `@dontmangle` jsDoc tag
- [x] Add blacklist config
- [x] Add whitelist config
2. Editor support
- [x] Building panel show whether to enable this functionality, default is false.

PR #18172, opened 2025-01-12
[v3.8.6] Optimize code size by adding helper functions to get data from pipelineSceneData.

dumganhar:
Re: #18056

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/3d/framework/mesh-renderer.ts
+++ b/cocos/3d/framework/mesh-renderer.ts
@@ -952,7 +952,7 @@ export class MeshRenderer extends ModelRenderer {
 
         if (mainLight.node.mobility === MobilityMode.Static) {
             const sceneGlobals = this.node.scene.globals;
-            const lightProbeInfoData = this.node.scene.globals.lightProbeInfo.data;
+            const lightProbeInfoData = sceneGlobals.lightProbeInfo.data;
             let forceClose = false;
             if (this.bakeSettings.texture && !sceneGlobals.disableLightmap) {
                 forceClose = true;
--- a/cocos/3d/framework/mesh-renderer.ts
+++ b/cocos/3d/framework/mesh-renderer.ts
@@ -44,6 +44,7 @@ import { SubModel } from '../../render-scene/scene';
 import { isEnableEffect } from '../../rendering/define';
 import type { Model } from '../../render-scene/scene';
 import type { ReflectionProbeManager } from '../reflection-probe';
+import { getPipelineSceneData } from '../../rendering/pipeline-scene-data-utils';
 
 const { ccclass, help, executeInEditMode, executionOrder, menu, visible, type,
     formerlySerializedAs, serializable, editable, disallowAnimation } = _decorator;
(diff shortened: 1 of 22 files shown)


cocos/3d/framework/mesh-renderer.ts
```diff
@@ -950,13 +951,13 @@ export class MeshRenderer extends ModelRenderer {
         if (!mainLight.node) { return; }
 
         if (mainLight.node.mobility === MobilityMode.Static) {
+            const sceneGlobals = this.node.scene.globals;
+            const lightProbeInfoData = this.node.scene.globals.lightProbeInfo.data;
```
minggo: Can use `sceneGlobals `.

dumganhar: Good catch. I will fix that.

dumganhar: Fixed.

cocos/rendering/pipeline-scene-data-utils.ts
```diff
@@ -0,0 +1,47 @@
...
+    (cclegacy.director as Director).root!.pipeline.pipelineSceneData.isHDR = v;
+}
+
+export function getShadowsInPipelineSceneData (): Shadows {
+    return (cclegacy.director as Director).root!.pipeline.pipelineSceneData.shadows;
+}
+
+export function getSkyboxInPipelineSceneData (): Skybox {
+    return (cclegacy.director as Director).root!.pipeline.pipelineSceneData.skybox;
+}
```
minggo: I think just export `getPipelineSceneData ` is enough. There are many data in `pipelineSceneData`. Export more help functions will reduce the size but it also make codes less maintainable.

dumganhar: Could you explain more about `it also make code less maintainable`?, perhaps some code snippet is appreciated.

minggo: What i mean is that. As pipelineSceneData can be got by a helper function, why should export its internal data? As they are easy to get by using pipelineSceneData.xxx.  If export skybox and hdr here, then why not export other data, such as fog or shadows? How to determine if need to export a helper function?

dumganhar: > If export skybox and hdr here, then why not export other data, such as fog or shadows? How to determine if need to export a helper function?

Yep, for best code size optimization, we need to export all method in PipelineSceneData to utils. I exported just some of them because they are used more frequently. So if we export all methods in PipelineSceneData to `utils` file, would it cause the maintaining issue?

Anyway, good suggestion, I will change the code. And there will be only one method `getPipelineSceneData` in `pipeline-scene-data-utils.ts`.

dumganhar: Updated

dumganhar:
@cocos-robot run test cases

minggo (approved):

dumganhar:
@cocos-robot run test cases

PR #18176, opened 2025-01-13
[v3.8.6] Strip JSB relevant code in ui-opacity.ts for non-native platforms

dumganhar:
Re: #18056

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/2d/components/ui-opacity.ts
+++ b/cocos/2d/components/ui-opacity.ts
@@ -61,7 +61,7 @@ function setEntityLocalOpacityDirtyRecursively (
     const uiOp = node.getComponent(UIOpacity);
     if (uiOp && stopRecursiveIfHasOpacity) {
         // Because it's possible that UiOpacity components are handled by themselves (at onEnable or onDisable)
-        uiOp._parentOpacity = parentOpacity;
+        uiOp._setParentOpacity(parentOpacity);
         return;
     }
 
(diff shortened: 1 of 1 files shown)


cocos/2d/components/ui-opacity.ts
```diff
@@ -56,10 +130,17 @@ export class UIOpacity extends Component {
      *
      * @zh
      * 父节点的opacity。
+     *
+     * @engineInternal
+     * @mangle
      */
-    private _parentOpacity: number = 1.0;
+    public _parentOpacity: number = 1.0;
```
minggo: Make it as public, then other engine module may use it too. I don't think it is a good idea. We can't just only consider package size.

dumganhar: But there are already lots of `@engineInternal` public properties or methods are allowed. How do we deal with those public methods or properties, should they be forbidden ?

WeChat64900c9bf2fc4747304e87d10656c4b3

dumganhar: node.ts

企业微信截图_3068ebbe-4d74-4484-8247-b657afab95f2

dumganhar: Added a new internal method `_setParentOpacity` and revert `_parentOpacity` to private.

qiuguohua (approved):

minggo (approved):

PR #18185, opened 2025-01-14
game.restart() return a promise on native platforms

minggo:
https://github.com/cocos/cocos-engine/issues/17924

Make the return type of game.restart() the same on all platforms.

### DIFF
--- a/platforms/native/engine/jsb-game.js
+++ b/platforms/native/engine/jsb-game.js
@@ -24,13 +24,18 @@
  ****************************************************************************/
 
 cc.game.restart = function () {
-    // Need to clear scene, or native object destructor won't be invoke.
-    cc.director.getScene().destroy();
-    cc.Object._deferredDestroy();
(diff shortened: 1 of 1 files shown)


platforms/native/engine/jsb-game.js
```diff
@@ -24,13 +24,18 @@
...
-    __restartVM();
+    return new Promise((resolve) => {
+        // Need to clear scene, or native object destructor won't be invoked.
+        cc.director.getScene().destroy();
+        cc.Object._deferredDestroy();
+        __restartVM();
+
+        // Resolve the promise to indicate completion
+        resolve();
+    });
```
dumganhar, line 35: Doesn't need to put the __restartVM to Promise, just return `Promise.resolve();` is ok.

minggo: Yep, i will test it.

PR #18189, opened 2025-01-14
make support iOS 11+ and macOS 14+

minggo:


native/cocos/renderer/gfx-metal/MTLUtils.mm
```diff
@@ -1149,38 +1057,30 @@ void main() {
 
 NSUInteger mu::highestSupportedFeatureSet(id<MTLDevice> device) {
     NSUInteger maxKnownFeatureSet;
-    NSUInteger defaultFeatureSet;
 #if CC_PLATFORM == CC_PLATFORM_IOS
-    defaultFeatureSet = MTLFeatureSet_iOS_GPUFamily1_v1;
     if (@available(iOS 12.0, *)) {
-        maxKnownFeatureSet = MTLFeatureSet_iOS_GPUFamily4_v2;
+        maxKnownFeatureSet = MTLFeatureSet_iOS_GPUFamily5_v1;
```
star-e, line 1062: Will upgrade `maxKnownFeatureSet` break compatibility?

minggo: It will not break compatibility. As the following code will use it to check if it is really support
```c++
for (int featureSet = static_cast<int>(maxKnownFeatureSet); featureSet >= 0; --featureSet) {
        if ([device supportsFeatureSet:MTLFeatureSet(featureSet)]) {
            return static_cast<NSUInteger>(featureSet);
        }
    }
```

star-e (approved):

PR #18192, opened 2025-01-14
don't link unneeded framework

minggo:


native/CMakeLists.txt
```diff
@@ -3465,8 +3464,6 @@ if(APPLE)
         endif()
 
         target_link_libraries(${ENGINE_NAME} PUBLIC
-            "-framework QuartzCore"
-            "-framework MetalPerformanceShaders"
```
minggo, line 3469: Duplicated. They are included above.

dumganhar (approved):

PR #18205, opened 2025-01-16
[v3.8.6] Mangle more static properties by configuration.

dumganhar:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/.github/workflows/package-size-check.js
+++ b/.github/workflows/package-size-check.js
@@ -42,7 +42,29 @@ console.log(`features: [ ${features.join(', ')} ]`);
         },
         wasmCompressionMode: 'brotli',
         inlineEnum: true,
-        mangleProperties: true,
+        mangleProperties: {
+            mangleList: [
+                'UITransform._sortSiblings',
+                'UITransform._cleanChangeMap',
+                'Node._findComponents',
+                'Node._findChildComponent',
+                'Node._findChildComponents',
+                'Node.idGenerator',
+                'Node._stacks',
+                'Node._stackId',
+                'Node._setScene',
+                'EffectAsset._layoutValid',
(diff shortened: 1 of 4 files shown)


.github/workflows/package-size-check.js
```diff
@@ -42,7 +42,29 @@ console.log(`features: [ ${features.join(', ')} ]`);
...
+                'Node._stacks',
+                'Node._stackId',
+                'Node._setScene',
+                'EffectAsset._layoutValid',
+                'EffectAsset._effects',
+                'ReflectionProbe.DEFAULT_CUBE_SIZE',
+                'ReflectionProbe.DEFAULT_PLANER_SIZE',
+                'WebGLDeviceManager.setInstance',
+                'WebGL2DeviceManager.setInstance',
+                'CanvasPool',
```
minggo, line 62: Why need to add the list?

dumganhar: They will be added to default mangle config list too.
They are `protected` properties that 99% people will not use.
If the 1% people want to use it ,just remove some lines in `engine-mangle-config.json`

minggo: As it is to mangle engine variables and functions, why not add `@mangle` in codes.

dumganhar: We could.
Just to put the default mangle configuration in code by `@mangle` or put them to `mangleList` in  template configuration.
As we discuss before, protected properties of `Component` should not been mangled has already been set in `dontMangleList` section in the template configuration.

minggo: Ok, got it.

PR #18282, opened 2025-02-06
refactor(components): enhance component accessor type safety with generics

9romise:
Re: #

### Changelog

- Convert component accessor methods to use generic types for precise type inference
- Improve developer experience with auto-completion and type checking

Before:
```
const clz = node.getComponent('clz') as CLZ | null
```

After:
```
const clz = node.getComponent<CLZ>('clz')
```

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/scene-graph/node.ts
+++ b/cocos/scene-graph/node.ts
@@ -877,7 +877,7 @@ export class Node extends CCObject implements ISchedulable, CustomSerializable {
       * var test = node.getComponent("Test");
       * ```
       */
-    public getComponent(className: string): Component | null;
+    public getComponent<T extends Component>(className: string): T | null;
 
     public getComponent<T extends Component> (typeOrClassName: string | Constructor<T> | AbstractedConstructor<T>): T | null {
         const constructor = getConstructor(typeOrClassName);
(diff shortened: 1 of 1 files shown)


dumganhar:
@cocos-robot run test cases

minggo (approved):

minggo:
@cocos-robot run test cases

dumganhar:
Hi, @9romise , @minggo 

This code will break the compatibilty in engine, for example, In spine/skeleton.ts.

```ts
debugDraw = debugDrawNode.addComponent('cc.Graphics') as Graphics;
```

ESLint report: This assertion is unnecessary since it does not change the type of the expression.

If I save the file, eslint configuration will automatically remove `as Graphics`, so the code will be
```ts
debugDraw = debugDrawNode.addComponent('cc.Graphics');
```

But it causes another TS compiler error:

```
Type 'Component' is missing the following properties from type 'Graphics': lineWidth, lineJoin, lineCap, strokeColor, and 101 more.
```

Yes, we could modify the code to 

```ts
debugDraw = debugDrawNode.addComponent<Graphics>('cc.Graphics');
```

to resolve the issue in engine. But if our users also write the same kinds of code, they will complain about the bad compatibilty.

And I think the two kinds of `addComponent` declaration doesn't have much differences.

```ts
debugDraw = debugDrawNode.addComponent('cc.Graphics'); // Old way and new way both return `Component | null` type
debugDraw = debugDrawNode.addComponent<Graphics>('cc.Graphics'); // New way in this PR, Return Graphics type
debugDraw = debugDrawNode.addComponent('cc.Graphics') as Graphics; // Old way, Return Graphics type
```

I suggest to revert this PR for better compatibility. What do you think?

minggo:
@dumganhar can we keep both of the two usage?

dumganhar:
@minggo It seems we could add the new overrode `addComponent` for the compatibilty, let me investigate more.

dumganhar:
Sent a PR for this:

https://github.com/cocos/cocos-engine/pull/18345

@minggo, @9romise , please review it.

PR #18306, opened 2025-02-10
Fix sprite inspector i18n configurations

dumganhar:
Re: #

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/2d/components/sprite.ts
+++ b/cocos/2d/components/sprite.ts
@@ -180,7 +180,6 @@ export class Sprite extends UIRenderer {
      */
     @type(SpriteAtlas)
     @displayOrder(4)
-    @tooltip('i18n:sprite.atlas')
     get spriteAtlas (): SpriteAtlas | null {
         return this._atlas;
     }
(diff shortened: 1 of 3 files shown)


cocos/2d/components/sprite.ts
```diff
@@ -180,7 +180,6 @@ export class Sprite extends UIRenderer {
      */
     @type(SpriteAtlas)
     @displayOrder(4)
-    @tooltip('i18n:sprite.atlas')
```
minggo, line 183: Why remote it?

dumganhar: Because it doesn't exist in i18n configuration (localization.js), but it has already been in `ui.js`
See: https://github.com/cocos/cocos-engine/blob/8ae4dcf8a3f62c57310e0e25fa9a7b0717b942b5/editor/i18n/en/localization.js#L333

I added the tooltip in ui.js which is more suitable since it just matches the property and doesn't need to configure it in the source code (sprite.ts).

minggo (approved):

knoxHuang (approved):

PR #14572, opened 2023-03-14
remove and ban @ts-expect-error

PPpro:
Re: https://github.com/cocos/3d-tasks/issues/15855

### Changelog

* remove and ban `@ts-expect-error`

-------

### Continuous Integration

This pull request:

* [x] needs automatic test cases check.
  > Manual trigger with `@cocos-robot run test cases` afterward.
* [ ] does not change any runtime related code or build configuration
  > If any reviewer thinks the CI checks are needed, please uncheck this option, then close and reopen the issue.

-------

### Compatibility Check

This pull request:

* [ ] changes public API, and have ensured backward compatibility with [deprecated features](https://github.com/cocos/cocos-engine/blob/v3.5.0/docs/contribution/deprecated-features.md).
* [ ] affects platform compatibility, e.g. system version, browser version, platform sdk version, platform toolchain, language version, hardware compatibility etc.
* [ ] affects file structure of the build package or build configuration which requires user project upgrade.
* [ ] **introduces breaking changes**, please list all changes, affected features and the scope of violation.

PR #18320, opened 2025-02-12
UIRenderer.updateMaterial should not be marked as `@mangle`

dumganhar:
This fix an error reported in daily build.

cocos-test-projects/assets/cases/rendertexture depends on this method, so it should not be marked as `@mangle` now.

`protected` is not equal to `@engineInternal + public`, because `protected` methods are also APIs exposed to developers,
For example, developers could implement a class which extends `UIRenderer` and call this method.
The mistake was merged in https://github.com/cocos/cocos-engine/pull/14572 , and it needs to be fixed in the future.

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/2d/framework/ui-renderer.ts
+++ b/cocos/2d/framework/ui-renderer.ts
@@ -471,8 +471,11 @@ export class UIRenderer extends Renderer {
     }
 
     /**
+     * cocos-test-projects/assets/cases/rendertexture depends on this method, so it should not be marked as `@mangle` now.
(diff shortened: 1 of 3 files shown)


minggo (approved):

dumganhar:
Testcase code:

https://github.com/cocos/cocos-test-projects/blob/v3.8.6/assets/cases/rendertexture/capture_to_web.ts#L36

PR #18328, opened 2025-02-13
Treeshake xr module import.

dumganhar:
Re: https://github.com/cocos/cocos-engine/issues/18222

### Changelog

*

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

### DIFF
--- a/cocos/ui/button.ts
+++ b/cocos/ui/button.ts
@@ -25,7 +25,7 @@
 
 import { ccclass, help, executionOrder, menu, requireComponent, tooltip, displayOrder, type, rangeMin,
     rangeMax, serializable, executeInEditMode } from 'cc.decorator';
-import { DEBUG, EDITOR, EDITOR_NOT_IN_PREVIEW } from 'internal:constants';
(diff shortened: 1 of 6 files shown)


dumganhar:
@cocos-robot run test cases

cocos/ui/button.ts
```diff
@@ -1057,12 +1068,14 @@ export class Button extends Component {
     }
 
     private _xrClick (): void {
+        if (!USE_XR) return;
```
minggo, line 1071: Why not do it like this
```typescript
if (USE_XR) {
    ...
}
```

Than all codes will be removed if USE_XR is false.

dumganhar: Their treeshaking results are the same.
`if (USE_XR) {}` will make the original code be intented by 4 space right and it will be more difficult to maintain.
Just adding **one line code** to treeshake the function is better since we avoid `{`, `}`, `4 spaces indentation`.

minggo: > Their treeshaking results are the same.

If so, then there is not problem.

minggo (approved):

PR #18411, opened 2025-03-05
If float filtering is not supported, then it falls back to using unfilterable settings

GengineJS:


star-e (approved):

cocos/gfx/webgpu/webgpu-device.ts
```diff
@@ -477,6 +477,10 @@ export class WebGPUDevice extends Device {
         defaultResource.descSet.update();
     }
 
+    get FloatFilterable (): boolean {
```
dumganhar, line 480: Property name should start with `lowercase`.

PR #18455, opened 2025-03-13
Fix the issue of inconsistent attachment sizes

GengineJS:


dumganhar (approved):

cocos/gfx/webgpu/define.ts
```diff
@@ -85,13 +85,39 @@ export class DefaultResources {
...
+            Format.DEPTH_STENCIL,
+            width,
+            height,
+            TextureFlagBit.NONE,
+            1,
+            1,
+            SampleCount.X1,
+            1,
+        );
+        return texInfo;
```
dumganhar: Do we need to `new TextureInfo` each time invoke `getDefaultDSInfo`? Could we cache the default value?

star-e (approved):

star-e:
@cocos-robot run test cases

PR #18467, opened 2025-03-15
fix spine renderDrawInfo not clear at onDisable

gaojie111119:
issue: https://forum.cocos.org/t/topic/166248
076fe7fa-6b6c-465f-8bcc-778bd544f691

After review code, I find that when call **SkeletonRenderer::render** 
`    
auto *entity = _entity;
entity->clearDynamicRenderDrawInfos();
_sharedBufferOffset->reset();
 _sharedBufferOffset->clear();
` 
be called; 
but when call **SkeletonRenderer::onDisable**, only clear _sharedBufferOffset but not clear _entity .

after clear _entity at SkeletonRenderer::onDisable. no crash happened.

### DIFF
--- a/native/cocos/editor-support/spine-creator-support/SkeletonCacheAnimation.cpp
+++ b/native/cocos/editor-support/spine-creator-support/SkeletonCacheAnimation.cpp
@@ -460,6 +460,9 @@ void SkeletonCacheAnimation::beginSchedule() {
 void SkeletonCacheAnimation::stopSchedule() {
     MiddlewareManager::getInstance()->removeTimer(this);
 
+    if (_entity != nullptr) {
+        _entity->clearDynamicRenderDrawInfos();
+    }
     if (_sharedBufferOffset) {
         _sharedBufferOffset->reset();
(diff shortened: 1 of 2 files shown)


gaojie111119:
补充：spine节点onDisable时必须将skeletonData置为null

dumganhar:
Cache 模式下，SkeletonCacheAnimation::stopSchedule 也需要修改下。

gaojie111119:
> Cache 模式下，SkeletonCacheAnimation::stopSchedule 也需要修改下。

I added it. Maybe Dragonbones (CCArmature) has the same problem

dumganhar:
> I added it. Maybe Dragonbones (CCArmature) has the same problem

@gaojie111119 Yes, it probably causes the same issue as which in Spine. 

But I think we should focus on Spine in this Pull Request. If there are issues in DragonBones adapter code, we need to send another Pull Request.

dumganhar (approved):

dumganhar:
@cocos-robot run test cases

PR #18511, opened 2025-03-26
提供spine可被自定义顶点数据格式的方法

NorhsArk:


cocos/spine/assembler/simple.ts
```diff
@@ -86,12 +91,21 @@ function _getSlotMaterial (blendMode: number, comp: Skeleton): MaterialInstance
 
 class Simple implements IAssembler {
     vCount = 32767;
+
+    public static customVfmts (customizedOneColorVfmt: Attribute[], customizedTwoColorVfmt: Attribute[]): void {
+        vfmtOneColor = customizedOneColorVfmt;
+        vfmtTwoColor = customizedTwoColorVfmt;
+        _byteStrideOneColorCustomized = getAttributeStride(vfmtOneColor);
```
bofeng-song: Does it need to recalculate _byteStrideOneColor?

NorhsArk: Yes, when the number of bit stride in the custom vertex format is different, it needs to be recalculated

bofeng-song: It's not recalculate here.

cocos/spine/assembler/simple.ts
```diff
@@ -95,8 +93,6 @@ class Simple implements IAssembler {
     public static customVfmts (customizedOneColorVfmt: Attribute[], customizedTwoColorVfmt: Attribute[]): void {
         vfmtOneColor = customizedOneColorVfmt;
         vfmtTwoColor = customizedTwoColorVfmt;
```
bofeng-song, line 95: As colorVfmt is customized, _byteStrideOneColor & _byteStrideTwoColor can not be delcared as const.
And its value must be recalculated here

cocos/spine/skeleton-cache.ts
```diff
@@ -28,12 +28,15 @@ import { vfmtPosUvColor4B, vfmtPosUvTwoColor4B, getAttributeStride } from '../2d
 import spine from './lib/spine-core';
 import { SkeletonData } from './skeleton-data';
 import { warn } from '../core/platform/debug';
+import { Attribute } from '../gfx';
 
 const MaxCacheTime = 30;
 const FrameTime = 1 / 60;
 const _useTint = true;
 const _byteStrideOneColor = getAttributeStride(vfmtPosUvColor4B);
 const _byteStrideTwoColor = getAttributeStride(vfmtPosUvTwoColor4B);
```
bofeng-song, line 37: _byteStrideOneColor may be changed after call customVfmts();

cocos/spine/assembler/simple.ts
NorhsArk: The _byteStrideOneColor & _byteStrideTwoColor in the realTimeTraverse function is always used because  The vertex data format obtained in spine.wasmUtel.wasm.heap8 is fixed. The custom vertex format byteStripe can be obtained from the renderData's allocator without the need for recalculation

cocos/spine/skeleton-cache.ts
NorhsArk: The _byteStrideOneColor in skeleton-cache.ts will always be used because the vertex data format obtained in spine.wasmUtel.wasm.heapu8 in updateRenderData is fixed. And the customized 'byteRideOneColorCustomized' only needs to be calculated in customVfmts

bofeng-song:
@NorhsArk Could you upload you demo? At this PR, it doesn`t inlude native's modify.
This will result in inconsistencies between native and web implementations.

PR #18520, opened 2025-03-27
gpu instancing sort

GengineJS:


cocos/rendering/define.ts
```diff
@@ -1121,3 +1121,21 @@ export function isEnableEffect (): boolean {
 }
 
 /* eslint-enable max-len */
+
+export function getPassPool (): RecyclePool<IRenderPass> {
+    return new RecyclePool<IRenderPass>((): {
```
dumganhar, line 1126: Do we need to return `new RecyclePool` each time invoking `getPassPool`?
If so, perhaps `createPassPool` is a better name.

native/cocos/renderer/pipeline/RenderInstancedQueue.cpp
```diff
@@ -42,11 +42,28 @@ void RenderInstancedQueue::clear() {
...
+    for (const auto& item : tempSorted) {
+        if (item->getPass()->getBlendState()->targets[0].blend == 0) {
+            _renderQueues.emplace_back(item);
+        }
+    }
+    for (const auto& item : tempSorted) {
+        if (item->getPass()->getBlendState()->targets[0].blend != 0) {
+            _renderQueues.emplace_back(item);
+        }
+    }
```
star-e: Here can sort `std::forward_as_tuple(lhs.blend, lhs.hash, lhs.shaderID) < std::forward_as_tuple(rhs.blend, rhs.hash, rhs.shaderID)`

cocos/rendering/instanced-buffer.ts
```diff
@@ -86,6 +101,11 @@ export class InstancedBuffer {
             shader = subModel.shaders[passIdx];
         }
         const descriptorSet = subModel.descriptorSet;
+        const hash = (0 << 30) | (subModel.passes[passIdx].priority as number) << 16 | (subModel.priority as number) << 8 | passIdx;
```
dumganhar: What's the purpose of (0 << 30) ? 
0 << 30 is still 0, and it doesn't affect following `|`  bit operations.

cocos/rendering/render-instanced-queue.ts
```diff
@@ -55,21 +55,16 @@ export class RenderInstancedQueue {
...
+                this._renderQueue.push(item);
             }
-            res = it.next();
         }
-        it = this.queue.values();
-        res = it.next();
-        while (!res.done) {
-            if (res.value.pass.blendState.targets[0].blend) {
-                this._renderQueue.push(res.value);
+        for (const item of sortedArray) {
```
dumganhar: `sortedArray.forEach` has better performance and smaller code size.

star-e (approved):

star-e (approved):

cocos/rendering/define.ts
GengineJS: This will not be called every frame, but only once at initialization.

dumganhar (approved):

PR #18556, opened 2025-04-03
Fix uv update

WenhaiLin:


dumganhar:
@WenhaiLin thanks for the bug fix.
Could you describe what the issue you got? 
At glance, the patch seems good. But I want to know what situation will cause the issue, then we could make a test case for bug.

WenhaiLin:
错误来自线上平台收集 已有其他PR处理这类问题  https://github.com/cocos/cocos-engine/pull/18584

finscn:
> 错误来自线上平台收集 已有其他PR处理这类问题 #18584

但是这个 pr 没处理 spirteFrame 不存在的情况.

Issue #16860, opened 2024-04-03
[3.8.2] When the sprite component renders a texture from RenderTexture, it cannot perform merged rendering. Merged rendering can only be achieved after using a custom material


zhefengzhang:
### Cocos Creator version

3.8.2

### System information

All

### Issue description

![image](https://github.com/cocos/cocos-engine/assets/35944775/def78d00-d9d7-4722-b6dc-bf81ec1126d5)

When a new material resource is created using the builtin-sprite shader and this material is set as the custom material for the sprite component, merged rendering can be achieved.
![image](https://github.com/cocos/cocos-engine/assets/35944775/42f6c0dd-9b98-40bc-bd2b-5b9cb942f9c5)

### Relevant error log output

-

### Steps to reproduce

-

### Minimal reproduction project

[NewProject.zip](https://github.com/cocos/cocos-engine/files/14846021/NewProject.zip)

PR #18587, opened 2025-04-10
fixed rt sprite cannot batch drawcall

zhefengzhang:
Re: https://github.com/cocos/cocos-engine/issues/16860

### DIFF
--- a/cocos/2d/components/sprite.ts
+++ b/cocos/2d/components/sprite.ts
@@ -27,6 +27,7 @@ import { ccclass, help, executionOrder, menu, tooltip, displayOrder, type, range
 import { BUILD, EDITOR } from 'internal:constants';
 import { SpriteAtlas } from '../assets/sprite-atlas';
 import { SpriteFrame, SpriteFrameEvent } from '../assets/sprite-frame';
+import { builtinResMgr } from '../../asset/asset-manager/builtin-res-mgr';
 import { Vec2, cclegacy, ccenum, clamp, warnID } from '../../core';
 import { IBatcher } from '../renderer/i-batcher';
 import { UIRenderer, InstanceMaterialType } from '../framework/ui-renderer';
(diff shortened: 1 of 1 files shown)


dumganhar (approved):

cocos/2d/components/sprite.ts
```diff
@@ -564,9 +565,14 @@ export class Sprite extends UIRenderer {
     protected _updateBuiltinMaterial (): Material {
         let mat = super._updateBuiltinMaterial();
         if (this.spriteFrame && this.spriteFrame.texture instanceof RenderTexture) {
-            const renderMat = new Material();
-            renderMat.copy(mat, { defines: { SAMPLE_FROM_RT: true } });
-            mat = renderMat;
+            const rtMatName = `rt-${mat.name}`;
+            var rtMat = builtinResMgr.get(rtMatName) as Material | null;
```
dumganhar: var -> let

dumganhar:
@cocos-robot run test cases

cocos/2d/components/sprite.ts
```diff
@@ -564,9 +565,14 @@ export class Sprite extends UIRenderer {
     protected _updateBuiltinMaterial (): Material {
         let mat = super._updateBuiltinMaterial();
         if (this.spriteFrame && this.spriteFrame.texture instanceof RenderTexture) {
-            const renderMat = new Material();
-            renderMat.copy(mat, { defines: { SAMPLE_FROM_RT: true } });
-            mat = renderMat;
+            const rtMatName = `rt-${mat.name}`;
```
dumganhar, line 568: Would the `mat.name` be empty? If it's empty, `rtMatName` will be `rt-` which will cause issues.

dumganhar: If it's possible empty, it's better to check the name, keep the old logic if it's empty.

zhefengzhang: The original code logic does not consider the possibility of mat being null, since built-in resources are generally not deleted.
In the engine's core code, there's no implementation that actively deletes assets from builtinResMgr.
I believe it's unnecessary to account for potential errors caused by engine modifications through unconventional channels.

dumganhar: That makes sense.

PR #18868, opened 2025-07-28
Fixed the problem that addSceneOfCamera function cannot draw UI

GengineJS:


star-e:

PR #18877, opened 2025-08-04
The CC_USE_2D macro definition of the material is changed from an instance object to a shared object

GengineJS:


star-e:
@cocos-robot run test cases

star-e:
reverted in https://github.com/cocos/cocos-engine/pull/18880

PR #19026, opened 2025-09-24
fixed particle2d fade-out error

zhefengzhang:
Re: https://github.com/cocos/3d-tasks/issues/19132

### DIFF
--- a/cocos/particle-2d/particle-simulator-2d.ts
+++ b/cocos/particle-2d/particle-simulator-2d.ts
@@ -51,7 +51,7 @@ function getWorldRotation (node): number {
 class Particle {
     public pos = new Vec2(0, 0);
     public startPos = new Vec2(0, 0);
-    public color = new Color(0, 0, 0, 255);
+    public color = { r: 0, g: 0, b: 0, a: 255 };
     public deltaColor = { r: 0, g: 0, b: 0, a: 255 };
     public size = 0;
     public deltaSize = 0;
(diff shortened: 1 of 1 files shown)


cocos/particle-2d/particle-simulator-2d.ts
```diff
@@ -51,7 +51,7 @@ function getWorldRotation (node): number {
 class Particle {
     public pos = new Vec2(0, 0);
     public startPos = new Vec2(0, 0);
-    public color = new Color(0, 0, 0, 255);
+    public color = { r: 0, g: 0, b: 0, a: 255 };
```
zhefengzhang, line 54: 因为 Color 类型对象使用 Uint8ClampedArray 类型数组存储 rgba 数值，此类型数组的元素值添加时会对数值进行四舍五入。
而在 particle-simulator-2d.ts 中计算每帧粒子的颜色数据需要 rgba 值与 dt 浮点值相乘，最终得到的 rgba 浮点值赋值给 color 的 rgba 值时会被四舍五入，导致最终显示结果不符合预期。

Because Color type objects use an array of type Uint8ClampedArray to store rgba values, element values of this array type are rounded when added.
However, calculating the color data of each frame particle in particle-simulator-2d.ts requires multiplying the rgba value by the dt floating point value. The resulting rgba floating point value assigned to the rgba value of color will be rounded, resulting in the final display result not meeting expectations.

PR #19118, opened 2025-12-29
Fix webgl1 and webgl2 vao implementation on ios

### LINKED ISSUE #19088: IOS 26 rendering problem (missing objects)
### Cocos Creator version

3.8.7

### System information

IOS 26, webgl1

### Issue description

This is a issue that affects all the glitches I've noticed in iOS 26, and they're all related to rendering. I think they're all interconnected.
There are several different problems:
1. Sprites with `UIOpacity` disappear when the orientation changes — this problem seems to be solved by updating parameters such as `opacity` after the orientation has changed.
2. `WebGL: INVALID_OPERATION: glDrawElements: Vertex buffer is not big enough for the draw call.`
I don't know exactly what causes it yet, but in my case, literally one Sprite caused the problem in the production code. If you hide it, everything is fine. I will try to recreate the test scene later, but I couldn't recreate it on a bare project.
This has already been discussed on the [forum](https://forum.cocos.org/t/topic/170848), but it didn't work for me and the bug didn't show up. If you manage to recreate it on a blank project, please send it to us.

### Relevant error log output

```
WebGL: INVALID_OPERATION: glDrawElements: Vertex buffer is not big enough for the draw call.
```

### Steps to reproduce

The following option was suggested for the forum, but it did not reveal the problem for me:
```
Reproduction method:
Attach an image resource (or an atlas) to a UI node.  
Then, in the code, create two sprites, for example:

const node = new Node();
node.addComponent(Sprite).spriteFrame = this._atlas.getSpriteFrame("icon_0");
this.node.addChild(node);

After that, toggle the visibility of these two nodes in turns, for example:

let flag = false;
this.schedule(() => {
    flag = !flag;
    node_0.active = flag;
    node_1.active = !flag;
}, 0.2);
```

_No response_

raduangelescu:
Sometimes the VAO on IOS would not bind correctly because of the async nature of IOS metal. To solve this I implemented vertex buffer versioning.

This should fix https://github.com/cocos/cocos-engine/issues/19088 . (tested it on the issue I was able to reproduce and it was fixed). 
### Changelog

*  This is a similar solution that Three.js would do but at a higher level. We do a hash of the vertex buffer and rebind VAO when the hash is different.

-------

### Continuous Integration

This pull request:

-------

### Compatibility Check

This pull request:

<h3>Greptile Summary</h3>

Implemented iOS-specific VAO versioning to fix rendering issues caused by async Metal backend behavior. When buffers are updated, their version counters increment, and VAOs are recreated if buffer versions changed since VAO creation. Additionally forces VAO rebinding on every draw call on iOS.

**Key changes:**
- Added `updateVersion` counter to buffer objects, incremented on each buffer update
- Added `vaoVersions` map to track buffer version hashes when VAOs are created
- On iOS, computes version hash from all vertex/index buffers and compares with cached version
- Recreates VAO if hash mismatch detected
- Forces VAO rebinding on every draw call on iOS (bypasses cache check)

**Issues found:**
- Hash collision vulnerability: weighted sum approach can produce identical hashes for different buffer update patterns
- Performance concern: forced rebinding on every draw may have overhead

<h3>Confidence Score: 3/5</h3>

- PR addresses real iOS rendering bug but hash collision risk could cause correctness issues
- The implementation successfully tracks buffer updates and recreates VAOs, which should fix the iOS Metal async issue. However, the weighted sum hashing approach has mathematical collision vulnerabilities that could prevent VAO recreation when needed, potentially leaving the original bug unfixed in edge cases. The forced rebinding provides a safety net but may impact performance.
- `cocos/gfx/webgl/webgl-commands.ts` and `cocos/gfx/webgl2/webgl2-commands.ts` need attention for hash collision fix

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| cocos/gfx/webgl/webgl-commands.ts | Added iOS-specific VAO versioning with buffer update tracking and forced rebinding; potential hash collision in version calculation |
| cocos/gfx/webgl2/webgl2-commands.ts | Added iOS-specific VAO versioning identical to WebGL1 implementation; same hash collision risk exists |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant App as Application
    participant Buf as Buffer Update
    participant Ver as Version Tracking
    participant Bind as Bind States
    participant VAO as VAO Management
    participant GPU as GPU/Metal

    App->>Buf: Update vertex/index buffer
    Buf->>Ver: Increment updateVersion++ (iOS only)
    Buf->>Buf: Unbind current VAO
    Buf->>GPU: Upload buffer data

    App->>Bind: Draw call - bind states
    Bind->>Ver: Compute current version hash
    Note over Ver: hash = Σ(bufferVersion[i] × (i+1))<br/>+ indexVersion × 1000
    
    Bind->>VAO: Get cached VAO for shader
    
    alt VAO exists and iOS
        Bind->>Ver: Get cached vaoVersion
        Ver->>Bind: Return cached version
        
        alt Versions match
            Bind->>VAO: Reuse existing VAO
        else Versions differ
            Bind->>VAO: Delete old VAO
            Bind->>VAO: Create new VAO
            Bind->>Ver: Store new vaoVersion
            VAO->>GPU: Configure vertex attributes
        end
    else No VAO exists
        Bind->>VAO: Create new VAO
        Bind->>Ver: Store vaoVersion (iOS only)
        VAO->>GPU: Configure vertex attributes
    end
    
    alt iOS platform
        Bind->>VAO: Force rebind VAO (every draw)
    else Other platforms
        Bind->>VAO: Bind if cache miss
    end
    
    Bind->>GPU: Draw call with VAO
```

### DIFF
--- a/cocos/2d/components/rich-text.ts
+++ b/cocos/2d/components/rich-text.ts
@@ -44,6 +44,7 @@ import {
     getEnglishWordPartAtLast,
     getSymbolAt,
 } from '../utils/text-utils';
+import { Sorting2D } from '../../sorting/sorting-2d';
 
 const _htmlTextParser = new HtmlTextParser();
 const RichTextChildName = 'RICHTEXT_CHILD';
@@ -504,10 +505,21 @@ export class RichText extends Component {
     protected _lineOffsetX = 0;
     protected declare _updateRichTextStatus: () => void;
     protected _labelChildrenNum = 0; // only ISegment
+    protected _currentSortingLayer = 0;
+    protected _currentSortingOrder = 0;
+    protected _sorting2d: Sorting2D | null = null;
 
     constructor () {
         super();
         this._updateRichTextStatus = this._updateRichText;
+    }
+
+    protected override __preload (): void {
+        this._sorting2d = this.getComponent(Sorting2D);
+        if (this._sorting2d != null) {
+            this._currentSortingOrder = this._sorting2d.sortingOrder;
+            this._currentSortingLayer = this._sorting2d.sortingLayer;
+        }
     }
 
     public onLoad (): void {
@@ -864,7 +876,11 @@ export class RichText extends Component {
         this.node.insertChild(labelSegment.node, this._labelChildrenNum++);
         this._applyTextAttribute(labelSegment);
         this._segments.push(labelSegment);
-
+        if (this._sorting2d != null) {
+            const childSorting = labelSegment.node.addComponent(Sorting2D);
+            childSorting.sortingOrder = this._sorting2d.sortingOrder;
+            childSorting.sortingLayer = this._sorting2d.sortingLayer;
+        }
         return labelSegment;
     }
 
@@ -1046,6 +1062,11 @@ this._measureText(styleIndex) as unknown as (s: string) => number,
             if (event) {
                 segment.clickHandler = event.click;
                 segment.clickParam = event.param;
+            }
+            if (this._sorting2d != null) {
+                const childSorting = segment.node.addComponent(Sorting2D);
+                childSorting.sortingOrder = this._sorting2d.sortingOrder;
+                childSorting.sortingLayer = this._sorting2d.sortingLayer;
             }
         }
     }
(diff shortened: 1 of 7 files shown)


lw404-ai:
@raduangelescu Is it necessary to also include macOS, as macOS is facing a similar issue now?

